PR #191, opened 2014-02-12
Overminddl1 inputchanges

OvermindDL1:
Added in support for raw keycode values in to the previously unused keysym field (now renamed raw).  Adds a P_RAW to the Urho3D variant field for keyboard calls.  Looks like extra mouse buttons are already supported but not tested yet, will later and if any issues will submit another pull request for those.

### DIFF
--- a/Source/ThirdParty/SDL/src/input/evdev/SDL_evdev.c
+++ b/Source/ThirdParty/SDL/src/input/evdev/SDL_evdev.c
@@ -610,9 +610,9 @@ SDL_EVDEV_Poll(void)
                     scan_code = SDL_EVDEV_translate_keycode(events[i].code);
                     if (scan_code != SDL_SCANCODE_UNKNOWN) {
                         if (events[i].value == 0) {
-                            SDL_SendKeyboardKey(SDL_RELEASED, static_cast<Uint32>(events[i].code), scan_code);
+                            SDL_SendKeyboardKey(SDL_RELEASED, (Uint32)(events[i].code), scan_code);
                         } else if (events[i].value == 1 || events[i].value == 2 /* Key repeated */ ) {
-                            SDL_SendKeyboardKey(SDL_PRESSED, static_cast<Uint32>(events[i].code), scan_code);
+                            SDL_SendKeyboardKey(SDL_PRESSED, (Uint32)(events[i].code), scan_code);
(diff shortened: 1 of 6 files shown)


weitjong:
I am this close to revert back the changes introduced in this PR in order to have easier future maintenance when upgrading SDL from upstream. As I understand it, the objective of this PR is to pass along the "raw" keyboard code from the platform-specific keyboard to the Urho3D input subsystem. This looks fine at first glance, however, IMHO there are at least two problems with it.

1. The "raw" code as its name implies is really raw and hardware/platform dependent. There is no easy way to consume the P_RAW at our end in a generic matter.
2. The changes are too intrusive to SDL as it touches a common SDL keyboard interface which forces this PR to patch all the platform-specific keyboard implementation for all the platforms SDL supports! And that's a lot, besides new platforms keep coming on-board in subsequent SDL releases.

IMHO, if we are in a situation where a raw code is not recognized yet by SDL, i.e. SDL cannot  translate it into a generic "SDL scan code" then the right course of action is to modify the translation map so that it becomes supported. App should never be exposed to platform-specific stuff.

If I do not hear any objections then I will proceed to revert the changes.

OvermindDL1:
As long as there is some way to get all key mappings.  My keyboard has a
large set of extras keys mapped to the xfree86 Linux scancodes, which SDL
does not support for some unknown reason.  Those keys are standard key
mappings, just SDL does not pass them on for some retarded reason.

Those keys include keys like volume, quick copy, calculator button, etc,
highly useful for games that need many key mappings.

weitjong:
I have a keyboard like that too! It's my t...rusty Logitech. Anyway, if that is the problem then we should fix the problem at its root. Either in the src/video/x11/SDL_x11events.c or src/core/linux/SDL_evdev.c (to be investigated).

OvermindDL1:
Lol, precisely that, my Logitech G15,and I would love values are for all of
XFree86,especially F13-F16 keys too.

JoshEngebretson:
I wonder if 2.04 changes anything up on keyboard handling.  On our CEF3 WebView, I have to convert from SDL codes back to native platform codes to send to it for offscreen rendering, since there is no abstraction.

weitjong:
The keyboard handling is more or less the same in 2.0.4 in this respect. The code maps the platform-specific code to SDL scan code and send the scan code out.

PR #333, opened 2014-04-28
Added Catmull-Rom spline type which interpolates through the control nod...

szamq:
...es

I found that it is very difficult to control current the curve with current splinepath class because slightly change of one node changes all curve. So, here is Catmull-Rom spline which interpolates through control nodes and position change of one node changes curve only locally(4 nodes). Note it needs 4 nodes minimum to actually work.
Preview camera flying through this spline.
http://gfycat.com/TediousHeartyEnglishsetter


cadaver:
If you rebase this to current, I believe it's OK to merge.

szamq:
Is it ok now? I tried to merge commits into one but it just fails

PR #436, opened 2014-08-26
A basic terrain editor that integrates with the Urho3d Editor. This is i...

luveti:
...n very early stages. It currently supports raising/lowering terrain, setting the height of the terrain, and smoothing the terrain (which needs some serious improvisation). This is all done using "Brushes", which you can create and edit (only the alpha of the brush affects the brushes effect on the terrain). Terrains are saved to the file they we're loaded from, and editing one will cause SetSceneModified to be called (therefore prompting you to save before exiting). This is WIP and should not be considered done by any means.


VinnyVicious:
Any news on this? @luveti, @cadaver?

VinnyVicious:
No love for this PR @cadaver? :(

cadaver:
It was left dormant for a long time. Would be good to hear from the author if it's now in the intended state for merging.

luveti:
I've yet to add undo / redo functionality, so I wouldn't say it's quite ready to merge. I can look into adding this though :)

hdunderscore:
Just tested it out, seems there may be some UI issues although it's possible I have made some errors. Mind posting a screenshot of what the UI should look like?

luveti:
I'll upload a photo as soon as I get a chance. I was actually having an issue with the UI as well. The window for the tool was expanding much more then it should have been.

luveti:
This is what it looks like for me right now, on the latest commit of Urho3D:

fcfa02d8843e115d8fb02d6a417e9ec4

Dragging the bottom of the window down a bit will force it to re-layout to the correct height, like so:

a20dce958425fccf516a2a4ed825c05f

I believe this has something to do with the BorderImage that holds the current tools description text. It also appears that some of the icons aren't loading anymore either..

hdunderscore:
Yeah that's what I am getting. The icons don't show up because the PR (at least recent commits) don't seem to include the changed icons.png. 

So far I've been able to use the first three tools of the terrain editor, I haven't put in significant effort to understand how the other tools are meant to work yet.

My first impression was that the UI would include some options such as brush size and the terrain height for the terrain height tool. The empty panel with the title in the UI also led me to believe that, but I guess that's not the case.

Functionally it works well enough for quick adjustments, which is good enough for me even without the undo functionality. A follow up PR could include undo functionality.

luveti:
Here's what it looks like with my recent commits:
eaaa59bbda5d9cfaf123e24af09b1ae9
And a short video: https://i.gyazo.com/0da50e6614451352c25f8154b467720a.mp4

I'm currently working on undo / redo functionality at the moment. I'm not sure when I'll be able to implement the other tools; Especially the tree and foliage ones, as I feel like Urho3D would require some sort of new sub system to support those properly.

The "Set Height" edit mode could defiantly use a height slider of some sort! So I'll go ahead and add that in!

Brush size shouldn't be hard to implement. I may hold off on undo / redo support and get custom brush sizes in first.

luveti:
I've implemented the "Settings" section. It gets updated based on the current tool you have selected. I've also improved the smoothing algorithm to be a bit faster. And I've added more functionality to the "Create Terrain" button. It now creates a new heightmap, which is always a power of 2 at the moment; then saves it to the disk as "heightmap-<milliseconds since epoch>.png".

I've also refactored and cleaned up this 2 year old code a ton :). Which includes alphabetizing it and updating it to follow the official coding standards.

Let me know what you guys think! The next big step is getting undo / redo going.

luveti:
Alright, undo / redo support has been added. It could surely be improved upon, as it currently adds a new EditAction on every call to EditorTerrain.Work. But this should probably be fine for the first version. I would say this could be merged into master now! @cadaver :)

9773c46cc122fccdc78701922c7137e5

1vanK:
Yoy can draw brush as circle (or any another shape)
https://i.ytimg.com/vi/xgwmZ03vZbI/hqdefault.jpg
http://i58.tinypic.com/2v7ypet.png
http://www.superku.de/TED.jpg

Just modify points of cyrcle and set its y coord to Terrain::GetHeight() + 0.01

cadaver:
Thanks! I will be sure to check this out soon.

luveti:
@1vanK I'll look into showing the brush tonight!

luveti:
@1vanK a circle is now drawn where the mouse is currently focused.

71324ea96816804822fb246e210ee99b

cadaver:
Appeared to work nicely based on not-so-thorough testing. Will merge now, can be fixed later if there's need. Thanks for seeing this through!

PR #501, opened 2014-10-26
asPEEK AngelScript debugger

JSandusky:
asPEEK debugger implementation. To use the debugger the build setting URHO3D_ANGELSCRIPT_DEBUGGING must be enabled and you'll have to grab a copy of the client from: https://bitbucket.org/saejox/aspeek/
To debug you're going to want to be running Urho in windowed mode.

I'm not terribly happy with the handling of C++ types, so that's likely to change heavily.


weitjong:
Time flies. Feel just like yesterday. @JSandusky, are you still plan to resubmit the PR? If not then I will put a final nail to sweep it under the closed 1.5 milestone. Thanks.

JSandusky:
Sweep it under. If binding generation comes to fruition then it'll be worth re-evaluating. Binding the as types backwards to c was/is too painful for general use. If it ever pops up again (such as wrapping all as bindings to gen. the inverse) there'll be little to nothing related.

PR #524, opened 2014-11-13
SDK Cleanup

hickscorp:
Hi!

I have refactored the work i have done, this time following every guideline i have been told:
- Each commit includes changes to only one library.
- Every time a library is being refactored, the first commit is for moving the files, the second commit is for changing code.

I have checked that there is no regression in any commit, it appears that the code is built on top of the very latest Urho3D version, as opposed to last time (Alex pointed me to some code being reverted by my changes, probably because of my editor settings... Still a mistery for me.).

Only the kNet source code is changed. A few files from the SDL are also changed, but most of them were generated from the SDL ./configure step, which means that they should be automatically reflecting those changes the next time a new version of the SDL is merged to Urho3D.

An ultimate change i have done is to allow all headers from Bullet to be installed in the SDK (As opposed to the vanilla version of Urho3D, which only installs the LinearMath headers). This is to allow programmers to use the whole Bullet environment since the library is statically linked, all symbols are present in Urho3D lib. This could be easily reverted if needed.

The result (After running make package) is a self-contained SDK that only requires one include path (SDKPath/include/Urho3D) instead of many.

Please tell me if everything fits.


weitjong:
Close the issue now as there is no more comment from the original pull requester and that now we have completed the "SDK cleanup" using other implementation.

PR #560, opened 2014-12-10
[Editor] inspector lock

friesencr:
I added a little lock to the top right of the inspector.  If you hit the button it keeps the inspector from switching to new nodes.  I would of just pushed this but I am shocked by how little code I had to right which obviously means  I am missing something.  All of the ui elements in the inspector held their component/node ids so there wasn't that much to do.


friesencr:
This is ready to go.  I didn't get any feedback whether anybody liked this or not.

hdunderscore:
I like it, add it already :D

PR #603, opened 2015-01-11
Rename Readme.txt to Readme.md

CrackedP0t:
Change the file extension of Readme.txt to .md so that GitHub displays it in Markdown, which looks much better. Also changed the history format so that it displays as a list and changed "Readme.txt" to "Readme.md" in the Rakefile.


hdunderscore:
Simple enough change, the result can be previewed: https://github.com/CrackedP0t/Urho3D/tree/patch-1

It enabled markdown support on the github front page preview of Urho3D. I think there is a visual improvement, however in its current state the history section is mangled.

I looked at using the table format to fix it, although it's quite limited: https://github.com/hdunderscore/Urho3D/tree/master .

This PR should probably be resolved one way or the other-- request to rename one file shouldn't take 30 days ;)

weitjong:
I see two concerns with this PR.

1. It is incomplete. The file "Readme.txt" is being processed by Travis-CI to update the credits/contribution list from that file to About.md in our main website. To rename this file, the corresponding Rakefile in the project root directory need to be changed as well.
2. Although in master branch we do not have README.md, but we do have it in the other CI-mirror branches. If this PR get merged in then in future we will have both Readme.md and README.md in the CI-mirror branches. Not sure how it will play out with this scenario.

If these two concerns are being considered and verified that nothing bad would happen than you get one vote from me :).

cadaver:
The history information is also duplicated in the Doxygen documentation, so if we rename the Readme, it could be removed at the same time. Probably it's better for the github repo frontpage if the text is shorter rather than longer, and also removes duplicate effort when making new releases.

CrackedP0t:
I fixed the history and Rakefile. The CI-mirror branches _should_ be OK.

hdunderscore:
You'll probably need to fix that commit message for this PR to be accepted.

cadaver:
Yes, please avoid using "fun" commit messages (even if they may seem fun at the time of writing), as they'll be stored as part of history forever and may not seem as funny later. I know, I had to rewrite all comments from one of my C64 games' source code before publishing it!

CrackedP0t:
Message more descriptive.

weitjong:
Although we do not recommend users to checkout our CI mirror branches, some did it anyway. I wonder what would happen if both Readme.md and README.md exists on a user local file system that is not case-sensitive. I am sure on the server side GitHub can handle it.

I think it is ok to remove the existing README.md in the mirror branches so that we don't leave it to chances. I will see to that.

hdunderscore:
Just to clarify the earlier suggestion of changing the commit message, the following messages associated to the earlier two commits will need to change:
https://github.com/CrackedP0t/Urho3D/commit/6eba728bb63c8cb9cb3bdba044be6e339088730f :

    Used fun regexes

https://github.com/CrackedP0t/Urho3D/commit/0a6b5a4e4945b4bc79cf799d61de3f11841c58e5 :

    I just changed the text; hopefully my changes to the actual file won't screw it up

Otherwise it seems ok from what I see. Thanks !

weitjong:
I hope you don't mind that I have taken the liberty to squash your commit messages into one.

PR #641, opened 2015-01-28
Added 2 Urho2D samples (isometric demo and platformer)

Mike3D:


cadaver:
The build fails on Windows due to PlaySound redirecting to the Win32 API function PlaySoundA. Can you add something like this to Sample2D, it should fix it:

#ifdef PlaySound
#undef PlaySound
#endif

friesencr:
Intense!  Looking forward to looking at it more closely.  Thanks!

Mike3D:
@cadaver, thanks for the tip, I was a bit lost with this issue :cold_sweat:
@friesencr, hope it remains simple enough, I went a little farther than what I intended in the begining

bjorn:
Just a tip about the TMX files: you really should be using the base64 + zlib layer data format, at least when Urho3D supports it. I noticed the files are currently in "XML" layer data format, which is a really inefficient way to store the data and it's a format I'd eventually like to deprecate.

cirosantilli:
These two examples are a must for any 2D game engine.

weitjong:
We are short-handed. As much as I agree with you, we need help from @Mike3D or other contributors to bring this PR up to speed again before we can review them. The fact that the PR remain open until now indicates we actually want to see that happens. Time flies though :)

Mike3D:
I can push AS & lua samples to master (and C++ a little later) if it can help

eugeneko:
Closed via #2176

PR #647, opened 2015-01-31
1.32 emscripten

badpixels:


hdunderscore:
I've submitted a pull request onto this fork to bring it up to date to master.

I've successfully tested this on Windows:

- Using cmake gui, load the emscripten toolchain
- Enter this into the CMAKE_CXX_FLAG field: -s ALLOW_MEMORY_GROWTH=1 --preload-file F:/Urho3D/UrhoEmscripten/Urho3D/Build2/bin/Data.pak@/Data.pak --preload-file F:/Urho3D/UrhoEmscripten/Urho3D/Build2/bin/CoreData.pak@/CoreData.pak

(note you'll want to set the location of your own data pak's. Also note the @ symbol, it allows you to remap a relative or absolute path-- [read more here](http://kripken.github.io/emscripten-site/docs/porting/files/packaging_files.html) )
- Disable Network, Angelscript, Lua, SSE -- enable Samples, Navigation, File watching, Physics, Profiling, Urho2D.
- After you generate the build files, `make Urho3D -j4` in the build dir.
- Now we get a bit crude and navigate to `Build/Source/Samples/[sample]/CMakeFiles/[sample]/link.txt` and change output extension from .js to .html
- in build root run, `make [sample] -j4` .

weitjong:
@hdunderscore IMHO, your last commit (for fixing infinity loop bug) should be submitted directly to the Urho3D master branch, just in case this PR could not be merged in its entirety.

In my local working repo, I have used different rebasing strategy than yours. But it should produce similar result than yours PR to badpixels. What I did.
```
git checkout -b badpixels-1.32-emscripten 1.32
git pull https://github.com/badpixels/Urho3D.git 1.32-emscripten
git rebase --onto master 1.32 badpixels-1.32-emscripten
```
And the resolve all the conflicts from there.

On the Linux host, the newly improved enable_pch() macro does not play well with em++. It appears em++ is not a complete drop-in for g++ as clang++ does. So, more ```if emscripten``` is needed.

Personally, I also think the current Emscripten/CMake toolchain file (cmake/Modules/Platform/Emscripten.cmake) is incorrect. Although its comment states it should be used as CMake toolchain file, however, it is rather strange that the file exists in cmake/Modules/Platform as if "Emscripten" is a platform. To me, "Emscripten" is a compiler toolchain and not a platform. I am attempting to write another new toolchain file. Basically, the toolchain file should configure a cross-compiling build using emscripten tools (em++, emarr, etc) and targeting a Linux platform instead, just like Android/CMake toolchain does. Actually, as of CMake 3.1 Android is already being considered as valid platform, but prior to 3.1 the toolchain file still configures the build tree to target Linux platform.

Sorry for the long comment.

hdunderscore:
True, I did rebase initially but noticed a bad commit slipped in (https://github.com/badpixels/Urho3D/commit/8e93eb764027dd9151929476de86176586a07bb9) despite my effort to avoid it, so I tried a quick-fix.

I definitely defer to you on all matters build, looking forward to not needing any hacks and having a clean master branch we can all work from.

friesencr:
This is kind of mess because me, hd_, and dahlia are working on it.

Let me try to distill the chaos.  

@weitjong  You have permission from @badpixels and @hdunderscore  to go nuts with the cmake stuff.  

We found that emscripten, when pulling sdl, uses the more common path of <SDL2/*>. When you run emcc you can run with -v and it will tell you the include paths of its cache folders.  One of the other commits from hd_ is to fix that, but it does not alter the include header install to the new SDL2 folder.   The sdl cmake script needs to be updated to deploy the files to the different folder.

Lets move all of this work to a new branch on urhos repo.  I am working on this and should be done soon.





@weitjong  I think @hdunderscore would agg


I propose we move all of @badpixels work to a branch on urho's repo.

friesencr:
This is a hard rebase for me with all the moved files.  Don't wait for me if you have something.  Sorry.

weitjong:
I have completed the rebase. I have not merged my own local changes into this branch yet nor the SDL2 refactoring done by @hdunderscore. At last we have a common ground to proceed.

@hdunderscore Is the change of the library name from SDL to SDL2 really necessary?

hdunderscore:
The issue we discovered in upgrading to the new build system is that emscripten seems to expect references to SDL2 in that manner. Some brief mention of issue was raised in this emscripten issue: https://github.com/kripken/emscripten/issues/2404

It can be worked around, eg, the 1.32-emscripten(badpixels) was somewhere having the build reference Urho's own sdl headers, so building didn't throw an error.

friesencr:
@weitjong thanks for sharing your branch with us :heart: 

Closing this is favor of the new branch.  We can open a new PR on that branch for further dialog if needed.  A lot of people will be happy to have an up to date branch of this on iirc.

PR #680, opened 2015-03-20
Render refactor

### LINKED ISSUE #270: OpenGL 3.x, OpenGL ES 3.x, Directx 9.x / 10.x (SM3) as minimun ?
Because 3.x versions are now 5-6 years old and have better performance than 2.x and more options...
(Urho3D Website - Direct3D9 or OpenGL rendering (Shader Model 2, OpenGL 2.0 or OpenGL ES 2.0 required as minimum))

OpenGL 2.x (~2004 - 2008) Directx 9.x ( * )
OpenGL 3.x (~2008 - 2010) Directx 10.x ( ** )
OpenGL 4.x (~2010 - 2014) Directx 11.x ( *** )
OpenGL 5.x (~2014 - 2015) Directx 12.x

The version of Directx 9.x is ok, because 10.x have bad performance in some cases.
*OpenGL 3.2, Directx 9.0c

cadaver:
- Adds D3D11 support (-DURHO_D3D11=1)
- Adds OpenGL3 support. This is dynamically detected at runtime, falls back to GL2 if not available
- Various rendering performance optimizations
- The half-pixel UV offset required by D3D9 only can be queried from Graphics, instead of using ifdef.
- OpenGL matrix multiplications are reversed to allow uploading uniforms without transpose
- On desktop APIs 8 continuous texture units are available for materials
- D3D9 SM 2.0 support is removed due to testing difficulty

Testing is welcome. When testing on D3D, you may need to delete all existing cached shaders from bin/CoreData/Shaders/HLSL/Cache, to ensure that shaders get recompiled.

Closes #270.


friesencr:
Thanks Lasse.  Just in time for some weekend fun :metal:

Mike3D:
Awesome work! :+1::+1::+1: What a pity I can't test.

Hevedy:
Awesome, will be here OpenGL 4.0 to equal the DX11 support ?

hdunderscore:
Seems to work in my tests -- although something that you may have mentioned somewhere in the docs is that you need to bring your own d3dcompiler_47.dll on Windows 7 for dx11 (eg, from Windows 8.1 SDK).

Great work !

aster2013:
thanks.

boberfly:
Love your work Lasse! Are UBOs used instead of uniforms in GL3.x? We could bump up the bone limit count on those platforms perhaps... :)

MonkeyFirst:
>We could bump up the bone limit count on those platforms perhaps
Yes, the bone limits is very low. Then i'm create rig for model i try to create do not over 64 bones.
I'm delete some additional bones and split few into one on middle and it's some kind limitation not friendly for artist who want add some additional features for his character. I know what we may add few skeleton under master in the node, but it's makes some additional work in 3d editor to splitting some character parts into separated.
I guess the limit around 128+ bones will be fine for desktop.

OldSnake22:
Thank a lot for this ! I am agree that 64 bones limit is a bit low

cadaver:
There is code for UBO support, but in my tests I've found them slower than uniforms, at least on a NVidia + Windows setup. It's a define in the Uniforms.glsl file whether you want to enable them or not. The bone count could be bumped up. But it has to be made a config option in the importers, so that they can still produce output that's compatible with eg. mobiles or the RPI.

Hevedy:
Its possible change/fix the read of d3dcompiler_47.dll from the .exe folder, to read directly from windows folder ?

cadaver:
It's not installed in any Windows OS folder, but only in the Windows SDK, where Windows doesn't know how to look for it automatically. Microsoft's recommendation is that applications ship with it. What could be made would be to make CMake copy it automatically, but of course that works only if it's installed in the default location.

Hevedy:
I got that file installed in C:/Windows/winsxs

MonkeyFirst:
i'm also add MaxBones property for blender-plugin. And already tested it with new fixes of this render-refactor brunch
https://github.com/reattiva/Urho3D-Blender/network

I guess that have an issue - not all bones are viewed(with white lines) in DebugRenderer of Urho3D Editor. anyway my BigBot model with 78 bones exported fine and play his animations also fine.

boberfly:
Making an OpenGL ES 3.x context might be cool also, lots of devices are coming out with support. Can that be dynamic on mobile if ES2.0 is the fallback? I'm guessing not.

cadaver:
If you have to link a different library (at least on Android) then it's hard to make dynamic. However it seems a lot of the 3.0 stuff could also be used as extensions.

cadaver:
From my perspective this is ready to go into master; I'm not planning further work at this point and an Intel driver update solved the OpenGL issues on Intel GPU + Windows.

boberfly:
Awesome! I didn't get a chance to test but I'm not running anything exotic to test it on...

weitjong:
If you like, we could run a full CI build tests on the branch before the merge.

cadaver:
Yes, that would be a good idea.

weitjong:
I am not in front of my workstation at the moment. You should be able to trigger the full CI build tests on the branch by submitting a commit to the branch with commit message containing these keywords yourself. "Your commit message. [ci rebase]"

The mechanism is already there for quite some time. It is the commit message that does the triggering. No additional setup required.

friesencr:
Congratz Lasse!!

ucupumar:
A bit late to the party, but well done Cadaver!! Thanks a lot!!

AGreatFish:
Thumbs up!

JoshEngebretson:
Fantastic work @cadaver!  I've been updating Atomic to Urho 1.4 (prerelease) and the rendering cleanups are really great to work with :+1:

kingofoz:
Can somebody tell me if OpenGL ES 3.0 is supported now?

PR #688, opened 2015-04-05
Mainly added Node transformative function overloads

Modanung:
Mainly added Node transformative function overloads in the header file.
Also removed a pair of round brackets that seemed obsolete in Object.h.


Modanung:
Sorry about that...

PR #692, opened 2015-04-06
Slight usability improvement of cmake shell scripts.

raould:
When I tried to run the script, I found it confusing & the online docs didn't really clarify it well. I had to look into the scripts themselves to see what was expected. So I'm trying to improve that a little bit.

### DIFF
--- a/cmake_clean.sh
+++ b/cmake_clean.sh
@@ -22,7 +22,8 @@
 #
 
 # Clean the CMake cache and CMake generated files in the build tree
-if [ "$1" ] && [ -f "$1"/CMakeCache.txt ]; then BUILD=$1; elif [ -f $(pwd)/CMakeCache.txt ]; then BUILD=$(pwd); else echo An error has occured, build tree has to be provided as the first argument OR call this script in a build tree itself; exit 1; fi
+source cmake_shutil.sh
+get_build_dir
 rm -rf "$BUILD"/{CMakeCache.txt,CMakeFiles}
 touch "$BUILD"/CMakeCache.txt
 
--- a/cmake_generic.sh
+++ b/cmake_generic.sh
@@ -22,7 +22,8 @@
 #
 
 # Determine source tree and build tree
-if [ "$1" ] && [[ ! "$1" =~ ^- ]]; then BUILD=$1; shift; elif [ -f $(pwd)/CMakeCache.txt ]; then BUILD=$(pwd); else echo An error has occured, build tree has to be provided as the first argument OR call this script in a build tree itself; exit 1; fi
+source cmake_shutil.sh
+get_build_dir
 SOURCE=$(dirname $0)
 if [ "$SOURCE" == "." ]; then SOURCE=$(pwd); fi
 if [ "$BUILD" == "." ]; then BUILD=$(pwd); fi
--- a/cmake_shutil.sh
+++ b/cmake_shutil.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+function get_build_dir() {
+	if [ "$1" ] && [ -f "$1"/CMakeCache.txt ]
+		then BUILD=$1
+	elif [ -f $(pwd)/CMakeCache.txt ]
+		then BUILD=$(pwd)
+	elif [ -e $(pwd)/.git ]
+		then BUILD=$(pwd)
+	else echo "Current directory doesn't appear to be a 'build tree'. Please provide a build tree directory as the first argument, OR call this script in a build tree itself."
+		exit 1
+	fi
+}


weitjong:
Thanks again for the PR. I think you already know from the Travis CI build results that your patch does not work as expected. Your fix does not work when configuring and generating the build tree for the very first time where CMakeCache.txt still does not exist. We also do not recommend user to generate the build tree in our project root where .git resides. In fact we try not to provide any default in our scripts nowadays. User has to pass the build tree path or the script should strictly just fail.

raould:
Thanks for taking a look! Overall I suggest to clarify what a "build tree" is for Urho. I just do not, as a new user, have any idea where I'm supposed to run the build scripts. The documentation says "root" and nothing about "build trees" so overall it is very confused/confusing/unclear.

I think: Ideally the user should not have to know much at all; Ideally I think they should be able to be in the same directory as the shell scripts, and run them with ./foo.sh and they should magically do the right thing. What would be the right way to get that sort of usability?

Thanks :-)

friesencr:
I do wonder how many people are bailing out because there is not 'double click' option.

raould:
p.s.

> In fact we try not to provide any default in our scripts nowadays. User has to pass the build tree path or the script should strictly just fail.

the scripts do already still currently have code in them to try to default / see if the current dir is acceptable, so they don't strictly just fail w/out argument, i think.

weitjong:
> the scripts do already still currently have code in them to try to default / see if the current dir is acceptable, so they don't strictly just fail w/out argument, i think.

No, that is different. The script needs just one mandatory thing from user and that is the build-tree path. If the build tree is not yet generated then user must provide it as the first argument or the script fails. However, if the build-tree is already generated and if happen that user already cd to this build-tree directory then the script is smart enough to detect that and use the current directory instead.

About the usability point you mentioned in the earlier comment - you may want to know that our script used to be like that before our recent refactoring. Then our scripts ```decides``` the build-tree path on user behalf, so they just worked when invoked without any arguments. But guess what, we had another camp of users who complain about this, saying that our out-of-source build-tree were not the real deal or something like that (I forget the exact word). But anyway, I do agree on one thing. The wording of the error message when the script fails out could have been clear. The online documentation could have been clearer too. I suppose you can guess it is not written by a native English speaker :).

raould:
(dunno if this will get broken by markdown interpretation.)

cmake_generic.sh explicitly checks if CMakeCache.txt is there as a file and if so automagically sets BUILD to pwd, no?

if [ "$1" ] && [[ ! "$1" =~ ^- ]]; then BUILD=$1; shift; elif [ -f $(pwd)/CMakeCache.txt ]; then BUILD=$(pwd); else echo An error has occured, build tree has to be provided as the first argument OR call this script in a build tree itself; exit 1; fi

raould:
yeah, it might be a language thing either way. :-) To me, " User has to pass the build tree path or the script should strictly just fail" really does conflict with "the script is smart enough to detect that and use the current directory instead." Figuring out how to make it all more clear would be great - some day.

raould:
Re: complaints leading to refactoring. Yeah, to me it all seems like CMake has certain culture and best practices, and perhaps how things worked before in Uhro it went against those ideals. Ok. But then I feel like it has gone too far towards "you have to already know *all about CMake culture* to be able to understand this documentation" which I think is not ideal. Yes ok for power users. But if possible, better to make the scripts "just work" but also the "right way" as CMake likes it.  Because the current stuff feels like there's no way I can figure out what is Right w/out doing a lot of CMake learning up front rather than being able to just start working on my game ;-)

weitjong:
You are welcome to reword this error message "An error has occured, build tree has to be provided as the first argument OR call this script in a build tree itself".

raould:
Other projects I've seen have documentation that says "1) mkdir build; 2) cd build; 3) cmake ..; 4) make" which I didn't see in Urho docs. Can I add that to Urho docs? (w/out step 4 since there are different step 4s depending on ide.)

cadaver:
Ah yes, the simplest possible example for a default build into the checkout tree, like 1) ./cmake_generic.sh . 2) make wouldn't hurt.

raould:
I am apparently only half-way to being the right person to do this. I am still confused about how CMake + Urhu work. I *think* I just did a build where I did (1) mkdir build (2) cd build (3) ../cmake_generic.sh (4) make and it seemed mostly ok...

...but the build/bin/ directory that was generated and populated in there was not complete e.g. the Data directory that exists in just bin/Data wasn't copied? So I'm very confused. Is that expected? is that not? Builds take so long (on my machine, anyway!) that it is hard / a little depressing to try to experiment with such things.

If somebody could clearly state what a simple, canonical set of steps would be I'm happy to then test it and document it. (I will try a clean build again and see what ends up in build/bin.)

friesencr:
I like the idea of the default build path.  On windows double clicking on the bats will result in in error that will immediately close.  Folders would get clobbered if the build folder is set to .

PR #702, opened 2015-04-15
Add unsigned long and unsigned long long support to String += operator

marynate:


weitjong:
Might as well add the ```unsigned long long``` version too while you at it. :smile:

cadaver:
Unsigned long is a bit nasty datatype, as its size can vary depending on the bitness and compiler. I don't oppose adding it in this case though. On the platforms Urho supports, unsigned & unsigned long long always have 32 & 64 bit fixed size respectively and will probably be most useful.

marynate:
Well, then unsigned long long it is. :)

PR #705, opened 2015-04-16
Fix PCH build with Clang

evolarium:
Clang doesn't use include path (-I) when looking for PCH files.  Use full path to header instead.  Tested on Ninja and Make builds using GCC and Clang.


weitjong:
Thanks for your PR. We have to modify your change to use back the relative path for Xcode build to keep it happy.

Issue #708, opened 2015-04-21
MinGW/Clang crash and premature deallocation using AngelScript in C++11 mode


Skrylar:
Environment:
MSYS
GCC 4.9.2
Clang 3.5.1
Windows 7

Situation:
Compiling under clang with C++11 enabled results in the AngelScript bindings being unusable.  After tracing around the debug and release builds under the same environment, it seems to be related with how strings are getting registered/allocated with AS.  Removing code that creates strings limps along a little further, so the errors seem to be focused around how strings/objects are being allocated.  At one point I also saw an error about an assertion volation where the refcount was off by one as well.

I'm not entirely sure what is causing this; it seems that calling functions works, as tracing through shows some methods do in fact work when the C++ and AS boundary is crossed.  It seems to be something related to constructors and reference counting, as though objects are being prematurely freed on their way out to AS  I did a third debug build under the native GCC and that particular build works, so it does appear to be compiler specific. Now the question is finding out exactly which part has to be dug around to fix _that_, since the rest of the codebase (including the samples written in C++) appears to work just fine in this setup.

Skrylar:
I just tested the latest AngelScript in a test project with the same compiler, and it seems string support works fine there when using the built-in std::string library.  They seem to mention that some code gets inlined by clang which is unhelpful, so there's some workarounds in the angel code. I'll have to check and see if this is the same problem affecting current versions of Urho as well.

weitjong:
I am not sure it's related to your issue or not. In my own Urho3D fork I have also attempted to use modern CPP standard and in doing so, one of the compiler (Clang if I recall it correctly) gave a number of compilation errors (or wanings, not very sure about that) in the AngelScript API binding code for assignment operator for a few classes. I am not in front of my workstation now so I could not double check that. I resolved them by using asMETHODPR macro instead of asMETHOD macro used by upstream Urho main repo. I also encountered similar compilation errors when I experimenting with Emscripten embind which uses "std=c++11" on a freshly cloned Urho3D repo.

So, a question for you. Did you encounter those compilation errors too when you use the new cpp standard? If yes, how did you resolve it?

Skrylar:
[...]

PR #711, opened 2015-04-28
Support MinGW 4.7 ABIs on Clang.

Skrylar:
Fix for  #708.  Adds a secondary check for recent Clang versions in addition to mingw, so AngelScript won't cause explosions.


cadaver:
Thanks! Will test on MinGW (ordinary MinGW without Clang, just to make sure nothing breaks) shortly, then merge. 

For future reference note the convention how we annotate modified third party sources: put "// Modified by xxx for Urho3D" near the top of the header, then "// Urho3D: xxx xxx" at the modification to describe what it does. This started to satisfy the zlib license requirement to mark any modifications, but is also helpful whenever merging a new upstream version.

PR #715, opened 2015-05-03
Implementation of DetourCrowd and DetourTileCache

JSandusky:
Added Components:
- DynamicNavigationMesh (uses DetourTileCache)
- Obstacle (interacts with DynamicNavigationMesh)
- DetourCrowdManager (uses DetourCrowd for crowd control)
- CrowdAgent (interacts with DetourCrowdManager)
- NavArea (marks area types)

Changes:
- Refactor NavigationMesh build process
- Addition of cylinder debug rendering
- Exposed Area Type Costs in navigation

CrowdNavigation sample for Lua, Angelscript, and C++ using
DynamicNavigationMesh, Obstacle, CrowdAgent, and DetourCrowdManager
components.


friesencr:
Congratz!

JSandusky:
Travis was really freaking me out there for a second. Until it just hit me that the exact case of the path doesn't match the CMakeLists.txt (Windows doesn't care) - just have to fix that without creating a mess of directory removed / directory added.

Mike3D:
Thanks :thumbsup:

cadaver:
Very solid work, thanks a lot! This should be merged soon.

cadaver:
Merged. I wrote a very brief documentation section on the crowd navigation (just lists the new components and refers to the example), feel free to make a new PR to amend it.

friesencr:
Wow!  That was very solid work!

Mike3D:
For NavArea, what do you think of renaming Get/SetBounds() and GetTransformedBounds() to Get/SetBoundingBox() and GetWorldBoundingBox() for consistency across classes?

cadaver:
Good idea, I changed those function names. Added also NavArea Lua binding.

Mike3D:
Thanks.
For same pattern purpose, we could also replace GetAgentCount() in DetourCrowdManager by GetNumAgents().

cadaver:
Yes, if you want to change that, please go ahead. This is something I don't feel that strongly about, as we also have e.g. GetVertexCount() in VertexBuffer.

weitjong:
I am wondering whether the CrowdAgent::updateNodePosition_ is really required? The base Component class already has Component::enabled_ flag. Is there a real use case of having a CrowdAgent component enabled but without updating the node position? In such case, the agent would participate in the crowd movement and even may send event that it has reached the target, without actually moving an inch! Or do I get it all wrong?

cadaver:
There could be a scenario where you don't want the crowd agent to manage the position for you automatically, but you for example want to use physics forces / velocity to try to reach the position it's giving you (it sends an event of each new position)

weitjong:
Ah, ok. It makes sense now. Thanks.

PR #720, opened 2015-05-07
Better CrowdAgent state handling. Consistent use of AreaID instead of…

JSandusky:
… AreaType+AreaID.

CrowdAgent now tracks its previous state and sends the
CrowdAgentStateChanged event when appropriate. CrowdAgentStateChanged
event now includes both the target and agent states.

Additionally:
- GetAreaCost in NavigationMesh, script bindings updated
- Fixed error spawning Jack nodes on top of boxes in Angelscript sample

### DIFF
--- a/Source/Urho3D/Navigation/CrowdAgent.h
+++ b/Source/Urho3D/Navigation/CrowdAgent.h
@@ -163,6 +163,10 @@ private:
     NavigationQuality navQuality_;
     /// Agent's Navigation Pushiness.
     NavigationPushiness navPushiness_;
+    /// Agent's previous target state used to check for state changes.
+    CrowdTargetState previousTargetState_;
+    /// Agent's previous agent state used to check for state changes.
+    CrowdAgentState previousAgentState_;
     /// Internal flag to ignore transform changes because it came from us, used in OnCrowdAgentReposition().
     bool ignoreTransformChanges_;
 };
--- a/Source/Urho3D/Navigation/NavArea.cpp
+++ b/Source/Urho3D/Navigation/NavArea.cpp
@@ -20,23 +20,27 @@
 // THE SOFTWARE.
 //
 
+#include "../Scene/Component.h"
 #include "../Core/Context.h"
 #include "../Graphics/DebugRenderer.h"
-#include "../Scene/Component.h"
-#include "../Scene/Node.h"
+#include "../IO/Log.h"
 #include "../Navigation/NavArea.h"
+#include "../Scene/Node.h"
+#include "../Container/Str.h"
 
 namespace Urho3D
 {
+    static const unsigned MAX_NAV_AREA_ID = 255;
     static const Vector3 DEFAULT_BOUNDING_BOX_MIN(-10.0f, -10.0f, -10.0f);
     static const Vector3 DEFAULT_BOUNDING_BOX_MAX(10.0f, 10.0f, 10.0f);
-    static const unsigned DEFAULT_AREA = 0;
+    static const unsigned DEFAULT_MASK_FLAG = 0;
+    static const unsigned DEFAULT_AREA_ID = 0;
 
     extern const char* NAVIGATION_CATEGORY;
 
     NavArea::NavArea(Context* context) :
         Component(context),
-        areaType_(0),
+        areaID_(DEFAULT_AREA_ID),
         boundingBox_(DEFAULT_BOUNDING_BOX_MIN, DEFAULT_BOUNDING_BOX_MAX)
     {
     }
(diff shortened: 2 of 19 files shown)


Source/Urho3D/Navigation/CrowdAgent.h
```diff
@@ -163,6 +163,10 @@ class URHO3D_API CrowdAgent : public Component
     NavigationQuality navQuality_;
     /// Agent's Navigation Pushiness.
     NavigationPushiness navPushiness_;
+    /// Agent's previous target state used to check for state changes
```
weitjong: Not really important but please add period for all the Doxygen comment lines in the future for consistency sake.

Source/Urho3D/Navigation/NavArea.cpp
```diff
@@ -53,12 +53,12 @@ namespace Urho3D
         COPY_BASE_ATTRIBUTES(Component);
         ATTRIBUTE("Bounding Box Min", Vector3, boundingBox_.min_, DEFAULT_BOUNDING_BOX_MIN, AM_DEFAULT);
         ATTRIBUTE("Bounding Box Max", Vector3, boundingBox_.max_, DEFAULT_BOUNDING_BOX_MAX, AM_DEFAULT);
-        ACCESSOR_ATTRIBUTE("Area Type", GetAreaType, SetAreaType, unsigned, DEFAULT_AREA, AM_DEFAULT);
+        ACCESSOR_ATTRIBUTE("Area ID", GetAreaID, SetAreaID, unsigned, DEFAULT_AREA_ID, AM_DEFAULT);
     }
 
-    void NavArea::SetAreaType(unsigned newType)
+    void NavArea::SetAreaID(unsigned newType)
```
weitjong: newID, perhaps?

Source/Urho3D/Navigation/NavArea.h
```diff
@@ -43,9 +43,9 @@ namespace Urho3D
         virtual void DrawDebugGeometry(DebugRenderer* debug, bool depthTest);
 
         /// Get the area id for this volume.
-        unsigned GetAreaType() const { return areaType_; }
+        unsigned GetAreaID() const { return areaID_; }
         /// Set the area id for this volume.
-        void SetAreaType(unsigned);
+        void SetAreaID(unsigned);
```
weitjong: Add the parameter name here.

Source/Urho3D/Navigation/NavigationMesh.cpp
```diff
@@ -890,7 +897,7 @@ void NavigationMesh::GetTileGeometry(NavBuildData* build, Vector<NavigationGeome
                 NavArea* area = static_cast<NavArea*>(geometryList[i].component_);
                 NavAreaStub stub;
                 //\todo is there an alternative to casting down? Attributes restricts area ID/type to being an unsigned int
```
weitjong: IMHO, you can remove this TODO comment or at least remove the reference to area "type" in the comment.

As I understand it, in theory there is nothing prevent us from having an ```unsigned char``` type attribute. Currently it does not compile if you try to do so because the attribute macros cannot find templates specialized for ```unsigned char``` and also currently Variant class does not have overload for this data type. But all these can be easily added if we really want to. In practice, however, I don't think it would make much performance difference with what you have done now.

If it really bothering you then perhaps you should change the data type for ```NavArea::areaID_``` private member variable to ```unsigned char``` and make the necessary adjustment to its getter/setter. So, the casting only happens once in the setter. In other words, the "property" is ```unsigned char``` although the "attribute" is ```unsigned```.

JSandusky:
This should take care of your noted conflicts, be more consistent (indentation of DynamicNavigationMesh, enums bound to script, periods, etc), and cover the "half way up the block" error while demonstrating handling the CrowdAgentFailure event that is fired on a failure either internally of the agent or of it's target point.

JSandusky:
No problem. Let me know if/when you want me to squash commits down.

weitjong:
@cadaver Aside from the case where an event is sent due to component attaching to or detaching from a node, is it safe to assume that we can call Component::GetNode() in the event handler to obtain the node pointer? I see that the physics events and now also navigation events include both node pointer and component pointer in the event parameter.

@JSandusky IMHO, no need to squash. I don't think Lasse concerns with that commit counts. You see I have littered his project commit history with those "For Travis CI stuffs" and I have not received a word from him.

cadaver:
GetNode() itself should be safe in all sane situations, including event sending / event handlers. Only problem could come from keeping its return value in a raw pointer and accessing it some unspecified time later (when the node might be already deleted), but that's a user error.

The only times I've wanted commits to be squashed is when the commit messages look unprofessional or experimental in a haphazard & unsure way. That isn't the case here.

weitjong:
I am thinking whether it makes sense to remove the P_NODE as parameter to save some bytes.

cadaver:
Though it's safe, I've used the principle to rather include "too much" information in events so that the user doesn't need to GetNode() himself. It could be removed, but then I think the corresponding physics event parameters should be removed too for consistency and that could break applications.  

As the event data maps are reused the map element allocator has likely already been "primed" at that point and no new allocation happens because of the node pointer.

weitjong:
Thanks for answer.

JSandusky:
Sorry about that, things got really nasty when I fixed that conflict.

PR #724, opened 2015-05-15
Updates to Navigation documentation

JSandusky:
Updates to navigation documentation and request for comments on what everyone feels is important to cover versus what's overkill.

### DIFF
--- a/Docs/Reference.dox
+++ b/Docs/Reference.dox
@@ -1699,9 +1699,33 @@ To query for a path between start and end points on the navigation mesh, call \r
 
 For a demonstration of the navigation capabilities, check the related sample application (15_Navigation), which features partial navigation mesh rebuilds (objects can be created and deleted) and querying paths.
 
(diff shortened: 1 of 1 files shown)


weitjong:
This PR has just missed the 1.4 tag.

Mike3D:
Thanks. I've given a look at Unity's documention and now things are way clearer relative to areas, costs, quality...

For now I am porting an existing project and it works great. With dynamic navmesh, rebuilds are not perceptible on mobiles and way easier to handle and building initial navmesh seems to never fail. Crowd agents work fine and make life really easy. I'm only experiencing segfaults when removing nodes containing a CrowdAgent component. I'll open an issue after investigation.
Finally I will play with off-mesh connections.
I will then document all these experiments and keep track of unanswered questions and potential improvements.

PS: as a feature request, it would be handy to have a boolean in CrowdAgent to instruct the agent to face its velocity.

PR #734, opened 2015-05-31
AssetImporter was not importing single keyframe animation correctly

ninjastone:


cadaver:
Looks like this could index zero size arrays without first checking the size. firstScale & firstRot seem unused otherwise so it looks like this is some test code that was left there?

ninjastone:
Yeah, looks like I forgot to remove test code. They are gone now.

PR #744, opened 2015-06-13
Fixed usage of deprecated GLSL function texture2DLod.

GKR:
It is now textureLod and textureLodOffset. Error message was: (0(857) : error C7616: global function texture2DLod is removed after version 140)


cadaver:
Thanks for the PR and bringing this to attention. This can be solved in another manner by adding these as definitions in Samplers.glsl. GL2 still needs the deprecated functions, I believe.

GKR:
Thanks, that sounds like a better solution. Glad it got fixed, regardless :-)

PR #745, opened 2015-06-14
Add `luaL_loadfile` support, with CMake option to prefer this strategy

feltech:
* Added `LuaScript::LoadRawFile` and `LuaScript::ExecuteRawFile` to load
files using `luaL_loadfile` rather than via the resource cache
(`luaL_loadbuffer`).  This strategy makes the source available to Lua
debuggers (e.g. eclipse LDT) for setting breakpoints, etc.
* Added a CMake option `URHO3D_LUA_RAW_SCRIPT_LOADER` to make
`luaL_loadfile` the default script loading strategy.
  - `LuaScript::ExecuteFile` will thus prefer the `luaL_loadfile`
strategy, aiding in debugging as above, but without needing to change
method calls.
  -  `LuaScript::Loader` will also prefer the `luaL_loadfile` strategy,
so that scripts `require`d in the main script are also amenable to
debugging.
  - If a script file cannot be loaded from the file system, then it will
fall back on the default resource cache `luaL_loadbuffer` strategy (e.g.
for packaged scripts).

### DIFF
--- a/CMake/Modules/Urho3D-CMake-common.cmake
+++ b/CMake/Modules/Urho3D-CMake-common.cmake
@@ -93,6 +93,20 @@ cmake_dependent_option (URHO3D_SSE "Enable SSE instruction set" ${URHO3D_DEFAULT
 if (CMAKE_PROJECT_NAME STREQUAL Urho3D)
     cmake_dependent_option (URHO3D_LUAJIT_AMALG "Enable LuaJIT amalgamated build (LuaJIT only)" FALSE "URHO3D_LUAJIT" FALSE)
     cmake_dependent_option (URHO3D_SAFE_LUA "Enable Lua C++ wrapper safety checks (Lua/LuaJIT only)" FALSE "URHO3D_LUA OR URHO3D_LUAJIT" FALSE)
+
+    set(LUA_RAW_DESC "Prefer loading raw script files from the file system before falling back on Urho3D resource cache. Useful for debugging (e.g. breakpoints), but less performant (Lua/LuaJIT only)")
+    if (CMAKE_BUILD_TYPE STREQUAL "Release")
+        cmake_dependent_option (
+            URHO3D_LUA_RAW_SCRIPT_LOADER
+            ${LUA_RAW_DESC} OFF "URHO3D_LUA OR URHO3D_LUAJIT" OFF
+        )    
+    else ()
+        cmake_dependent_option (
+            URHO3D_LUA_RAW_SCRIPT_LOADER
+            ${LUA_RAW_DESC} ON "URHO3D_LUA OR URHO3D_LUAJIT" OFF
+        )     
+    endif ()
+    
     option (URHO3D_SAMPLES "Build sample applications")
     cmake_dependent_option (URHO3D_TOOLS "Build tools (native and RPI only)" TRUE "NOT IOS AND NOT ANDROID AND NOT EMSCRIPTEN" FALSE)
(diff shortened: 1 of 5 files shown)


feltech:
For a little background, I wrote about this in http://urho3d.prophpbb.com/post6725.html .

Source/Urho3D/LuaScript/LuaScript.h
```diff
@@ -71,6 +71,12 @@ class URHO3D_API LuaScript : public Object, public LuaScriptEventListener
     bool ExecuteFile(const String& fileName);
     /// Execute script string. Return true if successful.
     bool ExecuteString(const String& string);
+    /// Load script file on file system (i.e. not from resource cache).
```
weitjong: Based on our coding convention, the comment on headers should be a one-liner to keep Doxygen happy. If the comment is really long (which is not in this particular case) then you could use the ```/// \brief``` at the beginning of the comment.

CMake/Modules/Urho3D-CMake-common.cmake
```diff
@@ -93,6 +93,20 @@ cmake_dependent_option (URHO3D_SSE "Enable SSE instruction set" ${URHO3D_DEFAULT
 if (CMAKE_PROJECT_NAME STREQUAL Urho3D)
     cmake_dependent_option (URHO3D_LUAJIT_AMALG "Enable LuaJIT amalgamated build (LuaJIT only)" FALSE "URHO3D_LUAJIT" FALSE)
     cmake_dependent_option (URHO3D_SAFE_LUA "Enable Lua C++ wrapper safety checks (Lua/LuaJIT only)" FALSE "URHO3D_LUA OR URHO3D_LUAJIT" FALSE)
+
+    set(LUA_RAW_DESC "Prefer loading raw script files from the file system before falling back on Urho3D resource cache. Useful for debugging (e.g. breakpoints), but less performant (Lua/LuaJIT only)")
```
weitjong: I think all these code can be made much simpler. I would just write it as:
```
if (CMAKE_BUILD_TYPE STREQUAL Release OR CMAKE_CONFIGURATION_TYPES)
    set (URHO3D_DEFAULT_LUA_RAW FALSE)
else ()
    set (URHO3D_DEFAULT_LUA_RAW TRUE)
endif ()
cmake_dependent_option (URHO3D_LUA_RAW_SCRIPT_LOADER "Prefer loading raw script files from the file system before falling back on Urho3D resource cache. Useful for debugging (e.g. breakpoints), but less performant (Lua/LuaJIT only)" ${URHO3D_DEFAULT_LUA_RAW} "URHO3D_LUA OR URHO3D_LUAJIT" FALSE)
```

Things to note:

- In CMake, we have an unwritten convention: use double quotes for string literal only when it is absolutely necessary. Don't quote "Release".
- CMAKE_BUILD_TYPE variable is only used by single-configuration generator, such as Eclipse CDT4. So, I have set the default to false also for the multi-configurations generator.

cadaver:
Thanks for the PR. I'd have two further suggestions:
- List the added CMake define in the table in Docs/GettingStarted.dox
- Add a brief paragraph on the feature in the LuaScripting page in Docs/Reference.dox
Otherwise the feature could become forgotten or unclear for those who haven't seen the forum thread.

feltech:
In the above commit I have implemented the review comments, and of course checked CMake still works and doxygen looks good.  Just let me know if there is anything more to add or change.

PR #746, opened 2015-06-15
Allow ray queries to return UV texture coordinates of ray intersection

nemerle:
This change adds a new RayQueryLevel - ```RAY_TRIANGLE_UV```
This will ask the query to return texture UV coordinates at Ray-Geometry intersection point.

Since this change is mostly geared towards integrating external UI toolkits, I've only implemented this functionality for StaticModel. _(And frankly doing it for AnimatedModels would be pretty resource intensive)_

This feature will allow us to render UI into textures, place them on the game world objects and properly translate and forward mouse/touch event coordinates to the external UI library.

I've also added a small change to Decals sample to exemplify the usage of this feature.

### DIFF
--- a/Source/Samples/08_Decals/Decals.cpp
+++ b/Source/Samples/08_Decals/Decals.cpp
@@ -24,7 +24,6 @@
 
 #include <Urho3D/Graphics/Camera.h>
 #include <Urho3D/Core/CoreEvents.h>
-#include <Urho3D/IO/Log.h>
 #include <Urho3D/UI/Cursor.h>
 #include <Urho3D/Graphics/DebugRenderer.h>
 #include <Urho3D/Graphics/DecalSet.h>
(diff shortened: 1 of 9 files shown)


cadaver:
Thanks for PR, this should not be a problem to merge. However, please check the coding convention, eg. use if ( instead of if( and do not put the starting brace on the same row with the if statement, and put the constant last in if statements. Also it would be good to add the new raycast level to script bindings.

nemerle:
on it :)

cadaver:
Also one more thing, unless you are going to edit the AngelScript & Lua decal examples to match, please remove the test / debugging code from the C++ decals example.

cadaver:
Thanks. Will probably rename texture_uv to textureUV as we've not used underscores in the middle of variable names.

nemerle:
Renamed :)

PR #760, opened 2015-07-02
OnSceneSet()

### LINKED ISSUE #748: Removing child node from scene does not remove drawables, if node is still alive post removal
It seems a node's drawables are being left in the scene, if the node survives being removed from a scene (ie. is referenced somewhere else)

In 04_StaticScene, add this after line 114: https://github.com/urho3d/Urho3D/blob/master/Source/Samples/04_StaticScene/StaticScene.cpp#L114

        // keep node alive, after removing node from scene
        mushroomNode->AddRef();

        // Removing components does get rid of drawables in scene, once node is removed
        //mushroomNode->RemoveAllComponents();

        // if node is still alive, drawables are still in scene
        scene_->RemoveChild(mushroomNode);

cadaver:
This is a refactoring which creates a new virtual function to Component: OnSceneSet(). This allows components such as drawables to be created outside a scene, to be moved into a scene later, or to be moved into another scene. It should also fix  #748.

Testing would be appreciated. Some things will not work 100%, for example if a RigidBody exists outside the scene and you set its linear velocity, it has nowhere to store it as the Bullet object does not exist yet.

### DIFF
--- a/Docs/Reference.dox
+++ b/Docs/Reference.dox
@@ -295,6 +295,8 @@ Whenever there is some hierarchical composition, it is recommended (and in fact 
 
 It is also legal to create a Node that does not belong to a scene. This is useful for example with a camera moving in a scene that may be loaded or saved, because then the camera will not be saved along with the actual scene, and will not be destroyed when the scene is loaded.
 
+However, depending on the components used, creating components to a node outside the scene, then moving the node to a scene later may not work completely as expected. For example, a RigidBody component can not store its velocities if it does not have access to the scene's physics world component to actually create the Bullet rigid body object.
+
 \section SceneModel_Update Scene updates
 
 A Scene whose updates are enabled (default) will be automatically updated on each main loop iteration. See \ref Scene::SetUpdateEnabled "SetUpdateEnabled()".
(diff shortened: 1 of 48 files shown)


friesencr:
After one pulls the node out of the scene is there any way of knowing it exists other than holding a pointer to it?

cadaver:
At that point it's like any C++ object, and the engine doesn't care of its existence (except if it's still a subscriber to events, but that mechanism uses WeakPtr's and doesn't keep it alive), so yes, you'd have to hold a SharedPtr or AngelScript handle to it to keep it alive, or use a WeakPtr to test when it has been destroyed.

friesencr:
Does this change nested Nodes at all?  Setting a node seems different then setting a scene.  what if something moves from on node to the next?  What if a parent node is removed from the scene?  I am still combing the source code but these are some of the things I am thinking to try to break it.

cadaver:
When a node is pulled into or out of a scene, the entire child node hierarchy is processed. For the most part this is simple, if there are e.g. multiple drawables in the hierarchy, they all are inserted/removed from the octree.

Where it will break is things where nodes or components are referred to with ID's, like the Constraint's "other body" attribute. Outside the scene there are no ID's, so the reference breaks.

Mike3D:
Does it allow to move a component from one node to another?

cadaver:
No.

JoshEngebretson:
I've had a chance to roll these updates in and they are working great!  Thanks :)

PR #765, opened 2015-07-06
OpenGL multi-context support

thebluefish:
Urho3D now properly supports multiple contexts with OpenGL. This now resembles the behavior allowed by DX9/DX11, in that a single process or thread can create and manage multiple Urho3D contexts.

### DIFF
--- a/Source/Urho3D/Graphics/OpenGL/OGLGraphics.cpp
+++ b/Source/Urho3D/Graphics/OpenGL/OGLGraphics.cpp
@@ -621,6 +621,8 @@ bool Graphics::BeginFrame()
 {
     if (!IsInitialized() || IsDeviceLost())
         return false;
+        
+        SDL_GL_MakeCurrent(impl_->window_, impl_->context_);
     
     // If using an external window, check it for size changes, and reset screen mode if necessary
     if (externalWindow_)
--- a/Source/Urho3D/Graphics/OpenGL/OGLGraphics.cpp
+++ b/Source/Urho3D/Graphics/OpenGL/OGLGraphics.cpp
@@ -2322,6 +2322,8 @@ void Graphics::Release(bool clearGPUObjects, bool closeWindow)
     if (!impl_->window_)
         return;
     
+    SDL_GL_MakeCurrent(impl_->window_, impl_->context_);
+    
     {
         MutexLock lock(gpuObjectMutex_);
         


thebluefish:
Example of use: http://imgur.com/JJUiwY9

Oddly when the SDL_GL_MakeCurrent call was being called in my application, there was some odd rendering glitches. However having it directly in the BeginFrame seems to produce no issues.

thebluefish:
Hm, it seems that there may be a few places we need to call this. Let's not yet merge until I can very all functionality is working right.

For example, closing one GL context may cause another to stop working properly.

weitjong:
I am about to ask the same thing. Why it needs to be called every frame? I suppose you have checked that call is inexpensive to happen every frame. Though, it appears to have an early out check internally, what concerns me is the internal call to SDL_SYS_GetTLSData() in order to do this internal check. It locks/unlocks before retrieving the thread local data. I am saying this just by browsing through the code and not actually profiling anything. So my worry could be unfounded.

Lasse could probably tell us where is the best event to subscribe to call this method.

BTW, could you follow our code convention. Use 4 spaces per indentation level.

thebluefish:
It could be expensive, I haven't been able to find much conclusive information in that regard.

I believe I found out the cause for the lost resources, it seems to be working properly now. However I am going to see if there is a performance penalty, and in what scenarios.

My apologies on the 4-spaces thing. Visual Studio default behavior is to use tabs, and I keep forgetting to change that back when I do a reimage.

cadaver:
I have worked in projects where multiple contexts were used and they were a pain. Another thing is possible broken driver support so I'm extremely hesitant to call Urho as supporting multiple contexts and to merge this PR.

IMO it should be left as the application's responsibility, meaning that if you create multiple contexts, then you also have to manage switching them manually, the only thing that matters to Urho is that when each instance is drawing graphics or loading resources, the correct context must be active.

thebluefish:
Hm, I have not heard of any driver issues related to multiple contexts.  I personally like the idea of enforcing a context switch any time GL functions are accessed. After-all, the GL stuff is supposed to be hidden away from the user. It would be rather hack-y to force someone to dig into Urho3D for functionality that already exists with DirectX.How about a Graphics-class function, MakeContextCurrent(), that would do the work? That way, the user would be responsible for calling the function anytime they are doing stuff with that specific context.

I will try some stuff out.

cadaver:
The expected majority use case when Urho is used to make e.g. a game is single context, so I would not like to see that case (as well as the maintainability of the GL backend) suffer just so that multi-context, which is a lot more esoteric, would work out of the box.

D3D has been designed more wisely as there is no global state in the D3D library/driver itself, instead everything goes through the device.

I'd consider the user-callable function that makes the context current an acceptable middle ground solution.

thebluefish:
I will probably have something up tonight or tomorrow. I went with the user-call function method and it seems to be working when modifying the Urho3D 1.4 release. However the latest master branch is giving me a nvoglv32.dll crash, even without edits. So I will need to figure that out before committing to anything. After that, I will create a sample and document it.

thebluefish:
It looks like no further changes to Urho3D are necessary for proper multi-context support beyond the previous PR.  ```SDL_GL_MakeCurrent``` can be called externally in all the right places, so even a user-callable function isn't entirely necessary. Thanks again for the info!

PR #774, opened 2015-07-12
Editor new node position

1vanK:
Another methods for determining positions for new nodes


1vanK:
It's my first pull request, sorry if it is wrong

MonkeyFirst:
maybe "3d cursor mode" also need to be added ? oh and it's own self - the 3d cursor ) like in blender

cadaver:
Thanks. Seemed to work well. Changed the editor code to use world positions throughout when spawning the node, to make sure parented nodes are positioned right.

Issue #629, opened 2015-01-22
Urho3DAll.h and/or Urho3D/<Submodule>/All.h headers for convenience


Stinkfist0:
Aforementioned header files would be handy, especially for PCH.

weitjong:
What is the difference in adding more stuffs into the already existing
"Urho3D.h" than creating yet another header file that user needs to include
again in their code?

Best regards,
YaoWT

cadaver:
In my experience it's rare that you are going to be using eg. all (or even most) of Graphics in a single .cpp file. Whether it's beneficial from a PCH standpoint would need measurements.

At least for building the Urho library itself IMO nothing should be added to the current Urho3D.h, which just lists the vital defines for export/import/deprecation.

weitjong:
What I wanted to say in my last comment was actually that I am not in favor of having yet another header file, be it being called as "Urho3DAll.h" or "Urho3D_fwd.hpp" or whatever. But this is just my humble opinion.

> At least for building the Urho library itself IMO nothing should be added to the current Urho3D.h, which just lists the vital defines for export/import/deprecation.

Good point.

Stinkfist0:
Was mainly thinking these headers for users of Urho, and agree that Urho itself doesn't need these internally. These files could be maybe automatically generated during Urho deployment, if the feature request is accepted and implemented.

v0van1981:
This is very useful feature, at the moment i use script (C#)

http://paste.ofcode.org/BZHxd2K9cahvPnuYH4DkxS

>> Compiling:
>> csc /target:exe /out:Program.exe *.cs

PR #776, opened 2015-07-13
Adding Urho3DAll.h

1vanK:


weitjong:
This is a duplicate of #629. IMHO, this is a bad practice. Sure it is convenient but it slows the build unnecessarily. It makes the source code depends on all the Urho3D headers that it really needs. The dependency is being checked on each build and when any single Urho3D header does change, your source code will be rebuilt totally instead of incrementally. Also, it would add burden to contributors to maintaining this header file manually when we change our code base.

1vanK:
Any user can use the old way if he wants to. But now he has an alternative :) And add a few lines is not so difficult for contributors.

MonkeyFirst:
I have something similar - common.h with most used urho's headers files within (for Precompiled)
it's speedup compiling.

weitjong:
Don't get me wrong. Adding a few more headers that you think are commonly used by your own project to your own project's Precompiled.h is fine. But adding **all** headers into a single header file just for saving the developer to include the required headers the code actually needs is another matter. It would not be any better even when this single header file is getting precompiled. You can do a quick google search on why this is a bad practice. But, hey, you can do what you want within your own project. For Urho3D project though, my original opinion still stands. Personally I would not promote bad practice. Just my two cents.

PR #783, opened 2015-07-16
Internationalization

1vanK:
Any suggestions?

### DIFF
--- a/Docs/AngelScriptAPI.h
+++ b/Docs/AngelScriptAPI.h
@@ -5668,6 +5668,40 @@ int weakRefs;
 int width;
 };
 
+class Localization
+{
+// Methods:
+String Get(const String&);
+String GetLanguage(int);
+int GetLanguageIndex(const String&);
+void LoadJSON(const JSONValue&);
+void LoadJSONFile(const String&);
+void Reset();
+void SendEvent(const String&, VariantMap& = VariantMap ( ));
+void SetLanguage(const String&);
+void SetLanguage(int);
+
+// Properties:
+/* readonly */
+StringHash baseType;
+/* readonly */
+String category;
+/* readonly */
+String language;
+/* readonly */
+int languageIndex;
+/* readonly */
+int numLanguages;
+/* readonly */
+int refs;
+/* readonly */
+StringHash type;
+/* readonly */
+String typeName;
+/* readonly */
+int weakRefs;
+};
+
 class Log
 {
 // Methods:
@@ -10416,6 +10450,7 @@ Array<Variant> attributeDefaults;
 /* readonly */
 Array<AttributeInfo> attributeInfos;
 Array<Variant> attributes;
+bool autoLocalizable;
 /* readonly */
 StringHash baseType;
 bool bringToBack;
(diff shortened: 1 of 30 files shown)


cadaver:
Looks nice on a cursory look. However, check the contribution checklist http://urho3d.github.io/documentation/HEAD/_contribution_checklist.html, which deals with things like adding script bindings for new features. Also, a UI text element which would automatically relocalize itself according to its translation key would be nicer than having to manually react to a language change, like the example is doing.

1vanK:
1) I do not understand why travis generates errors :(
2) Often, when you change the language you want to change not only the text data, but sprites and other aspects of the game. This implementation provides the user complete freedom to decide what to do when current language is changed.

xgundam05:
I might suggest, if possible, to rename it to Localization instead of Internationalization. Localization being the more standard term, and it makes it a bit more clear that you could have languages like "en-US", "en-GB", "en-CA", "fr-CA", etc. (which it appears this implementation is able to handle that, didn't see anything in there that wouldn't).

cadaver:
@1vanK: understood that one might need to do a lot more than to change text elements. However it's a common and repetitive task, so for example a new attribute "localizationKey" or similar, or a subclass of Text that knows to localize itself, will ease that repetitiveness.

Will be waiting for the script bindings, and the example implemented also in scripting languages (the necessary evil).

Would prefer not having the flag images in the repo, as text already demostrates the localization feature.

1vanK:
Any manual how to binding to LUA? xD

cadaver:
Looking better! One (possibly final) thing is to add documentation, since this is a pretty big addition. I recommend an added page in Reference.dox (it doesn't need to be overly long, just explain the basic concept of using localization, and maybe refer to the example), and add a link to the references list ("For further reference, see") in Urho3D.dox

1vanK:
My English is too bad for writing documentation xD

MonkeyFirst:
Write as you can, while you write it you are learning xD

JSandusky:
If you think it's really really bad, post it in your forum thread. I'm sure you'll get several rewrites, and rewrites of rewrites.

1vanK:
I finished writing the code, pls review. It remains to write the documentation. Russification of the editor can be completed later (presently russified menu).

weitjong:
I feel that putting all the texts into one json is creating an unnecessary hotspot file. I like how Android and iOS have done it, i.e. one language one file. It is easier for multiple localization teams to do the translation at the same time and no more additional work to put those translated string into one file and merge conflict.

1vanK:
It is possible to loading multiple files with different languages

weitjong:
> It is possible to loading multiple files with different languages

If that case then it is not too bad.

1vanK:
It will be good if a native speaker will correct my creativity xD

cadaver:
It looks like the documentation won't be a lot of work to fix, great!

I took a look at the subsystem itself, noticed the asserts which you can get to trigger quite easily from the subsystem public API if you haven't set a language first / loaded valid language files.

I would suggest rather returning some "friendly" error values (for example simply an empty string for no language) as asserts are typically used in Urho3D code for serious or low-level internal error situations, like array out of bounds indexing.

Stinkfist0:
Idea for later on: could be nice to have `String Localization::FormatNumer(int/float/etc.)` for formatting e.g. `1000000.123f`  -> `1 000 000,123`, `1,000,000.123, or `1.000.000,123`.

PR #784, opened 2015-07-17
Jft editor modifications

julyfortoday:
Some minor tweaks to the editor for more Blender like behavior

- center on selected object using HOME key
- setting for MMB to pan, and use MMB + SHIFT to rotate (currently works the opposite way)

Blender used to be MMB for pan, MMB+SHIFT to rotate for years. Then a couple years ago they switched it around. So the current editor matches fine, but for those of use who (greatly) prefer the original way, that is what this feature is for.

### DIFF
--- a/Docs/GettingStarted.dox
+++ b/Docs/GettingStarted.dox
@@ -720,6 +720,7 @@ Left mouse         - Select nodes or drag the node transform gizmo. Hold Shift t
                      select components instead. Hold Ctrl to multiselect.
 Right mouse        - Hold down and move mouse to rotate camera
 Middle mouse       - Hold down orbits the camera around selected objects
+Shift+Right mouse   - Hold down pans the camera
 Shift+Middle mouse - Hold down pans the camera
 
 WSAD or arrows     - Move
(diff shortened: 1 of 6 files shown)


MonkeyFirst:
Blender have a great hotkey functionality but:
I guess that centring on objects is key "dot" or "del on numpad"  in Blender
and Home key already busy by camera to view mode switch ortho or perspective same key in blender do same work 

The key functionality that I like and that I want to see in Urho editor it is the R(rotate)-S(scale)-G(grab) keys with same functionality for fastest level blocking and object manipulation, if this will be introduced in editor it will be a great !)
the problem is that the S key are already used by default camera for movements but I guess if maybe fixed with adding the qualificator for this, I mean that - WASD + RMB = camera move, and single S key will be free for use.

cadaver:
When you add something in the editor settings / preferences dialogs, remember to write settings load/save code for it.

cadaver:
Also, when you add new key shortcuts or functions in the editor, please document them in Docs/GettingStarted.dox (editor instructions page). Otherwise the added features might be easily forgotten.

julyfortoday:
Okay, added load/save code, and updated the documentation.

cadaver:
Thanks!

cadaver:
Will move the setting to the settings dialog, as that contains all other camera behavior toggles. Note for next time: enable whitespace and make sure you don't enter tabs instead of spaces, to match with the rest of the code & xml data.

PR #785, opened 2015-07-17
Material editor few model preview

MonkeyFirst:


cadaver:
I would be against removing established key shortcuts (Ctrl+1,2,3) from the Urho editor. Rather add the Blender shortcuts as an alternative.

Also, for the material preview model, why not just let the user pick one with the fileselector?

MonkeyFirst:
>Rather add the Blender shortcuts as an alternative.
Ok, i'm think about this, maybe add some option in editor's preferences.

>Also, for the material preview model, why not just let the user pick one with the fileselector?
You mean that to give a user ability to choice a model from "Models" directory ?

MonkeyFirst:
Note: this new hotkeys available only for blender hot keys mode, for standard mode it works as before

Ctrl + F - switch camera flying mode
if camera is not in flying mode then single object possibly to manipulate with
R - rotate 
R + ALT - reset rotate to Quat(0,0,0)
S - scale 
S + ALT - reset scale to Vector3(1,1,1)
G - grab (move) 
G + ALT - reset position to Vector3(0,0,0)

also have a F-key to Focus on the selected object
the H-key for hide in terms of Blender or Enable/Disable in terms of Urho
P + Alt - Unparent node
Z - Cycle switching Fill mode
X - Delete object

Stinkfist0:
BTW you got typo going on here: "standardt" should be "standard".

MonkeyFirst:
This feature allows to parent group of selected nodes into space of last selected node (make it as parent for others)

Additional keys:
for Standart key mode - U
for Blender key mode - P + Ctrl

Stinkfist0:
@MonkeyFirst "standard" is English, "standardt" is not.

MonkeyFirst:
Oh, I think about that usage the word "standarD" or "standarT" in some cases are wrong
Could you point to where i typed "standarDT" in sources ?

Stinkfist0:
Might have looked that wrong and even typoed myself, but in any case, the word should be "standard".

MonkeyFirst:
ok, i got it. i'm replace all "standart" to "standard"

cadaver:
>You mean that to give a user ability to choice a model from "Models" directory ?

Yes, just open the fileselector with .mdl extension and default it into the Models directory. Though thinking more, I don't think this is important, as usually you want to use simple shapes to preview a material. The hardcoded list just has to be updated if more basic shapes are ever added.

Please document the new shortcut keys to GettingStarted.dox under the "Editor Instructions" page, I believe then this should be possible to merge.

Testing from others is also appreciated, since this seems to touch a large amount of editor functionality.

MonkeyFirst:
>Please document the new shortcut keys to GettingStarted.dox under the "Editor Instructions" page, I believe then this should be possible to merge.

Ok, i'm try to figure out with documentation
Later i want do some polish commits with some additional functionality to Mask Layer Editor and after that I think it will be ready to merge.

MonkeyFirst:
little demonstration of it
https://www.youtube.com/watch?v=Y8M4hfKYd5M

MonkeyFirst:
Please, check out this. I think is ready to merged.

cadaver:
Have tested this, I had no problems with it but I'm not an expert on Blender so would be great to have confirmation from someone who is more experienced. Tested that it will also merge (with a minor conflict that's easy to fix) with the russification PR, so that can be merged first.

I noticed that the instructions shows "standard hotkeys" section twice, the lower should be Blender hotkeys, right? Also the existing instruction block about the MMB behavior should likely be moved next to the standard hotkeys section, as it doesn't have effect in the Blender mode. Anyway, these are rather minor things, and can be fixed post-merge.

MonkeyFirst:
>the lower should be Blender hotkeys, right?
Yep, but strange I'm just copy section with std keys and add some changes for blender keys place it under std keys. Maybe with text "EditorInstructions_Controls" are also need to do some changes? for example edit it into EditorInstructions_Controls2 ? It will be great if you create this all sections in right way and after that I'm later add or fix these all blender hotkeys.

cadaver:
The sections have to be uniquely named, so yes. for example EditorInstructions_Controls2.

MonkeyFirst:
Cool ! I guess that not last pull from me for supporting blender style working in UrhoEditor. 
In future i trying make pulls more smaller and about one feature thing ) thanks cadaver.

PR #790, opened 2015-07-21
add warnings to Localization::LoadJSON, partial editor russification

1vanK:


weitjong:
Thanks for the PR but I actually have a reservation of adding another Editor wrapper shell script/batch file just for switching to a particular language. Now that Urho3D has localization subsystem then it should automatically switch to use the right localized language based on the OS-specific L8n setting. On Unix-like OS, the application should read the ```LANG``` environment variable for this purpose. For example, this is what I have on my Linux workstation.
```
$ locale
LANG=en_US.utf8
LC_CTYPE="en_US.utf8"
LC_NUMERIC=en_US.utf8
LC_TIME=en_US.utf8
LC_COLLATE="en_US.utf8"
LC_MONETARY=en_US.utf8
LC_MESSAGES="en_US.utf8"
LC_PAPER=en_US.utf8
LC_NAME="en_US.utf8"
LC_ADDRESS="en_US.utf8"
LC_TELEPHONE="en_US.utf8"
LC_MEASUREMENT=en_US.utf8
LC_IDENTIFICATION="en_US.utf8"
LC_ALL=
```
and on my Mac OS X VM.
```
$ locale
LANG=
LC_COLLATE="C"
LC_CTYPE="UTF-8"
LC_MESSAGES="C"
LC_MONETARY="C"
LC_NUMERIC="C"
LC_TIME="C"
LC_ALL=
```
I could have my ```LANG``` env var set to ```ru_RU``` or ```zh_SG``` or any other languages. I believe on Windows side, they have something called regional settings as well. So, it would be great if Urho3D app can react to that automatically.

1vanK:
This is a temporary solution until the languages only two. Later, I will create switching of the language in the editor settings.

weitjong:
Thanks for that in advance. Having a language switcher in the Editor will be good. However, you still miss my point. IMHO, it will be even better if the Editor or any Urho3D app knows what is the default language it should be using initially.

cadaver:
Sometimes the automatic language detection may not be wanted. Easiest is probably to react to the initial languagechanged-event and restore the language the application wants to enforce (for example one loaded from the application's own config)

weitjong:
I agree this may be app-specific requirement. It may not even required to be localized. On other app, it may be required to have its own language setting overriding OS localization setting. In the latter case, I think the common sense would be to use OS L8n setting if there is no app-specific setting initially. And if OS LANG setting is not set then the app should default to en-US. In any case there is no need for additional launcher per language.

cadaver:
Is this ready to merge now, or will you add more still?

1vanK:
it is ready to merge

weitjong:
I wonder whether the localization logic could be moved to the String class so existing apps does not need to pepper the code with the ```localization.Get()``` calls and so the localization subsystem can be enabled/disabled as a build option similar to other optional subsystems. Granted that this won't work for a normal string literals in the code, but it would reduce the amount of work to get ready for i8n tremendously. Below is a simple use case for illustration.

```
// L8n subsystem loads the JSON language pak during initialization
String::SetLanguage("de_CH");    // New method to govern system-wide L8n language setting.

// ... later
String str("hello");
LOGINFO(str);        // Output: grüezi
```

1vanK:
Change language can affect not only the strings

weitjong:
Yes, I absolutely agree. I am not saying it is :smile: .

1vanK:
And not every string should be localized: String path = "c:\temp\file.txt";

1vanK:
But not every string shoul be translated. Inside engine lines are used everywhere - the path, IDs, the names of textures. Changing the language will need to inspect all strings into program.

Real example when i localized the Editor:
old:
```
    renderStatsText.text = String(
        localization.Get("Tris: ") + renderer.numPrimitives +
        localization.Get("  Batches: ") + renderer.numBatches +
        localization.Get("  Lights: ") + renderer.numLights[true] +
        localization.Get("  Shadowmaps: ") + renderer.numShadowMaps[true] +
        localization.Get("  Occluders: ") + renderer.numOccluders[true]);
```
new
```
    renderStatsText.text = String(
        localization.Get("Tris: ") + renderer.numPrimitives +
        localization.Get("  Batches: ") + renderer.numBatches +
        localization.Get("  Lights: ") + renderer.numLights[true] +
        localization.Get("  Shadowmaps: ") + renderer.numShadowMaps[true] +
        localization.Get("  Occluders: ") + renderer.numOccluders[true]);
```
In the new version will be
```
String a1 = String("Tris: ");
a1.SetAutoLocalizaed(true);
String a2 = String("  Batches: ");
a2.SetAutoLocalizaed(true);
String a3 = String("  Lights:");
a3.SetAutoLocalizaed(true);
String a4 = String("  Batches: ");
a4.SetAutoLocalizaed(true);
String a5 = String("  Shadowmaps: ");
a5.SetAutoLocalizaed(true);
    renderStatsText.text = String(
        a1 + renderer.numPrimitives +
        a2 + renderer.numBatches +
        a3 + renderer.numLights[true] +
        a4 + renderer.numShadowMaps[true] +
        a5 + renderer.numOccluders[true]);
```

Or we can add AutoLocaization Parametr to Constructor:
```
    renderStatsText.text = String(
        String("Tris: ", true) + renderer.numPrimitives +
        String("  Batches: ", true) + renderer.numBatches +
        String("  Lights:", true) + renderer.numLights[true] +
        String("  Batches: ", true) + renderer.numShadowMaps[true] +
        String("  Shadowmaps: ", true) + renderer.numOccluders[true]);
```

weitjong:
I am just throwing the idea here. I am sure there will be some difficult design decision to make to actually implement this. My short use case above is by no means showing all the changes required. IMHO, the E_CHANGELANG event should only be used for UI-related objects like Text or Text3D classes where they need to react immediately to the language changed event. For other Strings object, I suppose they just need to return the string in the new language at the time when they are needed. I could be wrong here but I won't subscribe String to E_CHANGELANG.

Specifically for the Editor's localization work, since it is using AngelScript and Urho3D has registered a string factory to use Urho3D::String, I believe all the string literals in the AngelScript are anyway instances of Urho3D::String class. I have quickly confirmed this by entering this test in the Editor's console.
```
Print("This is a test".length)
```
It produced 14 as output as expected. So, I don't think we need to change the string literal there. In my earlier comment I was referring to C++ string literal.

I think I have said it before but let me say it again. Thanks for your continued effort for this. Keep up the good work.

1vanK:
As a string should determine whether it needs to translate its value?

>> I could be wrong here but I won't subscribe String to E_CHANGELANG
The string must be stored translation and update it when  E_CHANGELANG. Or search translation in the dictionary for each access (it's heavyweight operation).

weitjong:
I actually don't have the answers for all the questions that may arise for implementing this idea. I would probably add a new instance variable in the String class to indicate whether the string object is ```immutable``` (borrowing the terminology from Java). Those that constructed directly from a string literal can be considered as immutable. Those that come from concatenation operation, for example, does not have this flag set. We can still have getter/setter for this flag to let user override it otherwise. Probably only those objects with immutable flag set need to be translated and eventually get translated when the language pak has the entry for it. Perhaps we should call this flag ```translatable``` instead? :smiley: 

I have already given my opinion on how to handle E_CHANGELANG in my previous comment. To give a concrete example. It is the Text class that reacts to E_CHANGELANG event. Its handler would get the new translated strings from the String objects that the Text object owns. String class does not know E_CHANGELANGE directly. It just knows that when it needs to return the string representation of itself then it would perform a lookup on demand. Probably it is a good idea to cache this lookup result and the LANG being used in the lookup. So, lookup is only being done again when the cache is invalidated. Thus, if I have 100 Strings in my app and only one being used in the Text object then only one String object would perform lookup when E_CHANGELANG arrives at Text object.

As for how the lookup can be done in an efficient manner, this is yet another implementation detail question. I think my guess of what might work best is a good as any of us here.

1vanK:
Sorry, I lost the general idea of it. Need automatically translatable strings or not? What do you mean?

1vanK:
I do not understand the general idea of it. But you can implement it yourself.

PR #793, opened 2015-07-23
Smart instance

MonkeyFirst:
This feature allows fast instantiate one model for StaticModelGroup.
You start with single node with one Static model within.
Then you do first pressing ( alt+d) it create a static model group with first instance of model you need to move it somewhere and  next pressing on alt+d just add new instances.
see this in work: https://youtu.be/YOGpdieVCUE

### DIFF
--- a/Docs/GettingStarted.dox
+++ b/Docs/GettingStarted.dox
@@ -790,6 +790,8 @@ Z                  - Cycle through solid, wireframe and point rendering
 M                  - Show/Hide LayerEditor
 F                  - Focus on selected object (LookAt)               
 X                  - Delete node or component
+Alt+D              - Smart Duplicate (note: select one of axises on gizmo to show direction for next instance)
+Key "." or Del     - View Closer selected node(s)  
 \endverbatim
 
 Press right mouse button in the 3D view if you want to defocus the active window without changing the object selection.
--- a/bin/Data/EditorStrings.json
+++ b/bin/Data/EditorStrings.json
@@ -610,5 +610,13 @@
 	"File":{
 		"en":"File",
 		"ru":"Файл"
+	},
+	    "Please, use only nodes with one StaticModel for instansing":{
+		"en":"Please, use only nodes with one StaticModel for instansing",
+		"ru":"Пожалуйста, используйте только ноду с одной статической моделью для инстансинга"
+	},
+	"  CameraFlyMode: ":{
+		"en":"  CameraFlyMode: ",
+		"ru":"  Камера в режиме полета: "
 	}
 }
--- a/bin/Data/Scripts/Editor/EditorActions.as
+++ b/bin/Data/Scripts/Editor/EditorActions.as
@@ -14,6 +14,48 @@ class EditActionGroup
     Array<EditAction@> actions;
 }
 
+class CreateStaticModelGroupIndexListAction : EditAction 
+{
+    uint nodeID;
+    uint drawableID;
+    uint instanceNodeID;
+    
+    void Define(Drawable@ drawable, Node@ instanceNode)
+    {
+        drawableID = drawable.id;
+        nodeID = drawable.node.id;
+        instanceNodeID = instanceNode.id;    
+    }
+    
+    void Undo()
+    {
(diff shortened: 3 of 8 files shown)


cadaver:
I don't have anything against this feature, because if you really need to use StaticModelGroup you want it to be as convenient as possible, but am just curious to know: have you profiled in practice that using StaticModelGroup is vital for your performance? It has downsides too: it creates a large combined bounding box, the instances will not LOD uniquely, and lighting is performed for the whole group at once. So I'd recommend to use it as little as possible.

MonkeyFirst:
>have you profiled in practice that using StaticModelGroup is vital for your performance?
i guess In most cases on solid geometry with simple material it produce less batches, and only if instances are using mats with alpha tech rendering the smg have same performance as Static Model. But I'm not sure about this at 100%. Maybe you're right and in some cases smg have a poor behavior then sets of static model.
In future this feature may have an additional functionality. For example you do this instancing and then in menu or by pressing on other hotkeys change this sets of nodes to Static Model set with common parent node.

cadaver:
When we have a case with a single light (forward rendering), StaticModel and StaticModelGroup should produce exactly same batch count.

When there are multiple lights (forward rendering too), you may notice StaticModelGroup producing less batches, but that's actually because it's lighting all the objects in the group (according to the group bounding box) even if some could be left unlit by some lights. So that may actually be increasing the GPU workload a lot, at the cost of less CPU work.

MonkeyFirst:
you mean that if it will be a StaticModel copies, it will be more better, and you wont need that users use sgm a lot ?
ok, am fix that. but in this case it no differect between duplicate and this feature, except adding the offset to next model.

cadaver:
It's hard to give a valid answer for every case as it depends on so many factors (GPU, CPU, lights, triangle count in meshes, object count) but my rule of thumb would be that StaticModelGroup should be an exception used for extreme cases (like you see in the HugeObjectCount test) and normal StaticModels for common cases.

Profiling is best: when you see profiler blocks under UpdateViews taking a lot of ms when looking at a direction where there's a lot of objects, it may be time to investigate StaticModelGroups.

MonkeyFirst:
checkout this please, maybe it's may to be merged
- in most cases this additional functionality for B.mode

cadaver:
Thanks, I'll check it out and likely merge today.

cadaver:
Two things I noticed, which would be nice to be addressed:

lastSelectedNode / lastSelectedComponent etc. These are strong AngelScript handles, so they will actually keep the node / component alive. Prefer to use WeakHandle for these or make sure they are set to null if the node / component in question is removed from the scene. It probably does not cause an issue because the engine will not find them in the scene anymore after removal, and they shouldn't e.g. render any more, but from editor code cleanliness standpoint it would be good. 

Noticed that the group node for smart duplicate instances seemed to always have a local ID. Would be preferable to be able to choose global or local. (or did I miss something?)

MonkeyFirst:
>lastSelectedNode / lastSelectedComponent etc
I'm introduce this because the selectedNodes, selectedComponents  or editNodes etc do not store node in path as they was selected. Moreover they may absolutely empty but selection of node in viewport are present.
>Prefer to use WeakHandle for these
I do not used this before, and it's looks as a lot additional writing with someVarName.Get() or mb also need do cast<>?
WeakHandle<Node@> lsn
Node@ node = cast<Node>(lsn.Get())
I try to figure out with this WeakHandle, and do some clearness )

> have a local ID.
>or did I miss something?)
you may unparent they if you wish later, but for "duplicate time" it will be better if they all placed in some "common branch/node" also many "garbage nodes"  in main scope are heavy to navigate with HierarchyWindow (it also need some "freezing ability" for avoid unrolling marked group of nodes every time when HW are updates)

cadaver:
Yes, that's WeakHandle use in a nutshell. You shouldn't need casts, as they're implicitly performed.
´´´
        WeakHandle nodeWeak = myNode;
        Node@ nodeStrong = nodeWeak.Get();
        Print(nodeStrong.name);
´´´

PR #803, opened 2015-07-28
French strings addon

christianclavet:
Added french strings, so the editor should be able to support french


christianclavet:
Not tested. Let me recheck if there is missing comma. I forgot some strings at the beginning of the file and just updated.

christianclavet:
This should be good now.

PR #805, opened 2015-07-29
hsv-color-wheel

MonkeyFirst:
Hey! i'm back ) with solved fork
Please, checkout this i'm think it ready to merge.
There is still the same color wheel as before for fast color change in view port such things as light, static mesh, zones by pressing hotkey on alt+w  (b.mode) In std mode you may add any hotkey for this.


cadaver:
This seems to contain a lot of hardcoding and string compare, which can create a maintainability problem: if I understand right it will not work automatically with new components that have color attributes, or material parameters other than the ones the code recognizes.

IMO ideally it should work generically on any COLOR type component attribute, and any material parameter that has 3 or 4 values (can be edited as rgb or rgba). Sometimes you might be able to mistakenly edit some naterial parameter with the color wheel that isn't actually a color (for example if it's rather a direction), but I don't think that's a big deal.

I'm not firmly opposed to merging this as is, as it's already usable, but consider if you want to rework it to a more generic fashion.

MonkeyFirst:
> hardcoding and string compare
it's menu@.names and it's not use the localization things as before, i'm just do not find better way to do that. maybe I should use menu@.vars["menuIndex"] = n and then compares this indexes? 

>IMO ideally it should work generically on any Color type component attribute, and any material parameter that has 3 or 4 values 

Yes, you are right, but I have not figured out these things yet (attribute inspector, material editor). It's maybe added in future in any time I guess.

cadaver:
Merged. Alt+W is used for the wheel in both hotkey modes. Changed the wheel image to uppercase file name to match the existing image files.

Btw. in the future be careful to consistently use the same brace spacing convention like used elsewhere in the code, for example: if (hotKeyMode == HOTKEYS_MODE_STANDARD)

PR #829, opened 2015-08-11
add debug icons

MonkeyFirst:
Added editor feature that allows show debug icons on Alt+I (or from menu)

Please, checkout this.


1vanK:
Nice!

MonkeyFirst:
Thanks )
add splinePath debug show

1vanK:
We can use the footprint for the path :)

http://tattoo-search.ru/uploads/posts/2011-05/1306475240_tatu-sled.jpg

MonkeyFirst:
Yes, especially this image  :)

cadaver:
Looking good! Will check this in detail later, hopefully today.

MonkeyFirst:
Yes this all temp, temp container - node and temp billboard sets.
It will be great if Debug renderer will supporting billboard sets and sprites for this.

And one question: i found one strange thing until work on this, what the Zone affecting on Billboards with DiffVColUnlitAlpha this normal behavior? because this "Unlit" material. I think that it mean - no color affecting on this from any other scene's object

cadaver:
If you use Unlit shader, it will still use the zone fog color and distance. Using the Basic shader would ensure the most basic operation possible - no fog.

If DebugRenderer would be able to render sprites / billboards, it would be duplicated functionality & code compared to BillboardSet, so I believe using BillboardSet is actually better in this case, and DebugRenderer remains for the very simple geometric rendering (lines etc.)

MonkeyFirst:
>so I believe using BillboardSet is actually better in this case

ok

>Using the Basic shader would ensure the most basic operation possible - no fog.

I'm just try it, do copy of DiffVColUnlitAlpha and change Unlit to Basic, but all billboards are gone, and not visible. i'm don't know why(
I think you may add for this proper shader like unlit just exclude Zone ambient and fog color affecting after merge. or add DiffVColUnlitAlphaWithoutZoneAffecting.xml and i'm rewrite icons materials.

Also am think add icon for Rigidbody with IsTrigger option - On (trigger volumes)
Any suggestion for Icon? I don't know what it might be

cadaver:
Yes, I can check out the shader & fix the fogging later, it isn't that critical.

For trigger, maybe just take the icon of zone and make it different, for example with dotted outlines. It's kind of hard to think of something suitable, unless we go to actual mechanical objects, like a lever / pushbutton, which may be odd compared to the rest.

MonkeyFirst:
> for example with dotted outlines. It's kind of hard to think of something suitable, unless we go to actual mechanical objects, like a lever / pushbutton

I'm looked into CollisionShape types there a lot std various forms, and I decided create something universal for each type: label with char - "T", this mean - Trigger. Anyway this icons are opened for modifying in future from any contributors.

cadaver:
Ok. Looked through this. There is a lot of repeating code that could be avoided by using an array of billboardsets, and putting the repeated stuff into functions. For example instead of debugIconsSetDirectionalLights you could refer to debugIcons[ICON_DIRECTIONAL_LIGHTS] where the array index is a defined constant. This way for example enabling/disabling all icon types would be a simple loop through the array.

Also, there are sometimes several empty rows used erratically. Please use 1 empty row inbetween code sections consistently.

After these 2 things it's OK to merge by me.

MonkeyFirst:
checkout this, now this uses array of BS .

cadaver:
Thanks. Will re-check.

friesencr:
Thanks @MonkeyFirst these are pretty rad!

MonkeyFirst:
No problem ) 
In future it may be modified with one common BS and altas texture with all types of Icons, but in this case it's will be not easy to add new icons to common texture, also texture do not have infinite size -> limited icon amount.

cadaver:
It'd be even nicer to have the icons update every frame when necessary (so that movement and sizing are smooth) which may need changes to the engine, for example BillboardSet auto-resizing, or moving into DebugRenderer after all. That can be revisited later, it's certainly also good as it is now.

MonkeyFirst:
Am do some tweak for smooth update I move the UpdateViewDebugIcons from EditorView.as (HandlePostRenderUpdate) into Editor.as (HandleUpdate)  and it's become more smooth on mouse rotation(look). You may do this fix in master.

PR #835, opened 2015-08-13
Added = operator

cosmy1:
Added assign operator since there is a copy constructor. (More consistent)


cadaver:
Please merge all "Added = operator" requests into one pull request, as there were several with a similar change, and they all are rather small.

PR #836, opened 2015-08-13
Added = operator

cosmy1:
Added assign operator since there is a copy constructor. (More consistent)


Stinkfist0:
Wonder why the manually written copy-ctors exist for these. For script bindings maybe?

cadaver:
@Stinkfist0: It's just a style to make explicit (simply from looking into the header) what happens when a math object is copy-constructed.

PR #838, opened 2015-08-13
Wrong Casting

cosmy1:
Fixed some casting warnings

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ generated/
 /Source/Urho3D/Urho3D.h
 /Source/Urho3D/Urho3D.pc
 /Source/Urho3D/librevision.h
+/Source/Urho3D/Precompiled.cpp
 
 # Compiled shaders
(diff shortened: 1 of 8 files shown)


weitjong:
Most of the CI build jobs failed, except those using Clang compiler. I think you have found a bug but not sure it belongs to ```ccache``` or ```GCC/MinGW``` or something wrong with our PCH setup; or in combination of those. I will try to sort this one out.

weitjong:
Clearing the cache seems to fix the problem. It is still interesting to find that replacement of datatype from ```String``` to ```const String&``` would have caused our build system to fall apart. I do not have time to investigate this further but it seems like ```ccache``` skipped some of the compilation units and used one of its already out-dated cached object, but it did so only because the GCC or MinGW had given it a wrong cue. Only Clang compiler pass in flying color here.

cosmy1:
Seems fixed now.

PR #840, opened 2015-08-14
Const references

cosmy1:
Fixed some parameters wrongly passed by value


cadaver:
In SplitRect() function it's important for the original rect to be a value, because the vector where it comes from is modified in the function, and this may happen several times while the original rect is still needed. Vector reallocation would invalidate the reference.

cosmy1:
I was doing the same thing, but yeah, perfect!

Issue #839, opened 2015-08-14
CollisionXXXXX2D without RigidBody2D


1vanK:
If the node does not have a RigidBody2D and create CollisionXXXXX2D component arise a warning
```
No right body component in node, can not create collision shape");
```
If then add RigidBody2D, physics still not working. Maybe we need to automatically create  RigidBody2D component? PhysicsWorld2D automatically created, if it is need.

cadaver:
Automatic generation of scene root level subsystem components is nice, but automatically creating node components is undesirable IMO, as sometimes it's intentional and the programmer intends to manually add the component later.

It's clear that it should be able to recover in case the RigidBody2D is created later, similar to how RigidBody & CollisionShape work.

@aster2013 hasn't been around that much for some time, so I feel the Urho2D part of the engine is in danger of falling unmaintained. If you (or anyone else) would like to work to fix on issues like this, you're most welcome.

1vanK:
I learn Box2D now and afraid to do terrible things in code :) But I can try :)

1vanK:
Question: Can the node to have more than 1 RigidBody2D?

cadaver:
By a preliminary look it seems like the collision shapes are already destroying and recreating themselves when their enabled flag is toggled (and storing all attribute data into the fixtureDef, so it doesn't depend on the existence of the actual Box2D object), so it looks like this can be solved by RigidBody2D iterating through collisionshapes in the node upon its creation, and calling CreateFixture() on them.

Multiple RigidBody2D's in a node makes no sense, since they would both be conflicting in setting the node's transform.

1vanK:
I don't know why, but it doesn't works
```
void RigidBody2D::OnNodeSet(Node* node)
{
	if (node)
	{
		node->AddListener(this);

+		PODVector<CollisionShape2D*> shapes;
+		node_->GetDerivedComponents<CollisionShape2D>(shapes);
+
+		for (PODVector<CollisionShape2D*>::Iterator i = shapes.Begin(); i != shapes.End(); ++i)
+		{
+			(*i)->CreateFixture();
+			AddCollisionShape2D(*i);
+		}
	}
}
```

1vanK:
Also when serialize/deserialise (e.g. in editor press start/stop with enabled RevertOnPause) warning arise every time (and console opening) and probably need to sort components...

PR #841, opened 2015-08-14
fix add RigidBody2D after CollisionShape2D

1vanK:
fix https://github.com/urho3d/Urho3D/issues/839


cadaver:
Good work, thanks! Will check out shortly.

PR #850, opened 2015-08-19
TmxFile2D: Fix determination of object type

bjorn:
TmxObjectGroup2D::Load was using the presence of the "width" and "height" attributes for deciding whether an object was a rectangle, ellipse or something else. This approach breaks for maps saved in more recent versions of Tiled. This is because the width and height are only saved when they differ from the default (0), and because the width and height are now also used to store the size of tile objects.

So for maps saved with Tiled >= 0.12, this change:

* Fixes identification of rectangle and ellipse objects with 0 width and height (which was leading to the entire map failing to load).
* Fixes identification of tile objects.

### DIFF
--- a/Source/Urho3D/Urho2D/TmxFile2D.cpp
+++ b/Source/Urho3D/Urho2D/TmxFile2D.cpp
@@ -163,47 +163,52 @@ bool TmxObjectGroup2D::Load(const XMLElement& element, const TileMapInfo2D& info
         if (objectElem.HasAttribute("type"))
             object->type_ = objectElem.GetAttribute("type");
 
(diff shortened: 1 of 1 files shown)


cadaver:
Thanks, nice to get a fix from the author himself!

bjorn:
We needed this fix for [LucKeyProductions/KO](https://github.com/LucKeyProductions/KO), a game my brother is working on and I was helping him integrate Tiled for level editing. :-)

I consider the issue here somewhat my fault, for making the TMX format a little confusing. It happened due to the continuous adding of unanticipated features. It's hard to describe all these details and unfortunately the sheer amount of TMX support libraries out there are almost consistently out of date. 

Maybe I'll have more changes later. :-)

PR #858, opened 2015-08-26
fix-m-typing

MonkeyFirst:
little fix then you typing text into LineEdit, do not show LayerEditor.


cadaver:
Thanks, merged. Again, please be careful with the editor coding convention: indents 4 spaces, and space between if and the parenthesis.

PR #862, opened 2015-08-29
Hierarchy window extras

MonkeyFirst:
Some fixes: for debugIcons, add splinepath len for scripts, delete by x fix.
And new few extras for hierarchy window: ID show option, oneClick collapse to root(scene) button.  
Please, checkout this.


cadaver:
I will check this in detail later. Two immediate things:

- Please also add the corresponding Lua binding for the added SplinePath length property
- Coding convention, please strive to be consistent so that no later fixes are needed. There's some missing spaces between if & parenthesis, as well as some brackets at the end of line (should be on a line of its own)

MonkeyFirst:
> add the corresponding Lua binding 
It will be little tricky for me because i never used Lua before. But I found some file with name - SplinePath.pkg and looking into it. and there i'm not found some kind of binding to GetSpeed() method. I see only speed property binding. And it's strange because splinePath.h have only float speed_ and actually it private variable and not public.

>between if & parenthesis
You mean that IF(onespace)(condition) are always need to keep/write one space?

cadaver:
There may be inconsistencies in the existing bindings. In Lua the tradition has been to expose both setter & getter functions, and properties, while in AngelScript usually the property only is exposed. But in this case the explicit getter & setter functions are missing. Feel free to add them while you're at it. The tolua property mechanism always uses getters/setters (in this case automatically) so it doesn't matter that speed_ is private.

Yes, there should never be if(something) in the code, but rather if (something)

cadaver:
Seems to work OK, the additions like show ID toggle are nice.

Two suggestions / questions:
- The hierarchy is now wider than inspector, can you try to trim the excess size (collapse and expand buttons, and some spacing). It doesn't matter if it stays slightly wider. Finally you could make the inspector's default width same as the hierarchy's new default width, it's probably too small for things like Vector3 attributes anyway
- Will you do the Lua change? Not a big deal if you don't, in this case, it's just something to strive for that each pull request keeps all script API's in sync, when something gets added.

MonkeyFirst:
>can you try to trim the excess size 

In last commit I try do this

> it's probably too small for things like Vector3 attributes anyway

yes probably LineEdit field of all vectors in Inspector are very wide. And I guess that values precision with allows show floats such as 0.000005 do not needed in text field actually. 0.001 is enough I guess. 

>Will you do the Lua change?

no

cadaver:
Thanks for answer, will merge now.

PR #869, opened 2015-09-01
Private section in ProfilerBlock

randfst:
There was no "private:" in ProfilerBlock before data section


cadaver:
Did you verify that the library builds after this change? It doesn't. Anyway, the ProfilerBlock being all public is intentional, as there are several data members, to which adding private and getters would be cumbersome, and it's practically only ever used by the Profiler.

PR #874, opened 2015-09-04
Added flag to enable/disable animations being auto-removed when weight is zero

hjmediastudios:
As the title says; I added a flag and associated functions to allow the `AnimationController`-controlled animations to be automatically removed (current behavior) or not as the user chooses. This is to aid functionality with some different animation techniques, especially with my implementation of an animation state machine where animations need to be persistent.


cadaver:
Good feature! Noticed you are missing the Lua binding changes, and there is some strange commits with the GraphicsAPI file (looks like Github site text). Can you add the Lua binding and squash the history?

cadaver:
Merged. Added the Lua bindings and some missing things (network replication, getter)

PR #883, opened 2015-09-10
Add cubemap generation capability to the Urho3D editor

JSandusky:
Addition of a command and associated settings for generating cubemaps in the Urho3D editor.

Cubemap sampling is performed at the position of the node containing a selected zone. This is deliberate as node position is considered to be the "true" position of the zone for sampling purposes. For other reasons a zone may need to reach substantially outside of this range (ie. into the sky), even though only a limited area is really relevant for cubemap generation.

During this process existing zones are disabled and duplicates are made matching their traits, this is deliberate to prevent doubling of anything that may depend on the Zone texture.

The added "GetImage" method to Texture2D in script is intended solely for editor usage. It is likely not sufficiently robust for general purpose usage.

Cubemaps are by default written to Data/Textures/Cubemaps/{SceneName}, and are named by the component ID of the zone. A name be prepended to the generated name for the purpose of generating multiple cubemaps for the same zones (night/day scenarios).

Settings for this process are available in the settings window.

### DIFF
--- a/Source/Urho3D/Script/GraphicsAPI.cpp
+++ b/Source/Urho3D/Script/GraphicsAPI.cpp
@@ -175,6 +175,23 @@ static Viewport* ConstructViewportSceneCameraRect(Scene* scene, Camera* camera, 
     return new Viewport(GetScriptContext(), scene, camera, rect, renderPath);
 }
 
+static Image* Texture2DGetImage(Texture2D* tex2d)
+{
+    Image* rawImage = new Image(tex2d->GetContext());
+    const unsigned texSize = tex2d->GetDataSize(tex2d->GetWidth(), tex2d->GetHeight());
+    const unsigned format = tex2d->GetFormat();
+
+    if (format == Graphics::GetRGBAFormat() || format == Graphics::GetRGBA16Format() || format == Graphics::GetRGBAFloat32Format())
+        rawImage->SetSize(tex2d->GetWidth(), tex2d->GetHeight(), 4);
+    else if (format == Graphics::GetRGBFormat())
+        rawImage->SetSize(tex2d->GetWidth(), tex2d->GetHeight(), 3);
+    else
+        return SharedPtr<Image>();
+
+    tex2d->GetData(0, rawImage->GetData());
+    return rawImage;
+}
+
 static bool Texture2DSetData(Image* image, bool useAlpha, Texture2D* ptr)
 {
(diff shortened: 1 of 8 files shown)


MonkeyFirst:
Great work! But i'm look into cubemap xml file and find this ) texture name it's same in all faces

<?xml version="1.0"?>
<cubemap>
    <face name="Data/Textures/Cubemaps//16777228_PosX.png" />
    <face name="Data/Textures/Cubemaps//16777228_PosX.png" />
    <face name="Data/Textures/Cubemaps//16777228_PosX.png" />
    <face name="Data/Textures/Cubemaps//16777228_PosX.png" />
    <face name="Data/Textures/Cubemaps//16777228_PosX.png" />
    <face name="Data/Textures/Cubemaps//16777228_PosX.png" />
</cubemap>

MonkeyFirst:
I try to fix this gaps between faces and now it looked better I guess.
video demo: https://youtu.be/cbivto_Au6E

also I turn-off the grid and debugIcons until it do baking. 
Maybe Gizmo are also need to be hided and physic debug stuff, actually I do not known.
there is source with fixes: http://pastebin.com/J2Wf6NjA

JSandusky:
Ahh, yeap, I goofed the name.

Thanks for the refinements on the camera for near distance and aspect ratio and debug/editor elements rendering in the cube sampling.

I think you might have gotten ahead of yourself though on the cubemap faces/rotation. Moving the increment of updateCycle_ would have discombobulated the HandlePostRender's saving of the images. HandlePreRender would have worked with one value while HandlePostRender would work with a different one (ie GetFaceForCycle(0) vs GetFaceForCycle(1)).

cadaver:
Checking this now. I'll modify the GetImage function slightly, as attempting to capture a float or 16bit texture would overflow Image's buffer. Got also an error to create the cubemap destination directory; this is actually a shortcoming of FileSystem::CreateDir() which doesn't create the intermediate directories (I was missing the Cubemaps directore under Textures). After these I believe it's ready to go in.

JSandusky:
Is there a preferred way you'd like the CreateDir issue to be handled?

On Windows there's the "MakeSureDirectoryPathExists" function in DbgHelp to do it all in one go. MS notes that it'll be present on all machines (but probably not up to date).

cadaver:
I fixed the CreateDir() function directly. Also made the path handling relative to the currently set scene resource path (instead of being a relative filesystem path, which is errorprone) and some minor fixes to the XML generation.

PR #896, opened 2015-09-16
Export OBJ files from the Urho3D Editor

JSandusky:
This pull request adds OBJ export of either the entire scene or just the selected drawables to the editor. Includes normals and texture coordinates (TEXCOORD1) if present in the drawable's geometry. Doesn't write a material file yet (having trouble grokking that format) and all drawable's are written using their highest LOD. Special reminder: Blender ignores OBJ vertex normals.

Example uses:

* Send scene geometry to Houdini to do voodoo like flowmaps and such
* Use to aid in modeling a "fitted" model 
* Author animation for the actual space (ie. dreaded ladder rungs)
* Run analysis tools on it (vertex density "hotness," etc)
* Use for offline rendering
* Not great for combining geometry into a single static geometry (bad workflow for that), but can be used for it
* etc

I haven't been able to come up with a reason for this to be used outside of the Editor scenario, so there's no bindings for Lua. That would require refactoring the functions out to someplace else. I don't have a problem with making those changes though, just don't see a reason.

### DIFF
--- a/Source/Urho3D/AngelScript/GraphicsAPI.cpp
+++ b/Source/Urho3D/AngelScript/GraphicsAPI.cpp
@@ -1837,7 +1837,7 @@ static CScriptArray* OctreeGetDrawablesFrustum(const Frustum& frustum, unsigned 
     PODVector<Drawable*> result;
     FrustumOctreeQuery query(result, frustum, drawableFlags, viewMask);
     ptr->GetDrawables(query);
-    return VectorToHandleArray<Drawable>(result, "Array<Node@>");
+    return VectorToHandleArray<Drawable>(result, "Array<Drawable@>");
 }
 
 static CScriptArray* OctreeGetDrawablesSphere(const Sphere& sphere, unsigned char drawableFlags, unsigned viewMask, Octree* ptr)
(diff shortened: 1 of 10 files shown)


cadaver:
IMHO it's somewhat dangerous to put significant unique functionality to the AngelScript bindings implementation. The files might go away if we ever automatize the bindings. The Texture2D::GetImage() was OK since it was something you could do in C++ with a few lines when you can do your own GetData() to it, but this is talking about over a hundred lines.

I'd suggest rather refactoring these functions into Drawable (export itself) and Octree (export every object in it)

299299:
add a new api to scene ？ something like "scene.exportToObj(const Array<uint>& export_node_ids) ? ", so what ever lua or angel script can both share this function.

cadaver:
Putting the whole scene export in Scene is fine too. Octree would just have Drawables directly.

aster2013:
May be write another tool use Assimp to export Urho3D scene or model to Collada, OBJ ect. is a better solution.

JSandusky:
Dug into assimp, it's output format support is quite limited and its' Collada output support is the bare minimum so I don't really think that it'd be worth the effort and usage hassle.

I'm going to go with the middle of the road, scene asking Octree for all drawables when doing whole scene export and a utility method for writing a list of drawables to OBJ (that the scene method would call) to cover both bases.

As to physics shapes, bullet already has it's own serialized output format and unless I'm mistaken there's plugins for most modeling packages to load that in different capacities. I'd think that'd be the best route for a physics shape export.

friesencr:
I was going to use a uv unwrap/packer tool that supported a 'multi obj' file.  I don't know if it has a specification or not.

cadaver:
Assimp is also a huge dependency (you can see that AssetImporter is roughly as large as Urho3DPlayer even though it doesn't include the whole engine) so I certainly agree with not using it for this purpose. The engine itself shouldn't be made to depend on it, and a new tool (or even expanding AssetImporter) just for this task would be clumsy, as it would require the input scene as a file, instead of being able to dump an in-memory scene.

JSandusky:
Sorry about that, rebasing didn't it cut it for resolving the conflicts with the renaming of Script to AngelScript.

weitjong:
Sorry about that. The renaming is necessary to simplify some of the construct in the CMakeLists.txt for Clang-tools development work. It should be the only thing that would have direct impact on library user's code that uses AngelScript for scripting, or so I hope. About the rebase, I have done more complicated rebase before in the past and I can say from my experience that it should be doable to resolve a conflict caused by directory/file renaming using git in CLI.

cadaver:
Merged.

PR #897, opened 2015-09-16
Update assimp

cadaver:
This updates assimp to head revision from 15th September ( https://github.com/assimp/assimp/commit/e159728307f17888b8b3ab4f23cf2143562c1c10 )

Testing with real-world data would be appreciated. If there's a case where Assimp *still* doesn't work correctly, we can't do much, but I'd like to hear at least if something used to work but stopped working with this new version.

On a technical front, assimp no longer seems to use PCH. I didn't attempt to bring this back in.


Source/ThirdParty/Assimp/CMakeLists.txt
```diff
@@ -24,702 +24,728 @@ ADD_DEFINITIONS( -DASSIMP_BUILD_BOOST_WORKAROUND )
...
+  SET( SOURCE_FILES ${SOURCE_FILES} ${C4D_SRCS})
+  INCLUDE_DIRECTORIES(${C4D_INCLUDES})
+ELSE (ASSIMP_BUILD_NONFREE_C4D_IMPORTER)
+  add_definitions(-DASSIMP_BUILD_NO_C4D_IMPORTER)
+ENDIF (ASSIMP_BUILD_NONFREE_C4D_IMPORTER)
 
 set (TARGET_NAME Assimp)
-list (APPEND SOURCE_FILES ${CONTRIB_FILES} ${PUBLIC_HEADERS} ${COMPILER_HEADERS})
-
 set_source_files_properties (contrib/clipper/clipper.cpp contrib/irrXML/irrXML.cpp PROPERTIES NO_PCH TRUE)
```
weitjong: In that case we don't need this line as well.

OldSnake22:
So far so good, no issues on my side.

PR #901, opened 2015-09-18
Add missing teleport ability to CrowdNavigation sample

EgorBo:
It said: "CTRL+LMB to teleport main agent\n" but nothing happened when I clicked it.


cadaver:
Thanks for your contribution, however note that the functionality is consistently missing from all versions of the sample (C++, AS, Lua.) Teleporting is already demonstrated in example 15 so the easier solution is just to remove the misleading text.

PR #907, opened 2015-09-28
Update Angelscript to 2.30.2

JSandusky:
This pull request updates Angelscript to 2.30.2.

All angelscript samples have been checked. Editor functions but is likely not covered in entirety.

**Notes:**
* Binding ref/value casts has changed, will effect any custom bindings that do not use APITemplates methods
* new protected scope
* private scope changed, now truly private (was previously really protected)
* anonymous methods
* bug-fixes


cadaver:
Thanks. Will do a manual verify for Android & iOS to double-check the native calling convention hasn't been broken, then merge.

JSandusky:
There's a typo in 40_Localization.as of using Color::BLACK instead of BLACK, which just appears to be something overlooked in porting C++ to angelscript. 

It's a single line fix (compile error), would you like that added to the PR? Trying to avoid adding anything extraneous to it, CrowdManager binding fixes were because the bindings led to crashing, so they were fairly vital.

cadaver:
I've already merged the PR internally so it's easiest if I just correct that before pushing. Thanks for notifying.

cadaver:
Merged.

1vanK:
What is the problem of Color::Black? I seen CrowdAgentState::CA_STATE_INVALID in 39_CrowdNavigation.as

PR #923, opened 2015-10-05
Add support for detecting ETC1 and PVRTC extensions in Emscripten/WebGL.

juj:


cadaver:
This alone doesn't unfortunately enable ETC1 / PVRTC, you still need to patch Graphics::GetFormat() to return a nonzero value. Currently these are only ifdeffed in for GLES2, using its extension-specific format enums.

PR #930, opened 2015-10-08
Bullet 2.83.6

### LINKED ISSUE #929: Consider upgrading to Bullet 2.83.6?
As part of investigating https://github.com/bulletphysics/bullet3/issues/482 , I upgraded Urho3D Bullet from 2.82 to 2.83.6. I could submit that as a pull request, but before I do so, has there been some specific reason to stick to Bullet 2.82? (it's from October 2013) Would a PR to upgrade to Bullet 2.83.6 be desirable?

juj:
This resolves #929.

Cherry-picked the changes from git history to Bullet. The first commit in this pull request is a baseline, which only contains files from upstream bullet and replaces all contents in the previous Source/ThirdParty/Bullet/ directory altogether. The remaining commits play back the different local changes I found either marked with `// Urho3D` in the source tree, or found with git history of Source/ThirdParty/Bullet/.

I do not know if this PR might bring in new regressions from upstream Bullet. Let's hope not. :/


cadaver:
Thanks!

cadaver:
This is missing some of the local mods (constraint breaking threshold infinity fix, apply gravity on each substep) and some of the notes of files being modified, which is needed by the license. Will probably be easier if I close this and update independently; that way I'll also take full responsibility if something breaks.

juj:
Oh sorry, I must have misinterpreted that those fixes would have already existed in upstream and no action would have been needed on them. Thanks for finishing this in the other commit!

PR #933, opened 2015-10-09
ANGLE_instanced_arrays

juj:
Add support for GPU hardware instancing on Emscripten. In WebGL 1 it is exposed by the ANGLE_instanced_arrays ( https://www.khronos.org/registry/webgl/extensions/ANGLE_instanced_arrays/ ) extension, and in WebGL 2, it is enabled in core.

### DIFF
--- a/Source/Urho3D/Graphics/OpenGL/OGLGraphics.cpp
+++ b/Source/Urho3D/Graphics/OpenGL/OGLGraphics.cpp
@@ -69,6 +69,18 @@
 #define glClearDepth glClearDepthf
 #endif
 
+#ifdef __EMSCRIPTEN__
+// Emscripten provides even all GL extension functions via static linking. However there is
+// no GLES2-specific extension header at the moment to include instanced rendering declarations,
+// so declare them manually from GLES3 gl2ext.h. Emscripten will provide these when linking final output.
+extern "C"
+{
+    GL_APICALL void GL_APIENTRY glDrawArraysInstancedANGLE (GLenum mode, GLint first, GLsizei count, GLsizei primcount);
+    GL_APICALL void GL_APIENTRY glDrawElementsInstancedANGLE (GLenum mode, GLsizei count, GLenum type, const void *indices, GLsizei primcount);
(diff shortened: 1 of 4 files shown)


Source/Urho3D/Graphics/OpenGL/OGLGraphics.cpp
```diff
@@ -785,7 +797,7 @@ void Graphics::Draw(PrimitiveType type, unsigned indexStart, unsigned indexCount
 void Graphics::DrawInstanced(PrimitiveType type, unsigned indexStart, unsigned indexCount, unsigned minVertex, unsigned vertexCount,
     unsigned instanceCount)
 {
-#ifndef GL_ES_VERSION_2_0
+#if !defined(GL_ES_VERSION_2_0) || defined(__EMSCRIPTEN__)
```
juj: Note: Before this pull request, I found it odd that the function `Graphics::DrawInstanced()` existed even when `GL_ES_VERSION_2_0` is defined, but it compiled down to an empty function body. That could possibly lead to silent failure behavior if some code accidentally calls the no-op `Graphics::DrawInstanced`? It might be more desirable to completely `#ifdef` out the whole `Graphics::DrawInstanced()` function from existing in the first place when instancing is known to be not supported at compile-time. This is probably a non-concern when/if instancing support is added to GLES2 in the future.

Source/Urho3D/Graphics/ShaderPrecache.cpp
```diff
@@ -119,7 +119,11 @@ void ShaderPrecache::LoadShaders(Graphics* graphics, Deserializer& source)
 
         // Check for illegal variations on OpenGL ES and skip them
 #ifdef GL_ES_VERSION_2_0
-        if (vsDefines.Contains("INSTANCED") || (psDefines.Contains("POINTLIGHT") && psDefines.Contains("SHADOW")))
+        if (
```
juj: Found this a bit clumsy, but couldn't figure a cleaner way without introducing annoying code repetition..

PR #935, opened 2015-10-09
sse_matrix

juj:
SSE optimize matrix-matrix multiplications when URHO3D_SSE is enabled. These constitute 7.82% of 06_SkeletalAnimation sample with 2000 Jacks in it. After SSE optimizations applied, matrix multiplication takes only 2.13% of total time in AMD CodeXL profile. This change also seems to have the effect that in scalar version, VS2015 did not dare to inline Matrix3x4 multiplication in AnimatedModel::UpdateSkinning(), but after SSE, the instruction count is small enough that it happily does so.

The code uses explicit unaligned loads and stores as opposed to aligned loads, so that memory layout does not need to change.

Only tested on 06_SkeletalAnimation sample, which uses  `Matrix3x4 operator *(const Matrix3x4& rhs)`. It is possible that the other variants are not tested. Tried to carefully check that the conventions are the same and that the functions are correct - they're adapted from MathGeoLib where they are unit tested.

### DIFF
--- a/Source/Urho3D/Math/Matrix3x4.h
+++ b/Source/Urho3D/Math/Matrix3x4.h
@@ -24,6 +24,10 @@
 
 #include "../Math/Matrix4.h"
 
+#ifdef URHO3D_SSE
+#include <xmmintrin.h>
+#endif
+
 namespace Urho3D
 {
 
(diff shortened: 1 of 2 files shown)


juj:
Thanks! I'm looking at some optimizations, profiling first, mainly with the motivation of seeing how far the Emscripten builds can be optimized when running in a web browser. AMD CodeXL and Gecko Profiler do give a few other arithmetic operations moderately up in the list as well, and perhaps those could benefit from some SIMD as well.

PR #936, opened 2015-10-09
approx_quat_slerp

### LINKED ISSUE #914: Optimization opportunity: Math.acos() and Math.sin() are relatively slow in Quaternion::Slerp() on Emscripten.
Looking at a profile of 06_SkeletalAnimation with the number of skinned models increased to 2000, it looks like browser built-ins `Math.acos` and `Math.sin` called by `Quaternion::Slerp` take up a total of 11.6% of total CPU time spent.

Unfortunately it is difficult to optimize those functions in browser side, one issue being that the browser's hands are tied due to precision guarantees, and that the generic variant must assume unbounded input number, and do range reduction on the input.

On the engine side however, it would be possible to do better manually by exploiting the fact that in slerp, `acos` is ever called with a number in the range `[0,1[`, which allows writing a custom version (LUT, or Lagrange interpolating polynomials are common choices) that's faster for this specific purpose.

juj:
Optimize Quaternion::Slerp() to remove the costly trigonometric functions and replace them by approximations instead. The approximations have been profiled to generate a maximum interpolation error of 0.38 degrees (see test Quat_Slerp_precision in MathGeoLib). In native 64bit Windows 8.1 VS2015 builds, testing in 06_SkeletalAnimation sample with 2000 Jacks, the time taken in Quaternion::Slerp() is reduced by -17.4%, and in Emscripten build of the same sample running in Firefox 41, the time taken by Quaternion::Slerp() is reduced by -44.7%. Closes #914.

### DIFF
--- a/Source/Urho3D/Math/Quaternion.cpp
+++ b/Source/Urho3D/Math/Quaternion.cpp
@@ -238,31 +238,36 @@ Matrix3 Quaternion::RotationMatrix() const
 
 Quaternion Quaternion::Slerp(Quaternion rhs, float t) const
 {
-    float cosAngle = DotProduct(rhs);
-    // Enable shortest path rotation
-    if (cosAngle < 0.0f)
+    float angle = DotProduct(rhs);
+    float sign = 1.f; // Multiply by a sign of +/-1 to guarantee we rotate the shorter arc.
+    if (angle < 0.f)
     {
-        cosAngle = -cosAngle;
-        rhs = -rhs;
+    angle = -angle;
+    sign = -1.f;
     }
 
-    float angle = acosf(cosAngle);
-    float sinAngle = sinf(angle);
-    float t1, t2;
-
-    if (sinAngle > 0.001f)
+    float a;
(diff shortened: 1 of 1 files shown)


Mike3D:
Very interesting :+1:
Do you have knwoledge in inverse kinematics?

juj:
Thanks - I have written a Cyclic Coordinate Descent and a Jacobian Transpose solver some ~7 years ago, been a long time since I worked on that. No doubt the same type of slerp operation is useful there as well, this could speed up some of Urho3D ragdoll, if it implements them manually (did not check).

cadaver:
For consistency, as the math code doesn't use approximations elsewhere, I'll #ifdef __EMSCRIPTEN__ this. Users can of course choose to remove it and use the optimized version regardless of platform.

juj:
Yeah, that sounds good. I was pondering whether to do this Emscripten only, since the performance impact is most in Emscripten side, but then thought to propose it unconditionally for determinism across platforms.

PR #938, opened 2015-10-10
opt_node_markdirty

### LINKED ISSUE #931: 42.73% of time is spent in Node::MarkDirty() in 06_SkeletalAnimation sample.
Profiling a native Win64 build of `06_SkeletalAnimation` in AMD CodeXL, with the modification that the scene consists of 2000 animated Jack models, the profile looks slightly surprising:

06_skeletalanimation

Most of the runtime is spent in marking scene nodes dirty. Taking a closer look, the function profile gives this view:

06_skeletalanimation_2

which indicates that the weak pointer handling is causing cache misses. What is the listener mechanism used there for? Skinning <-> Scene node interaction? or Skinning <-> Bullet interaction? Or something related? Are weak pointers necessary?

(Sidenote: how large can one expect the `listeners_` array to grow? The `listeners_.Erase(i);` behavior in this function has a O(n^2) complexity when erased in this manner, since `Erase()` shifts all elements back by one, i.e. calling `MoveRange`. although in the captured profile, that does not seem to be an issue here. Using the swap-with-last-element-and-erase-last trick could be used to avoid the quadratic complexity if the order of elements in `listeners_` is not important)

In the `12_PhysicsStressTest` sample with 3000 boxes, the `Node::MarkDirty()` function does show up in the profile as well. Some mitigating strategies if the performance is a concern:
   - prefetch the weak pointers in the iteration (if `listeners_` is long)
   - pool the weak pointer control blocks and/or listener objects
   - implement `Node::MarkDirty()` as iterative instead of recursive, since some time (10% in the profile) is spent in recursively calling child nodes
   - remove the weak pointer dependency and make it a strong ref dependency, synchonously/imperatively maintaining pointer removals on both sides of the link?

Not sure if this is a hotspot in any real world scenes, it is possible that 2000 Jacks is just a contrived scenario (feel free to close if so!), but when looking to replace 10kCubes with 10kJacks... ;)

juj:
Optimize `Node::MarkDirty()` implementation by

a) relying on the invariant that all children of a dirty node must also be dirty, and that all parents of a clean node must also be clean,
b) avoid double-dereferencing Component WeakPtr in listeners,
c) use the swap-with-last erase trick to avoid O(n^2) behavior in removing listeners, and
d) perform tail call optimization to avoid excessive recursive function calls when marking children dirty.

This optimizes the `Node::MarkDirty()` time in 06_SkeletalAnimation scene with 2000 Jacks from 42.73% down to 16.43%, a net saving of -61.5% less time spent in that call site. After this, `Node::MarkDirty()` is no longer the hottest function in the profile, so can say this closes #931.

### DIFF
--- a/Source/Urho3D/Scene/Node.cpp
+++ b/Source/Urho3D/Scene/Node.cpp
@@ -558,23 +558,50 @@ void Node::SetOwner(Connection* owner)
 
 void Node::MarkDirty()
 {
-    dirty_ = true;
-
-    // Notify listener components first, then mark child nodes
-    for (Vector<WeakPtr<Component> >::Iterator i = listeners_.Begin(); i != listeners_.End();)
+    Node *cur = this;
+    for (;;)
     {
-        if (*i)
-        {
-            (*i)->OnMarkedDirty(this);
-            ++i;
-        }
-        // If listener has expired, erase from list
-        else
-            i = listeners_.Erase(i);
-    }
+        // Precondition:
+        // a) whenever a node is marked dirty, all its children are marked dirty as well.
+        // b) whenever a node is cleared from being dirty, all its parents must have been
+        //    cleared as well.
+        // Therefore if we are recursing here to mark this node dirty, and it already was,
+        // then all children of this node must also be already dirty, and we don't need to
+        // reflag them again.
+        if (cur->dirty_) return;
+        cur->dirty_ = true;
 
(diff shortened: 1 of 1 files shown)


cadaver:
Thanks! We used to have the behavior of not dirtying if already dirtied, but it was removed for the consistent behavior of the TransformChanged() script callback. However the speed gain from this optimization can not be ignored. I'll just add a note that the callback should read the world transform to clear the dirty flag and allow the next dirtying callback to trigger.

cadaver:
Merged.

PR #942, opened 2015-10-10
unsynchronized_jacks

juj:
Make Jacks not walk in synchronized lockstep timing in 06_SkeletalAnimation sample.


cadaver:
It would be preferable to make the same change in AngelScript & Lua versions of the samples too.

juj:
Thanks, updated. However I don't know how to run the AngelScript and Lua versions, does the API look correct for them?

JSandusky:
Just tested angelscript version. It's correct.

cadaver:
To run the scripts: urho3dplayer <path_to_scriptfile>. Both versions seemed to be OK.

PR #946, opened 2015-10-12
hot_sse

juj:
SSE optimizes hot arithmetic functions in 06_SkeletalAnimation sample with 10000 Jacks in it. This improves the performance of that sample by roughly +19%.

### DIFF
--- a/Source/Urho3D/Math/BoundingBox.cpp
+++ b/Source/Urho3D/Math/BoundingBox.cpp
@@ -126,6 +126,38 @@ BoundingBox BoundingBox::Transformed(const Matrix3& transform) const
 
 BoundingBox BoundingBox::Transformed(const Matrix3x4& transform) const
 {
+#ifdef URHO3D_SSE
+    __m128 minPt = _mm_set_ps(1.f, min_.z_, min_.y_, min_.x_);
+    __m128 maxPt = _mm_set_ps(1.f, max_.z_, max_.y_, max_.x_);
+    __m128 centerPoint = _mm_mul_ps(_mm_add_ps(minPt, maxPt), _mm_set1_ps(0.5f));
+    __m128 halfSize = _mm_sub_ps(centerPoint, minPt);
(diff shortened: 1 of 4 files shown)


MonkeyFirst:
cool, thanks!

10000 Jacks = 19%
1000 = 1.9
100 = 0.19
10 = 0.019
1 = 0.0019
am right ?

Is it also speed-up octree nodes update calculation ?

cadaver:
It's a pleasure to merge these PR's :)

juj:
Thanks!

@MonkeyFirst : Are you trying to estimate what the speedup might be for skinning a single Jack? 19% of 10000 is still 19% of, say, 100, or 19% of 1, so percentages do not scale down like that. However absolute times do, and when in scalar version before this PR, a scene with 10000 Jacks would take 102.01 milliseconds to render, and after this PR, it took 86.6 msecs to render, the improvement being -15.41 msecs per frame on 10000 Jacks, it means that on 1000 Jacks, the theoretical best improvement is -1.541 msecs per frame, and on 100 Jacks, it is -0.1541 msecs per frame. Not much of a change unless one is doing big crowd simulations, but gains often come in small increments!

Now, I say "theoretical best improvement" since the workload doesn't scale linearly with the number of Jacks, but only affinely at best (i.e. `timePerFrame = someFixedCostPerFrame + numberOfJacks*timeToProcessOneJack` as opposed to `timePerFrame = numberOfJacks*timeToProcessOneJack`), so with smaller workloads, `someFixedCostPerFrame` starts to trump and the gains end up being smaller.

After this PR and a few other changes that will be separate PRs, the profile of 10000 Jacks looks like follows:

urho3d_profile

That profile is done with `-DURHO3D_THREADING=0` to streamline the work that is done. Octree nodes processing (`Urho3D::Octree::Update()`) does not show up there as a hot function, but instead in the top functions, cache misses are dominating the performance. I did some pipeline profiling on those with Intel VTune, however they seem to be quite tough ones to crack at the moment.

PR #948, opened 2015-10-12
allow_enable_sse_on_emscripten

juj:
Enable setting -DURHO3D_SSE=1 on Emscripten, but leave it disabled by default.


juj:
Thanks!

PR #949, opened 2015-10-13
no_boundingbox_bool

juj:
This PR shows how to get rid of the `defined_` boolean in `BoundingBox` class, and instead use `min = INF` and `max = -INF` denote the case when a bounding box is degenerate. This helps reduce branching and SSE'ifying BoundingBox. Using this scheme, it might be possible to further reduce `if (box.Defined())` checks in the codebase, since given how the math falls out, such arithmetic can naturally yield a desired negative result (no hit, outside, etc.) depending on the case.

Also SSE optimizes some hot BoundingBox functions and adds dummy padding floats to `BoundingBox` structure to help SSE loads and stores, which seems safe to do here.

There is also a small (pedantic) logic fix for `BoundingBox::Define(const Vector3* vertices, unsigned count)` when `count == 0`, which should probably define a degenerate box, instead of keeping the old one around(?).

The performance improvement here is rather small, about 3% on the 10k Jacks scene, and this carries an API change that extends to the scripting languages, so it is understandable if there is a concern merging this.

If this looks interesting, I noticed that classes `Rect` and `Sphere` also use such a `defined_` scheme, which could be removed using the same techniques.


cadaver:
Looks good. I don't think this is a problem interface -wise. Even if the performance gain is minor, it reduces memory footprint of drawable classes, and is therefore beneficial.

PR #950, opened 2015-10-13
sse_opt_quat

juj:
SSE optimizes functions in Quaternion.h.

### DIFF
--- a/Source/Urho3D/Math/Quaternion.h
+++ b/Source/Urho3D/Math/Quaternion.h
@@ -36,39 +36,59 @@ class URHO3D_API Quaternion
 {
 public:
     /// Construct an identity quaternion.
-    Quaternion() :
-        w_(1.0f),
+    Quaternion()
+#ifndef URHO3D_SSE
+       :w_(1.0f),
         x_(0.0f),
         y_(0.0f),
         z_(0.0f)
+#endif
     {
+#ifdef URHO3D_SSE
+        _mm_storeu_ps(&w_, _mm_set_ps(0.f, 0.f, 0.f, 1.f));
+#endif
     }
 
     /// Copy-construct from another quaternion.
-    Quaternion(const Quaternion& quat) :
-        w_(quat.w_),
+    Quaternion(const Quaternion& quat)
+#ifndef URHO3D_SSE
+       :w_(quat.w_),
         x_(quat.x_),
         y_(quat.y_),
         z_(quat.z_)
+#endif
     {
+#ifdef URHO3D_SSE
+        _mm_storeu_ps(&w_, _mm_loadu_ps(&quat.w_));
+#endif
     }
 
     /// Construct from values.
-    Quaternion(float w, float x, float y, float z) :
-        w_(w),
+    Quaternion(float w, float x, float y, float z)
+#ifndef URHO3D_SSE
+       :w_(w),
         x_(x),
         y_(y),
         z_(z)
+#endif
     {
+#ifdef URHO3D_SSE
+        _mm_storeu_ps(&w_, _mm_set_ps(z, y, x, w));
+#endif
     }
 
     /// Construct from a float array.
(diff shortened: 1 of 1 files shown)


amerkoleci:
Please take look here: http://codesuppository.blogspot.it/2015/02/sse2neonh-porting-guide-and-header-file.html
Would be nice to add ARM NEON instricts too.

juj:
Most of the SSE functions I have added in these PRs are taken from my MathGeoLib library ( https://github.com/juj/MathGeoLib ), where there exists also NEON variants of the same code. However, NEON is a completely different beast than SSE, because it has separate pipelines for ARM and NEON instructions, and a long pipelining latency in transferring data from NEON back to ARM. See http://infocenter.arm.com/help/topic/com.arm.doc.ddi0344k/ch16s05s02.html and http://infocenter.arm.com/help/topic/com.arm.doc.ddi0344k/Babffgig.html .

In general in SSE, it is quite safe to add a SSE variant of scalar code, and statically reason that it won't be any slower at least than the scalar counterpart. In NEON however this is a much harder problem because of the 20 cycle stall, and in general the experience with MathGeoLib has shown that one must be very judgemental about which code to run as NEON and what not to, since mixing a lot of ARM and NEON instructions can starve both pipelines when data is being often shuffled over. This has the effect that very short instruction sequences are usually not worth it for NEON, and for example, copying a quat over to another quat (as done here https://github.com/urho3d/Urho3D/pull/950/files#diff-e864d823c7004b7e6349c35262edc54bR62), generally loses performance in NEON, if the copied quat is immediately needed afterwards.

In MathGeoLib, I have taken an approach to generously implement identical code in NEON as has been done with SSE (and MGL has its own NEON-SSE abstraction layer at https://github.com/juj/MathGeoLib/blob/master/src/Math/simd.h, much like the linked John Ratcliff's page. However people have reported that sometimes the NEON path in ARM is slower than the scalar path, which is most likely attributable due to the pipelining latency.

In summary, adding any piece of NEON code calls for much more precise profiling than with SSE. My focus here is to experiment with Urho3D and SIMD.js in Emscripten, so I am not planning to visit NEON. If someone does have interest for this, feel free to experiment with taking NEON variants of the algorithms from MathGeoLib source tree, and profiling in different test cases to see what the performance effect is. I believe the contents of this PR would not be faster with NEON, whereas perhaps the code in #946 might be, since those deal with longer sequences of SIMD code at a time.

weitjong:
I have experimented with SSE2NEON header and added a new URHO3D_NEON option which when switched on will use the header to map the SSE intrinsic coded by Jukka to NEON equivalent. Of course this approach would not produce high quality NEON code, but on the other hand we don't have to maintain them because there aren't any. I have got into a point where everything built and run fine. However, I could not say I could detect any difference when running the SkeletalAnimation with and without NEON on my Galaxy Tab 3. I don't have the right instrumentation tool to measure/profile the impact. If I understand Jukka's point correctly then the naive/short NEON conversion will not benefit at all. I have tried to avoid that by only using the SSE2NEON mapping on block of longer SSE block. Still, without a way to instrument it, it is hard for me to push this changes into master branch. Having said that, I have one or two bug fixes for the build system in my current local repo in relation to SSE, so if I just push this to topic branch then I have to extra work to push those fixes to master directly.

@cadaver What's your opinion? Should I just push this to master now and let others who are more knowledgeable in this matter or has better tool to help to improve/fix the code incrementally.

weitjong:
I have an update. With some luck I am able to switch on the BT_USE_NEON on my Android build (had a few compilation and assembly errors but I managed to fix them, I think). However, I just could not tell the performance differences with my eyes. At the very least I can say it does not regress and it still runs :).

billyquith:
Have you seen [Ne10](https://github.com/projectNe10/Ne10) ? This does most of the heavy lifting for lots of ARM optimisation.

PR #953, opened 2015-10-14
Multi line

markomiz:
Not perfect yet, better text highlighting is needed in future as well as other little bits. SetMultiLine(bool enable) toggles multi line functionality.

### DIFF
--- a/Source/Urho3D/UI/LineEdit.cpp
+++ b/Source/Urho3D/UI/LineEdit.cpp
@@ -53,7 +53,7 @@ LineEdit::LineEdit(Context* context) :
     cursorMovable_(true),
     textSelectable_(true),
     textCopyable_(true),
-	multiLine_(false)
+	multiLine_(true)
 {
     clipChildren_ = true;
     SetEnabled(true);
--- a/Source/Urho3D/UI/LineEdit.cpp
+++ b/Source/Urho3D/UI/LineEdit.cpp
@@ -43,6 +43,8 @@ LineEdit::LineEdit(Context* context) :
     lastFont_(0),
     lastFontSize_(0),
     cursorPosition_(0),
+	cursorPosition_x(0),
+	cursorPosition_y(0),
     dragBeginCursor_(M_MAX_UNSIGNED),
     cursorBlinkRate_(1.0f),
     cursorBlinkTimer_(0.0f),
@@ -50,7 +52,8 @@ LineEdit::LineEdit(Context* context) :
     echoCharacter_(0),
     cursorMovable_(true),
     textSelectable_(true),
-    textCopyable_(true)
+    textCopyable_(true),
+	multiLine_(false)
 {
     clipChildren_ = true;
     SetEnabled(true);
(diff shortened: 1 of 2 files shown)


cadaver:
Tested quickly, basic operation seemed solid.

However it may make more sense to create a new class, for example TextEdit, and keep LineEdit as it is for simple one line editors. The multi-line editor should preferably inherit from ScrollView so that it can show scroll bars (both horizontal and vertical) for large text.

Also, note the code convention! Spaces instead of tabs, brackets on their own line.

JSandusky:
markomiz, would you mind if I picked this up (pushing PRs to your branch)?

markomiz:
Sure, keep me updated.

SirNate0:
Another possible name would be TextArea (from html), but I somewhat favor TextEdit in keeping with LineEdit.

Zamir7:
in Urhosharp from Xamarin upper version 1.0.410,  LineEdit on android is crashed, you know what reason? And multiLine is very needed option... thank you.

markomiz:
If anyone is interested I'm currently using this - MultiLineEdit class I made - I've added some functionality from the commit above. The files attached can be included in your project and don't have to be built with the rest of the game engine.

[src.zip](https://github.com/urho3d/Urho3D/files/339413/src.zip)

OldSnake22:
Hello markomiz, why don't you make a pull request with this new class ? Could be nice to have this in the master branch of Urho :)

Zamir7:
unfortunately I do not know c ++, and use Urhosharp (Urho on c#), i try to translate in C # this class) thank you for great work Markomiz!

weitjong:
The attached code seems to be a little bit dated. We now use key constants in lower cases.

markomiz:
weitjong, in addition to changing names of key constants, having spaces instead of tabs and putting brackets on their own lines are any other changes you would like me to make to make it suitable for a pull request? (I will also move the one function defined in .h file to .cpp)

weitjong:
I think I also agree with @cadaver regarding from where which it should be derived from.

markomiz:
I'm not sure I understand the benefit of inheriting from ScrollView when it is very easy to just set a MultiLineEdit as a content element of the ScrollView?

weitjong:
Very true. I suppose it is from the "ease of use" point of view.

m-roberts:
I must admit, seeing as nothing else appears to inherit from ScrollView, it seems to be that convention is to 'wrap' the scrollable element(s) within a ScrollView.

I'm not sure that it is inherently more sensible from an 'ease of use' point of view to change this in one instance.

cadaver:
Closed as superseded by #1443.

weitjong:
@m-roberts I beg to differ. See http://urho3d.github.io/documentation/HEAD/class_urho3_d_1_1_list_view.html.

cadaver:
ListView inherits from ScrollView. Inheriting ScrollView makes IMO sense from ease of use perspective when there is no anticipated usecase for having composition inside the content element. However, in case we guess wrong and there is indeed a weird usecase, then not inheriting is more flexible.

markomiz:
I use MultiLineEdit sometimes to have several sections of writing, some editable and some not editable, in the same scroll view. In this case I see it as making less sense to inherit from ScrollView. Admittedly though, this is not a very normal use case.

PR #956, opened 2015-10-15
sse_micro_opts

juj:
Optimize away some uses of _MM_TRANSPOSE4_PS(), to remove two shuffles, two movs and replace shufps with the lighter movehl/lh and unpacks. Also do vector loads and stores manually in BoundingBox::Transformed() to avoid VS2015 from generating dirty asm code for the function. This has a few percent improvement on the heavy BoundingBox::Transformed() function call.


cadaver:
Merged.

PR #962, opened 2015-10-17
fixed the wrong behavior of spriter animation

ijsonlin:
1, When we has only one texture in scml file, the SpriteInfo has wrong x and y values.
2, When we use SpriterPacker without the option "-trim", it will cause wrong frameWidth and frameHeight values
3, The newest version of Spriter use attribute name "object_type" instead of "type"


friesencr:
I wrote the sprite packer and you are right there are unhandled permutations where an uninitialized value would is bad.   Thanks for the pull request!
:+1:  for sprite packer fixes.

PR #965, opened 2015-10-19
Precise quat sse math

juj:


juj:
Thanks! There was a failed Travis CI run at https://travis-ci.org/urho3d/Urho3D/jobs/86140587 that pinged me now. It looks like the failure was due to a timeout of not being able to reach an Android device via ADB, and not related to the code itself.

weitjong:
It was a CI build "error" and not a "failure". Those are two different build statuses. The former indicates the technical error is from Travis CI side, while the latter is due to the software being built/tested. You can safely ignore the error.

PR #966, opened 2015-10-19
Support navigation obstacle transfer between scenes

JSandusky:
Despite the branch name (that's what I set out to deal with), there's nothing about detour crowd here.

Changelog:

* Obstacles seek upwards in the scene hierarchy to find their DynamicNavigationMesh rather than downwards from the scene root
* Obstacles now properly clear their ownerMesh_ when scene is cleared/reset, enabling moving them from one scene to another
* Addition of Node::GetParentComponent, Node::GetParentDerivedComponent with script bindings for Angelscript and Lua

### DIFF
--- a/Source/Urho3D/AngelScript/APITemplates.h
+++ b/Source/Urho3D/AngelScript/APITemplates.h
@@ -532,6 +532,11 @@ static Component* NodeGetComponentWithType(const String& typeName, bool recursiv
     return ptr->GetComponent(typeName, recursive);
 }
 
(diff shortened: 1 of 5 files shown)


cadaver:
Is it intentional that the obstacle uses false (default value) for the fullTraversal parameter, meaning it only searches one parent?

JSandusky:
Nope. That's a bug that didn't rear its' head while testing 39_CrowdNavigation.

PR #969, opened 2015-10-20
Add URHO3D_API specifier to Swap template function.

carnalis:


cadaver:
Have you encountered an error case from not having the export specifier? We have several other inline and template functions that don't have the specifier.

carnalis:
With Urho3D built as shared library, I get undefined reference to Swap<String, String>.

Perhaps there are better potential solutions:
Move the implementations into the headers, or explicitly instantiate the possible templates in the .cpp.

cadaver:
Thanks for the clarification. Yes, indeed the specializations would fail. In this case the potential better solution IMO is to export just those specializations. I'll do that.

PR #973, opened 2015-10-21
OBJECT -> URHO3D_OBJECT

1vanK:


cadaver:
For consistency's sake, I believe similar treatment should then also be given to EVENT, ATTRIBUTE, HANDLER etc. macros. Opinions?

1vanK:
Maybe I forgot something.

cadaver:
Thanks for the work! Found a couple more: the logging macros like LOGINFO, LOGWARNING etc. and IS_ASSET in File.h (that one is pretty much only internally used)

1vanK:
What about defined in CMake, like ASSET_DIR_INDICATOR?

cadaver:
Let's leave those for now, so that we don't break intricacies of the build. Having those clash is extremely unlikely.

cadaver:
I am ready to merge this imminently in a few hours so if any objections, please voice them now :) The logging macros become kind of verbose / unwieldy, but on the other hand it would not be surprising to conflict with some other code that also defines logging, and the consistency is nice.

PR #976, opened 2015-10-22
ResourceCache memory usage display in DebugHud

JSandusky:
Adds an additional table display that outlines:

* Resource type
* Number resources of that type loaded
* Average resource size
* Largest size of those resource (ie. size of the single biggest texture)
* Budget for that resource type
* Total memory usage for that resource type

Implemented in the samples as F3 key, to toggle on/off.

An image should explain it best:
http://i.imgur.com/pguB10g.png

### DIFF
--- a/Source/Samples/Sample.inl
+++ b/Source/Samples/Sample.inl
@@ -199,7 +199,21 @@ void Sample::HandleKeyDown(StringHash eventType, VariantMap& eventData)
 
     // Toggle debug HUD with F2
     else if (key == KEY_F2)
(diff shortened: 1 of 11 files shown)


MonkeyFirst:
Good, but I guess you forgot add the row of "total for all types". No?

cadaver:
Will implement the total row and do some minor refactorings. Unsigned long is a bit awkward type since it can be 32 bit on some platforms and 64 bit on some. Unsigned or unsigned long long are deterministic in that regard.

JSandusky:
I'll keep that in mind.

PR #992, opened 2015-10-29
Add function to control 2D physics update

cosmy1:


cosmy1:
Fixed

cadaver:
I would recommend creating a feature branch for each PR, and ensuring it doesn't contain merge commits (use rebase). In this case I'll probably cherrypick the actual feature.

cadaver:
Merged and implemented the same in PhysicsWorld.

PR #1004, opened 2015-11-02
Add missing world transformation in 2D particle emitter

cosmy1:


cadaver:
This breaks at least the 25_Urho2DParticle sample. The node world position is already taken into account when emitting new particles, so this would add it doubly.

cosmy1:
Yes you right; Noticed now. This fix a bug but introduce another one.

There is a bug with ParticleEmitter2D which you can see in the attached pic:
z

Basically here you can notice flowers are placed at z=0, particle = z=-2, animated coin z=-1
But particle ignore the Z axis and keep it on z = 0. My PR fix this but as you said introduce another one.

Here is my PR visually :
fix

cadaver:
Yes, the proper fix is to add proper Z coordinate math to ParticleEmitter2D. Closing this one.

Issue #713, opened 2015-04-30
WebP Textures


Skrylar:
How do we feel about supporting the loading of webp textures?

It seems possible to take just the decoding part of the reference implementation and hook it up to the existing image loading API, toss it in the third party source directory and hook up a "WANT_WEBP" flag in cmake.  I'll probably do this to my local copy regardless, but if we're okay merging this in to Urho I'll patch it in proper.

cadaver:
I have no strong feelings for or against, if it integrates cleanly a pull request would be nice.

Our CMake build flags begin with URHO3D_ so it should be for example URHO3D_WEBP. Though we don't conditionally compile other image formats support, so if the reference decoder is small (comparable to stb_image) and does not pull in other dependencies, a flag would not even be strictly necessary.

Skrylar:
Its a little bit bigger than stb_image (though stb_image is quite small, so that isn't saying much.)  I don't think it has any external dependencies, though.

cadaver:
In that case you can add the build flag but set it enabled by default. People with interest in absolutely minimum runtime size can then disable it, similarly like physics or Urho2D can be disabled.

carnalis:
This issue can be closed.

PR #1008, opened 2015-11-04
Add WEBP image format support

cosmy1:
Saw this as feature request so I decided to implement it.

### DIFF
--- a/Source/Urho3D/Resource/Image.cpp
+++ b/Source/Urho3D/Resource/Image.cpp
@@ -34,6 +34,11 @@
 #include <STB/stb_image.h>
 #include <STB/stb_image_write.h>
 
+#ifdef URHO3D_WEBP
+#define WEBP_EXTERN(type) type
+#include <webp/decode.h>
+#endif
+
 #include "../DebugNew.h"
 
 extern "C" unsigned char* stbi_write_png_to_mem(unsigned char* pixels, int stride_bytes, int x, int y, int n, int* out_len);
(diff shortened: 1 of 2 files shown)


cosmy1:
Tested with libwebp-0.4.4 binary version.
libwebp doesn't support CMake build system so I don't know how to deal with CMake scripts.

cadaver:
The library looks like quite huge for just the support of one (relatively little-known) image format, so I don't think I'd be comfortable including it in Urho3D in any case. 

In this case you could make Urho3D's build detect if the developer has libwebp (via CMake's find_package function) and if so, set the URHO3D_WEBP define and include the functionality.

cosmy1:
I don't know CMake scripting so if someone want to deal with it would be cool so this PR can be merged and close #713 .

JSandusky:
It'd be a nice courtesy to print the expected chunk ID when it fails to find WEBP (similar to "Sound"s declaration that it was trying to load a WAV file) or enforce the use of the webp extension for webp loading.

ANI and AVI are both valid uses of the RIFF container that are plausibly loadable into Image and RIFF is still pretty heavily utilized as a container for proprietary data as well. Logging the expected chunk ID will at least indicate that the wrong resource type was assigned or that something was overlooked.

cosmy1:
Ok. Now it prints expected and found header chunk ID.

cosmy1:
Closing this as is not a very common format for game development

Issue #1014, opened 2015-11-05
String::Split doesn't emit empty strings for start/end/adjacent separators


barrettcolin:
Reproducible with 1.4 (it's not convenient for me to try latest code); for example:

    Vector<String> splits = String::Split(";", ';');
    splits = String::Split(";;", ';');

In both cases 'splits' is empty. This pattern arises in an XML scene if a model has empty material slots (for instance, create a node with the ninja model in the editor (which has two material slots), pick a material for the *second* slot, save the scene and reload - the material has jumped to the first slot because of the Split behavior - the XML attribute looks like this:

`<attribute name="Material" value="Material;;Materials/NinjaSnowWar/Ninja.xml" />`

)

weitjong:
Some of the time the extra empty strings in the result is a nuisance. Some of the time though the empty string is expected, especially when the original string is pre-formatted. Another example for the latter use case is the recently revised resource prefix paths engine parameter. I have to use
```
" ;../share/Urho3D/Resources"
```
(note that a space character have to be inserted before the first semicolon) in order to get the expected "empty" string as the first split result.

Probably we need to Split() functions or just one function but with another extra boolean parameter to switch on/off the deterministic behavior.

barrettcolin:
From a quick test, both Javascript and Python split a string like ";" into two empty strings.

Stinkfist0:
Typically string splitting functions tend to have `removeEmptyStrings` bool or enum parameter which can be used to control this behavior. That would be good addition for Urho's string class too.

cadaver:
Practically the correct default behavior is that which is required for the material list serialization case. Then, like said, the possibility to omit the empty ones.

barrettcolin:
Yes, that sounds good

PR #1017, opened 2015-11-07
New String::Split implementation

barrettcolin:
Keeps empty splits by default, optionally omit-able using splitMode parameter - see #1014

### DIFF
--- a/Source/Urho3D/Container/Str.cpp
+++ b/Source/Urho3D/Container/Str.cpp
@@ -558,9 +558,9 @@ String String::ToUpper() const
     return ret;
 }
 
-Vector<String> String::Split(char separator) const
+Vector<String> String::Split(char separator, StringSplit splitMode) const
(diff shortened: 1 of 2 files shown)


barrettcolin:
This will break code that assumes the split result of an empty string is a Vector<String> of size 0 - the SPLIT_KEEP_EMPTY behavior in that case is to return a Vector<String> containing one empty string. At least the code handling 'resourcePackages' in Engine.cpp suffers from that.

cadaver:
From the perspective of API correctness it would be indeed nice to have the default value as it is. From perspective of existing code and verification that it works correctly, not so (especially that we have a release coming :) ), so will reverse the default value and only use the keep empty -mode in the identified problem spots (prefix path, resource ref lists)

barrettcolin:
That sounds fine. Shall I leave it with you?

PR #1020, opened 2015-11-08
Add protected check to script variable population

JSandusky:


JSandusky:
Checking up on this, I'm not sure why Travis never launched that one remaining task. If there's some sort of system message about it an admin received it'd be great to know if so.

weitjong:
I have force-started it again. Sometime a CI build may encounter an internal error and Travis may kill the overall build job and re-queue it. I don't know why for your case, the job just stays in the queue.

PR #1021, opened 2015-11-08
Turbo Badger integration with demo

Lumak:


Lumak:
Still trying to fix build errors for non-Windows build.

weitjong:
It actually failed on VS build too with Urho3D shared library type.

I just had a quick look on the PR. IMHO, it still needs some more cleanup to do even when it passes our battery of build tests. First of all, you should not assume everyone would want to build with TB. The usage of ```URHO3D_TB_DEMO``` build option does not go inline with our current design, it seems that it only controls whether the TB demo is included while the TB library itself is being included regardless.

Lumak:
I have only built vs2013 build on my PC but will take a look at the build log and see what needs to be corrected.  I may have a way to test on Linux tomorrow.
I actually wasn't sure if this integration is something that you wanted in Urho3D, but if I were to move the UTBDemo to the sample folder and replace URHO3D_TD_DEMO with URHO3D_TD to include TurboBadger library then would you actually accept this PR?

JSandusky:
You don't need to use linux to confirm. Just use MingW to compile for GCC. Occasionally issues will pop up it won't catch, but Travis will tell you enough about those.

Unfortunately your errors also suggest issues with your changes to the CMake commands.

Also you've "overrided" a non virtual method. "AddQuadInternal" is not virtual, you can't override that. In all honesty, your implementation is atrocious.

Lumak:
I changed AddQuadInternal() func to virtual in my original 1.4 integration.  Did I not do that for 1.5 to add the UV offset hack?  I'll have to take a look.

Lumak:
I think you're confused about this PR with my https://github.com/Lumak/Urho3D-1.4-TurboBadger integration, where I specifically note that the AddQuadInternal() func is virtual.
That's what you're looking at, right?

JSandusky:
Personally, I don't consider that an acceptable edit of an external library. You aren't dealing with an intrinsic issue, you're sidestepping around the library itself.

Lumak:
So you're talking about another repository which has nothing to do with this PR?

JSandusky:
Edits to libraries are made within reason. Your edit is not within reason.

So I solve your DX9 problem for you and you act flippant? You're effectively filtered and will receive nothing from me from this point on.

Lumak:
I'm confused as to how I was being flippant.  I have shown gratitude for your help with the DX9 issue and have nothing but respect for your contribution to Urho3D community as I have said in one of your forum threads.  Maybe I didn't convey that information well.  

But let's end this discussion, as it's getting off the topic about what I need to do about this PR.

JSandusky:
I agree, it is off topic. I'll apologize for any inappropriate tone. Keep it to the metal per se.

weitjong:
> I actually wasn't sure if this integration is something that you wanted in Urho3D, but if I were to move the UTBDemo to the sample folder and replace URHO3D_TD_DEMO with URHO3D_TD to build TurboBadger library then would you accept this PR? And I'm not sure what other "cleanup" would be.

I just had a quick look as I have commented earlier. I did not try to build or test it yet. On the surface I can only tell the glaringly obvious things to me. The usage pattern of build option is one thing. The way the include dirs are being setup for this 3rd party lib is another, also as of 1.4 we have started to use angle bracket include for the libs. I didn't look into C++ implementation yet, so I have no comment on that.

What I have just said is my personal opinion. As usual please take it with a grain of salt. Lasse is the one that decides the merge-readiness of a PR. I guess we are all entitled for an opinion. Just stay cool until you hear from him.

BTW, thanks for the PR.

BlueMagnificent:
@Lumak  Because a PR might not be merged does not mean its worthless, the idea and effort put in are great. You shouldn't have deleted the repository, nonetheless, I'm glad I got a copy before it was removed

yushli:
So this PR to add turbobadger support is going on or not? It is nice to have at least the option to enable turbobadger. Lumak is contributing to the Urho3D project. So if the implementation needs improvement, just improve it. He simply should not be treated like that. In all honesty, some comments in this thread is atrocious.

Lumak:
I thought this integration could be beneficial for the community and the reason for my effort. I had just fixed the turbo badger lib, similar to how box2d done it, when I took a break from that and saw this:
http://urho3d.prophpbb.com/post9022.html#p9022
I decided that I don't have to contribute anything to the community.
I will continue to use the engine bc I think it's great. Good luck.

yushli:
Hi, Lumak, Please don't go away and delete your integration of turbobadger to Urho3D because of that. I personally would really appreciate to have it enabled in Urho3D.

weitjong:
> @Lumak Because a PR might not be merged does not mean its worthless, the idea and effort put in are great. You shouldn't have deleted the repository, nonetheless, I'm glad I got a copy before it was removed

I second to that. The PR could still be worked on until it is ready to be merged. For those that want to check out the work, you can still get it from Urho3D repository: 
```
git fetch origin pull/1021/head:TB_integration_Lumak
git check out TB_integration_Lumak
```
This assumes ```origin``` is the remote pointing to Urho3D remote repository.

JSandusky:
Even as the asshole that went full on 'roid rage on you (quite literally, it's really hard being a programmer + meathead), keep it around (you nuked it, but I'm sure you kept a local copy) - some people will want that and options are good to have even if they're of the same library.

Angelscript debugging was something that never made it in (maintenance burden too high), but has been used far more than I ever expected (get a ton of PMs about it, especially when Urho3d/master changes in ways that break it ... someone actually tracked down my phone number a couple months ago). 

Back then I didn't do much differently than you now (deleted that initial version) only that I was absolutely infuriated at GCC's default of C99 standard.

You didn't deserve that. You really didn't, and even when I chimed in to help out, it was because I saw effort. In raging at you I basically corrupted everything I did in helping you.

PR #1032, opened 2015-11-11
Urho3D Editor: Spawn on selected node in Spawn Editor

VladimirPobedinskiy:
Image
It allow to spawn only on selected node, ex. grass on terrain through
 trees, also not spawn on previous inserted nodes.


cadaver:
It would be preferable to use feature branches for PR's, because a master branch feels like something that might at any time be updated with more work which may not be related. Not saying you would do that, but you're placing quite a restriction on your fork with this workflow.

VladimirPobedinskiy:
Ok.

PR #1041, opened 2015-11-14
Build SDL with PulseAudio if available

evolarium:
Using ALSA sound on Linux results in 100% CPU usage for one of the threads.  PulseAudio behaves better on my machine.  The sample 14_SoundEffects goes down to around 20% CPU usage when PulseAudio is used instead.

### DIFF
--- a/Source/ThirdParty/SDL/CMakeLists.txt
+++ b/Source/ThirdParty/SDL/CMakeLists.txt
@@ -22,6 +22,7 @@
 
 # Define target name
 set (TARGET_NAME SDL)
+include (FindPkgConfig)
 
 if (EMSCRIPTEN)
     # Emscripten has its own SDL2 port, so just create a custom target to:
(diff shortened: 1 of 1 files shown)


friesencr:
I had problems with ubuntu 15.04 with alsa.  It flooded my logs bad.  I think the fuss went away with 15.10.  It seems like a good idea to have some options on linux.

weitjong:
> It seems like a good idea to have some options on linux.

Totally agree to that. Although I have a perfectly working PulseAudio out the box from my Fedora distro, I think if you search the internet then you can easily find quite a number of PulseAudio distractors out there. In most modern Linux distros nowadays, the audio system has ALSA at the lower stack and PulseAudio at the higher stack (i.e. PA depends on ALSA). The distractor or those that for some reasons having problem with PulseAudio may opt to uninstall the PulseAudio stack and configure their sound applications to interface with ALSA directly. I am not one of them. It is always good to have more options to choose from.

cadaver:
Merged.

PR #1060, opened 2015-11-26
Improving Particle Editor

ucupumar:
I'm trying to improve particle editor with these changes:
- Adding grid with option to hide it
- Gizmo now always visible in the left corner
- Change particle preview viewport camera to behave more predictably

particle_editor_grid

There are some hacks while I implement this, like:
- Aspect ratio of preview viewport is always returning value of 1 in the first frame, so I decided to manually set aspect ratio.
- Camera node euler angle is returning vector of zero in first frame, so I decided to hardcoded global camera rotation variable in first frame.

I'm not sure if those are bugs, maybe it's because particle editor isn't immediately visible when starting Urho3D editor.

What do you guys think?


cadaver:
In auto aspect ratio mode it is normal for camera aspect ratio to be 1:1 before rendering for the first time (unless not manually set)

cadaver:
Changes look good. Some minor things which I can fix while merging: we generally don't use underscores in variable names, also prefer not leaving commented out lines, which creates the feeling as if one is testing things and not sure of the code.

cadaver:
Merged.

ucupumar:
Okay, thanks for fast merging. The comments I leaved are line of codes expected if hacks aren't necessary. You can always get rid of them. About underscores in variable names, my bad, it's because my habit in Python, you can change them too :D

PR #1077, opened 2015-12-08
Added instantaneous rigidbody collision check

hjmediastudios:
Added a `GetRigidBodiesInstantaneous` method (plus associated Lua/Angelscript APIs). This function is similar to `PhysicsWorld::GetRigidBodies` but actually performs an instantaneous collision check, instead of relying on cached collision data. This, for instance, allows for checks on static geometry that otherwise do not return the expected results.

### DIFF
--- a/Source/Urho3D/AngelScript/PhysicsAPI.cpp
+++ b/Source/Urho3D/AngelScript/PhysicsAPI.cpp
@@ -280,6 +280,13 @@ static CScriptArray* PhysicsWorldGetRigidBodiesBody(RigidBody* body, PhysicsWorl
     return VectorToHandleArray<RigidBody>(result, "Array<RigidBody@>");
 }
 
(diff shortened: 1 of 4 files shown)


hjmediastudios:
The Appveyor test seems to be failing because of `IsSimulating` not being a member function of `RigidBody`, not because of my pull request-related changes.

cadaver:
Thanks! I will meditate on this a bit, and may make changes on merging. Seems to me that GetRigidBodies() would be preferable to return an instant result (since it is always correct, and also because the other overloads also do instant tests) and then just make only RigidBody::GetCollidingBodies() return the last frame cached result, and document that difference.

cadaver:
Arguably the most proper fix would be to store also static and non-signaled collisions to the currentCollisions map, but this would reduce performance. So in that sense your solution is preferable. Will test a bit.

cadaver:
Adapted your code instead of merging, since your commits did something to the IsSimulating flag (possible bad merge?)

PR #1083, opened 2015-12-09
Made CrowdAgent emit events the attached Node as well

hjmediastudios:
Made CrowdAgent emit events from the attached Node as well as the
associated CrowdManager, to allow for simpler reaction of other
components on the same Node to crowd agent events.

### DIFF
--- a/Source/Urho3D/Navigation/CrowdAgent.cpp
+++ b/Source/Urho3D/Navigation/CrowdAgent.cpp
@@ -292,7 +292,7 @@ int CrowdAgent::AddAgentToCrowd(bool force)
             crowdManager_->SendEvent(E_CROWD_AGENT_FAILURE, map);
             Node* node = GetNode();
             if (node)
-                node->SendEvent(E_CROWD_AGENT_FAILURE, map);
+                node->SendEvent(E_CROWD_AGENT_NODE_FAILURE, map);
 
             // Reevaluate states as handling of event may have resulted in changes
             previousAgentState_ = GetAgentState();
@@ -529,7 +529,7 @@ void CrowdAgent::OnCrowdUpdate(dtCrowdAgent* ag, float dt)
             map[P_ARRIVED] = HasArrived();
             map[P_TIMESTEP] = dt;
             crowdManager_->SendEvent(E_CROWD_AGENT_REPOSITION, map);
-            node_->SendEvent(E_CROWD_AGENT_REPOSITION, map);
+            node_->SendEvent(E_CROWD_AGENT_NODE_REPOSITION, map);
 
             if (updateNodePosition_)
             {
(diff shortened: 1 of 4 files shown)


weitjong:
Objects can already receive any events when they are registered as event subscriber of an event (i.e. non-specific event receiver), don't they?

hjmediastudios:
This is just so that a component can subscribe to the node so that it  can handle only agent changes associated with that node; not, for example, having a navigation-related component attached to a node that has to filter through 100+ events not related to its own node in order to perform an action related to a CrowdAgent component on that node.

weitjong:
I understand from where you are coming from. Currently all the crowd agent event types have the P_NODE parameter, so if your event handler for nodes is generic (it is the same for all the nodes) then I believe it is easy to use that node parameter to trigger action(s) only on the associated node or on its components. Your proposed changes would only make sense to me when the event handler is non-generic, i.e. you could have different event handlers for different nodes. And if this use case is more prevalent then I would probably just dispatch the event only once using node as the sender instead of dispatching it twice as proposed. This is just my personal opinion. As always, take it with a pinch of salt.

hjmediastudios:
You're right as the motivation, but I think in practice there's probably more use cases for a non-generic handler as opposed to a generic one. There are plenty of components that might need to react to changes in a single CrowdAgent- the impetus for this PR is the application I'm developing, which has a custom rigidbody-based AI character component that uses the CrowdAgent and needs to react to state changes and failures, which is a pain to deal with in the current case as it needs to filter through every other node's state changes in order to act on a lone node. A component attached to one node concerned solely with that one node, as in my case, does not need to filter out dozens of unrelated events, and the purpose of the event and component system is somewhat compromised when all classes like this need to hook into a central manager from the user's point of view. Unless I'm wrong, it seems to me that use cases like mine are more prevalent than those like the trivial examples presented in the samples.

weitjong:
You are correct to say that the crowd navigation demo that we have has a trivial logic. In fact it does not care which object is actually the event sender. But that also means that in any cases, the event should only be dispatched only once or otherwise for those simple use case then the event handler would be double-triggered.

The generic event handler I mentioned in my previous comment does not necessarily mean all the node logic would be have the same. For example, each node may have a ScriptInstance component. The event handler may invoke an action or pass another event or what have you to the node's ScriptInstance component in a generic manner. However, you can have different script loaded into the component and so although the event handler is generic, the behavior is still unique per node.

hjmediastudios:
I'm not saying that it can't be done to have unique behavior per-node; simply that making nodes emit events themselves in addition to the CrowdManager is the way to go. Consider the collision handling right now- it's possible for a node to subscribe to all collisions in the scene, but it's also possible to subscribe to collisions of only one node, as from a user standpoint it's pointless to filter out potentially hundreds of other nodes.

cadaver:
I would recommend using another event type for those that are sent from the node, that way there should not be confusion. Similar to PHYSICSCOLLISION (from physicsworld) and NODECOLLISION (from node) in physics. Sending the same type of event twice has potential to be fishy.

hjmediastudios:
Pun intended? Anyway, that's a good idea; I'll implement it a bit later today and see how that looks.

hjmediastudios:
Pun intended? Anyway, that's a good idea; I'll implement it a bit later today and see how that looks.

hjmediastudios:
All right! I separated the events into the existing events, which remain emitted by the CrowdManager, and node-specific events that have the same parameters but are emitted by the node only.

PR #1084, opened 2015-12-09
Urho3D Editor: Preview drag-moving rotation problem are fixed.

neat3d:
The preview drag-moving did not work properly while you re-click the mouse button to rotate an object, it always reset the rotation to zero.

Fixed...

### DIFF
--- a/bin/Data/Scripts/Editor/EditorMaterial.as
+++ b/bin/Data/Scripts/Editor/EditorMaterial.as
@@ -355,10 +355,10 @@ void RotateMaterialPreview(StringHash eventType, VariantMap& eventData)
     
     if (materialPreview.height > 0 && materialPreview.width > 0)
     {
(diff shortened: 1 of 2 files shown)


cadaver:
Didn't feel like much of an improvement compared to the original. I particularly didn't like the fast spinning when drag distance increased, therefore not merging this now.

neat3d:
The original version just can't not rotate properly.
For example, you can't rotate 30.0 degree first, and then rotate 20.0 degree again, you can't get 50.0 (30+20) degree rotation, the first step is good to work, but the second step would rotate back to 0.0 degree and rotate 20.0 degree, you only got 20.0 degree rotation.

And the spinning speed isn't slow down, it is the same as the original one, if I keep the parameter as 90.0, actually it will spin too fast.

PR #1093, opened 2015-12-13
Added PhysicsWorld::RaycastSingleSegmented()

Enhex:
By segmenting a ray into several sub-rays segments, the physics engine has to check for intersection with much less rigid bodies (I assume Bullet uses AABB test).
It has big performance gain with large scenes that have a lot of rigid bodies in them, and that's quite a general case.

In my use case it more than doubled game framerate.

ss 2015-12-13 at 03 57 03

It's missing Lua API.
A RaycastSegmented() would also be useful, and perhaps for the graphical raycast if it uses AABBs too.


cadaver:
Remember spaces (not tabs) in Urho code! Will apply this change and add missing Lua binding.

cadaver:
Merged.

PR #1094, opened 2015-12-14
Fix narrowing conversion from int to unsigned char in jo_jpeg

prshreshtha:
Xcode 7 would not compile Urho3D because of the narrowing conversion. Changes are semantically equivalent so I don't think it would break anything.


cadaver:
Interesting if that was the only change required.

cadaver:
Merged.

PR #1098, opened 2015-12-17
Variance shadow mapping

darwikey:


friesencr:
Thanks for your contribution!  There are merge conflicts with your pull request against our master branch.  Can you please update your branch?

Thanks,
Chris

darwikey:
Ok I will do that

darwikey:
I have merged, everything seems fine but Travis is not happy

weitjong:
I think on Travis it failed at test running the editor.  That's usually
caused by mistake in angelscript binding. Thanks for the PR.

MonkeyFirst:
Awesome! Thanks darwikey!

darwikey:
I don't understand  the utility of the file  "bin/Data/NinjaSnowWarShaders.xml". It seems to be used to precache some shaders, but shader defines depend of the hardware. For example with some hardware, SHADOWCMP will be add to the define list.

cadaver:
The precache file isn't obviously going to be valid for all situations. In that case shaders will be compiled instead of precached. However, its utility is greatest for OpenGL since Direct3D already supports binary shader loading.

cadaver:
Note: I'm doing independent fixes (for example properly fix the editor settings dialog) so I won't be taking those last commits. Looking good, will only need to test mobiles.

darwikey:
Ok perfect

cadaver:
Merged with adaptations.

1vanK:
Rotation of the camera by a few degrees
http://savepic.su/6822187.png
http://savepic.su/6871338.png

Same scene with old shader
http://savepic.su/6826285.png
http://savepic.su/6817069.png

Also new shader VERY laggy for point light. For which cases is a new shader?

1vanK:
In some foreshortening the shadow is missing
Scene:
https://raw.githubusercontent.com/1vanK/MyFiles/master/Temp/test%20new%20shadow.xml

cadaver:
VSM is a whole new can of worms, it's not automatically better working, but may need tuning. Probably in most cases I'd personally continue to use the old shadows. Hopefully @darwikey stays aboard to improve where possible.

darwikey:
In this screenshot : http://savepic.su/6826285.png, you can see raw big pixels. Maybe cascade parameters don't have the good values ? And so when you blur the shadow map, it will hide the shadow.

MonkeyFirst:
Is it use this sort of correction ?
Light Space Perspective Shadow Map (https://www.cg.tuwien.ac.at/~scherzer/files/papers/LispSM_survey.pdf )


also I think to solve this issue (on counters of shadows) probably need fix shader (Chebyshev) with adding this lines:
~~~
    // To combat light-bleeding, experiment with raising p_max to some power
    //p_max = pow( p_max, 20 );
~~~


~~~
float Chebyshev(vec2 Moments, float depth)  
{  
    //One-tailed inequality valid if depth > Moments.x  
    float p = float(depth <= Moments.x);  
    //Compute variance.  
    float Variance = Moments.y - (Moments.x * Moments.x); 

    float minVariance = cVSMShadowParams.x;
    Variance = max(Variance, minVariance);  
    //Compute probabilistic upper bound.  
    float d = depth - Moments.x;   
    float p_max = Variance / (Variance + d*d); 
    // Prevent light bleeding
    p_max = ReduceLightBleeding(cVSMShadowParams.y, p_max);
    
    // To combat light-bleeding, experiment with raising p_max to some power
    p_max = pow( p_max, 4 );
    
    return max(p, p_max);
}
~~~

darwikey:
Hum I will try this method. The function "ReduceLightBleeding" just clamps and interpolates p_max.

MonkeyFirst:
Yes, this pow kill this light bleeding.

Do you read this topic on gamedev,com? 
http://www.gamedev.net/topic/485290-extending-vsm---cheap-pcss--alternate-representation/

~~~
So I store depth and encoded = 4.0*(depth - depth^2). I think this gives me better precision over the entire range, but I'm not entirely sure, if any math gurus wish to verify this I would be most grateful! (it seems to work, so just thought I'd share). Then to recover:
 sigma2 = avg_depth - 0.25*encoded - avg_depth*avg_depth;
~~~

also (I do not know very well english but) I guess what he wrote what he store something in mips to math penumbra of shadow

darwikey:
Actually he used 16 bit buffers to store depth and squared depth, it's faster but it can cause imprecision. My implementation uses 32 bit floating point buffer and so I don't need to encode / decode depth.

MonkeyFirst:
Even with 32bit buffers you got some kind of shadow fighting in place where objects start cast his shadow on floor. Actually I found what floor plane or terrain also must be cast shadows(rendered into SM) to properly shadow rendering on it from other objects (above)

darwikey:
"you got some kind of shadow fighting in place where objects start cast his shadow on floor" -> you can play with the first value in Renderer::SetVSMShadowParameters(). It clamps the variance, by default it's equal to 0.0000001f, you can try to add or remove a zero.

darwikey:
I have tried to use p_max = pow( p_max, k ); to reduce lightbleeding but it's not very efficient. The two parameters of SetVSMShadowParameters() seem to be enough.

MonkeyFirst:
Where in editor I may to change this SetVSMShadowParameters() ?

darwikey:
Yep minVariance is not for lightbleeding.
Your function "smoothStep" is exactly the same as "ReduceLightBleeding" ;)

MonkeyFirst:
Ok I find this works   
 //cVSMShadowParams.y = 0.2f;
p_max = ReduceLightBleeding(0.4, p_max);

I think this should be part of light component parameters in attribute inspector and various directional light may have own settings of this

darwikey:
Yep blurred vsm uses a 7x7 gaussian blur and it can be a bit expensive. Maybe we can propose 3x3 blur or something like that.

You can easily implement your own blur for shadow map with :
void Renderer::SetShadowMapFilter(Object* instance, ShadowMapFilter functionPtr);

You can look at the function "void Renderer::BlurShadowMap(View* view, Texture2D* shadowMap)" to see how to implement a filter.

PR #1102, opened 2015-12-19
Editor: customizable default scene

1vanK:
Like Blender. It also looks more attractive for new users than black screen :)

PR #1107, opened 2015-12-21
Editor: customizable default scene 2

1vanK:
old revision: https://github.com/urho3d/Urho3D/pull/1102

### DIFF
--- a/bin/Data/Scenes/DefaultScene.json
+++ b/bin/Data/Scenes/DefaultScene.json
@@ -1,0 +1,143 @@
+{
+	"id": 1,
+	"attributes": {
+		"Name": "",
+		"Time Scale": 1,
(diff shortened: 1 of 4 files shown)


cadaver:
I don't agree with changing the camera startup angle or loading a scene by default at all, however I'm happy to include the scene you have provided.

1vanK:
Often the creation of a new scene begins from loading blank (for example a scene containing a customized lighting). if u need empty scene on start - just remove "Sample" node and press "Save scene as default". It also looks more attractive for new users than black screen. This approach is implemented in Blender. :)

MonkeyFirst:
видно не судьба ) есть повод призадуматься... 

mb try add this feature with empty default scene ?

PR #1112, opened 2015-12-22
added node tags

svifylabs:
Implemented a simple node tag system.
Nodes can set a stringhash as a tag and the scene caches the nodes with the same tags. If you want to get the string of the tag you have to register the name in the scene. 

Angelscript bindings are implemented.

TODO:
- Lua bindings
- Example


1vanK:
This method scans each node. When the scene is large it may be slow. It may be better to store the table. "tag -> list of nodes" ?

cadaver:
This is a nice feature. Didn't evaluate the performance characteristic in-depth, but I can say that defining the name as a StringHash attribute will make editing (in the editor) hard. I'd rather recommend just having the node tag be stored as a String attribute, and also store as hash when assigned.

This would make editing work directly, and make the whole RegisterTag / UnregisterTag mechanism + saving/loading the registered tags unnecessary. Obviously this would cause some more dynamic allocations, in case tags are used, but otherwise an empty string is just 12 bytes more for each node.

If you intend to continue using StringHash attribute, then this PR would also need the editing part implemented (registering, unregistering and choosing tags for nodes), either by you or by someone else.

Another way to sidestep the editing issue, and still avoid the per-node String, is to provide SetTag and GetTag String overloads, which are used for the attribute definition, and which register and unregister tags to the scene automatically behind the scenes. Meanwhile, hashes are still used for the queries.

svifylabs:
@1vanK you can use the scene for getting the cached nodes for a specific tag Scene::bool GetNodesWithTag(PODVector<Node*>& dest, StringHash tag)  const; that should be fast, but you get nodes for the hole scene ...
Or you can use the slower node function Node::GetChildrenWithTag(PODVector<Node*>& dest, StringHash tag, bool recursive = false) const; it does the same as GetChildrenWithComponent ... 

Is this what you had in mind ? 

@cadaver added a tagstring attribute into the node. you can now use Node::SetTagString and Node::GetTagString.

1vanK:
> Is this what you had in mind ?

Oh, yes. Sorry, I did not notice it :)

cadaver:
Thanks for the update. Will take a more deeper look in the near future.

1vanK:
Tags would also be useful for UIElements. Sometimes it is needed to show or hide a group of elements depending on the state of the game.

cadaver:
Now looking through this. For future reference, set your text editor to spaces instead of tabs and set it to show whitespace so you notice the difference (this seems to be common in pull requests)

cadaver:
Seems that you can "desync" the tag string and tag by calling SetTag() on your own, which doesn't update the string. It would be preferable to have a public API that doesn't allow you to make mistakes, so perhaps make the tag public API only operate on strings and use the hashes internally for queries. Also now RegisterTag / UnregisterTag

1vanK:
What about more than one tag per node? Each node can belong to different groups. And methods for node: AddTag(String), RemoveTag(string), Vector GetTags(), IsTagged(String)

cadaver:
This was forgotten for a while. Now that the UI tags PR is already in, I can re-check and unify this with that system.

svifylabs:
ok. refactored it to use a stringvector (like the UI tags). So Node::GetChildrenWithTag will be a bit slow but the scene caches the tagged nodes. As long as you need all specificed tagged nodes, Scene::GetNodesWithTag can be used. ...

1vanK:
When hierarchy of nodes reparented from another scene, hash table is updated?

svifylabs:
caching of the tagged nodes is done in Scene::NodeRemoved and Scene::NodeAdded so reparenting should work.

cadaver:
I'll be doing some code convention edits (use spaces not tabs, no underscores in variable names) and minor readability or logic cleanup changes. Functionality itself looking good. Will rename some of the UI tag functions to match.

cadaver:
Merged.

PR #1113, opened 2015-12-23
JSON value fix

hjmediastudios:
Fixed bug that caused a crash when setting a JSONValue to a VariantVector; previous behavior attempted to assign to indices beyond the bounds of the array. Now appropriate space is reserved and accessed.


cadaver:
Merged as squashed. Thanks!

Issue #1065, opened 2015-11-29
(FEATURE REQUEST) Writing all config definitions to Urho3D.h


1vanK:
We can use COMPILE_DEFINITIONS for it: string (REPLACE ";"...
It allows users to manually create new projects without using CMake.

Also we can add libs to header file like
```
#if define(_MSC_VER)
#    if defined(_DEBUG)
#        pragma comment(lib, "Urho3D_d.lib")
#    else
#        pragma comment(lib, "Urho3D.lib")
#    endif
#endif
```
I think it's much easier to create new projects. Just create empty project in VS and include "Urho3D.h" in source file.

PR #1119, opened 2015-12-25
Write all CMake definitons to header file

1vanK:
https://github.com/urho3d/Urho3D/issues/1065

It is a first try (there are some issues). Any feedback?

Sample result (Config.h)
```
#pragma once

#define URHO3D_MINIDUMPS
#define URHO3D_FILEWATCHER
#define URHO3D_PROFILING
#define URHO3D_LOGGING
#define URHO3D_THREADING
#define URHO3D_ANGELSCRIPT
#define URHO3D_LUA
#define TOLUA_RELEASE
#define URHO3D_NAVIGATION
#define URHO3D_NETWORK
#define URHO3D_PHYSICS
#define URHO3D_URHO2D
#define _CRT_SECURE_NO_WARNINGS

#ifdef _MSC_VER
#    ifdef _DEBUG
#        pragma comment(lib, "Urho3D_d")
#    else
#        pragma comment(lib, "Urho3D")
#    endif
#    pragma comment(lib, "user32")
#    pragma comment(lib, "gdi32")
#    pragma comment(lib, "winmm")
#    pragma comment(lib, "imm32")
#    pragma comment(lib, "ole32")
#    pragma comment(lib, "oleaut32")
#    pragma comment(lib, "version")
#    pragma comment(lib, "uuid")
#    pragma comment(lib, "ws2_32")
#    pragma comment(lib, "winmm")
#    pragma comment(lib, "dbghelp")
#    pragma comment(lib, "opengl32")
#endif
```

### DIFF
--- a/CMake/Modules/GenerateConfigHeader.cmake
+++ b/CMake/Modules/GenerateConfigHeader.cmake
@@ -1,0 +1,22 @@
+set (FILE_NAME "${CMAKE_CURRENT_BINARY_DIR}/Config.h")
+
+file (WRITE ${FILE_NAME} "#pragma once")
+
+get_directory_property (URHO3D_COMPILE_DEFINITIONS COMPILE_DEFINITIONS)
+list (REMOVE_ITEM URHO3D_COMPILE_DEFINITIONS HAVE_STDINT_H GLEW_STATIC GLEW_NO_GLU URHO3D_IS_BUILDING ODBC_3_OR_LATER ASSET_DIR_INDICATOR)
+string (REPLACE ";" "\n#define " URHO3D_COMPILE_DEFINITIONS "#define ${URHO3D_COMPILE_DEFINITIONS}")
+file (APPEND ${FILE_NAME} "\n\n${URHO3D_COMPILE_DEFINITIONS}")
+
+define_dependency_libs (${TARGET_NAME})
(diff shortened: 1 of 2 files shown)


weitjong:
Thanks for PR. Besides the build issue flagged up by AppVeyor (VS 2015) and Travis-CI (MinGW) for SHARED lib type, there is also problem with your setup even when it works because the ```Config.h``` being newly generated each time when CMake rerun and since it will be included by our export header then it will trigger a rebuild unnecessary when its content actually does not change. The main issue with your approach is you are baking the compiler defines into a new header file based on the effective compiler defines in the current scope, since the new header file is being included in the same build then you will always get macro redefinition warnings. Also, since we already now have, IMHO, a better way to bake the compiler defines and also a way to auto-discovering them for the downstream projects, I don't think why we need another mechanism. In the master branch currently we only bake-in a very small number of compiler defines but that is by our own choice, not because we cannot.

1vanK:
It seems that "winmm.lib" is doubled in Urho3D-CMake-common.cmake

1vanK:
```
if (WIN32)
            list (APPEND LIBS user32 gdi32 winmm imm32 ole32 oleaut32 version uuid)
```
```
if (WIN32)
            list (APPEND LIBS winmm)
```

PR #1127, opened 2015-12-29
Transfer some defines to Urho3D.h

1vanK:


1vanK:
Maybe it would be better to rename <Urho3D/Urho3D.h> into <Urho3D/Urho3DDefines.h> (or <Urho3D/Urho3DConfig.h>) ?

weitjong:
Thanks again for the PR. I have in the past commented in other places that I have expressed my reservation about baking compiler defines into our export header file, so I would not repeat it here again. However, I do have some other points I want to make. There are implication when the compiler defines are being baked this way so we don't want to do it indiscriminately, or otherwise we would have done it wholesale in the first place. IMHO, we should consider baking the compiler define on case by case basis.

I think I can understand that baking the compiler defines make the user who does not use CMake one step easier to configure their project in the IDE. However, that is not our main use case and IMHO we should not support it a the expense of our main use cases, i.e. via CMake. Without using CMake, those users would have to worry about more things than just the compiler defines.

As always, all the above is just my personal opinion. Take it with a grain of salt.

1vanK:
The problem is not in using Cmake. The user must use Cmake for building of engine in any case. The problem is that when he uses the compiler to create applications, he must exactly repeat Cmake settings, which he used when compiling engine.

weitjong:
Ok, I thought that was your main reason because you did not mention it in the description. Aside from those already baked in the export header currently, I don't think the other compiler defines needs to be set exactly the same as Urho3D library did. We could still have missed out a few though, so if you can give a valid reason on case by case basis why a define need to be baked then it can be considered but not wholesale like this. For example, I could have built my library with both AS and Lua enabled, but in my own downstream project, I may just want to use AS only or Lua only. Or another example, I could have built my library with logging turned off, but in my own downstream project I want to turn it back on, so I get the log entries just from my app.

1vanK:
> However, that is not our main use case and IMHO we should not support it at the expense of our main use cases, i.e. via CMake. Without using CMake, those users would have to worry about more things than just the compiler defines.

I have researched this issue. All required libs can be merged into one Urho3d.lib / Urho3d_d.lib.
All additional include paths can be used directly in the code (perhaps in bullet lib we need replace angle brecet form for includes)

weitjong:
Not sure I am understanding you, but it would be best to open this discussion in another issue or in the forum. But not to preempt anything, I want to inform you that we had a long discussion before we did  we last major refactoring of the build system regarding how the include search path should be configured and what form of the C/C++ include statement we would like to use for all our 3rd-party libraries. What you see at the present day is the result of that work.

1vanK:
Urho3d used modified versions of third party libraries. But used its as external libraries (#include <... >) I think is incorrect, since such libraries can be installed in system.

1vanK:
> Not sure I am understanding you

I mean remove /I"${includedir}/Urho3D/ThirdParty" /I"${includedir}/Urho3D/ThirdParty/Bullet from generated projects but used relative paths in sources

weitjong:
That's a valid point but that's also why we don't simply install those 3rd party library headers at the top level (/usr/local/include) and instead they are all installed under our project umbrella (/usr/local/include/Urho3D), so they don't get mixed up with other packages which may also try to install Bullet in the filesystem.

1vanK:
And in the Windows?

weitjong:
> I mean remove /I"${includedir}/Urho3D/ThirdParty" /I"${includedir}/Urho3D/ThirdParty/Bullet from generated projects but used relative paths in sources

In theory it can be done and someone has attempted to do it in the past also. I think we even has a PR submitted but it got rejected because the changes required was too much. What we have now is not nice, I agree but it is the most practical without changing the 3rd party code by too much.

weitjong:
> And in the Windows?

If I recall correctly, it should c:\program files\Urho3D\include\ThirdParty\.....

1vanK:
Constraint.cpp

```
#include <Bullet/BulletDynamics/ConstraintSolver/btConeTwistConstraint.h>
#include <Bullet/BulletDynamics/ConstraintSolver/btHingeConstraint.h>
#include <Bullet/BulletDynamics/ConstraintSolver/btPoint2PointConstraint.h>
#include <Bullet/BulletDynamics/ConstraintSolver/btSliderConstraint.h>
#include <Bullet/BulletDynamics/Dynamics/btDiscreteDynamicsWorld.h>
```
Assume i haw installed original Bullet in my compiler/includes dir
Which path will be used?

weitjong:
The answer depends on how you setup the include search path in the project.

1vanK:
I do not set paths, CMake generate project :)

weitjong:
Well, if you use our custom CMake modules correctly then that part is being taken care of already.

weitjong:
I will end this discussion now as it is off topic for this issue and not to agitate others who happen to watch the repo communication because we are now spamming their emails :)

1vanK:
Ok, compiler has standart include dir and additional include dir. Both have bullet subdir. What will be used?

PR #1144, opened 2016-01-11
Direction support for particle/billboard

ucupumar:
This pull request will add particle direction inspired by Unity stretched billboard particles. If any of you have tried that on Unity, stretched particle can create pseudo 3d effect despite still using billboard. It can be really useful for sparking effects and many other stuffs. 

Screenshot:
partikel_1

You can see this particle in action at example 'Particle/Burst.xml' or just see this youtube video: https://www.youtube.com/watch?v=zR7KtF6qtaU

About the implementation, I modify the billboard object to store direction. This data only works if new face camera mode FC_DIRECTION is used. If this mode is used, billboard will rotate facing camera with direction as axis, so it will create pseudo 3D effect. Because every billboard has different rotation, Urho needs to calculate rotation for every billboard. For now, CPU is doing those calculation, but I have plan to move them to the GPU.

What do you guys think?

### DIFF
--- a/Source/Urho3D/AngelScript/GraphicsAPI.cpp
+++ b/Source/Urho3D/AngelScript/GraphicsAPI.cpp
@@ -1319,12 +1319,14 @@ static void RegisterBillboardSet(asIScriptEngine* engine)
     engine->RegisterEnumValue("FaceCameraMode", "FC_ROTATE_Y", FC_ROTATE_Y);
     engine->RegisterEnumValue("FaceCameraMode", "FC_LOOKAT_XYZ", FC_LOOKAT_XYZ);
     engine->RegisterEnumValue("FaceCameraMode", "FC_LOOKAT_Y", FC_LOOKAT_Y);
+    engine->RegisterEnumValue("FaceCameraMode", "FC_DIRECTION", FC_DIRECTION);
 
     engine->RegisterObjectType("Billboard", 0, asOBJ_REF);
     engine->RegisterObjectBehaviour("Billboard", asBEHAVE_ADDREF, "void f()", asFUNCTION(FakeAddRef), asCALL_CDECL_OBJLAST);
     engine->RegisterObjectBehaviour("Billboard", asBEHAVE_RELEASE, "void f()", asFUNCTION(FakeReleaseRef), asCALL_CDECL_OBJLAST);
     engine->RegisterObjectProperty("Billboard", "Vector3 position", offsetof(Billboard, position_));
     engine->RegisterObjectProperty("Billboard", "Vector2 size", offsetof(Billboard, size_));
+    engine->RegisterObjectProperty("Billboard", "Vector3 direction", offsetof(Billboard, direction_));
     engine->RegisterObjectProperty("Billboard", "Rect uv", offsetof(Billboard, uv_));
     engine->RegisterObjectProperty("Billboard", "Color color", offsetof(Billboard, color_));
     engine->RegisterObjectProperty("Billboard", "float rotation", offsetof(Billboard, rotation_));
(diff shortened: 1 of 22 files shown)


weitjong:
Nice two! :)

cadaver:
Good feature, can you clean up / squash the history as there's quite a bit of merge commits? (we can also squash on merging, but in that case your commit messages & author info will be lost)

hdunderscore:
Very cool effects you can get from that !

ucupumar:
Thanks for all the comments! :)
@cadaver okay, I'll try that, I never squash commits before

1vanK:
I'll try that, I never squash commits before
```
git rebase -i HEAD~14
```
then opened VIM editor
prees "i" to switch insert mode
for each string except the first replace command (firs word in string) to f
press "esc" to end insert mode
press ":" and "w" to save
press ":" and "q" to exit
```
git push origin master --force // you need to specify the branch
```
to send changes to repository

ucupumar:
I think I made a mistake here. The command 'HEAD~14' will effect about 1800 commits altogether. Maybe it's because I've been using this fork since 2014. I tried to squash those 1800 commits but kept failed.  :cry:

What do you think should I do?

weitjong:
Make sure you have a backup plan before you proceed further.

Assuming commit ```a361d2f``` is the initial commit for this PR as indicated by the log message and also assuming you are still in the ```particle-editor``` branch then I think what you should do are:

- In case you have not done it, abort your failed rebase attempt.
```
git rebase --abort
```
- In case you have not added it, add a new remote pointing to urho3d/Urho3D in your local repo.
```
git remote add -f upstream https://github.com/urho3d/Urho3D.git
```
- Rebase your changes onto upstream/master, resolve any conflicts (if any) as you go.
```
git rebase --onto upstream/master a361d2f HEAD
git checkout -b particle-editor-rebased
```
- Rename the branches and verify the logs before you force push again.
```
git branch -m particle-editor particle-editor-old
git branch -m particle-editor
git log
```
- Finally force push, assuming origin is pointing to your fork repo.
```
git push -f -u origin particle-editor:particle-editor
```

I type all these commands as I write this comment. They are not verified, so please adjust accordingly.

weitjong:
BTW, it seems there is still one more merge commit after the ```a361d2f````. So, may be before the final force push, you should do a ```git rebase -i HEAD~7``` as suggested by 1vanK earlier but with lower number. Semoga sukses!

ucupumar:
Thanks for the answer!
I got a lot of conflicts at ```git rebase --onto upstream/master a361d2f HEAD```
Can I safely just skip them? Or should I resolve and commit one by one?

weitjong:
I think the conflicts are most come from that one merge commit I mentioned earlier. The short answer is no, you cannot simply skip them. But actually we are just being stupid. The rebase is usually used when we want to somehow preserve  some of the commit history and squash others. In your case, however, we want to squash all. So, I think we can take another approach. Create a diff of your current branch HEAD against upstream/master branch. Then create a new branch based on upstream/master. And apply the diff on top of this new branch.
```
git diff upstream/master HEAD >patch.txt
git checkout -b particle-editor-new upstream/master
patch -p1 -i patch.txt
git commit -am "squashed hurray ^_^"
```
If it works then followed by branch rename and force push as I shown earlier. Again, I just type this without validating those commands.

ucupumar:
Finally...
But, I ended up using TortoiseGit serial patch. I can't seems to use command line patch properly on my Windows machine. Because of this, the commits only squashed to six.
Is it still okay?

ucupumar:
Okay, I just realized that I can just do rebase again, now all my commits already squashed into just one commit.

cadaver:
Nice! Is this missing GLSL shader code apart from Transform.glsl?

ucupumar:
@cadaver Nothing is missing, the only GLSL code changed is Transform.glsl. That's because unlike HLSL, GLSL doesn't need a lot of preprocessor when declaring attribute. 

Anyway, I pushed new commit that move rotation computation from CPU to GPU. Now particle with direction should be give faster performance. I tried to input 200k of particles, but I think Urho limit it's particles to only some thousands. Below is screenshot of them.

partikel_4

Surprisingly, Radeon R9 270x can run it at only 450Mhz clock speed. :+1:

cadaver:
Ah, thanks for clearing that up. Right, everything happens in Transform.glsl.

ucupumar:
I see, let me check that out

ucupumar:
I have created pull request to fix them

weitjong:
Terima kasih!

ucupumar:
Sama-sama :)

PR #1156, opened 2016-01-18
Add view matrix

hjmediastudios:
Added a view matrix uniform to the shader code. The impetus for this is to allow me to convert world-space normals into view-space normals to perform SSAO calculations in the rendering pipeline.


cadaver:
Shouldn't be a problem to merge. There are a couple of added hashmap lookups every time the camera changes, but the impact shouldn't be significant unless one is changing shaders or camera for every drawcall, which would be bad anyway.

ucupumar:
Nice! I've been looking for default view matrix for quite some time. This will be huge time safer for some of custom shader creation.

MonkeyFirst:
I want to mention one thing.
the matrix
const Matrix3x4& view = camera->GetView(); is already inversed in 
```
const Matrix3x4& Camera::GetView() const
{
    if (viewDirty_)
    {
        // Note: view matrix is unaffected by node or parent scale
        view_ = GetEffectiveWorldTransform().Inverse();
        viewDirty_ = false;
    }

    return view_;
}
```
so your graphics_->SetShaderParameter(VSP_VIEWINV, view.Inverse()); 
is actually double inversing I guess

cadaver:
Using simply GetEffectiveWorldTransform() should work and save the work of double-inverse. Can make that change when merging.

Btw. after merging this the separate CameraPos & CameraRot uniforms could be removed, since they can be derived from the view inverse.

PR #1157, opened 2016-01-20
Add tags to UIElement

1vanK:
For now without bindings and attributes. Any notes?

### DIFF
--- a/Source/Samples/37_UIDrag/UIDrag.cpp
+++ b/Source/Samples/37_UIDrag/UIDrag.cpp
@@ -73,6 +73,9 @@ void UIDrag::CreateGUI()
         b->SetSize(300, 100);
         b->SetPosition(IntVector2(50*i, 50*i));
 
+        if (i % 2 == 0)
+            b->AddTag("SomeTag");
+
         SubscribeToEvent(b, E_DRAGMOVE, URHO3D_HANDLER(UIDrag, HandleDragMove));
         SubscribeToEvent(b, E_DRAGBEGIN, URHO3D_HANDLER(UIDrag, HandleDragBegin));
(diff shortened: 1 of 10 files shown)


1vanK:
It is possible to make acessor_atribute for Vector<String> ?

cadaver:
This looks OK. Actually we could follow the same approach for the node tags; judging from this it would be OK to be a straightforward string-based multi-tag operation instead of being extremely optimized. The idea being that if you don't use tags, you don't pay cost for them (except for storage of the tags vector in each node, which is lightweight if empty)

Vector<String> is not directly convertible to a Variant. You can use dedicated functions for the attribute get/set access which aren't actually part of the API (see functions like SetMaterialsAttr, GetMaterialsAttr in StaticModel for example) and convert the string vector to a VariantVector.

Another possibility, which would be good for editing, would be to mark the attribute as a string and convert/deconvert it to a to a comma- or space-separated list in the accessors.

1vanK:
Squashed

1vanK:
In Editor the tags are separated by ';'

cadaver:
Seems to work great also in practice. I'll rename IsTagged() to HasTag() to be more typical of the existing API convention.

cadaver:
Merged.

1vanK:
> This looks OK. Actually we could follow the same approach for the node tags; judging from this it would be OK to be a straightforward string-based multi-tag operation instead of being extremely optimized. The idea being that if you don't use tags, you don't pay cost for them (except for storage of the tags vector, which is lightweight if empty)

Unlike a interface, scenes usually has many childred in root. So without hash table "tag -> nodes" it can be slowly (unlike GetChild(name) we need look ALL nodes, but not first founded). But with this hashtable it is necessary to look all children when node is parented / unparented (some cost even without tags). Also nodes has components and vars, so in my small practics i yet never needed tags in scene.

PR #1169, opened 2016-01-25
Tags menu

MonkeyFirst:
Merged into one commit, checkout this.
+ tags settings added


MonkeyFirst:
So, I close this for another rebase trying.

weitjong:
Have you tried a simpler approach? Use ```git diff``` and redirect to an output file then reapply the output file as a patch? See #1144.

MonkeyFirst:
Thanks, but I guess find more easily way to do this. Maybe I try this soon :)

PR #1170, opened 2016-01-25
C89 compliance in SDL_windowsevents.c when compiling under MSVC2012

TheComet93:
The error message is:
```
error C2275: 'SDL_Mouse' : illegal use of this type as an expression SDL_windowsevents.c    426    1    SDL
```
This happens because MSVC will compile all ```*.c``` files using the C compiler in C89 mode, meaning all variables must be declared at the beginning of each scope. A change previously made by an Urho3D developer violated this constraint (see diff for more info).

GCC and clang will by default ignore such errors and only generate a warning.


cadaver:
Thanks!

PR #1186, opened 2016-02-04
Added Matrix2 class

TheComet93:
A Matrix2 class seemed to be missing from Urho3D.


cadaver:
Traditionally 3D engines need only 3x3, 3x4 & 4x4 matrices, though I could see this useful in 2D applications. Have you used this class to solve an actual problem you had? Also, can you add script bindings?

TheComet93:
Yes, I used the Matrix2 class to rotate the direction of the joystick's control stick (which is a Vector2) by the camera's Y angle, so that the direction the player heads is relative to the 3rd person camera's angle. Code:
```Vector2 targetDirection = GetControlStickDirection();

// rotate input direction by camera angle using a 2D rotation matrix
targetDirection = Matrix2(Cos(cameraAngle_), -Sin(cameraAngle_),
							  Sin(cameraAngle_), Cos(cameraAngle_)) * targetDirection;```

As to script bindings, I'll need to learn how to do that first, but yes, I'll do that too!

TheComet93:
That's strange... That last commit compiles fine on my machine.

cadaver:
The PR seems no longer mergable to master branch due to conflicts, so those need to be resolved. I can check if it's something trivial.

cadaver:
The conflict is the addition of a few lines MathAPI, which can be resolved as "yours". Will merge shortly. Thanks for the work!

PR #1194, opened 2016-02-12
Allow implicit conversions of SharedPtr objects.

AntonPetrov83:
Cannot see any reason to avoid implicit upcasting of SharedPtr.
This allows not to use `StaticCast` where silent upcasting is possible.
Besides I have changed comparision operators to allow mixed comparisions.
Some testing was performed.

This is my first pull request to test Urho3D collaboration :) so hello everyone!

### DIFF
--- a/Source/Urho3D/Container/Ptr.h
+++ b/Source/Urho3D/Container/Ptr.h
@@ -47,6 +47,13 @@ public:
         AddRef();
     }
 
+    /// Copy-construct from another shared pointer allowing implicit upcasting.
+    template <class U> SharedPtr(const SharedPtr<U>& rhs) :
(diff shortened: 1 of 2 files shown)


AntonPetrov83:
@cadaver, how are you!
What do you think on implementing move semantics?
I can do this for some Container classes.

I have experience of using C++11 in a cross-platform project (a game of course) and there were no problems compiling the code to all modern platforms including emscripten.

cadaver:
Yes, this change is nice and shouldn't be unsafe under any circumstances. 
Of move semantics, Urho itself still doesn't actually require C++11 (and there still are some hardcore users who for some reason use pre-C++11 compilers) so at least implementing them optionally would be fine.

PR #1204, opened 2016-02-19
Urho2d fixing

OvermindDL1:
Fixes a couple of Urho2D issues that are of more immediate need.  Adds setters.


.gitignore
```diff
@@ -10,6 +10,7 @@
 /bin/tool/
 /include/
 /lib*/
+/build/
```
weitjong: Since we don't have a convention in naming the build tree anymore, i.e. user can use a subdir name whatever they like for their build tree and can put that subdir wherever they like too (including putting it inside the source tree, like in your case), ignoring just the 'build' subdir is amounted to hard-coding. Either move your build tree out or use the source tree itself as build tree to solve the problem. Alternatively,  we can agree on a naming convention for a build tree and ignore the subdir matching that naming convention. See how 'configure*' shell scripts are being ignored,  for example. In short no hard coding. 

Thanks for the PR. Glad to have a new 2D subsystem maintainer.

OvermindDL1: Ah, should I remove that?  I still have a lot of old scripts around that still uses that old convention from older Urho3D days.  ^.^

OvermindDL1:
Aaand what's say I actually build it as just quick-edits is not working for me...  Gimme a bit...  >.>

.gitignore
OvermindDL1: Removed the .gitignore line.

OvermindDL1:
Should be ready for pulling now if it meets with approval, just some basic setters added.  Really need to overhaul a lot of Urho2D though.

cadaver:
Thanks for this work! Will review in detail in a short while.

cadaver:
Functionality itself certainly meets approval. However I would suggest squashing the history, as it contains merges and back-and-forth adding and removal of things. Commit messages don't need to describe things yet missing or other shortcomings, it's sufficient to just describe what's added or modified in each commit.

Also, the code convention does not fully match at places, for example spaces in if statements or compare statements, and pointer types (the existing code uses TmxLayer2D* instead of TmxLayer2D *). This may at times seem anal but it helps to keep the code uniform looking.

cadaver:
Merged with changes.

OvermindDL1:
Er, sorry for the delay, did not mean for you to do it yourself, was doing
other code. Thanks though. :-)

Issue #917, opened 2015-10-05
The default behavior of hiding mouse cursor in windowed mode in Emscripten builds is troublesome.


juj:
In several samples built for Emscripten, the mouse behavior is a bit quirky. 

What currently happens:
    - mouse cursor is hidden when mouse is moved over the canvas.
    - mouse cursor is restored to be visible again when mouse exits the canvas.
    - moving the mouse over the canvas generates camera mouselook rotate events.
    - moving the mouse outside the canvas does not rotate the camera.
    - mouse cursor is not constrained to stay inside the canvas rectangle, but the mouse cursor is able to freely exit the canvas rectangle.
    - when mouse enters/exits the canvas rectangle, there is an abrupt "discontinuous" snap of the camera lookat direction.

The combination of these behaviors makes looking around the scene by rotating the camera with mouse move a bit difficult, and one can't easily do e.g. a 360 rotation of the . In fullscreen mode, this problem does not occur since the mouse cursor is not able to exit the canvas.

In the 08_decals sample, there is an additional issue that the software mouse cursor rendered by the engine does not match with the location of the actual mouse cursor.

If I might suggest, some ways to come out of the issues:
   - remove the behavior of hiding the mouse cursor by default and moving the camera by default, and make it e.g. a "hold down right mouse button to drag" type of gesture.
   - In 08_decals sample, one could avoid hiding the mouse cursor over the canvas, but instead let the real mouse cursor stay visible, and not render a software mouse cursor in Urho3D. Then use absolute mouse cursor positions provided by the input api to mark the mouse cursor location instead of calculating them incrementally/relative inside the engine.
   - an alternative could be to use the pointer lock api (`emscripten_request_pointerlock()`) when the user clicks on the canvas. Pointer lock hides the mouse cursor on the web page altogether, and changes it to a relative movement mode. See here for a demo http://mdn.github.io/pointer-lock-demo/

hdunderscore:
Most of the main issues should be addressed with the recent pull request. The online samples aren't updated yet though.

Issue #1200, opened 2016-02-16
Possible problem at DetourCrowd **integration** causing random segfault at 64-bit build using MSVC compiler toolchain


weitjong:
Since we have enabled AppVeyor (VS2015) to test run all the sample apps as part of the CI jobs, we have observed random segfault on 64-bit build of 39_CrowdNavigation for both STATIC and SHARED lib types. The problem appears at random and only manifests itself once in a blue moon. 32-bit MSVC builds do not exhibit this problem. CI jobs using other compiler toolchains (Clang and GCC) also do not exhibit this problem in any build configurations and ABIs.

cadaver:
Since I don't think DetourCrowd uses threads, it could be an uninitialized memory issue.

weitjong:
If this issue cannot be fixed at the soonest then probably we should temporarily move the 39_CrowdNavigation out from the default AppVeyor CI and into a new CI-mirror branch by itself temporarily so that it does not raise false alarm to PR submitter.

weitjong:
The prevalence rate of this issue on VS2015 seems to increase lately, i.e. it crashes more often than not now on AppVeyor, which is actually a good thing. If it constantly crashes instead of random crash then I think VS user would eventually notice it too instead of just AppVeyor, and hopefully it is easier for them then to study the stacktrace while it happened and fix the issue permanently.

cadaver:
Yes, no problem moving it away from the CI build. A while ago I tried reproducing this on VS2015 64bit but no luck so far.

JSandusky:
No shit it's broken, you radically changed your build scripts and expect things to work the same? WTF is wrong with you - the last time that got a 64bit check was by me x 1024? The library has degenerated into filth, it's 2016 and you still don't have Geometry/Hull/Domain/Compute shaders or even Texture arrays because you couldn't pull the stick out of your own ass without tearing it.

Don't pass the bullshit off on crowds or detour. I may have no interest but what you have all just expressed are fighting words. Don't drag me into this, when you call it a generic crowd thing - that's what you're doing.

And do some fucking research so you can see that UE is suffering the same problem. Fucking idiots talking like they know some shit.

weitjong:
I have locked this issue for internal member discussion only. But I will keep the comment from Sinoid as an anti pattern.

weitjong:
@JSandusky. One more unprovoke insult and you will block out from our GitHub repo for good.

weitjong:
[...]

PR #1207, opened 2016-02-20
Input mousemode updates

hdunderscore:
Some fixes to the input issues that have popped up on the Web platform, and an attempt to make the mouse modes compatible with the Console.

Related issue: https://github.com/urho3d/Urho3D/issues/917

```
- Added a way to suppress mouse mode and mouse grab changes (SetMouseMode(mode, suppress = false), 
SetMouseGrabbed(grab, suppress = false)), and reset them to a previous unsuppressed state (ResetMouseMode, 
ResetMouseGrabbed).
- Updated Console to react to mouse modes, similarly to mouse visibility previously.
- Console will utilize UI cursor if available.
- Fixed a bug with mouse modes not calling SetMouseGrab correctly.
- Updated samples to use different mouse modes where necessary.
- Updated samples to request pointer lock on mouse down event (Web platform).
- Updated samples that show the UI cursor to show the OS cursor if pointer lock (ie, mouse mode MM_RELATIVE) 
isn't acquired. This conceals a visual bug in UI cursor position (Web platform).
```

### DIFF
--- a/Source/Samples/01_HelloWorld/HelloWorld.cpp
+++ b/Source/Samples/01_HelloWorld/HelloWorld.cpp
@@ -51,6 +51,9 @@ void HelloWorld::Start()
     // like the ScreenMode event sent by the Graphics subsystem when opening the application window. To catch those as well we
     // could subscribe in the constructor instead.
     SubscribeToEvents();
+
+    // Set the mouse mode to use in the sample
+    Sample::InitMouseMode(MM_FREE);
 }
 
 void HelloWorld::CreateText()
--- a/Source/Samples/02_HelloGUI/HelloGUI.cpp
+++ b/Source/Samples/02_HelloGUI/HelloGUI.cpp
@@ -71,6 +71,9 @@ void HelloGUI::Start()
 
     // Create a draggable Fish
     CreateDraggableFish();
+
+    // Set the mouse mode to use in the sample
+    Sample::InitMouseMode(MM_FREE);
 }
 
 void HelloGUI::InitControls()
@@ -181,7 +184,7 @@ void HelloGUI::CreateDraggableFish()
 
 void HelloGUI::HandleDragBegin(StringHash eventType, VariantMap& eventData)
 {
-    // Get UIElement relative position where input (touch or click) occured (top-left = IntVector2(0,0))
+    // Get UIElement relative position where input (touch or click) occurred (top-left = IntVector2(0,0))
     dragBeginPosition_ = IntVector2(eventData["ElementX"].GetInt(), eventData["ElementY"].GetInt());
 }
 
@@ -198,7 +201,8 @@ void HelloGUI::HandleDragEnd(StringHash eventType, VariantMap& eventData) // For
 
 void HelloGUI::HandleClosePressed(StringHash eventType, VariantMap& eventData)
 {
-    engine_->Exit();
+    if (GetPlatform() != "Web")
+        engine_->Exit();
 }
 
 void HelloGUI::HandleControlClicked(StringHash eventType, VariantMap& eventData)
--- a/Source/Samples/03_Sprites/Sprites.cpp
+++ b/Source/Samples/03_Sprites/Sprites.cpp
@@ -54,6 +54,9 @@ void Sprites::Start()
 
     // Hook up to the frame update events
     SubscribeToEvents();
+
+    // Set the mouse mode to use in the sample
+    Sample::InitMouseMode(MM_FREE);
 }
 
 void Sprites::CreateSprites()
--- a/Source/Samples/04_StaticScene/StaticScene.cpp
+++ b/Source/Samples/04_StaticScene/StaticScene.cpp
@@ -63,6 +63,9 @@ void StaticScene::Start()
 
     // Hook up to the frame update events
     SubscribeToEvents();
+
+    // Set the mouse mode to use in the sample
+    Sample::InitMouseMode(MM_RELATIVE);
 }
 
 void StaticScene::CreateScene()
--- a/Source/Samples/05_AnimatingScene/AnimatingScene.cpp
+++ b/Source/Samples/05_AnimatingScene/AnimatingScene.cpp
@@ -67,6 +67,9 @@ void AnimatingScene::Start()
 
     // Hook up to the frame update events
     SubscribeToEvents();
+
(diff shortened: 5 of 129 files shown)


cadaver:
Nice update!

hdunderscore:
I think that can work, when absolute + hidden mouse, we request pointerlock, when absolute + visible mouse we leave pointerlock. Then it would be consistent with desktop and it should be quite seamless after the initial request.

I am not able to reproduce that appveyor segfault, and only one thing comes to mind about the potential source (in samples-> InitMouseMode called before CreateConsoleAndHUD), so I'll shuffle that around.

Anyone able to test these changes on mobile? I haven't had a chance to do so yet.

weitjong:
> I am not able to reproduce that appveyor segfault, and only one thing comes to mind about the potential source (in samples-> InitMouseMode called before CreateConsoleAndHUD), so I'll shuffle that around.

The random crash of crowd navigation sample app is a known issue (#1200), so you should not be concerned about it.

cadaver:
A quick test of NinjaSnowWar on Android at least didn't reveal anything to be broken.

cadaver:
Hmm, I quickly tested the 06_SkeletalAnimation sample in Emscripten, windowed mode. The behavior seemed like before (mouse cursor is hidden, but it's not confined to the window, and pointer lock request dialog did not turn up)

cadaver:
Emscripten samples indeed seemed to work nicely now. But on Windows I got a crash when I pressed right mouse button down in the editor. Didn't investigate further yet.

hdunderscore:
Thanks for testing it out! That crash is concerning, hopefully it's easy to track down.

I've been slowly working at the MM_ABSOLUTE behavior, haven't quite got it nailed down yet.

cadaver:
The crash is due to AngelScript bindings not being updated when the SetMouseMode function gained an additional parameter. Because the mousemode is a property which can only have one parameter, you need to create a wrapper function. Probably best to also expose SetMouseMode as a function so that user can supply "true" to the suppressEvent parameter.

hdunderscore:
Nice catch, I'll update that.

TheComet93:
The shared build (DLL) on Windows x64 seems to be very fragile. I had the same sample (39_CrowdNavigation) segfault on my last push, but it worked fine before that and I really didn't change anything other than renaming a symbol.

Double check you're not sharing memory across DLL boundaries. If you are, make sure you're correctly exporting it. That seemed to help.

weitjong:
The crash could happen on both STATIC and SHARED 64-bit build. For future discussion on this issue may I suggest to use #1200. I have just unlock it for public discussion.

hdunderscore:
Just an update, I'm inching closer to solution on this, it's a bit of a moving target as I find new little issues.

The latest change I've made is to turn off the event polling loop for emscripten and make use of SDL_AddEventWatch-- this means if you request pointerlock from an Urho input event with my upcoming changes, it will activate right in that event from the browsers perspective.

Previously, since we were polling for the events, regardless of where you requested pointerlock it would always occur in the next input event. This was all occurring very subtly due to this: https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/Input/Input.cpp#L132  The second argument (deferUntilInEventHandler) concealed the issue.

This meant for example, if you wanted to switch into MM_RELATIVE on mouse down, and switch out on mouse up, it wouldn't work before because the request in mouse down would be deferred to the next input event (commonly mouse up).

cadaver:
Ah, makes sense, thanks for continued looking into this!

hdunderscore:
I have pushed my changes-- I still need to update the script samples and perhaps polish some things off (also just spotted some whitespace issues).

To simplify testing, I have made the mouse controls slightly different in some of the samples to provide test cases and identify bugs. I need help with testing, let me know if you notice any issues.

- One known issue is LoseFocus/GainFocus behaviour is a bit different now, mouse modes don't immediately re-engage on GainFocus.
- On web, right click can sometimes open context menu.
- A possible issue could exist with this change too (I removed the if-guard on the delta changes, so they could be used by UI cursor): https://github.com/urho3d/Urho3D/commit/157d22b09caef0ec43679417b29860f80b7c3947#diff-eeea0f0ea09c3078b218adad26e10f10L1583
- This should probably be flipped back to true for the general case: https://github.com/hdunderscore/Urho3D/commit/157d22b09caef0ec43679417b29860f80b7c3947#diff-eeea0f0ea09c3078b218adad26e10f10R164

cadaver:
Excellent! I should be able to test the web examples shortly.

cadaver:
Tested on Windows desktop first. No crashes :) Editor and NinjaSnowWar worked great.

However the control scheme seemed changed on some examples, and it didn't seem to always work consistently. For example in 06_SkeletalAnimation I need to hold down right mouse to move with WASD (and debug geometry toggling doesn't work), while in 08_Decals I don't need to hold RMB to move

If the Emscripten pointer lock logic works OK in absolute mode now, could even all the samples be left unmodified compared to master branch?

hdunderscore:
Thanks for testing !

We can definitely keep the samples unmodified after we test across platforms and don't spot any issues-- I noticed on the forums etc that it's hard to predict how people will use the mouse modes so I wanted to test different cases, and I was definitely able to spot and squash issues while developing the PR with the different tests.

hdunderscore:
Actually thinking about it a bit more, some of the sample changes are needed, eg:
- Need a consistent way to request pointerlock, otherwise if the user misses the first prompt they will have to reload the page. This can be changed to be more consistent with existing sample style though.
- It can take a while to load an online html5 sample, so limiting the 'exit engine' logic is important to prevent confusion.
- The escape key logic moved to key up was needed to prevent a full exit of pointerlock with the console open.

cadaver:
Ah ok, naturally those changes are necessary.

cadaver:
Didn't have time to test everything yet, but at least StaticScene, AnimatingScene, SkeletalAnimation & Decals worked great, and similarly on web & desktop.

hdunderscore:
Just rebased with master to try with SDL 2.0.4, I haven't done any significant testing yet.

hdunderscore:
I made some changes and I think it's ready for merge, if there are no further issues identified.

cadaver:
Tested examples & script examples quickly on desktop and didn't notice anything off.

cadaver:
It's now in. Thanks again!

PR #1208, opened 2016-02-21
Update rapidjson to Version 1.0.2

svifylabs:


Stinkfist0:
The Urho3D-specific modification(s) are lost and should be preserved.

weitjong:
We are slowly moving toward using "git subtree" to manage all our future
3rd-party library updates. We already have a set of rake tasks to aid this
migration process. However, they are designed for the Urho3D project
members with push privilege. The objective of this migration is exactly to
avoid our modification for Urho3D getting lost.

Best regards,
YaoWT

svifylabs:
hmm ok so i should close it ? 
what are the modifications ? 
tested it with the scene load example using json to save and load the scene.

Stinkfist0:
The modifications should be only three lines, look for "Urho3D" in the originals files (prettywriter.h might be the only file with modifications).

Issue #305, opened 2014-04-12
2D aspect ratio distorted when changing device orientation


Mike3D:
2D display aspect ratio is no longer preserved when switching between landscape/portrait (changing device orientation) on Android.

Steps to reproduce:
- deploy 27_Urho2DPhysics on device
- change device orientation

Issue #230, opened 2014-03-06
Units per pixel in 2D components is misleading


cadaver:
In the Sprite2D component, units per pixel actually means pixels per unit (increasing it decreases sprite size) while in ParticleEmitter2D increasing it makes the effect even bigger.

I propose to fix this by renaming it PixelsPerUnit in the Drawable2D base class and making the behavior consistent. I can do this before the release.

PR #1209, opened 2016-02-23
Urho2D PIXEL_SIZE removal

gogoprog:
Hello,

I see no point in keeping this variable so here is a PR that removes it.
I also updated the samples to match the previous coordinates.

What do you think about this?

### DIFF
--- a/Docs/AngelScriptAPI.h
+++ b/Docs/AngelScriptAPI.h
@@ -14433,7 +14433,6 @@ int M_MIN_INT;
 uint M_MIN_UNSIGNED;
 float M_PI;
 float M_RADTODEG;
-float PIXEL_SIZE;
 int QUALITY_HIGH;
 int QUALITY_LOW;
 int QUALITY_MAX;
--- a/Docs/LuaScriptAPI.dox
+++ b/Docs/LuaScriptAPI.dox
@@ -8102,7 +8102,6 @@ Properties:
 - float M_RADTODEG
 - unsigned NUM_FRUSTUM_PLANES
 - unsigned NUM_FRUSTUM_VERTICES
-- float PIXEL_SIZE
 - int QUALITY_HIGH
 - int QUALITY_LOW
 - int QUALITY_MAX
--- a/Docs/Reference.dox
+++ b/Docs/Reference.dox
@@ -2088,7 +2088,7 @@ C++:
     Node* cameraNode = scene_->CreateChild("Camera"); // Create camera node
     Camera* camera = cameraNode->CreateComponent<Camera>(); // Create camera
     camera->SetOrthographic(true); // Set camera orthographic
-    camera->SetOrthoSize((float)graphics->GetHeight() * PIXEL_SIZE); // Set camera ortho size (the value of PIXEL_SIZE is 0.01)
+    camera->SetOrthoSize((float)graphics->GetHeight()); // Set camera ortho size
 \endcode
 
 AngelScript:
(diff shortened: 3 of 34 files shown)


cadaver:
I believe this to be a good call, as it's a fairly arbitrary adjustment to be honest. Will just need a bit of testing.

Mike3D:
This is related to issue #305, at that time it was mandatory in this particular situation.

cadaver:
I understand that without PIXEL_SIZE, the camera window would just need to be scaled differently, but the constant in itself doesn't help with orientation changes.

Like I said I believe this change to be in the right direction purely engineering-wise, however it would break compatibility of already written Urho2D applications. Opinions on this?

1vanK:
When I started learning engine, I hardly understood that PIXEL_SIZE just random number xD
> it would break compatibility of already written Urho2D applications
We often do this :)

gogoprog:
A SetPixelSize method would avoid breaking already written apps but I saw in the history that this method did exist but was removed.

By the way I'm not against breaking compatibility :)

cadaver:
@1vanK: yes, at this stage it's almost like an expectation that when you upgrade to next Urho version, your program will break :)

Will probably merge this for testing locally first and play around a bit in the editor & examples.

hdunderscore:
It's a confusing constant that doesn't appear to have too much meaning. A changeable pixel size doesn't seem like much more than a poor-mans zoom (which I believe we already have through camera).

If it can be removed safely, I say go for it.

gogoprog:
I just noticed I didn't update the Lua samples yet

cadaver:
Playing around in the editor I noticed there is a kind of justification for this. If we compare to the builtin box, sphere etc. 3D objects, their unscaled mesh is about 1x1x1 world units. However, sprites scale their mesh according to the pixel size of the image, so a 100x100 image would be 100x100 world units unscaled, and the PIXEL_SIZE constant reduces the sprites to a manageable size. It's actually the same as in Unity by default, but in Unity it can be changed individually in each sprite image resource.

Naturally you can downscale the scene node, but having to use small scale values seems awkward.

weitjong:
I think it is related to this discussion #230.

weitjong:
@aster2013 Any comment on this issue?

weitjong:
There is a possibility that we could reach Aster because of the "Great Firewall of China". The People's Republic of China does not like GitHub, although I should not have said this out loud.

weitjong:
Got a reply through Aster's personal email account regarding the removal of the PIXEL_SIZE conversion. Don't kill the messenger.

> 现在一个Pixel就是一个Unit，对于以后TileMap来说，可能会引起float的精度问题，
对于大一些的TileMap来说，可能会导致画面的抖动。

Basically, not in favor. Use google translate to translate it. It may cause float resolution problem on future tilemap. On bigger tilemap, it may cause screen jitters.

cadaver:
Ok. Since things work now (though the constant is "arbitrary"), I believe I'm going to close this one, sorry for the wasted work! This could be reworked into making a per-drawable or per-resource pixels-per-unit setting. Will possibly take a look into this in the future, but no hard promises.

gogoprog:
Well I'll continue to maintain this branch because I'm not convinced by the issues (yet).

Plus I'm using Urho from code and it is way more straightforward to set a 32x32 collision box on a 32x32 pixels sprite. Same for the camera size.

No problem about this closing ;)

1vanK:
> Box2D is tuned for MKS units. Keep the size of moving objects roughly between 0.1 
and 10 meters. You'll need to use some scaling system when you render your 
environment and actors. The Box2D testbed does this by using an OpenGL viewport 
transform. DO NOT USE PIXELS.

32 pixels is 32 meters in Box2D (with the corresponding behavior of the body), so PIXEL_SIZE need for Box2D also

PR #1217, opened 2016-02-29
STB update with basic HDR loading

dragonCASTjosh:
Urho3D is now able to open .HDR image types.


dragonCASTjosh:
I know the cause of the problem and will look to fix

dragonCASTjosh:
No idea what it is this time

hdunderscore:
Only the crowd navigation failed the appveyor test this time, which is an on-going problem (https://github.com/urho3d/Urho3D/issues/1200). Not sure about travis-ci though.

I'll let @weitjong comment on the changes that were made to combat the CI errors.

I did a quick test, HDR files can be loaded now, although they will be loaded as LDR (low dynamic range). Taking a look through Urho, there is a good amount of work required to get to the end goal, but that's a matter for another PR.

weitjong:
> I'll let @weitjong comment on the changes that were made to combat the CI errors.

PR is in general setup to only uses primary Travis CI and AppVeyor CI builds. Since the offending unit test is just being moved out from the primary AppVeyor CI build very recently in the master branch, your PR branch most probably does not have the change yet. Do a git rebase against the master branch and it should be good.

The Travis CI build shows there is still problem with the last commit in this PR with Clang and MinGW SHARED builds.

dragonCASTjosh:
so AppVeyor now passes but no idea what could be causing the symbols not to be found

dragonCASTjosh:
@weitjong it looks like everything compiled correctly so it should be ready to merge if it meets all the requirements :)

weitjong:
Thanks for the hard work! Usually it is Lasse who makes the important decision of whether to merge or not a contributed PR. The others, including myself, just try to help him on other more mundane tasks. Although we do have a license to kill or merge :smiley: . But since you are asking, I do have my reservation on your changes.

dragonCASTjosh:
That's fair :) I tried to keep changes to a minimum when possible, other then the fun I had fixing the CMakeLists.txt the only other changes was defines and 1 removal from image.cpp I believe it was a prototype but I'd have to check to be sure :)

cadaver:
We did have a non-standard STB setup with both .c & .h files (I believe some of these were manually prepared) so it would be good to get in line with the official structure for easier future upstream updates.

However, I see a lot of unused STB libraries being added. We don't use the STB tilemap editor or line edit for instance.

Also, we did modify stb_image for the SSE / Neon SIMD usage check, so it would be preferable to port that change to the new version.

dragonCASTjosh:
i did plan to remove to the unused libraries before but i guess it drifted over my head when i was fixing things :) .  They should be removed now unless i missed some. I also added the Urho SSE / Neon checks to STB. Let me know if there is anything else.

dragonCASTjosh:
@cadaver Everything should be good for a merge now to my understanding

cadaver:
Yes, I tested it too and it's ready to go in imminently.

PR #1220, opened 2016-02-29
MathDef templates

TheComet93:
This is a change that affects quite a lot of subsystems and files, so I'll try to explain what precisely was changed and why. The main motivation behind this change is the ability to use doubles because these days, on desktop computers, there is no performance advantage for using floats over doubles. This is not the case for Android afaik.

As mentioned on the forums ([http://urho3d.prophpbb.com/topic1903.html](http://urho3d.prophpbb.com/topic1903.html)) all of the math functions force you to use floats.

Overview of changes
----------------------------
* All math functions (except for Random()) are now templated. The default implementation calls sin(), cos(), etc. but there exist template specializations for floats which call sinf(), cosf(), etc.
* AngelScript and Lua bindings expose ```<float>``` specializations (same as before). It might be worth considering a ```URHO3D_USE_DOUBLES``` option to have scripts use doubles instead of floats.
* The math constants M_INFINITY, M_EPSILON, M_MIN_INT, M_MAX_INT, M_MIN_UNSIGNED, M_MAX_UNSIGNED were moved to a "private" file in ```Urho3D/Math/Internal/Script``` and defined using a new template ```Limits<T>``` which acts as a wrapper around ```std::numeric_limits<T>```. All occurrences of these constants were replaced with Limits<T> **except** in script bindings. This change was made because the values of epsilon and infinity depend on the type (float, double, or long double), and the values of integer ranges depend on the type's width.
* AngelScript bindings still expose the constants M_INFINITY, M_EPSILON, etc. However, they are defined using ```Limits<float>```, ```Limits<signed>``` and ```Limits<unsigned>```. The definitions of these "old" constants reside in files ```Urho3D/Math/Internal/Script/ScriptSpecificMathDefs.h``` and ```Urho3D/Math/Internal/Script/ScriptSpecificMathDefs.cpp```.
* The same is true for Lua bindings.
* Due to an issue with MSVC, I've added /DNOMINMAX as a global definition to disable the macros ```min()``` and ```max()``` on Windows, as they conflicted with ```std::numeric_limits<T>::min()/max()```.
* A lot of explicit casting between signed/unsigned types were causing compilation errors with the new templated functions. This was fixed, and is the main cause for the massive amount of files changed.


TheComet93:
Updated original comment with new changes.

weitjong:
Thanks for the hard work. I am sure you have well acquainted with Travis and AppVeyor by now :smiley: . I have canceled the backlog in the AppVeyor just now and only leave CI job from the last commit in the queue. The last commit is looking good so far.

@cadaver As this PR has changed a few fundamental classes, I think it needs to be tested first with the rest of CI jobs for other platforms. That is, merge the PR to a topic branch first instead of master branch.

TheComet93:
Hah, yeah! CI is great.

I will confess I'm not 100% happy with this. It's a step forward, but I've also introduced a slight divergence between the C++ API and the script APIs (```M_MIN_INT``` and co. are defined in scripts but don't exist in C++ any more).

I think I should also rename ```M_LIMITS<T>``` to ```Limits<T>``` since it is a class.

I'll wait to see what @cadaver has to say.

cadaver:
Thanks for this work, indeed it will need to live in a feature branch. Arguably the use of std::numeric_limits is more proper, but the consequence is less readable code, compared to the simple builtin constants. Will meditate on whether to merge this partially, or completely; I believe the main benefit is to stop compile errors for cases like attempting to Min() or Max() unsigned ints, like it was before.

TheComet93:
No problem. Let me know if you want to revert back to the simple builtin constants. They still existed in commit 7d84004 but I'd have to go over the script bindings again.

cadaver:
As long as you don't squash/rebase your branch, I should be able to cherrypick whatever is necessary  and fix the bindings, in case I'd decide for a partial merge.

cadaver:
I took your 2 first commits and did my own mods / cleanup, allowing e.g. Min() & Max() to take 2 template parameters. The work is now in template-math branch but there seemed to be no major issues and so could be merged to master.

cadaver:
Merged.

PR #1225, opened 2016-03-05
Additive animation blending

1vanK:
I try to make additive blending for animations. It works, but flickering every cycle, and I do not know why. Test scene: https://github.com/1vanK/MyFiles/raw/master/Temp/AnimationBlendTest.zip (need add CoreData and Data folders)

### DIFF
--- a/Source/Urho3D/AngelScript/GraphicsAPI.cpp
+++ b/Source/Urho3D/AngelScript/GraphicsAPI.cpp
@@ -1252,6 +1252,10 @@ static void RegisterAnimatedModel(asIScriptEngine* engine)
     RegisterRefCounted<AnimationState>(engine, "AnimationState");
     RegisterStaticModel<AnimatedModel>(engine, "AnimatedModel", false);
 
+    engine->RegisterEnum("AnimationBlendingMode");
+    engine->RegisterEnumValue("AnimationBlendingMode", "ABM_OVERRIDE", ABM_OVERRIDE);
+    engine->RegisterEnumValue("AnimationBlendingMode", "ABM_ADDITIVE", ABM_ADDITIVE);
+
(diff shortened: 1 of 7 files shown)


cadaver:
I'm not sure if subtracting quaternions to get a rotation delta is correct. However I don't have an answer what should be done in its place.

1vanK:
It seems that
```
            if (channelMask & CHANNEL_ROTATION)
            {
                Vector3 delta = keyFrame->rotation_.Slerp(nextKeyFrame->rotation_, t).EulerAngles() - stateTrack.bone_->initialRotation_.EulerAngles();
                Vector3 newAngles = node->GetRotation().EulerAngles() + delta;
                node->SetRotationSilent(Quaternion(newAngles.x_, newAngles.y_, newAngles.z_));
            }
```
is working. But there are too many conversions. This can cause problems?

cadaver:
The Euler conversion might not be unambiguous, or it may throw the rotation to different axes at singularity points. I recommend checking if Ogre has some code you would be able to take inspiration from, as its animations are all additive.

1vanK:
I found the solution here http://forum.unity3d.com/threads/quaternion-how-to-compute-delta-angle-for-each-axis.242208/
```
Quaternion delta = keyFrame->rotation_.Slerp(nextKeyFrame->rotation_, t) * stateTrack.bone_->initialRotation_.Inverse();
node->SetRotationSilent(delta * node->GetRotation());
```
It works. But in Node::Rotate() I see normalization:
```
rotation_ = (delta * rotation_).Normalized();
```
I tried both with and without normalization and do not see any difference. Is normalization the necessary?

cadaver:
It's a safeguard so that the quaternion won't eventually become skewed (non-normalized) after many successive rotates.

1vanK:
also updated test scene

hdunderscore:
Nice !

cadaver:
Looks good! Do you plan to add the change also to AnimationController? It's OK if you don't, it can also be added later now that the base work exists.

cadaver:
Merged with some changes. Added network serialization of blend mode to AnimationController.

1vanK:
Thank you!

PR #1239, opened 2016-03-13
Urho2D Sprite2D border fix

OvermindDL1:
Two commits, the first adds a member variable to Sprite2D that defines an amount to offset a texture 'inwards' to fix texture bleeding issues, which is absolutely paramount when not using full textures, like via a SpriteSheet, and should consequently also resolve the need to set textures to CLAMP'd mode.

The second commit removes the Sprite2D code and comment that force overrides a materials UV mode to CLAMP as that issue should not be resolved as well.  However if this is not wanted then state and I can delete it, but I left it as a secondary commit so it is obvious in the git logs in case a rollback if ever wanted on just that part.

### DIFF
--- a/Source/Urho3D/Urho2D/Sprite2D.cpp
+++ b/Source/Urho3D/Urho2D/Sprite2D.cpp
@@ -169,8 +169,8 @@ bool Sprite2D::GetTextureRectangle(Rect& rect, bool flipX, bool flipY) const
     rect.min_.x_ = rectangle_.left_ * invWidth + edgeOffset_;
     rect.max_.x_ = rectangle_.right_ * invWidth - edgeOffset_;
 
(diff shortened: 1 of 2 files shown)


cadaver:
This has the effect of not sampling the whole image, which is probably not always wanted either. It's also somewhat unclean to have a property in Resource which it cannot load on its own, but must be programmatically set. However it's no problem for me to take this in and set the default value for the border to 0, therefore retaining the original behavior.

Also, this has no script bindings. Generally it's preferable to have PRs that can go in with no further work needed from the core devs, but in this case it's not that big deal as this has to be modified anyway.

Finally I must note that I can't bring myself to care that much of the Urho2D classes overall, but if this helps you, then good.

OvermindDL1:
I am starting to agree about the Urho2D part, it needs an overhaul if it is to exist...

cadaver:
Merged with changes. Thanks!

PR #1243, opened 2016-03-15
Fix ToolTip border; EditorHierarchyWindow: tabs to spaces, fix ResetButton ImageRect

1vanK:
In different xml files uses different chars for indent (tabs or spaces), so I don't know which style is correct. In this case, my editor decided for me :)


cadaver:
When the UI editor has last been used to save a layout, then it would use tabs (just to save on filesize) so this is where it differs from the code's convention.

1vanK:
I could return the Table, if necessary

cadaver:
There are mixed tabs & spaces (mixed editing in either text editor or the layout editor) in the UI directory anyway, so it's no problem.

PR #1245, opened 2016-03-15
Application wide global variable implementation

BlueMagnificent:
This extends the AngelScript scoped global variable implementation (globalVars) to cover the entire application


weitjong:
Nice!

1vanK:
But variantMap has more, than two methods :)

1vanK:
In C++ I just register my own subsystem, if I need global variables and functions

weitjong:
This PR is still nice to bring the feature parity to Lua scripting environment and C++.

BlueMagnificent:
@1vanK , the idea leans more towards being a global variable storage than a global VariantMap. As it is, the application has no idea of how the global variable is implemented within the context, all it sees are the getter and setter methods

cadaver:
Good work, thanks! Will compare this to the node / UIElement vars implementation, there the functions are called SetVar() / GetVar(), so the global var access function names could be shortened too.

1vanK:
Also UIElement has
```
const VariantMap& GetVars() const { return vars_; }
```

cadaver:
True. In AngelScript it's exposed as a modifiable VariantMap, since in practice that doesn't do harm to allow direct non-const access, so the same could be done for these, as a global property.

BlueMagnificent:
Cool. Would do some refactoring right away and push the commits.

cadaver:
Please go ahead. We can also do further adjustments before merge, if necessary.

BlueMagnificent:
Quite a couple of changes have been effected. Shortened function name to GetGlobalVar() and SetGlobalVar(). That seems to be the shortest I could get that conveys the meaning :( . globalVariables_ was also renamed to globalVars_ and exposed as a global property to AngelScript as rightfully suggested.

PR #1248, opened 2016-03-15
Editor fixes

1vanK:


1vanK:
Need "fr" and "it" localizations

1vanK:
>[ 79%] ../../../lib/libUrho3D.a(Addons.cpp.o): In function `Urho3D::RegisterString(asIScriptEngine*)':
Addons.cpp:(.text+0xa012): undefined reference to `Urho3D::String::NPOS'

It is strange,but VS compiling it successfully and I don't know what went wrong

cadaver:
Since it's initialized in the class header, it may not have an address on GCC (treated similarly to enum.) Probably easiest workaround is to have an additional static const variable in Addons which is initialized to the same value.

1vanK:
reminded https://qph.is.quoracdn.net/main-qimg-c5c88a9decac1c00ff4822ae520d1815 xD

cadaver:
Merged.

PR #1253, opened 2016-03-17
Applied patch to fix undefined reference error when compiling for Wayland on Linux.

GKR:
Applied patch to fix undefined reference error when compiling for Wayland (1.10) on Linux.

This has been fixed in SDL, but no SDL release has been made that includes this fix.

The SDL fix can be found here:
https://hg.libsdl.org/SDL/rev/330f500d5815

In the mean time, this fixes Urho3D compilation on Linux.

```
[ 74%] Linking CXX executable ../../../bin/Urho3DPlayer
../../../lib/libUrho3D.a(SDL_waylandvideo.c.o): In function `display_handle_global':
SDL_waylandvideo.c:(.text+0x593): undefined reference to `wl_proxy_marshal_constructor_versioned'
SDL_waylandvideo.c:(.text+0x5d4): undefined reference to `wl_proxy_marshal_constructor_versioned'
SDL_waylandvideo.c:(.text+0x61d): undefined reference to `wl_proxy_marshal_constructor_versioned'
SDL_waylandvideo.c:(.text+0x684): undefined reference to `wl_proxy_marshal_constructor_versioned'
SDL_waylandvideo.c:(.text+0x6cc): undefined reference to `wl_proxy_marshal_constructor_versioned'
../../../lib/libUrho3D.a(SDL_waylandvideo.c.o):SDL_waylandvideo.c:(.text+0x6fb): more undefined references to `wl_proxy_marshal_constructor_versioned' follow
Source/Tools/Urho3DPlayer/CMakeFiles/Urho3DPlayer.dir/build.make:95: recipe for target 'bin/Urho3DPlayer' failed
make[2]: *** [bin/Urho3DPlayer] Error 1
CMakeFiles/Makefile2:1282: recipe for target 'Source/Tools/Urho3DPlayer/CMakeFiles/Urho3DPlayer.dir/all' failed
make[1]: *** [Source/Tools/Urho3DPlayer/CMakeFiles/Urho3DPlayer.dir/all] Error 2
Makefile:149: recipe for target 'all' failed
make: *** [all] Error 2
```

### DIFF
--- a/Source/ThirdParty/SDL/src/video/wayland/SDL_waylanddyn.h
+++ b/Source/ThirdParty/SDL/src/video/wayland/SDL_waylanddyn.h
@@ -79,6 +79,7 @@ void SDL_WAYLAND_UnloadSymbols(void);
 #define wl_proxy_get_user_data (*WAYLAND_wl_proxy_get_user_data)
 #define wl_proxy_add_listener (*WAYLAND_wl_proxy_add_listener)
 #define wl_proxy_marshal_constructor (*WAYLAND_wl_proxy_marshal_constructor)
+#define wl_proxy_marshal_constructor_versioned (*WAYLAND_wl_proxy_marshal_constructor_versioned)
 
 #define wl_seat_interface (*WAYLAND_wl_seat_interface)
 #define wl_surface_interface (*WAYLAND_wl_surface_interface)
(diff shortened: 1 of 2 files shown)


cadaver:
@weitjong What is the best approach now? To cherrypick this commit from the SDL mirror and update the subtree, rather than applying this PR directly?

weitjong:
I actually did not have problem building with wayland enabled on my Fedora. I will double confirm that again later tonight.

weitjong:
Ok. I just checked why I don't have issue with my build with Wayland enabled. My Fedora 23 still uses Wayland 1.9. Wayland 1.10 was just released last Feb. It looks like we will need this fix sooner or later. I don't think there will be newer SDL release by the time 1.10 version reaches mainstream Linux distros in a few months time, besides user brave enough could already get 1.10 now.

@cadaver Back to your original question. It really does not matter. You can commit to the subtree and do a ```subtree pull``` to pull it into master branch; or make a commit in the master branch by applying the PR and then later do a ```subtree push``` to push those changes in the Source/ThirdParty/SDL back to the subtree. Git will handle that transparently for us. In this case, I think it would be easier for core team to just always commit to master branch as per normal. Let the SDL package maintainer worries about the ```subtree push``` at the time he/she fits. Probably the push can be done just before every SDL upgrade exercise.

GKR:
Right, this applies to Wayland 1.10 which is what e.g. ArchLinux ships from their current package repository.

cadaver:
Ok, thanks for the clarification.

Issue #1260, opened 2016-03-19
DebugHud/Profiler average and frame values are weird


damu:
Something is wrong with the average and the frame values.
The frame value is always the same as average:  
![](http://i.imgur.com/Xautuy0.jpg)  
What is it supposed to display?
I'm optimizing a function and at first it looked like optimizing that function made other functions slower.
This is the normal version of ClearImage (Image::Clear): [http://i.imgur.com/Xautuy0.jpg](http://i.imgur.com/Xautuy0.jpg)
And this my optimized one: [http://i.imgur.com/rhRPt0m.jpg](http://i.imgur.com/rhRPt0m.jpg)
It looks like the optimization made SetTextureData (directly above ClearImage) much slower, which can't be. I'm just setting the bytes a bit differently. Also tried that outside Clear() and by measuring with my own timer, only the clearing gets much faster and SetTextureData does not change on a per call basis.

damu:
Oh my own timer actually reported ~15ms for the unoptimized Clear() which fits to the ~15ms displayed in the profiler. It show something <1ms for the optimized one which also fits to the 0,678 in the profiler. So these values seem to be correct. But why is SetTextureData seemingly getting slower in the profiler but not in my own measure function? Could it be called multiple times outside my measurement due to the slightly different Clear()?

weitjong:
Are using vsync which locks the engine at 60 frames per second?

damu:
My driver was configured to always use VSync. When switching that off and also in Urho to off I get around 200FPS and the SetTextureData is below 0.6ms. As soon as Urho's or the drivers VSync is on the FPS are around 60 and SetTextureData 13-16ms.
Weird.

weitjong:
Even 200 is an artificial cap. Try running with -nolimit and you could even get pass that with really simple scene. As for your SetTextureData thingy, I have no idea without looking into it deeper.

damu:
With -nolimit the FPS are ~450 and SetTextureData is around 0.47 ms. Oh maybe a SetTextureData is measured as measure point before sleeping due to the FPS limit and the measurement of that point is only stopped in the next frame? So that the "FPS-wait" gets added to the SetTextureData time?

cadaver:
The values are calculated as follows. There is a certain measurement interval, which is 1 second by default, so when a function is called once per frame, avg and frame (which really means avg per frame) end up being the same:

                float avg = block->intervalTime_ / block->intervalCount_;
[...]

PR #1262, opened 2016-03-19
Optimized Image::ClearInt() for the four component case (ARGB). In my ca…

damu:
…se this optimized ClearInt() only takes ~0.362ms compared to 14.853ms before, see http://i.imgur.com/8O7V77o.jpg vs http://i.imgur.com/qjhkaWw.jpg.
~~There could be an issue, see #1260.~~

### DIFF
--- a/Source/Urho3D/Resource/Image.cpp
+++ b/Source/Urho3D/Resource/Image.cpp
@@ -1114,9 +1114,20 @@ void Image::ClearInt(unsigned uintColor)
         return;
     }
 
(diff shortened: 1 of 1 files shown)


damu:
#1260 doesn't seem to be related to this change. That seems to be a Profiler bug related to FPS limitation.

cadaver:
Good work! Can you please fix the code convention to match?

damu:
Like that?

PR #1266, opened 2016-03-20
Convenience method Context::CreateObject<T>()

r-ku:
So we can create objects like:

    context_->CreateObject<MyType>()

Just like creating components.


1vanK:
what about SharedPtr(new MyType()) ?

PR #1270, opened 2016-03-21
Removed unnecessary check from Serializable::SaveJSON, closes #1265

### LINKED ISSUE #1265: Serializable::SaveJSON dest.IsNull()
I'm not sure if `if (dest.IsNull())` check is necessary in `bool Serializable::SaveJSON(JSONValue& dest) const`. It could be copied from SaveXML method by mistake. One has to assign a dummy object to the JSONFile root in order to use SaveJSON() method, because it returns false otherwise.
Also possibly code could be changed so that JSONFile acts more similarly to XMLFile, allowing to set the default node name.

rikorin:


cadaver:
Already committed a fix, which also includes similar remove from UnknownComponent.

rikorin:
I did a search, but for some reason missed that one.

PR #1271, opened 2016-03-21
BillboardSet::UpdateVertexBuffer optimization

damu:
Made the sine calculations there much faster in two steps. That brought the big loop there down from 22.03ms to 8.25ms and the total FPS from 20 to 24 in my test scene.


cadaver:
Excellent! But please add the spaces as in the rest of Urho code.

damu:
Those four spaces? Thought I had enough but missed those.

cadaver:
Thanks!

1vanK:
It is great! Maybe place sincosf_fast to Math and using somewhere else?

cadaver:
Yes, that makes sense, since it is likely to be needed by other code in the future. Can do that when merging. Is there btw. any explicit license stated for code on that page?

damu:
There's no license on that page. It's a public blog post so one could assume public domain and also that bit might not be enough to fulfill the "Threshold of originality" criteria anyway.

Also I'm not sure if the inprecision might be a problem. Seems fine in that case but might not be in others. That page had also a slightly slower version (30% slower) with a way higher precision.

damu:
Oh BTW I was really surprised that those two sinf and cosf calls pulled that memory intense loop so much down. For testing I replaced them with 1 and 0 and that whole loop in total per frame went from 24.4 ms to 2.8ms.

hdunderscore:
Can't assume permissive when it comes to licenses, wouldn't hurt to ask. Related: http://blog.codinghorror.com/pick-a-license-any-license/

1vanK:
> Without a license, the code is copyrighted by default. People can read the code, but they have no legal right to use it. To use the code, you must contact the author directly and ask permission.

;(

1vanK:
On the other hand it is not pateded algorithm and we can REWRITE the code :)

cadaver:
Ah, now I understood that sincosf is not part of the standard so it isn't found in e.g. VS2015, and thus we'd rather need the optimized variant. Please ask permission from the original author so that we can move forward with this.

weitjong:
I don't think Michael is the original author of the idea. In his blog he clearly stated that he got the idea from http://www.devmaster.net/forums/showthread.php?t=5784 (which is a broken link unfortunately). However, in my quick "research", I have a reason to believe that the original author of t=5784 was a guy named Nick and he has even improved the algorithm further here http://forum.devmaster.net/t/fast-and-accurate-sine-cosine/9648. In the improved version the branch has been eliminated. The constants in this PR is also being explained there, so they could be more properly defined using 4/PI instead of just a magic number. It seems that the extra precision worth the extra effort to get the better approximation as it just reuses the terms that already computed in lower precision version.

Arguably, we could say this is just math. We could approximate the sine function with quadratic equation ourselves and come up with similar conclusion. We could also approximate it with cubic equation and get another one (see http://allenchou.net/2014/02/game-math-faster-sine-cosine-with-polynomial-curves/).

Stinkfist0:
FWIW, MathGeoLib for example has SinCos() implementation and suitable (?) license https://github.com/juj/MathGeoLib/blob/ff2d348a167008c831ae304483b824647f71fbf6/src/Math/MathFunc.cpp

cadaver:
The MGL code should be good when SSE is enabled, and afaik that isn't an approximation, but just using the SSE2 instruction set for a speed-optimized result. I could profile that + the approximation approach later today.

cadaver:
Tested on VS2015 by setting vsync on and modifying sample 07 to use a large amount of billboards. Here are the timings of BillboardSet::UpdateGeometry

MGL SSE SinCos frame 0.5ms interval 30ms
Approx SinCos (the code from PR) frame 0.57ms interval 34ms
sinf + cosf 0.45ms interval 27ms

So on MSVC it would seem that just calling sinf + cosf separately is already fastest. We could make a function SinCos that would use sincosf if available and separate sinf + cosf on MSVC. This way it's also just as accurate as the standard library provides, instead of our manual approximation.

1vanK:
maybe damu tested debug version?

1vanK:
My test: http://pastebin.com/6MDvkyza

Result:
MSVC: 2276
FAST: 1763

So, approximated version is more fasted

1vanK:
This is debug version

Release version:
FAST: 17
MSVC: 17

Hard to testing :)

1vanK:
Modified version fo release version: http://pastebin.com/bpykwSNs

MSVC: 1718
FAST: 1708

So, difference is only in debug version

1vanK:
It seems, so MSVS just remove unused code from cycles in release configuration so the results are the same

Correct test for release conf: http://pastebin.com/XfUz3mnz

MSVC: 1865
FAST: 1029

Approximated version is more fast

cadaver:
It would look more credible that the approx. version (when testing just the math speed) is indeed faster. My timings included the vertex buffer update.

damu:
I didn't test on Debug and I also benchmarked the sincosf vs sincosf_fast alone outside of Urho, 
Did both on MinGW. CMake build it with -O3 -DNDEBUG -ffast-math -msse2.
Oh in the 1vanK pastebins you used clock() which only counts the process time and I'm not sure if sincosf involves some system (library) call(s) which would not count into clock().
I tested it again:
0.437025        <- sincosf
0.122007        <- sincosf_fast
http://pastebin.com/BcpAdm02

I also worked on making my own cubic equation that directly uses degrees instead of PI. That would solve any question regarding copyright and save one multiplication and one division. It's not finished yet though, as I seemed to have made a small error somewhere in my calculations by hand, the cubic function is close though, just the amplitude is a bit to small. Tried online solvers but none worked.
It's surprisingly hard to get back into math stuff like equations done so long ago.

Haven't yet looked into those links weitjong found.

damu:
On http://allenchou.net/2014/02/game-math-faster-sine-cosine-with-polynomial-curves/ "Allen Chou" says that VS's fmod is as slow as sin which would explain cadavers result.

cadaver:
No, we don't (at least currently) require the user to input angles in a specified range.

weitjong:
I wonder how good the approximation is. Instead of creating a program that step through from 0 to pi and then compare the real sine(x) and sine_approx(x) to get the average error or max error, I figure I could just calculate the net are under the curve represented by the function in question using integral calculus and compare the result.

**baseline**
image

**quadratic (Nick)**
image

**squared quadratic (Nick)**
image

**cubic (Allen)**
image

The winner is clear.

damu:
Finished my cubic function that works with degrees instead of radians. Haven't finished testing yet but looks good so far:
```
1.7361  <- sinf cosf
0.595034        <- sincosf
0.19001         <- sincosf_fast
0.170009        <- SinCosfFast (with degrees instead of radions)
0.149008        <- SinCosfFast2 (with degrees instead of radions and better branching)
```
I also came up with some branching optimizations.

damu:
Pushed a further optimization. The loop in UpdateVertexBuffer is now around 7.78ms from the previous 8.25ms.
Math benchmark (similar to the one before) results with `for(float degree=-720.0f;degree<=720.0f;degree+=0.0001f)`:
```
1.7231s    sinf cosf
0.598034s  sincosf
0.192011s  sincosf_fast this is the previous one
0.171009s  SinCosfFast  this is improved by using degrees and not having to convert deg->rad
0.132007s  SinCosfFast2 this is the just pushed one with additionally restructuring
```

weitjong:
With all due respect but I think it is not a valid approximation anymore. It is easy to prove that the error is too high now compare to the actual sine function. Those trigonometric functions love pi and radians for a very good reason, and I think it is also the case with their approximation knock-offs. Nice try though.

damu:
The same approach can be done with the higher precision version that that guy described. I'll do that but that precision difference shouldn't really matter in that cause because it's just rotating the billboard. I just compared some values and the highest difference was 0.06.

weitjong:
Try 45 degree. That one point alone has error almost reaching half of the
net error (from 0 to pi) of the original quadratic approximation.

Best regards,
YaoWT

weitjong:
I guess it is not fair though because that point is the where the max error
would occur ;-).

Best regards,
YaoWT

damu:
Hm 45 degree is here 0.75 with all my fast functions and with sinf and cosf 0.707107.
A higher difference is for example 18 degree with  0.36 vs 0.309017 (both sin).

weitjong:
> Oh and about pi and radians: this degree version is practically just prescaled/premultiplied which does not affect the precision but saves the degree->radians conversion.

My original intuition told me that that should not work and I just cross-check on 45 degree and then calculate the net area under the curve as I did before and found it to be unacceptable. But I take back what I just commented earlier about the validity of the approximation. I was erred on reading the integral calculation. I forget to account for the domain change (radian to degree). And with that adjusted, your net error is about the same as original quadratic approximation, as you have expected. My apology.

weitjong:
> Hm 45 degree is here 0.75 with all my fast functions and with sinf and cosf 0.707107.
A higher difference is for example 18 degree with 0.36 vs 0.309017 (both sin).

It goes to show that I could not trust my intuition.

damu:
>> Hm 45 degree is here 0.75 with all my fast functions and with sinf and cosf 0.707107.
>> A higher difference is for example 18 degree with 0.36 vs 0.309017 (both sin).

> It goes to show that I could not trust my intuition.

I thought you used your integral calculations to calculate that.
I only looked at my `for(float i=-720;i<=720;i+=90.0f/15)` list of values that I use to check and compare.
I also remember that integral stuff from school but I'd like to save that work as I always had my fight with it. ;-)

cadaver:
I have made a branch "sincos" which uses sincosf when available. Will merge this soon. After that billboard performance in MinGW and MSVC should be about the same speed. I don't think I'll be merging sin/cos approximation functions for now. But thanks to @damu for the initial work and bringing the performance into attention.

damu:
... too imprecise for your taste?

I started working on trying the high precision version from Michael. My math skills are failing me for the newest degree version and I have no idea how he found those values. The speed though is 0.168009s in my math benchmark, so >10 times faster as sinf and cosf separately.

With taking his values (radians) the precision looks quite good: http://i.imgur.com/b8LEPUY.jpg
The speed is 0.230013s compared to 0.616035s with sincosf or 1.7321 with separate sinf and cosf.

cadaver:
We don't really use approximations in Urho so far (for example fast sqrt as in Quake) so it's a question of consistency, and also considering that the performance of fmodf can vary on different compilers, so we could end up with actually worse performance.

cadaver:
Also one thing that comes to mind is to just pass the angle in the vertex buffer and let the shader calculate sin & cos.

Modanung:
To get fast sines I use a Vector of floats as a look-up table. It is generated when the program starts - but I think this should be possible during compile time - and after that you get pretty accurate (depending on the resolution) and fast results. Quarter of a period would be enough, which could be mirrored to fill in the rest.
Would this be a good way to have fast sines built into Urho? Since sines are pretty versatile and useful for game world control.

damu:
I also thought of using a look-up table with a resolution of for example 1 degree. It would have to be interpolated. I have the feeling it would be slower as one of the variants from/based on Michaels work but it is worth to benchmark and to compare the precision. With caching it might actually be faster.

But it would be also an approximation (and require fmod) which cadaver might not approve?

In general a speedup there and maybe in other places would be good of course. It's a design decision and I also thought of the possibility of making such faster but more inaccurate versions optional via Urho CMake options. That may not be that important for this case but there might be other places where an inaccurate approximation makes a higher performance impact which might be desired.

Oh I also saw a talk (here: https://youtu.be/Nsf2_Au6KxU?t=41m57s) by a Valve developer who worked at "Left for Dead" and he showed a fast and approximated sincos using SIMD, so it seems that Valve uses such things. I also heard of Unreal having various fast approximations.

Modanung:
Moved the discussion to [the forums](http://urho3d.prophpbb.com/topic2012.html), since it's a bit off-topic.

weitjong:
I am using GNU Octave to plot those curves to visualize the approximation error. The first graph covers the range from -pi to pi. It is easier to see why the approximation approach is only good for the range from 0 to pi and has to mirror the other half programatically.
screenshot from 2016-03-28 22-59-00
The second graph zooms in on this positive range and one can easily see the squared quadratic approximation is right on the money.
screenshot from 2016-03-28 23-01-34

PR #1272, opened 2016-03-21
AssetImporter can import a subset animation frames.

neat3d:
Example:
c:/>AssetImporter model xxx.fbx xxx.mdl -split 10 40

You will get a .ani file only with subanimation from frame 10 to frame 40.


ffpwserver:
Nice work guy ^_^

cadaver:
Merged, thanks! Btw. for future reference, enable showing whitespace characters in your editor / IDE and use spaces not tabs to match the existing Urho3D code.

PR #1274, opened 2016-03-23
Moved container begin and end from std to Urho3D namespace

Modanung:
This makes range based for loops work for me.
When begin and end are in the std namespace I get compiler errors saying they have not been declared.

### DIFF
--- a/Source/Urho3D/Container/HashMap.h
+++ b/Source/Urho3D/Container/HashMap.h
@@ -719,11 +719,6 @@ private:
     unsigned Hash(const T& key) const { return MakeHash(key) & (NumBuckets() - 1); }
 };
 
-}
-
-namespace std
-{
-
 template <class T, class U> typename Urho3D::HashMap<T, U>::ConstIterator begin(const Urho3D::HashMap<T, U>& v) { return v.Begin(); }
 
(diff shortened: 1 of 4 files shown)


1vanK:
I think it owerride std::begin() and std::end() for **Range-Based For** if user code not in Urho3D namespace 

http://www.drdobbs.com/cpp/auto-types-and-range-based-for-statement/232900460

Modanung:
It would probably have been better to have made these changes in a topic branch, wouldn't it?

damu:
It's supposed to overwrite begin and end with the Urho container types as parameters like
`begin(const Urho3D::HashMap<T, U>& v)`
Range based for loops work by looking up t::begin() and t::end() and if those don't exist it looks for begin(const T&) and end(const T&).
They should work when in a namespace that is being used like `using namespace std;` (I think). Putting them outside a namespace is better though as then you don't have to use `using namespace NAMESPACE;`.

Modanung:
From [StackOverflow](http://stackoverflow.com/a/31457319/3618748):
> The way to make a for(:) loop work on your type X is now one of two ways:
> - Create member X::begin() and X::end() that return something that acts like an iterator
> - Create a free function begin(X&) and end(X&) that return something that acts like an iterator in the same namespace as your type X.

Modanung:
I'll create a new pull request using a branch.

PR #1280, opened 2016-03-24
Update pugixml from 1.5 to 1.7

cosmy1:


cadaver:
Thanks! Will have to see if the Urho3D modifications to the previous version are no longer necessary. If it was only the xpath_node copy constructor, then it seemed to be already included.

However, when you update a library, please take the time to update Urho documentation referring to version numbers at the same time: README.md and Docs/Urho3D.dox.

cosmy1:
Didn't know, I'll do it right now

PR #1281, opened 2016-03-24
Add missing Italian localization entries

cosmy1:


cadaver:
Can you remove the pugixml update commit, since this is then easier to apply?

cosmy1:
Someone can translate those to French?

PR #1282, opened 2016-03-24
Add missing Italian localization entries

cosmy1:


cosmy1:
Someone can translate those to French?

@cadaver I did a mistake and removed the commit from the other PR so I made a new one

cadaver:
There are commas missing after the Russian translations, making it not work as a JSON array as intended.

cosmy1:
Yes, right! Should work now

PR #1283, opened 2016-03-24
Fix signed/unsigned mismatch in FontFaceFreeType.cpp

cosmy1:


cadaver:
When casting numbers, the existing Urho code has preferred simple C-style casts.

cadaver:
Thanks!

PR #1284, opened 2016-03-25
Moved container begin and end from std to Urho3D namespace

Modanung:
This replaces #1274, now using its own branch.


cadaver:
Tested on both GCC and MSVC (and the method you use should be just as valid according to standard) so looks safe to merge. Thanks!

cadaver:
Merged.

Modanung:
Cool!

PR #1287, opened 2016-03-26
(Web) Urho3DPlayer lua sample from query string

hdunderscore:
(Web) Added ability to select a lua sample in Urho3DPlayer via the browser address query string, eg: Urho3DPlayer.html?sample=4 will load the LuaScripts/04_StaticScene.lua sample. By default and in error cases, 01_HelloWorld is shown.

### DIFF
--- a/Source/Tools/Urho3DPlayer/Urho3DPlayer.cpp
+++ b/Source/Tools/Urho3DPlayer/Urho3DPlayer.cpp
@@ -38,6 +38,10 @@
 
 #include <Urho3D/DebugNew.h>
 
+#ifdef __EMSCRIPTEN__
+#include <emscripten/emscripten.h>
+#endif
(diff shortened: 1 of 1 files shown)


weitjong:
I have a reservation for this PR. Firstly, the list of samples should not be hard-coded into the source code. It is a maintenance headache. Secondly, as it is now the Urho3DPlayer is already capable of doing the query parameter parsing. Try running this locally:
```
firefox 'Urho3DPlayer.html?LuaScripts/06_SkeletalAnimation.lua&-w'
```
For some reason this does not work on our online web sample, however. I don't have time to look into detail yet but I think it is because our server is static.

hdunderscore:
Ah thanks for pointing out the existing option! In that case I'll close this.

PR #1288, opened 2016-03-26
Initialize correctly all DetourCrowd and DetourTileCache members

cosmy1:


hdunderscore:
Nice catch, do you know if this was the cause of the build test crashes?

cosmy1:
To be honest I don't know if this was the cause, maybe @weitjong can check and tell us, as the sample causing the segfault seems disabled from AppVeyor CI builds on x64 platforms.

weitjong:
The test in question is only being moved from the primary AppVeyor CI job to so-called secondary CI job. The crowd navigation demo still occasionally crashing till today in the secondary CI job. Interestingly it just happened on 64-bit build using MSVC compiler and not others. Some of us has tried including Lasse has tried to reproduce the segfault locally, but to no avail. So, I recon that fixing the segfault is not the main reason why you have made this PR. Nevertheless, who knows it might actually fix it. Thanks for the PR.

weitjong:
Still segfaulting on crowd demo. https://ci.appveyor.com/project/weitjong/urho3d/build/737/job/26s05id10m5hxxa7

cosmy1:
Tried different ways to reproduce it locally but without success.
There are 2 samples that use the navigation mesh but just CrowdNavigation fails.

My idea is to cut code lines from sample till we found the issue.
If you want, I can make few CrowdNavigation samples for you to test on AppVeyor.

weitjong:
Be my guest. I wonder if the issue only manifests itself when the host is under stress or some kind of resource constraint. Not sure how to simulate such conditions on a Windows system.

cosmy1:
Are tests executed in a precise order? If yes, what happens if you swap order and put that sample to be the first tested.
Maybe the host is under stress, what are AppVeyor server specs?

weitjong:
The tests are executed in the precise order. In the secondary AppVeyor CI jobs, the setup has been altered such that all the rest samples are being excluded. What remains are Urho3DPlayer and 39_CrowdNavigation. I don't think changing the order of execution matter much though.

The AppVeyor VM specs is stated here. http://www.appveyor.com/docs/build-configuration. We are using the Open-Source (Free) configuration. Initially our free account only run in shared instance with 1.7 GB and it was so slow but thanks to Feodor Fitsner it has been upgraded to full 1 core with 3.5 GB.

cosmy1:
Here is the first sample:
https://www.dropbox.com/s/j91lsvoyay3aa3w/39_CrowdNavigation_01.zip?dl=0
Its basically the Navigation sample. If this fails, there is an external factor that make the test to fail, if not, we will proceed with the next sample. 
I just want to be sure the code in CrowdNavigation have something that triggers the segfault.

weitjong:
Thanks. Will have a closer look later.

PR #1292, opened 2016-03-29
Aggregate initializer constructors for container types

r-ku:
Discussed [here](http://urho3d.prophpbb.com/topic1988.html). This includes only constructors, `SendEvent()` with const `VariantMap` is not included. They are only active if urho3d is built with `-DURHO3D_C++11` and are disabled on compilers that do not use c++11.

Also piggy-backed few small commits that seemed not worthy of their own PR:

* Added `.directory` files to `.gitignore` as KDE/plasma stores folder settings in there.
* Added `URHO3D_API` to `ValueAnimationInfo` class declaration so it is exposed in shared library.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,3 +94,4 @@ configure*
 *.log
 *.bak
 Thumbs.db
+.directory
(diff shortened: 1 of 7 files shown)


cadaver:
Merged. Thanks!

PR #1293, opened 2016-03-30
Fix URHO3D_SSE definition to allow toggling SSE in Bullet

cosmy1:


cadaver:
URHO3D_SSE is already defined in Urho3D.h, but naturally Bullet code doesn't see it there.

@weitjong What do you think as the preferable way to fix this?

cosmy1:
I think including Urho3D.h into Bullet is a bad practice, but ofcorse is more an aesthetic thing than functional. There is a also the possibility to add URHO3D_SSE specific to the Bullet project by adding the definition to its CMake script.

If you want I can move URHO3D_SSE to Bullet CMake.

weitjong:
This is a regression issue. The compiler define was set globally in the previous version up to 1.5. After 1.5, we have introduced the so-called auto-discovery feature by baking some of the compiler defines into our export header but we have overlooked that some of Urho3D-specific may have been used in 3rd-party library source codes as well. I have just done a quick scan on all the 3rd-party library source code to see how bad the situation is. Fortunately I could only find this.
```
[weitjong@igloo ThirdParty]$ git grep URHO3D_ |grep -v CMakeLists.txt
Bullet/src/BulletCollision/Gimpact/gim_memory.h:#if defined(URHO3D_SSE) && defined(_WIN32) && !defined(__CYGWIN__)
Bullet/src/LinearMath/btScalar.h:#elif (defined (URHO3D_SSE) && defined (_WIN32) && (_MSC_VER) && _MSC_VER >= 1400) && (!defined (BT_USE_DOUBLE_PRECISION))
Bullet/src/LinearMath/btScalar.h:    #if defined (URHO3D_SSE) && (defined (__i386__) || defined (__x86_64__))
Bullet/src/LinearMath/btScalar.h:    #elif defined (URHO3D_NEON) && defined( __ARM_NEON__ )
STB/stb_image.h:// Urho3D: do not use SIMD instructions if both URHO3D_SSE and URHO3D_NEON are disabled
STB/stb_image.h:#if !defined(URHO3D_SSE) && !defined(URHO3D_NEON)
```

We need to care of both URHO3D_SSE and URHO3D_NEON compiler defines in Bullet and STB. Adding URHO3D_SSE to the Bullet's CMakeLists.txt could be the last resort but I will first try something else, hopefully later today. Thus we will probably not merge this PR as it is. Thanks for bringing this to our attention though.

cosmy1:
Closing this as will be done in a different way.

cosmy1:
Btw, STB headers are directly used into Urho3D so they have access to URHO3D_SSE and URHO3D_NEON, just Bullet is the issue.

weitjong:
I am considering also baking the URHO3D_NEON, but you are right the impact is not the same on STB.

weitjong:
Update: instead, I have removed the URHO3D_NEON build option all together as it was ill-conceived. It was meant to capture "user intention" similar to URHO3D_SSE, however, in this case it is totally unnecessary because users already indicates their intention by specifying which ABI they want to use. Besides, it is not possible to turn the NEON to off for iOS platform.

cosmy1:
So what is your plan for Bullet SSE option?

weitjong:
it is still in the refactor-simd branch.

cosmy1:
Ok got what you had on mind. Waiting for it to be merged.

weitjong:
The branch is ready to be merged. Can you help to double check it on your side. Thanks.

cosmy1:
Ok, going to test it later. I'll let you know.

cosmy1:
Tested and these are my results under Visual C++ 2015 compiler.
Targeting x86 platform CMake shows option to toggle SSE but its not effective on Bullet as BT_USE_SSE and BT_ALLOW_SSE4 are always enabled.
Targeting AMD64 platform CMake doesn't show SSE option and BT_USE_SSE and BT_ALLOW_SSE4 are enabled.

weitjong:
Thanks for your time. The SSE on 64-bit is now always on. So, no more build option is the new expected behavior. As for 32-bit VS, I will double check. I haven't done any tests on VS, to tell you the truth.

cosmy1:
The issue is _M_IX86_FP. Here is the MS doc for it:
https://msdn.microsoft.com/en-us/library/7t5yh4fd.aspx

To force SSE off /arch must be set to IA32.

weitjong:
Yes, I have already read about that and therefore the code looks like this in the branch. https://github.com/urho3d/Urho3D/blob/5876e9d4a23c6dd928b42539100362bfc0f85bd3/CMake/Modules/Urho3D-CMake-common.cmake#L559-L569

cosmy1:
Interesting...
immagine

You are right but for some strange reason I still see it enabled.

weitjong:
Don't trust VS IDE for that :smile:

I just concluded my test and I cannot reproduce the problem. I am using ```#error``` to verify which code path is being selected. Switching SSE on and off on my 32-bit VS build tree (I have to regenerate the build tree for some reason after switching), but otherwise the test shows everything is as it should be.

cosmy1:
Thanks MS for your great Intellisense. Apparently on Visual C++ 2013 the IDE shows the correct path so yes you can merge it. Thanks for your work and sorry for that inconvenience.

PR #1295, opened 2016-03-31
Complete PBR work

dragonCASTjosh:
Included PBR Rendering for OpenGL and DirectX. Supported renderpaths include Forward and Deferred. Finally there is a new example scene included.

screenshot_thu_mar_31_20_26_27_2016
screenshot_thu_mar_31_20_26_35_2016


cadaver:
Thanks for the work! Some quick notes:

- Have you used tabs instead of spaces? Existing code uses spaces (view whitespace to be sure)
- Default deferred renderpath seems to be modified (buffer formats), though there also exists a separate PBR deferred renderpath. Is this the intention?
- Added variables, like light width & depth, don't have much explanation in the header comments. The explanation doesn't have to be mega-verbose but even something like "light width for PBR rendering" would give some context.
- Some shader files seem to be completely rewritten in the diff (whitespace or line ending differences?) which makes hard to compare what has actually changed
- There is a PBR ifdef in shaders such as LitSolid. Would be preferable to have a separate shader, as the shader code is complex enough already.
- The PR seems to revert some of the recent changes to e.g. editor (I guess you've used the "mine" option to resolve conflicts). This is not OK.

dragonCASTjosh:
- i used visual studios default i believe its 4 space tabs.

- i will have to look into that i may be left over from when i had deferred pbr in the default deferred render path

- i must of forget to remove then, they where original for area lighting but i have removed it from this commit but i think i need to clean up the code i guess

- thats because i had issues with commits from before the branch some how got messed up in the merge so i had to copy and past the shaders into a new branch this seams to of caused issues. not sure what i can do about that

-i will look into moving pbr stuff into a new LitSolid shader.

-I am not sure how i added the shaders from a fresh commit about an hour ago then made a PR

JoshEngebretson:
Given the history (https://github.com/urho3d/Urho3D/commit/0fa75740a6ede15cc9e5ee470547b2c58e12a597), is there any question as to the usability/copyright on portions of this code?

dragonCASTjosh:
The [Licence](https://github.com/dragonCASTjosh/Urho3D/blob/master/bin/CoreData/Licence.PNG) included in the PR shows Jonathan providing me full access to the last public commit to use as i see fit. If people are still uncomfortable with me using parts of Jonathan PBR implementation im happy look into changing it :)

hdunderscore:
Testing it out, it seems you have reduced the performance issues with the importance sampling, and the reduced textures also allows me to run the samples. I think the realtime importance sampling is a good way to go since it reduces the zone textures we need, and it seems to give an interesting effect.

It's hard to tell how accurate (or at least visually appealing) the implementation is due to a few issues with the sample assets being used. Either that the samples were intended for a different PBS setup or that the shaders aren't hitting the mark.

If you turn off the dynamic lights you can see the IBL is still weak, which is understandable because we didn't complete the hdr texture loading issue. One solution I've seen when handling LDR IBLs is to provide a fake-up power factor. The IBL should provide a lot more lighting, but the sample masks the issue with a powerful directional light.

Looking at the scene without the dynamic lights, it seems the roughness values aren't linearly interpreted by the IBL (0.0 and 0.5 roughness look very similar, most of the change happens between 0.5 and 1.0). Also, the concrete no longer resembles concrete and I can't figure out if the 'panel' texture is meant to be metallic or not (these could be issues with using sample assets not designed for these shaders).

dragonCASTjosh:
The performance issue with importance sampling is linked to using high resolution textures along with high resolution cubemaps, this combination is not likely to happen and when it is the developer would probably be targeted modern hardware.

The sample selection is part of some left over code i found in Jonathan's past work that i just re-purposed so i agree the samples may be a little off.

i think the biggest problem with IBL is there is no energy correction, its something that could easily be solved when im done with some of the other requested changes.

the problem with the roughness values is finding the sweet spot on the types of materials we wish to use, Cryengine uses Roughness <sup>3</sup> , Frostbite3 uses Roughness <sup>2</sup>. refer to page [14](http://www.frostbite.com/wp-content/uploads/2014/11/course_notes_moving_frostbite_to_pbr.pdf)

cadaver:
The license is dated before JSandusky requested his name to be struck down from the contributor list (which happened 15th December). So if you can get in touch with him again I'd recommend asking if he is fine with his name being reattached to the project via the license image. He himself cannot answer here as he has been banned from the github project due to repeated hostile behavior.

dragonCASTjosh:
Ill try contact him but if i am unable to then the licence looks to me that i will still be able to use what i  have for PBR

cadaver:
Yes, I would interpret it that way too.

dragonCASTjosh:
@cadaver any idea how to fix the issues with files showing as complete re-write

cadaver:
You could check with a hex editor (for example xvi32) if the line endings or something else in the file content is different, and use a tool (such as sfk) to change them if necessary.

dragonCASTjosh:
it looks like its linked to me re-indenting the shaders i changed indentation to make it more readable but never revered the change

JoshEngebretson:
I appreciate the work @dragonCASTjosh has done on this PR, for us downstream a redacted png image isn't going to cut it for license.  

I am not a lawyer and have no idea if the redacted png is legally binding, even so don't want to introduce a graphic of an email to the license for forks and end users.  The need of which being a history of hostility and pulling code down, which makes it impossible to even communicate with the person.  

If the license and code can be sorted, cool, otherwise this is going to introduce significant merge issues at least for us.  How much of the code in this PR does this represent, including shader code?

1vanK:
How much is left of the original code? Probably it is best to rewrite the pieces and abandon any relicensing?

dragonCASTjosh:
the original code is mostly getting the roughness and metallic, and the fresnel and GGX

cosmy1:
If you want, I can help you cleaning it and doing a new PR so it will have just the PBR feature.

dragonCASTjosh:
i believe most of the stuff has been cleaned up. im working on glsl stuff at the moment, other then that the only issue is in DebugRenderer so ill need to look into that

cosmy1:
I noticed also some tabs which need to be converted to whitespaces.

dragonCASTjosh:
You should be able to make PR to my master branch

cosmy1:
Tomorrow I'll clone your repo and continue to submit PRs or if you want give me write permission so I can push commits directly would be faster.

dragonCASTjosh:
personally id prefer to keep it as PRs so i can track changes myself

dragonCASTjosh:
@hdunderscore Give the newest commit a try it may have fixed some of your issues with IBL

dragonCASTjosh:
@cadaver im not sure what is happening with line endings, im using the github client and i copy and pasted the shaders from the master yet it made no diffrence

cadaver:
I would recommend using command line git / or tortoisegit instead, as they should have default settings for sane line endings conversion, but if most of the shader work will end up being in separate files (e.g. a separate LitSolid for PBR) the issue is not as large. We can perform manual compare ignoring line endings once the work is ready for merging for real.

dragonCASTjosh:
i moved PBR into its own shaders, therefore LitSolid, DeferredLight and Lighting can all be droped from the PR

hdunderscore:
The IBL appears to treat metallic opposite to what might be expected (eg, 0.0 should show diffuse colour, 1.0 should mute it -- just like it does with direct lighting).

dragonCASTjosh:
that is currently how it works, 0.0 has diffuse and 1,0 has minimal diffuse.

dragonCASTjosh:
screenshot_fri_apr__1_21_20_56_2016

hdunderscore:
Metallic = 1.0 + IBL + Direct Light:


Metallic = 1.0 + IBL (No Direct Light):


Metallic = 0.0 + IBL + Direct Light:


Metallic = 0.0 + IBL (No Direct Light):



There appears to be a visual discrepancy, especially if you compare the material preview window which only shows the direct light case.

dragonCASTjosh:
not entirely sure what cased that, could be the boost i performed to IBL. as for the material editor, that is likely due to the material editor not having a zone with a cubemap, therefore there is no reflection for metallic materials.

dragonCASTjosh:
so i found the problem and its kinda intentional to the affect, IBL provides lighting in the form of reflections. The IBL is applied prior to the material diffuse, a side effect of this is metallic materials materials are not affected by diffuse so there is no darkening of the material

hdunderscore:
Hmm I was assuming that the direct light case was correct and IBL was incorrect, but now comparing to other physical shaders it could be that the IBL is nearer to OK and the direct light isn't quite right. The metallic specular colour should take on a colour in direct light, but in these shaders the specular spot is always white.

dragonCASTjosh:
the specular takes on the color of the light

hdunderscore:
In other metallic/roughness setups, the specular also factor in the diffuse colour. If I recall correctly from the pbs theory, perfect (clean) metals don't have a diffuse reflection, but a specular one only.

Metal = 1.0:


Metal = 0.0:

dragonCASTjosh:
ill look into changing this so metallics specular change based on color

dragonCASTjosh:
this issue is proving harder to solve then i expected, its like im missing something

hdunderscore:
Thanks for looking into it !

I think part of the issue is that I have been testing the DX9 build and you might have been looking at the GL build, and the issue didn't show in the GL version. After taking a look, I found the offending line in the hlsl (float vs float3): https://github.com/dragonCASTjosh/Urho3D/blob/master/bin/CoreData/Shaders/HLSL/PBRLitSolid.hlsl#L229 . Looking at the glsl code, it is structured a little differently and the glsl drops the f0 term completely. There is a visual difference when you use f0 and specColor-- which is right? which is better? Small differences like that make it harder to use assets from different tools.

Ideally the hlsl and glsl should be structured as close as possible and should definitely use the same calculations.

I'm going to try set up the same scene in another tool for comparison. Do you have links to the source of textures you used?

dragonCASTjosh:
The F0 was something I was testing, they both do the same thing just slightly different, I believe iv settled with spec color although the specular is still white on metallics next t sure there is an easy fix.

The only reason gl and dx may be different is I mainly develop under dx11. So gl may be behind.

As for textures I get them from https://share.allegorithmic.com/ although you will need substance painter and will need to set up a Urho3D export

hdunderscore:
I put together the same scene in Unity, and have set up a few scenes files with different lighting set ups to compare: https://drive.google.com/file/d/0B8btTTaJ0buXTnlkQTRFMnVKd00/view?usp=sharing

This gives me some clues to work with, so I am going to take a closer look at the shaders and see if I can make some changes.

dragonCASTjosh:
I think metallics in unity look a little odd, as for the specular color im still trying to fix it

dragonCASTjosh:
@hdunderscore it looks like the problem does not show on metallics with a roughness of at least 0.4

hdunderscore:
I'm not 100% sure if the remaining white spot is a big issue or not, the best course would be to double check that the calculations match the theory (don't add fudge factors :o).

I'm going to share some progress soon. I've made some changes and the results are much closer to the Unity version, but there are some artifacts at grazing angles that might be a result of reverting the importance sampling method to the one epic described.

The biggest issue I noticed is that these shaders didn't have a diffuse IBL, so I added one by simply reusing the ImageBasedLighting funciton with roughness = 1 and specular = diffColor. The results seem in the ballpark of what might be expected.

Result:
pbr

dragonCASTjosh:
@hdunderscore my aims with the results where something between unreal 4 and frostbite 3. personally im not a fan of the unity implementation

dragonCASTjosh:
so after some more testing in Unreal 4 i found that UE4 does not tint the specular color on metallic surfaces, they remove the specular on 100% metallics and 90% metallic has a small white (light color) specular
highresscreenshot00000

hdunderscore:
You mean you don't want a diffuse IBL? If you look in the areas unlit by direct light in your shaders, they look metallic. That's a result of not having a diffuse IBL.

What do you mean UE4 doesn't tint the specular? You can still see the red specular spot in the UE4 image. I am guessing that UE4 clamps the output, because even on a very bright light at grazing angle the specular spot doesn't turn white, but in Substance Designer 4 (which also implements the method described by epics paper) and Unity they do show a white spot at grazing angles.
ue4specular

dragonCASTjosh:
maybe i had a bad scene when i tested. send me to line of code for the diffuse ibl so i can test how you did it, also i still cant find where the specular has gone wrong

dragonCASTjosh:
Solved
screenshot_sun_apr__3_22_32_11_2016

hdunderscore:
Change this line:
https://github.com/hdunderscore/Urho3D/commit/fe1aa4b35e34caafcfb3572ab4b93979dd703bba#diff-c606c7147d8c716146d1a14126cbec22R135

and these:
https://github.com/hdunderscore/Urho3D/commit/fe1aa4b35e34caafcfb3572ab4b93979dd703bba#diff-8db6b8cc2e0f35d118ed28db9d082c23R375

Notice that we don't need to boost the IBL anymore with the change.

dragonCASTjosh:
@hdunderscore  can you make a PR of that on my master branch

dragonCASTjosh:
once this gets merged it should be good to merge

dragonCASTjosh:
NOTE: OpenGL is currently broken

hdunderscore:
I look forward to trying the changes when I get home !

dragonCASTjosh:
Let me know your results

hdunderscore:
I'm working on some changes, I found a better way for the diffuse IBL.

dragonCASTjosh:
awesome, can you make a PR of it that include the spelling changes

dragonCASTjosh:
@hdunderscore feel free to make any changes you like to the IBL system as i would love for the IBL to look higher quiality on medium roughness materials. i was avoiding pre-integrated IBL as most people would use ZoneTextures in a real game so the engine would have to create the Mips itself and i didnt feel i had the time to do this.

dragonCASTjosh:
@hdunderscore hows progress on the improvements

dragonCASTjosh:
above commit credited to @hdunderscore

dragonCASTjosh:
DO NOT USE COMMIT ABOVE, errors where introduced in the merge and commits where reverted

dragonCASTjosh:
lets hope thats a working commit

dragonCASTjosh:
@cadaver if you are happy with this build i personally feel its ready to merge just keep in mind DebugRenderer.cpp/h, DeferredLight.hlsl/glsl, Lighting.hlsl/glsl and LitSolid.hlsl/glsl will need to be removed from the commit

Results:
screenshot_wed_apr__6_01_44_54_2016

hdunderscore:
I'll share a PR soon.

One thing to consider before merging is that there are a few large, unused assets included. The `Data` folder is significantly heavier in this PR vs master. Deleting and then committing changes will still leave a heavy weight on the git history so it's a bit problematic.

I suggest using just the one skybox texture for the zones (and if anyone wants to use a captured zone, they can generate one in the editor themselves), and also removing the unused extra textures.

cadaver:
When we merge this, we can IMO just forget the history, at least for the asset files, then anything unused shouldn't end up to the repo. Just get the branch to the shape that you want to be. 

Btw. merging someone's master branch (as opposed to a proper feature branch) always feels a bit iffy, but it's OK if you're not going to work with anything else before this is merged.

dragonCASTjosh:
It's currently in a state I'm happy with also the only reason it is the master branch is because i thought I would test my git clients PR features. I will keep it in mind going forward.

hdunderscore:
I posted my progress as a PR, changes only to hlsl though so they need to be ported to glsl.

Just in case you don't like the changes I made to the IBL I left the old version in, although I believe the old version has errors both visually and mathematically.

My changes bring the shaders back closer to the theory, that way anyone who wants to make changes in the future will be able to more easily orient themselves and experiment with different factors and BRDFs.

These shaders were nice to work with.

dragonCASTjosh:
@hdunderscore currently some issues with the PR, images attached to a reply on the PR

JSandusky:
After really really looking this over, it's basically all you. You're probably more than able to take on the PBR deferred path yourself now (I gave you shit in the past for messing with it, but by now you've probably got it). As far as I'm concerned this project is entirely unrelated to my previous work. This is all you, I revoke any claim on the base.

After the spacing tweaks and such, this has definitely become a definite goto reference in my opinion for PBR without frills. Good work, you took it in a direction towards quality where I refused to go (on account of my own hardware typically used) and it looks pretty solid.

Off-topic: I did read forum posts on IBL, I would suggest adding to the cube map baking in the editor the ability to run a tool afterwards, pick one tool as a baseline, but it's totally viable to bake from the Urho3D editor using a separate tool.

dragonCASTjosh:
@JSandusky i really appreciate the support, allot of the process for this involved me learning hlsl and glsl but i feel i have learned alot.

i felt quality was important for PBR as its the main reason for implementing the feature, im glad you think the code is so useful :)

Currently i dont have the time free to add cubemap baking into the editor although i feel it would solve all issues with the IBL used therefore making results that are much better. If someone wants to pick up development of such a features then i would be willing to merge it :)

and from what i can tell you are being very positive  :)

cadaver:
Yeah, contributing to Urho works easiest when there's only a single "link" of contributor to a Urho pull request, because then the contributor can just agree to the Urho's policy of assignment under Urho's existing MIT license. When there are multiple "links" involving code taken from someone (who may not agree) you can't readily do that. But I'm glad this specific situation was resolved amicably.

Btw. some open source libraries require a contributor to actually mail/fax a contributor's copyright co-assignment agreement, we don't go that far, hopefully it doesn't bite us in the ass :)

hdunderscore:
@cadaver any chance you can check out the PBRDeferredHWDepth in GL2/3/DX11 ? I hadn't used the HWDepth before so I'm unfamiliar with whether they are meant to work or not, but I had issues on those three. GL didn't seem to fill the g-buffers and DX11 outright crashes.

GL3 also doesn't render the specular spot on my end.

Otherwise I think these shaders are now a good base for future PBR work.

cadaver:
I will check them before merging. Crash sounds like an engine bug (a renderpath can be wrong and ineffective, but it shouldn't crash)

cosmy1:
I think there is a bug in Material.cpp in ResetToDefaults() call.
You hardcoded the PBRNoTexture.xml instead of NoTexture.xml, not sure if this is correct.

dragonCASTjosh:
That's was for ease of testing, I'll let @cadaver pick if he wants PBR as default

cadaver:
@dragonCASTjosh It doesn't sound good to change the default empty technique. Maybe you should add it as a choice to Renderer class (ie. SetDefaultTechnique / GetDefaultTechnique)

cadaver:
I've had an initial look at this, most changes go to the dedicated PBR directories (e.g. techniques and textures) which is very nice. The engine changes don't look bad as there's so little of them, only the default technique setting needs to be added instead of only modifying Material.

If you want to make our lives easier and ensure that nothing actually needed gets deleted by mistake, please clean up any files that shouldn't be there. For example I noticed a 12 MB output.dds in Textures folder, is it used? At present, blindly merging this would grow the Data folder from 16 MB to 100 MB, which seems a bit on the extreme side.

dragonCASTjosh:
ill look into cleaning that stuff up

cadaver:
Thanks!

dragonCASTjosh:
@cadaver i think im going to leave the default technique option for now. is there anything else that you need before merging

cadaver:
I'll tell you if I come across something critical, at the moment I don't need anything else.

dragonCASTjosh:
if it remains that way then its ready to be merged

dragonCASTjosh:
Ideally i would want a demo that generates a 10x10 grid of spheres with transitions metallic/roughness values, but i had issues setting this up so i moved on.

cadaver:
Ok.. a few more questions / comments.

- Are there going to be copyright problems with any of the images?
- Any problem if I convert the large images such as the skybox to DDS format?
- Seemingly same skybox was included twice, which should be OK to remove?
- The old skybox is overwritten. It would be preferable to not touch that and include another skybox material for the PBR scene.

dragonCASTjosh:
all the images are under creative commons.
that should be fine.
that should be ok to remove.
sorry about that, it should be ok to undo that change. i only made the change as there was issues with creating a new skybox material

cadaver:
Ok, thanks. Didn't see added license statement/attribution files, should there be? (I'm not that familiar with CC, I just know the fonts included in Urho require attributions)

dragonCASTjosh:
i think there will need to be a link to  [https://share.allegorithmic.com/](https://share.allegorithmic.com/) but other then that it should be fine.

for reference:
[http://creativecommons.org/licenses/by/4.0/](http://creativecommons.org/licenses/by/4.0/)

dragonCASTjosh:
@cadaver  i believe that licence should be enough

evolarium:
To be careful, you might want to include a link and user name for each asset included.  The attribution requirement says:

> If supplied, you must provide the name of the creator and attribution parties, a copyright notice, a license notice, a disclaimer notice, and a link to the material. CC licenses prior to Version 4.0 also require you to provide the title of the material if supplied, and may have other slight differences.

cadaver:
The attributions can also be amended in a separate PR later, so it's not a blocker. Of course it's good to have them sorted as soon as possible.

dragonCASTjosh:
@cadaver whats needs to be done for a merge. i can sort out the licence it shouldnt take long

cadaver:
Necessarily nothing, I just haven't have had time. Hopefully will have today. If you want to make my life easier convert all TGA's to something else as appropriate, for example skybox textures should be fine as DXT1 DDS, and the color grid texture (forgot the name) could be png. The texture formats are kind of all over the place as there's png, jpg etc. which are not good for runtime loading performance, but I'd guess in this case it's best to aim for minimum filesize (and thus minimum repository size).

dragonCASTjosh:
ill look into doing that, hopefully i beat you to the merge, if not ill make it as a separate PR. The reason i went with jpg is Urho wasnt loading .png normal maps, it was something along the lines of it being more then 8bit.

cadaver:
I have taken the branch as of now and made a squashed merge. Will push probably in a few hours, after which further work can be done. One thing at least missing is C++ version of the PBR materials example, which is not that big of a deal.

dragonCASTjosh:
ill make a PR of that at some point soon

cadaver:
Merged. I fixed the renderpath crash, that was trying to assign a depth-stencil texture as a color rendertarget (now the D3D11 renderer is safeguarded against that.)

Issues or needs for further development I found:
- The grainy reflections on the objects only seemed to work on OpenGL. Didn't look at that in more detail yet.
- Lua PBR example was also missing (in addition to C++ example), didn't notice that.

It's excellent to have PBR in Urho at last, however I'd hope for cleaner PR's in the future, since this took quite a bit of post-editing, particularly unnecessary added files, and one texture was deleted and needed to be added back.

Note that now you may need to force-push Urho master branch to your repo to continue the work, since I squashed and rewrote the history.

dragonCASTjosh:
ill get into these fixes as my next set of projects

1vanK:
I have not tested DirectX, but Opengl version does'nt work for me now (all balls looks the same)

1vanK:
I see difference only when the balls are illuminated by the sun, but in shadow they just reflect cubemap without any bluring

hdunderscore:
Looks like some issues popped up, I'll look into it.

1vanK:
for optimization, we can replace at several places
```
... / M_PI
```
to
```
... * M_INV_PI
```

hdunderscore:
OK I think I've spotted the main issue, it's that we might have removed the skycube with the mip maps in the cleanup? In any case it seems that we aren't getting the desired blurring/mip map when sampling the IBL.

Just as an important note, there is a by-default difference between the GL and DX versions because I left this change in: https://github.com/urho3d/Urho3D/blob/master/bin/CoreData/Shaders/GLSL/IBL.glsl#L3

Without that, GL is too slow (at least on my machine) at sampling the IBL but it also means there is a quality reduction. I think for development purposes it is a reasonable thing to leave on, meanwhile we can either find a solution that will allow GL to work faster with image sampling or we can submit a PR for pre-filtered cubemaps, which really shouldn't be much more than a few lines extra (although pre-filtering has it's own well-accepted artifacts vs performance trade).

When you remove that PBRFAST define, then you can see that DX and GL look the same. If you enable that define on DX you get the speed boost too, and in DX11 you still have a very good quality too (as long as we have a zonetexture with mipmaps?).

hdunderscore:
The pre-filtering workflow would require extra preparation effort external to urho. To include a pre-filtered solution in urho would require more than a few lines ;)

hdunderscore:
Another option is of course to simply increase the number of importance samples (and adjust sample kernel), but that's a big performance vs quality trade that needs to be considered carefully before releasing a game/app with it.

dragonCASTjosh:
@hdunderscore  maybe make it scale with the engines material quality setting

weitjong:
Slightly off topic but all the web samples do not work now. I suspect it is caused by the increase of data pak size due to this merge commit. We do not use memory auto grow feature in Emscripten for performance reason, so if the increment is greater than what we have reserved then the web samples won't work. We need to re-estimate the required total memory and adjust that manually to fix this. Alternatively, the build system could probably be improved to either a) estimate the required runtime memory automatically and use that during Emscripten build, b) smart enough to exclude assets/resources that are not needed in the build. At the moment the packager just blindly pak everything. I guess this is also the reason why the iOS packaging build failed after this merge commit.

hdunderscore:
Another way would be to make a third package directory, `LargeData`, and just not include that in the web build ?

weitjong:
I have running out of idea on how to free up more space on the Travis OSX build environment.

cadaver:
Since we have the Autoload feature which allows specifying optional data directories, I put the PBR assets there. Also generated mips to the PBR scene skybox. The cubemap "137" wasn't used in the scene after all, so I deleted it. It's still in the history if it's needed.

@hdunderscore If you have further changes / improvements in mind, feel free to work directly in master, no need to go through pull requests

@weitjong If you want to rewrite history at this point to get rid of the unnecessary / changed assets and reclaim repository size, please go ahead. I find my git ninja skills lacking to dare to attempt this.

cadaver:
Since we have the Autoload feature which allows specifying optional data directories, I put the PBR assets there. Also generated mips to the PBR scene skybox. The cubemap "137" wasn't used in the scene after all, so I deleted it. It's still in the history if it's needed.

@hdunderscore If you have further changes / improvements in mind, feel free to work directly in master, no need to go through pull requests

@weitjong If you want to rewrite history at this point to get rid of the unnecessary / changed assets and reclaim repository size, please go ahead. I find my git ninja skills lacking to attempt this.

weitjong:
> If you want to rewrite history at this point to get rid of the unnecessary / changed assets and reclaim repository size, please go ahead. I find my git ninja skills lacking to dare to attempt this.

Since you have squashed the commits before merging, I don't think those commit history in the PR branch itself got pulled into our master branch. I don't have any concerns with the repo size actually as the result.

cadaver:
Ok, good. I was thinking of the "137" cubemap and the unmipped -> mipped skybox, which were post-squash.

weitjong:
I see. I can do that.

weitjong:
It is done. Those pannels still remains as pannels.

cadaver:
Thanks!

weitjong:
No problem, but I concur with 1vank. The balls looks the same now to me too. I am using OpenGL API.

weitjong:
@cadaver , I think I made a small mistake. I did not realize you have changed the binary file at the same time for those skybox. I don't have a copy of your mipped version of the skybox anymore. If you send me those by email, I will redo it again.

weitjong:
Nevermind. I got it and rebased again.

hdunderscore:
Nice @weitjong, thanks for that !

Did you try disabling the PBRFAST define? Let me know what kind of FPS you get, maybe the performance isn't terrible on all machines.

1vanK:
maybe it will be helpful http://brabl.com/ibl/

weitjong:
@hdunderscore The PBRFAST define changes the FPS quite drastically on my machine. With it, I got about 120 FPS when all metalic balls are in view and about 90 FPS when all red/rough balls are in view. Without it the FPS dropped below 20 for both. However, there is a significant artifact on the red/rough balls with PBRFAST define when there were not lit.

1vanK:
miplevels created manually? I think need use something like AMD CubemapGen for it

hdunderscore:
I think it should be possible to create mipmaps that won't show those seams in GL. In dx11, those seams don't show with PBRFAST defined, it makes me wonder if there is a shader trick that can be used.

1vanK:
> create mipmaps that won't show those seams in GL
I downloaded last version with mipmaps and compiled DX version :)

I was faced with the problem of seams when did the game https://github.com/1vanK/FlappyUrho when I blurred cubemap and CubemapGen helped me

https://github.com/1vanK/FlappyUrho

dragonCASTjosh:
i wonder what specificly is causing the issues in OpenGL

dragonCASTjosh:
hmm after some testing on my machine removing PBRFAST on my machine made no difference in frame time it was 5ms either way

1vanK:
Hmm, I don't see any visual difference between 1 and 16 samples, but fps X4

1vanK:
post-process effects are added to mask artifacts? :)

ImportanceSampling WITHOUT postprocess and with 1 sample





I like more ImportanceSamplingSimple - this does not create bright spots on the perimeter.

1vanK:
ImportanceSamplingSimple without postprocess and 1 sample

1vanK:
I propose to focus on performance rather than on mathematical precision

please test it

PBRLitSolid.glsl
```
        #ifdef IBL
           float mipLevel =  GetMipFromRougness(roughness);
           finalColor.rgb += textureCube(sZoneCubeMap, reflection, 9.0).rgb * diffColor.rgb;
           finalColor.rgb += textureCube(sZoneCubeMap, reflection, mipLevel).rgb * specColor.rgb;
        #endif
```

It is visually very similar, but the performance is ultra high and without any artefacts

dragonCASTjosh:
@1vanK  could you post an image of this approch using a a material with 1 metallic and 0.6 roughness.  because this is where most PBR systems fall apart most.

1vanK:
> could you post an image of this approch using a a material with 1 metallic and 0.6 roughness. because this is where most PBR systems fall apart most.

Skybox2.xml from repo



my skybox wrong too

hdunderscore:
@1vanK , have you worked with PBR in other engines or tools ? What are you comparing your results against? Earlier I posted a link to the scene in unity, I personally used that to test against.

@1vanK you have been doing all your testing with the PBRFAST define on GL (which I already mentioned three times now), so no doubt it looks like the number of samples isn't doing anything (because it isn't). I was anticipating people would complain about the performance issue on GL so I included a simple hack (PBRFAST) and enabled it by default, but now I see there are complaints about both the visual and performance issue..

Since you are working on top of your own edits, you need to specify more detail about issues you are having because it's harder to understand which changes I need to make to replicate them. I would also like to see comparisons to other PBR implementations or theory that back up the issues, otherwise we can keep taking visual shortcuts until it no longer performs as a flexible PBR solution. Eg, I can go to blender and bake a scene and load that right into Urho with an unlit shader-- it would probably even look better than most PBR implementations and it's as fast as possible!

I don't agree with what you have pointed out as an artifact, but rather it's a result of using an input image that has a very bright surroundings for the IBL, and at the same time not using and ambient occlusion to control it. If you neutralize the specular brightness right at the shader, instead of changing the image input, then every scene will be dull and lifeless. Comparing this to Unity, our shaders are even a touch darker on the edges with increasing roughness.

EG, compare the concrete texture with ImportanceSampling and ImportanceSamplingSimple. Can you tell what that texture is meant to be with ImportanceSamplingSimple? For a long time I thought it was meant to be metal, and I didn't understand until I implemented ImportanceSampling, that the shiny spots are meant to be puddles of water. That's a big difference.

One thing to keep in mind, right now Urho only support LDR images, so anything that's 'white' in the input image is going to be 'maximum brightness' (eg, the white sun in the input image is only as bright as a white cloud or the white crashing waves), so can you now understand why it seems like the results could be too bright on the edges?

The purpose of the post-processing is not there to mask any artifacts, but to bring the image back into a gamma corrected colour space and add a little extra niceness. UE4 and Unity load their editor with the same effects by default (although more polished) - bloom, anti-aliasing, gamma-correction. If you are going to turn off the gamma correction, you need to adjust the texture .xml's to disable the srgb setting, otherwise it's obvious the output is too dark. Doing this though, all the lighting is done on top of gamma corrected image sources, which isn't typically what you want in PBR unless you authored your assets in that way.

Indeed there are many tricks you can use to speed up the IBL calculation, and in my previous shader effort I also used a prefiltered IBL exclusively. But there are three issues with that approach:

1. It takes a lot more effort on the user to prepare the prefiltered textures (or it would take a lot of effort on our part to integrate the process automatically, although I do want to do that eventually)
2. It's not a perfect solution since it makes some assumptions and loses out on some easy wins
3. You need to squeeze in room for an extra cubemap sampler

I do agree in adding support for prefiltered IBL as an option (and I always intended to), but I also agree in having the importance sampled version implemented first because it gives you a comparison to work with and lets you develop scenes more quickly, without spending time preparing your textures.

These shaders are easy enough to modify that almost anyone can do so in their project, so I am personally against including modifications that are obviously, visually inaccurate into Urho master and then declaring Urho3D has physically based shading and it's that reason that I was personally not impressed by the initial state of this PR. I do insist in following the theory, or at least maintaining visual consistency with Unity and/or UE4 because that's what new users might expect and it's what the art asset tools being developed out there are targetting.

As a final point, this PR introduces the base for future PBR work. That's what I said much earlier, because I know that there are many things that can be added, improved and optimised. If someone wanted to start developing a game in urho with these shaders, they won't be disappointed by the visual results compared to other popular implementations. As time passes and more contributions are made, they won't be disappointed by the performance and variety of shader options either (eg, hair shader, terrain shader, clear coat options, etc). But even if there are no further contributions made, the work here follows the theory so all you need to do is read the widely available references and make simple tweaks for big wins.

1vanK:
seriosly, I will not be pleased that a few pixels appear differently if FPS in game = 10 :)

hdunderscore:
It's true, it's all about making the right sacrifices for reasonable performance. I also protested in the forum, image sampling isn't the best way to go for performance and it's unfortunate that in GL it's so slow, but it does give us a good reference.

I'll make a change to support a prefiltered input (eg, from cubemapgen) and post it when I've finished testing it. I'll probably also just reduce the number of samples to 1 when PBRFAST is enabled like you suggested. 

In the meanwhile you can reduce the number of samples while you develop or use DX which performs much better, or one of the solutions you posted.

That last screenshot you posted, did you happen to check the seamless cubemap option when exporting that image? Try without it, or otherwise you need to make the shader cubemap-lookup change to make that work.

1vanK:
Opengl automatically change miplevels depending on the distance to the model and
```
textureCube(sZoneCubeMap, reflection, mipLevel).rgb
```
return incorrect value. How to fox it?

dragonCASTjosh:
@1vanK The final result you just posted is why i didnt use the technique originally, it makes the cubemap look extreamly low resolution.

The only reason i didnt implement pre-filtered IBL is because in most game's using PBR the cubemaps will be captured in engine using the Zone capture functionality. If pre-filtered maps where implemented i believe you would have to have it within the engine so the captured maps are filtered automaticly 

@cadaver  "our shaders are even a touch darker on the edges with increasing roughness." This is a result of using Burley diffuse, i believe if you swapped to Lambert, smooth materials will darken and rough will lighten, most of this is detailed in frostbites paper on PBR

1vanK:
I think we need it:

https://www.opengl.org/wiki/Cubemap_Texture

Under the standard filtering rules for cubemaps, filtering does not work across faces of the cubemap. This results in a seam across the faces of a cubemap. This was a hardware limitation in the past, but modern hardware is capable of interpolating across a cube face boundary.

To globally enable this, use glEnable(GL_TEXTURE_CUBE_MAP_SEAMLESS)​.

1vanK:
GL_TEXTURE_CUBE_MAP_SEAMLESS works not ideally, but gone sharp seams when "ege fixup options" is disabled

metallic = 1, roughness = 0.6

dragonCASTjosh:
i think we need to set up a secene that does a 10x10 grid do demonstrate all combonation of roughness and metallic, it would be useful in testing

1vanK:
```
      float GetMipFromRougness(float roughness)
      {
          float smoothness = 1.0 - roughness;
          return (1.0 - smoothness * smoothness) * 10.0;
      }
```

may be just 
```
      float GetMipFromRougness(float roughness)
      {
            return roughness * 10.0;
      }
```

dragonCASTjosh:
@1vanK  the old technique i believe was gained from Frostbites paper on PBR

1vanK:
mipmap 8 - 4x4 pixels
mipmap 9 - 2x2 pixels
mipmap 10 - 1 pixel

and it for rougt 0.6...

dragonCASTjosh:
@1vanK can you apply a single pass of ggx sampling to the sphere in the last image you pasted to try and smoother the low quality look of the relfection

hdunderscore:
glEnable(GL_TEXTURE_CUBE_MAP_SEAMLESS); seems to work well for me, using the texture that's already included in Urho. Can you confirm that @1vanK without your custom texture?

seamless

1vanK:
ok, you are right :) with ggx sampling it's not working

1vanK:
> Can you confirm that @1vanK without your custom texture?

I use texture from repo too

hdunderscore:
@1vanK , can you submit a PR for the GL_TEXTURE_CUBE_MAP_SEAMLESS change. It seems to be a relatively new feature, not sure if there should be a check for support before enabling.

PR #1297, opened 2016-04-02
Added Texture2DArray.

reattiva:


damu:
Great! That's one of my most wished features!
Haven't tested it yet. Is this already usable?
The array seems to be defined via an XML file like:
```
<xml>
  <layer name="textures/grass.png" \>
  <layer name="textures/stone.png" \>
  ...
</xml>
```
Is this correct? The name of the root (here "xml") doesn't seem to be relevant, is that correct?
Does anything else have to be done to get a texture array working?
Is something additionally planned regarding this new feature?

Found some information on how to use that in a shader in case someone else is interested:
GLSL: https://www.opengl.org/wiki/Array_Texture
HLSL: https://msdn.microsoft.com/en-us/library/windows/desktop/bb509695%28v=vs.85%29.aspx

reattiva:
It should work on DX11 and OpenGL.
The root name in the xml is relevant, it needs to be "texturearray" (case insensitive).
Here an example:
https://github.com/reattiva/Urho3D/tree/tex_arrays_sample/Source/Samples/EXT_TextureArraySample

cadaver:
Sweet! Will test this soon.

cadaver:
Right now is not the time to make the change, but a lot of the code would be simplified if there was just one texture class that would act all the modes (2d, 3d, cube, array) according to the data that's fed to it.

reattiva:
Indeed, starting from the cube code the changes I've done were really a few, so one texture class makes sense.
On another note, to use these arrays as render targets, geometry shaders are needed (or at least they can simplify the work), what do you think about it?

damu:
One texture class? As in one mightier base class as Texture or as in one giant class that does everything? I have the feeling that the later would require more runtime checking at critical parts and cost therefore more performance as the current polymorphic approach (even with the vtables). Are you sure that's a good idea? Could alternatively be more stuff moved into the base class to have a hybrid between the two approaches?

cadaver:
I believe there has been independent work toward geometry shaders on OpenGL already before (boberfly?) so if they can just be reasonably implemented in the existing API, I'd have nothing against them. Though I'm not sure how they will fit into the existing 3D high level rendering (Renderer / View), except simply by having a material/technique refer to a geometry shader, and in some cases the performance can be lackluster.

@damu The critical operations are usually binding textures, which can happen 1000's of times per frame. All textures regardless of type typically produce a handle or resource view, which can be blindly bound to the graphics API without checking the type or branching, so that's not a performance concern. Type checking is only necessary for the SetSize() / SetData() / GetData() type of operations, which are slow in any case as they actually create or manipulate GPU resources.

Also note that currently there aren't no virtual functions for texture data manipulations, and the parameters are different, rather you must explicitly know what kind of texture you're using.

damu:
Ah, ok so no relevant performance issues. So It's only a question of usability and internal structure? Wouldn't a more dynamic/unified texture type make the usage of Texture more complicated for the user as he has to check the type or watch out for the right type all the time? Like in a dynamically typed language. It the unification has advantages internally but makes the library usage more difficult: this could be fixed by a "wrapping hybrid" to have (possible optional) type safety via Texture2D and similar. At least the SetData and SetSize functions look tempting to misuse, like a 3D version on an actually 2D texture. This would also make the transition to the new approach easier.

cadaver:
Typical usecases of textures are either such that they're loaded as part of a material, in which case the user is just interested that the material works properly, or creating a texture in code from ground up, in which case you know what kind of texture you're creating. Furthermore the texture SetData() / GetData() API for all types could be generalized into a 3-dimensional access where the 3rd coordinate could mean cubemap face, 3D depth slice, or 2D array index, so in case you want to just set the first 2D "face" of whatever type texture, the parameters would be the same.

friesencr:
When I implemented the stb voxel renderer in Urho I had to hack some things in there to get it to work.  Texture Arrays, Texture Buffers, attribute arrays, and a different vertex format.  It's really cool to see Texture Arrays merged!  Attribute arrays were solved some months ago! It sounds like vertex formats are on the table.  Since there is some discussion about textures I thought I would bring up texture buffers.  It isn't as useful in gl2 land. 

I had made a texture type TextureBuffer which takes an IndexBuffer.  It gets weird, here are the highlights.
https://github.com/friesencr/Urho3D/blob/voxel/Source/Urho3D/Graphics/OpenGL/OGLTextureBuffer.h#L64
https://github.com/friesencr/Urho3D/blob/voxel/Source/Urho3D/Graphics/OpenGL/OGLTextureBuffer.cpp#L152
https://github.com/friesencr/Urho3D/blob/voxel/Source/Urho3D/Graphics/OpenGL/OGLTextureBuffer.cpp#L117
https://github.com/friesencr/Urho3D/blame/voxel/Source/Urho3D/Graphics/OpenGL/OGLGraphics.cpp#L896

You can really feel the texture abstraction fall apart in this use case.

Really though it wasn't a lot of code to get TextureBuffers into Urho, but the code is silly.  In the case of my voxel terrain I was able to render an incredible amount of data in a single uint32 and some lookup tables.

I have never used a game engine api that handled this scenario and have no ideas in representing a serializable/deserializable universal format description between gl and dx.  Our gl2/dx9 abstraction is really darn good and hits the major features, works on desktop, mobile and works in a browser \m/.

damu:
@cadaver:
Ah, cool. You've convinced me, sounds like a good idea now ;-).
The only other use case that came to my mind is manipulating existing textures or materials, less typical though.

The third coordinate for cubemaps is the face (0-5)?
I've also read about CubemapArrays which actually have four coordinates. So a SetData (and similar) should IMHO have several functions for up to three or even four coordinates (if Urho has some four coordinate texture type some day). I guess you had that in mind anyway. Would be odd to have to give XYZ coordinates for a 2D texture, which I actually had to do in a weird image library.

damu:
Yay it's working: https://youtu.be/22KI_TyXi10
I've adapted my terrain material test with three textures. Each has a diffuse, a normal and a specular map and different diffuse and specular settings per texture. It now load these 9 images as a texture array.
Testproject: https://github.com/damu/Urho_Material_Test (It's currently OpenGL only)
The shader does also have support for this "height based texture blending" but it's currently commented out as I don't have a height value. This could be added by either having an additional height map image or by using an alpha channel.

Should the samples (or a sample) be changed to (also) show this new feature? Is there a way to make an alternative material for DirectX9 for this purpose?

PR #1301, opened 2016-04-03
C++11 event binding (lambdas/std::bind).

r-ku:
And removed redundant EventHandler() constructor adding default value 0 for `userData`. Less code, does exactly the same. If there was good reason for two constructors to exist please let me know, ill be happy to fix PR in case this is unwanted change.

Now it is possible to:

    SubscribeToEvent(..., [&](StringHash type, VariantMap& args) {
        // win
    });

    void Class::SomeEventHandler(StringHash type, VariantMap& args)
    {
    }

    SubscribeToEvent(..., std::bind(&Class::SomeEventHandler, this)));

    void Class::SomeOtherEventHandler(VariantMap& args)
    {
    }

    SubscribeToEvent(..., std::bind(&Class::SomeOtherEventHandler, this, _2)));

There is an ugly workaround - `EventHandler` asserts that `receiver_` is not `0`. This is not relevant in case of c++11 as receiver is captured by `std::function<>` therefore i set `receiver_` to `0xDEADBEEF`. This should not cause any problems as long as person is aware of what he is doing and is not trying to get received from callback bound via c++11 methods. In case of a mistake it should quickly become clear that something is amiss due to very visual value of receiver pointer.

### DIFF
--- a/Source/Urho3D/Core/Object.cpp
+++ b/Source/Urho3D/Core/Object.cpp
@@ -181,6 +181,18 @@ void Object::SubscribeToEvent(Object* sender, StringHash eventType, EventHandler
     context_->AddEventReceiver(this, sender, eventType);
 }
 
+#if URHO3D_CXX11
+void Object::SubscribeToEvent(StringHash eventType, const std::function<void(StringHash, VariantMap&)>& function, void* userData/*=0*/)
+{
+    SubscribeToEvent(eventType, new EventHandler11Impl(function, userData));
+}
+
+void Object::SubscribeToEvent(Object* sender, StringHash eventType, const std::function<void(StringHash, VariantMap&)>& function, void* userData/*=0*/)
+{
+    SubscribeToEvent(sender, eventType, new EventHandler11Impl(function, userData));
+}
+#endif
+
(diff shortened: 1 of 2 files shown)


cadaver:
Minor issue: we prefer space before and after equals sign, so rather userData = 0. Can also do that myself upon merge.

r-ku:
Added spaces so you dont have to ;)

weitjong:
@r-ku Could you kindly convert the usage explanation in your comment above into an actual doxygen documentation? Perhaps we need a new section or page even for all things C++11 in one place.

r-ku:
@weitjong not sure how that should look. Could you point me to an example?

weitjong:
You could get some inspiration here.
```
git log Docs/GettingStarted.dox
```

r-ku:
Made a patch with relevant c++11 features documented:

[0001-c-11-docs-for-containers-and-event-binding.patch.txt](https://github.com/urho3d/Urho3D/files/205913/0001-c-11-docs-for-containers-and-event-binding.patch.txt)

Hope its what you were after.

weitjong:
I think it is perfect.

weitjong:
The patch is now merged.

PR #1312, opened 2016-04-07
fix-areaID

MonkeyFirst:
Fix for AreaIDs. 
Checkout this please.

### DIFF
--- a/Source/Urho3D/Navigation/NavigationMesh.cpp
+++ b/Source/Urho3D/Navigation/NavigationMesh.cpp
@@ -584,6 +584,20 @@ void NavigationMesh::FindPath(PODVector<NavigationPathPoint>& dest, const Vector
         pt.flag_ = (NavigationPathPointFlag)pathData_->pathFlags_[i];
         pt.areaID_ = pathData_->pathAreras_[i];
 
+        for (unsigned j = 0; j < areas_.Size(); j++)
+        {
+            NavArea* area = areas_[j]->GetComponent<NavArea>();
+            if (area && area->IsEnabledEffective()) 
+            {
+                BoundingBox bb = area->GetWorldBoundingBox();
+                if (bb.IsInside(pt.position_) == INSIDE)
+                {
+                    pt.areaID_ = (unsigned char)area->GetAreaID();
+                    break;
+                }
+            }
+        }
+
         dest.Push(pt);
     }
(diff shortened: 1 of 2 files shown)


cadaver:
Storing raw pointers of nodes at navmesh build time is unsafe, as they could get deleted. Maybe best way would be to store WeakPtr's of NavArea components directly, and nullcheck later. This would also be better for performance than repeating Node::GetComponent() during the pathfind operation.

MonkeyFirst:
so I add sort by distance to bb center if we have overlapped areas
also add weakPtr array instead PODVector of nodes*
in my testing scene if have such code to check areaIDs:
~~~
void BotWalker::FollowPath(float timeStep)
{
    if (currentPath2_.Size())
    {
        static int t = 0;
        Vector3 nextWaypoint = currentPath2_[0].position_; // NB: currentPath[0] is the next waypoint in order
        unsigned char areaID = currentPath2_[0].areaID_;
        NavigationPathPointFlag f = currentPath2_[0].flag_;
       
       Color c;
        switch (t)
        {
        case 1:
            c = Color(1.0f,0,0);
            break;
        case 4:
            c = Color(0, 1.0f, 0);
            break;
        case 5:
            c = Color(0, 0, 1.0f);
            break;
        case 2:
            c = Color(1.0f, 0, 1.0f);
            break;
        case 3:
            c = Color(1.0f, 1.0f, 1.0f);
            break;
        default:
            c = Color(0, 0, 0);
            break;
        }

        Node* bl = GetNode()->GetChild("bl",true);
        if (bl)
            bl->GetComponent<Light>()->SetColor(c);

        float move = 25.0f * timeStep;

        float distance = (node_->GetPosition() - nextWaypoint).Length();
        if (move > distance)
            move = distance;

        node_->LookAt(nextWaypoint, Vector3::UP);
        node_->Translate(Vector3::FORWARD * move);

        // Remove waypoint if reached it
        if (distance < 0.1f) 
        {
            t = currentPath2_[0].areaID_;
            currentPath2_.Erase(0);           
        }
    }
}
~~~
video: https://youtu.be/7CgXIQIJ2PQ

cadaver:
Note that GCC doesn't like >> there must be a space inbetween.

cadaver:
Merged with small changes.

PR #1319, opened 2016-04-12
Shader animation bug.

neat3d:
Changed the conditions when unsubscribes the shader animation events.
Or it will unsubcribe all the events if you call SetShaderParameterAnimation function twice. (For example, you need to animate texture U and V )


neat3d:
Hello, I think this would be a bug, and I modified it.
If not, can any one tell me how to animate more than one material properties, thanks!

Issue #707, opened 2015-04-20
Arbitrary vertex declarations


cadaver:
Port the vertex declaration code from Turso3D, which allows defining vertex elements by semantic, index and format; for example you can define an arbitrary number of texcoords.

Retain compatibility with the old "element mask" system to still allow defining predefined semantic+index+format combos easily.

The binary model format will need a compatibility mechanism, for example specifying a zero element mask, which means that arbitrary vertex declarations follow.

After this is done, AssetImporter (and possibly OgreImporter) can be made to handle source data vertex declarations more accurately.

aster2013:
It is better add a version number into file format after file identity.

weitjong:
Yes, I agree. In my ```animation-refactor``` branch, I have something like this for loading the animation file. We could do the same for the model file too.
```
bool Animation::Load(Deserializer& source)
{
    PROFILE(LoadAnimation);

    // Check ID
    if (source.ReadFileID() != "UANI")
    {
        LOGERROR(source.GetName() + " is not a valid animation file");
        return false;
    }

    unsigned memoryUse = sizeof(Animation);

    // Read the format version, if not available then fallback to version 1
    String versionString(source.ReadFileID());
    unsigned version = 1;
    if (versionString.StartsWith("v") && versionString.EndsWith("#"))
        version = ToUInt(versionString.Substring(1, 2));
    else
        source.Seek(4);     // Return the source position before the format version read

    // Read name and length
    animationName_ = source.ReadString();
    animationNameHash_ = animationName_;
    length_ = source.ReadFloat();
    tracks_.Clear();
```
Based on the version number that the code could handle both old format and new format.

cadaver:
Yes, that looks rather clean, and it's extremely unlikely that the following binary data in an old format model would accidentally yield a version string. On Android, I believe seeking backward induces a reset of the compressed asset stream, requiring to read from the beginning, but since it's so close to the file start it should not be a big deal. And it will only happen for the legacy files.

aster2013:
Change load function like this.
```
bool Model::BeginLoad(Deserializer& source)
{
    // Check ID
    if (source.ReadFileID() != "UMDL")
    {
        LOGERROR(source.GetName() + " is not a valid model file");
        return false;
    }
    
    geometries_.Clear();
    geometryBoneMappings_.Clear();
    geometryCenters_.Clear();
    morphs_.Clear();
    vertexBuffers_.Clear();
    indexBuffers_.Clear();
    
    // Version
[...]

PR #1320, opened 2016-04-13
Arbitrary vertex declarations

cadaver:
This implements arbitrary vertex declaration support (issue #707), where you can decide the order, type and semantic of your vertex elements. This allows up to 4 UV coordinate sets.

For compatibility, you can still use the "legacy" vertex element masks when defining vertex buffers. The need to specify buffer binding element masks is eliminated altogether, rather if you want to override some elements with another buffer (for example positions from a morph vertex buffer) the override buffer simply has to use a higher index in Geometry or Graphics::SetVertexBuffers().

Multiple input element semantics in vertex shaders are zero-based, e.g. iTexCoord would be the first (index 0) texcoord and iTexCoord1 would be second. D3D defines semantics explicitly but in GLSL it's based on string compare of the vertex attribute name.

A new model format (UMD2) which uses vertex element declarations instead of element mask is introduced. AssetImporter will save using this format.

Testing is appreciated. I'll still have to test morphs, which could have bugs yet, and whether Android / iOS / Emscripten works, but I'm opening this for early review in any case.


cadaver:
OBJ export is presently also commented out. That will be rectified shortly.

weitjong:
Since SEM_POSITION is expected to be the first element, is there a difference between calling```VertexBuffer::GetElementOffset(,,SEM_POSITION) != 0``` and ```VertexBuffer::HasElement(,,SEM_POSITION)```?

PR #1321, opened 2016-04-13
Added ResourceCache::ReleaseResource<T> similar to GetResource<T>

damu:
Added template version of ReleaseResource similar to the GetResource function.
```
template <class T> void ResourceCache::ReleaseResource(const String& name, bool force)
```
This makes unloading resources easier and similar to the "common" loading.


cadaver:
Thanks, merged.

damu:
Uhm there was actually a compiler error in there. The static cast has to go. I had fixed that but in the wrong file (the copied header in the build dir).

damu:
I think I managed to fix it and win my fight with Git: https://github.com/urho3d/Urho3D/pull/1322

PR #1324, opened 2016-04-15
Fix miplevels for Skybox2

1vanK:


1vanK:
some levels is wrong...

PR #1326, opened 2016-04-17
Fix cubemap seams for OpenGL version

1vanK:


cadaver:
I believe this has to be ifdeffed for desktop GL only. Also, would be preferable to do at context initialization time and not every frame. Finally, have you compared to D3D9 / 11 behavior, does it sample across cube faces or not?

1vanK:
> I believe this has to be ifdeffed for desktop GL only. Also, would be preferable to do at context initialization time and not every frame.

ok

> Finally, have you compared to D3D9 / 11 behavior, does it sample across cube faces or not?

I did not see difference with DX

1vanK:
Maybe you insert this to correct place himself? I can again insert it the wrong place xD
Whether I should take to attention the lost of context?

cadaver:
ResetCachedState() or Restore() (where it sets the pixel alignment) are good places to do one-time context inits. But I can do that.

hdunderscore:
A tweet convo with john carmack suggested this feature was introduced with dx10 hardware and dx10+ enable it by default. It does appear to be true for our dx11 implementation, I'm not sure where that leaves dx9.

PR #1328, opened 2016-04-18
Enable Ninja for Windows Systems

fire:
This enables the Ninja build tool to work on Windows again on both Mingw and MSVC.

If the generator is Ninja disable precompiled headers. <- This commit disables precompiled headers if you are using Ninja. It is slightly hacky. 

Please suggest improvements.

### DIFF
--- a/CMake/Modules/Urho3D-CMake-common.cmake
+++ b/CMake/Modules/Urho3D-CMake-common.cmake
@@ -855,17 +855,26 @@ macro (enable_pch HEADER_PATHNAME)
 
         if (MSVC)
             get_filename_component (NAME_WE ${HEADER_FILENAME} NAME_WE)
+            get_filename_component (PCH_HEADER_PATH ${HEADER_FILENAME} PATH)
             if (TARGET ${TARGET_NAME})
                 foreach (FILE ${SOURCE_FILES})
                     if (FILE MATCHES \\.${EXT}$)
                         if (FILE MATCHES ${NAME_WE}\\.${EXT}$)
                             # Precompiling header file
-                            set_property (SOURCE ${FILE} APPEND_STRING PROPERTY COMPILE_FLAGS " /Fp$(IntDir)${PCH_FILENAME} /Yc${HEADER_FILENAME}")     # Need a leading space for appending
+                            if(NOT ${CMAKE_GENERATOR} STREQUAL "Ninja")
+                                set_property (SOURCE ${FILE} APPEND_STRING PROPERTY COMPILE_FLAGS " /Fp$(IntDir)${PCH_FILENAME} /Yc${HEADER_FILENAME}")     # Need a leading space for appending
+                            else()
+                                set_property (SOURCE ${FILE} APPEND_STRING PROPERTY COMPILE_FLAGS " /Fp${PCH_FILENAME} /I\"${PCH_HEADER_PATH}\" /Yc${HEADER_FILENAME}")
+                            endif()
                         else ()
                             # Using precompiled header file
                             get_property (NO_PCH SOURCE ${FILE} PROPERTY NO_PCH)
                             if (NOT NO_PCH)
-                                set_property (SOURCE ${FILE} APPEND_STRING PROPERTY COMPILE_FLAGS " /Fp$(IntDir)${PCH_FILENAME} /Yu${HEADER_FILENAME} /FI${HEADER_FILENAME}")
+                                if(NOT ${CMAKE_GENERATOR} STREQUAL "Ninja")
+                                    set_property (SOURCE ${FILE} APPEND_STRING PROPERTY COMPILE_FLAGS " /Fp$(IntDir)${PCH_FILENAME} /Yu${HEADER_FILENAME} /FI${HEADER_FILENAME}")
+                                else()
+                                    set_property (SOURCE ${FILE} APPEND_STRING PROPERTY COMPILE_FLAGS "  /Fp${PCH_FILENAME} /I\"${PCH_HEADER_PATH}\" /Yu${HEADER_FILENAME} /FI${HEADER_FILENAME}")
+                                endif()
                             endif ()
                         endif ()
                     endif ()


fire:
How to create a Ninja build.
```bat
REM Download ninja https://github.com/ninja-build/ninja/releases/download/v1.6.0/ninja-win.zip
mkdir build
cmake .. -GNinja -DCMAKE_INSTALL_PREFIX:PATH=..\install
REM copy ninja.exe to build folder
cd build
ninja
REM if you're in a admin prompt
ninja install 
```

weitjong:
Thanks for the PR, but I don't quite get it. Our buildsystem has supported ninja(-build) for quite some time now. I have just tested the CMake/Ninja generator using both native GCC and MinGW as compiler on my Linux host system with all the default build options (PCH is on and with symlink creation, etc) and the test went on successfully (ignoring CMP0058 CMake warning for dev). I don't see why it will be of any difference on Windows host system. Your comment sounds like it was working for you before and now it is broken. However, the changes that you have made in the PR does not explain it or address the bug itself (if there is indeed a bug).

fire:
The patch solves the problem on Windows where the command line length is exceeded in Ninja builds.

It disables symlink creation because those do not exist on Windows in normal non-admin accounts.

fire:
Example Mingw failure on master. The one for MSVC 14 2015 Win64 looks similiar. https://gist.github.com/fire/8dd6e9368bd0a35c24a62e62465df7fc

fire:
See description of problem that was mitigated in cmake, but isn't here because of Urho3d macros.

https://github.com/ninja-build/ninja/issues/53

weitjong:
Ok, I guess then we must fix the root cause by using response file. It has been done before for other use case. As for the symlink or mklink, you can have it sorted out just by a quick googling. I am not using admin account and I am able to mklink just fine.

weitjong:
I don't have ninja on my Windows10 partition and I don't think I will install it because of limited space I have in the partition, it is on a small SSD triple-booting. As it is, I don't think this PR is merge-ready. You should fix the problem at its root cause and resubmit again. And put the symlink/mklink issue change out.

fire:
There is bug hidden in the pch removal: 


```cmake
if (FILE MATCHES \\.${EXT}$)
                        if (FILE MATCHES ${NAME_WE}\\.${EXT}$)
                            # Precompiling header file
                            set_property (SOURCE ${FILE} APPEND_STRING PROPERTY COMPILE_FLAGS " /Fp$(IntDir)${PCH_FILENAME} /Yc${HEADER_FILENAME}")     # Need a leading space for appending
                        else ()
                            # Using precompiled header file
                            get_property (NO_PCH SOURCE ${FILE} PROPERTY NO_PCH)
                            if (NOT NO_PCH)
                                set_property (SOURCE ${FILE} APPEND_STRING PROPERTY COMPILE_FLAGS " /Fp$(IntDir)${PCH_FILENAME} /Yu${HEADER_FILENAME} /FI${HEADER_FILENAME}")
                            endif ()
                        endif ()
                    endif ()
```

$(IntDir) <- The $ needs to be escaped via $$ in Ninja.

weitjong:
> $(IntDir) <- The $ needs to be escaped via $$ in Ninja.

I am not doubting your observation or anything, but if that is true then it is quite strange. I just told you in one of my comment that I don't have any issue with PCH enabled.

weitjong:
I think the chunk that you pasted is for MSVC. That could explain why I don't have issue. I only tested ninja with MinGW and GCC, not MSVC.

fire:
```
if (ARG_USE_FILE_SYMLINK AND NOT WIN32 AND NOT MINGW)
```

If you look at the mingw log at the end, you can see that it's using a response file,

```
FAILED: cmd.exe /C "cd . && "C:\Program Files (x86)\CMake\bin\cmake.exe" -E remove Source\ThirdParty\kNet\libkNet.a && C:\tools\mingw64\bin\ar.exe qc Source\ThirdParty\kNet\libkNet.a  @CMakeFiles/kNet.rsp && C:\tools\mingw64\bin\ranlib.exe Source\ThirdParty\kNet\libkNet.a && 
```
```
@CMakeFiles/kNet.rsp <-
```

However, FALLBACK_TO_COPY https://github.com/scorvi/Urho3DIDE/blob/master/CMake/Modules/Urho3D-CMake-common.cmake#L439 causes concatenation of the terms.

weitjong:
The last issue in your comment looks like a different one than the PCH ```$$``` escape issue. If I understood correctly then the last issue could be a bug belongs to CMake itself. From the code that you posted, it is clear that it is not our script that concatenating the command together. Instead, we just tell CMake that we want to perform a fallback copies as part of a post-build step. In other CMake generators that I have seen, the post build step is implemented differently and not just "&&" the whole things together after the archiving/linking command.

fire:
There is a bug report. http://www.gccxml.org/Bug/view.php?id=14597&history=1

fire:
I've traced through the code. 

1. Maybe a workaround is to use MKLINK on Windows for the TOO LONG command line problem. The `-DURHO3D_MKLINK=ON` codepath works.

2. $$ Escaping problem: According to this post /Fd may be removed? I don't know what it does. https://lists.gnu.org/archive/html/freetype-devel/2015-10/msg00108.html 

`cmake .. -GNinja -DCMAKE_INSTALL_PREFIX:PATH=..\install -DURHO3D_MKLINK=ON`

weitjong:
For issue 1, I think we both agree that it is CMake/Ninja generator bug. The Ninja generator should have handled the post-build command like other generators do. There is no easy way to workaround this bug, even if we really want to. Fortunately though, the fallback copy in post-build would only take place when your Windows account has mklink problem. So, the easiest way to fix this without changing anything in the build system is to get your Windows account sorted out to have mklink privilege. You cannot switch the URHO3D_MKLINK on/off externally. It is set to either on or off automatically/internally based on your Windows account capability.

For issue 2. We are using /Fp here for PCH on MSVC. It is needed. But, at least I understand why this error happened. We also have to escape literal '$' as '$$' on Unix Makefile, so this is not just specific to Ninja. It just that we never have this kind of problem before in our build system because no one have tried to use MSVC compiler using a Unix Makefile generator before. In fact I also curious how you setup your build to use MSVC compiler using Ninja generator.

fire:
MSVC compiler using Ninja generator:

1. Ensure that mingw is not in PATH
1. Open Command Line Tools for the MSVC compiler
1. Use ninja in cmake and as standalone.

MINGW Compiler with Ninja Generator

1. Ensure mingw is in PATH
1. Use ninja in cmake and as standalone.

`cmake .. -GNinja -DCMAKE_INSTALL_PREFIX:PATH=..\install`
`ninja`

weitjong:
I think I should have asked it differently. Aside from the PCH issue, i.e. after you have disable the PCH, did your ninja-build actually work at all? Or this was just the tip of the iceberg of the problems. The reason I am saying this because our build system assume MSVC compiler is only being used with "msbuild" (generated using CMake/VS generator). As I commented before, MSVC has not been tested to work with Makefile or ninja-build.

weitjong:
@fire Any more update on this?

fire:
Currently busy. I'll see what time I have over the weekend.

fire:
For problem 1:

Reasoning: Switch to execute process rather than how cmake chains the commands.

```patch
       elseif (${ARGN} STREQUAL FALLBACK_TO_COPY)
         if (SLASH_D)
           set (COMMAND COMMAND ${CMAKE_COMMAND} -E copy_directory ${ABS_SOURCE} ${ABS_DESTINATION})
 -             else ()
 -                 set (COMMAND COMMAND ${CMAKE_COMMAND} -E copy_if_different ${ABS_SOURCE} ${ABS_DESTINATION})
 -             endif ()
           # Fallback to copy only one time
           execute_process(${COMMAND})
 -             if (TARGET ${TARGET_NAME})
 -                 # Fallback to copy everytime the target is built
 -                 add_custom_command (TARGET ${TARGET_NAME} POST_BUILD ${COMMAND})
 +       else ()
 +         execute_process(COMMAND ${CMAKE_COMMAND} -E copy_if_different ${ABS_SOURCE} ${ABS_DESTINATION})
         endif ()
       else ()
         message (WARNING "Unable to create symbolic link on this host system, you may need to manually copy file/dir from \"${SOURCE}\" to \"${DESTINATION}\"")
```

For problem 2 disable pch for Ninja:

Reasoning, ninja compiles fast enough on my computer to not require pch.

```patch
   # Typically, user should indirectly call this macro by using the 'PCH' option when calling define_source_files() macro
   macro (enable_pch HEADER_PATHNAME)
       # No op when PCH support is not enabled
 -     if (URHO3D_PCH)
 +     if (URHO3D_PCH AND NOT CMAKE_GENERATOR STREQUAL "Ninja")
           # Get the optional LANG parameter to indicate whether the header should be treated as C or C++ header, default to C++
```

fire:
I can't enable mklink on my computer. Research says if you're an admin, you can not mklink (as a normal non-admin action).

weitjong:
For 1)

No, that is not a proper fix. Without it the post-build step, the headers in the build tree will get stale pretty quickly. With mklink/symlink we avoid the stale problem all together. 

For 2)

> Result is it appears to work with Ninja and MSVC.

Interesting. I thought it would at least failed in one more place where we have used literal '$' without escaping. See https://github.com/urho3d/Urho3D/blob/91b799d666c453b990420cc9bba31dcd427d9ca0/Source/Urho3D/CMakeLists.txt#L218-L244. I suppose you have never tried SHARED lib type yet. And I was expecting more thing to break, but apparently I am wrong. If you could properly escape those '$' with '$$' then we may consider to support Ninja-build + MSVC use case. What you need to do is something like:
```
if (NINJA)
  do some CMake command with parameter using '$$', i.e. replace all single '$' with its double version.
else ()
  do some CMake command with parameter using single '$' as per before
endif ()
```

You should not disable PCH just because it is NINJA or because you don't need it. The code in the common module is used by all Urho3D downstream projects. Other projects may rely on the PCH to speed up the build. Anyway, the above fix is very simple and work much better than your workaround. Besides, you cannot workaround for the other place that I mentioned but the fix can be implemented similarly there.

BTW, there is something wrong with your commit. The changes is not viewable in GitHub Web interface. Probably you have had a bad merge.

fire:
Replacing $ with $$ gives me.

```ninja
[565/1065] Building CXX object Source\Urho3D\CMakeFiles\Urho3D.dir\Audio\SoundStream.cpp.obj
FAILED: C:\PROGRA~2\MICROS~1.0\VC\bin\amd64\cl.exe   /nologo /TP -DHAVE_STDINT_H -DTOLUA_RELEASE -DURHO3D_ANGELSCRIPT -DURHO3D_FILEWATCHER -DURHO3D_IS_BUILDING -DURHO3D_LOGGING -DURHO3D_LUA -DURHO3D_LUA_RAW_SCRIPT_LOADER -DURHO3D_MINIDUMPS -DURHO3D_NAVIGATION -DURHO3D_NETWORK -DURHO3D_PHYSICS -DURHO3D_PROFILING -DURHO3D_THREADING -DURHO3D_URHO2D -D_CRT_SECURE_NO_WARNINGS -ISource\Urho3D -IS:\w\Urho3D-imgui\Source\Urho3D -Iinclude\Urho3D\ThirdParty -Iinclude\Urho3D\ThirdParty\Bullet -Iinclude\Urho3D\ThirdParty\Detour -Iinclude\Urho3D\ThirdParty\Lua /DWIN32 /D_WINDOWS /W3 /GR /EHsc /D_DEBUG /MDd /Zi /Ob0 /Od /RTC1     /Fp$(IntDir)Precompiled.h.pch /YuPrecompiled.h /FIPrecompiled.h /showIncludes /FoSource\Urho3D\CMakeFiles\Urho3D.dir\Audio\SoundStream.cpp.obj /FdSource\Urho3D\CMakeFiles\Urho3D.dir\ /FS -c S:\w\Urho3D-imgui\Source\Urho3D\Audio\SoundStream.cpp
S:\w\Urho3D-imgui\Source\Urho3D\Audio\SoundStream.cpp: fatal error C1083: Cannot open precompiled header file: '$(IntDir)Precompiled.h.pch': No such file or directory
ninja: build stopped: subcommand failed.
```

I'll see what I can do for the other problem.

weitjong:
I am neither very familiar with ```cl.exe``` (MSVC) nor ```MSBuild.exe```, but the new error you just got tell me that the fix actually solves the problem you had earlier with ninja-build. It now reads the ```$$``` correctly, escapes it and then passes it as a single literal '$' to the underlying command. However, it is the cl.exe itself does not understand the meaning of '$(IntDir)'. This indicates the '$(variable)' is specific to MSBuild.exe and should only be used when using VS generator. And if so, you have to figure out in ninja-build where the Precompiled.h.pch is being generated and then adjust the relevant CMake command accordingly.

fire:
Please look at these changes:

1. Instead of postbuild in Ninja it just copies every time.
2. Not sure what /I does but it seems to work.
3. There's white space differences that I can't get remove.
4. Tried response files, but I don't know how to tell cmake to use response files.

```cmake
file(WRITE ${CMAKE_BINARY_DIR}/symlink.rsp " copy_directory ${ABS_SOURCE} ${ABS_DESTINATION}")
file(WRITE ${CMAKE_BINARY_DIR}/symlink.rsp "\"${CMAKE_COMMAND}\" copy_if_different ${ABS_SOURCE} ${ABS_DESTINATION}")
add_custom_command (TARGET ${TARGET_NAME} POST_BUILD @\"${CMAKE_BINARY_DIR}/symlink.rsp\")
```

fire:
CI broke. So I'll revise PR.

weitjong:
Thanks for working at this.

> Instead of postbuild in Ninja it just copies every time. // It executes the ${COMMAND} everytime.

No, this is still wrong. The ```execute_process()``` only executes everytime when CMake is being run  but it does not execute again when you build your project. Like this, you are forcing ninja user to run CMake from time to time to refresh the build tree.

> There's white space differences that I can't get remove.

To make the review easier and also to make your PR merge-ready, it would be great if you follow the same coding convention. 4 spaces instead of 2. But don't worry too much about them now, as long as your changes are viewable in GitHub Web interface.

> Tried response files, but I don't know how to tell cmake to use response files.

Don't go there. You cannot do that. Earlier in my comment I mentioned using response file to solve the command line buffer overflow problem because I thought we have forgotten to turn it on, which is not the case. You cannot simply generate a response file and pass it to CMake because only certain build tools understand it.

weitjong:
Instead of response file, you can probably change the extension to .bat. Then revise the post build step to invoke cmd.exe to execute the bat file.

fire:
```
string(REGEX REPLACE "COMMAND" "" COMMAND ${COMMAND})
file(APPEND ${CMAKE_BINARY_DIR}/symlink.bat ${COMMAND})
add_custom_command (TARGET ${TARGET_NAME} POST_BUILD
COMMAND ${CMAKE_BINARY_DIR}/symlink.bat)
```

Something similar to this doesn't work. It still complains about too long command lines.

fire:
I give up. Will use mklink on windows as a user account. The patches to other bits can be reviewed. Perhaps it's possible to stop the build if it's Ninja and without mklink.

Not sure what I can do until https://cmake.org/Bug/view.php?id=14597 is fixed.

weitjong:
It is hard for me to see your problem from here without actually seeing the generated files in the build tree or the error message/log itself. All I can say is, it is technically possible. And this approach has been done before elsewhere in one of our CMakeLists.txt to beat the Windows's command line buffer overflow problem. See https://github.com/urho3d/Urho3D/blob/2151426adca48e58a424b8ebe32ea99fd8c200b1/Source/Urho3D/CMakeLists.txt#L363-L367. So, using the .bat file approach is definitely valid for consideration.

fire:
I'll look into it.

fire:
How do the other parts of the patch look?

weitjong:
I have been thinking about this. If MSVC compiler toolchain could work in a ninja-build then it must be able to work in any other build tools too like: mingw-make and Unix make (e.g. the one shipped in Android NDK). IMHO, although your change is a step in the right direction, it is still incomplete. I would rather only merge the changes is complete.

weitjong:
I just quickly tested the build using Unix Makefiles generator. It has the same PCH issue with your ninja build, so I might be able to help since I can now reproduce the issue. And yes, when URHO3D_PCH=0 then the static build went on successfully. I haven't tested shared build yet in my quick test.

fire:
It is great that you can help resolve the issue. Hopefully you can think of another approach. I've posted my response file approach that didn't work a few comments ago. https://github.com/urho3d/Urho3D/pull/1328#issuecomment-213635539

weitjong:
I do not intend to install ninja on my Windows 10 partition because of space constraint, so I probably won't be able to help much on second issue.

fire:
Ah, the response file approach is actually to solve the URHO3D_PCH=0 issue. Everything else works.

weitjong:
I have enhanced the buildsystem to support MSVC compiler toolchain to be used in other non-VS generator (i.e. non msbuild). I only tested it with "Unix Makefiles" generator using "make.exe" from Android NDK. So far both STATIC and SHARED with PCH enabled built successfully in my test. YMMV with ninja-build, however. Let me know the result in any case.

BTW, as far as I understand, the response file approach is to address the CMake/Ninja generator bug where it wrongly "&&" all the post-build commands. So, if you still do not have MKLINK setup correctly then you will still be hit by this other issue. As this kind of setup is really niche within niche, I will not attempt it myself (not unless I have extra SSD to install more build tools to verify and try). The "Unix Makefiles" + "MSVC" combo should work with or without MKLINK, because CMake/Unix-Makefiles generator does not have the same bug (i.e. it does not naively "&&" post-build commands together in the generated Makefile).

weitjong:
Feel free to reopen if you have more updates.

PR #1329, opened 2016-04-18
imgui integration for Urho3d

fire:
Request for comments.

Based on:
- @svifylabs (Please comment on contributing your code)
- https://github.com/ocornut/imgui

Fonts in the imgui class are not Font resources in Urho3d. I am not familiar with those system. Please advise.


cadaver:
Thanks for the work!

Without script bindings, this is a bit of a departure compared to other Urho features and I would not be directly comfortable including this. However I'm not necessarily saying that you should go into the effort of writing those, as all IMGUI features may not even be easily adaptable.

I also necessarily don't want to be the guy who always shoots down ideas so if other core devs feel that this should definitely be included then it can be. My view is that this could just as well work as a well-maintained external project that users can include into their Urho applications if they want.

JoshEngebretson:
@cadaver  I don't think I recall seeing anything ever shot down, though understand how it could get to feeling that way.  Pull request pressure is a real thing!  You do a great job curating, going as far as to clean up PR's so they CAN be landed!

The script coverage requirements for Urho3D could maybe be relaxed if the C++ code is high quality and useful in general?  Requiring both before landing could slow down contributions for functionality that can be added in followups or by other committers?  Autobind script tools help on this front definitely... 

IMGUI is a great kit for editor UI's especially and has good traction.  I've thought about using it for Atomic's editor and keeping the TurboBadger UI for game usage.

cadaver:
There is certainly a balance, it's unreasonable to demand a PR to be perfect in all respects, however we also want to avoid a situation where code is just dumped to become our maintenance responsibility. I'm not saying this case is either.

I don't actually remember anything 3rdparty being directly bound to the script API; these are usually harder to bind as they can have their own structs, string classes etc. UI libraries certainly rank on the top of binding effort required, if for nothing else then the sheer amount of classes that need to be exposed.

On a more philosophical level is the question how many UI libraries an engine should integrate :)

JoshEngebretson:
I'd venture that tooling UI and game UI could be and often are independent systems, so a theoretical limit to number of UI integrations of two?
 
UI binding can be especially horrific, yeah.  IMGUI is a bit different in the immediate mode usage (so less event handling, no lifetime management which is a BIG one for UI bindings, etc).

I really hear you on the maintenance part on the PR side of things, we're in similar boats.

cadaver:
The lack of events is indeed a big plus.

If wanting to take this into a scriptable direction, I believe it would benefit to build a wrapper Urho subsystem around it, to ensure Urho classes like String, Vector2 etc. are being used.

fire:
I'll look into wrapper Urho3d subsystems, but I can appreciate suggestions. 

When I have a spare moment, I'll report what I found.

fire:
Will see if I can work on this this weekend.

fire:
I'm going to probably resubmit this because it has gone stale.

Ptitpopcorn:
Thanks for the integration fire.

rokups:
Would it not be more worthy to integrate nuklear UI if such immediate mode UI is desired? ImGui is cool and all but it's use would be limited only to tools as it it's skinning capabilities are quite limited. nuklear on the other hand while being very similar to ImGui provides full skinning support so it immediately becomes useful as game UI too. It also has no hidden global state and api is way cleaner. Other than that both libs are more or less equal.

OvermindDL1:
I am curious, how do imgui and nuklear UI's work?  Do they setup an 'immediate-style' interface for a GUI, but in the backend it makes a diff between the last and current 'states' and only applies those changes?  If so that would be highly efficient so either could be quite useful.  I cannot think of any other backend style that would let an immediate-style interface be at all performant though as even heavy caching would potentially not equal in speed.

rokups:
Upon every frame UI is rebuilt and rendered in form of index/vertex buffers. Only parts that are visible cause code to be executed. It is very efficient, doubt you can squeeze out more juice by some bad ass diffing algorithms that likely would be even slower.-------- Original message --------

OvermindDL1:
Sounds very immediate-mode rather than a more efficient version there-of.  Wouldn't that get very costly especially the more complex a UI is?

rokups:
I bet you would have to try real hard to make it slow.. It is blazing fast really, but i bet we all could benefit from actual benchmarks. *This is left as exercise for the reader ;))*

OvermindDL1:
Heh yep.  I've always thought of an 'Elm'ish style UI for a game might be interesting, been playing with the idea for a few weeks now.  I should probably just make something.

fire:
Can someone help me figure out why the shared builds are breaking?

1vanK:
You forgot URHO3D_API ?

cadaver:
Some quick comments:
- There appear to be tabs in the added Urho code. Urho convention is to use spaces, please use them. Note that you don't have to convert any 3rd party code.
- I'd prefer the IMUI subsystem to default off, at least as long as it lacks scripting support and circumvents the resource system for fonts.
- Perhaps add a CMake build option similar to URHO3D_NETWORK etc. so that if someone does not use the library, it can be not compiled in to reduce binary size
- A brief documentation section in Reference.dox would be appreciated.

fire:
I am having trouble with linking:

https://gist.github.com/fire/0a2a32a5330bcc779b6d2959ca75e9e6

It doesn't seem to be working with shared libraries.

fire:
Posted my attempt at the shared library problem.

cadaver:
To fix the shared library Windows builds, IMGUI_API should be declared dllexport when building Urho and dllimport otherwise. But only if Urho is built as shared library. Similar define logic exists for Urho itself in Urho3D.h, but you will need to setup it separately for Imgui library build in CMake.

sabotage3d:
Hi I quickly tested this on OSX and it works fine but I did another test on Android and the imgui doesn't display at all.

cadaver:
Good catch, that's likely because imgui is trying to access resources directly (at least for the fonts), and it has no idea that on Android the SDL rwops / Urho's File class should be used, so that files inside apk can be accessed.

fire:
I'm migrating over to nuklear ui. Haven't had much time to work on imgui.

yushli:
@fire Any chance that you will share the integration with nuklear ui?

PR #1331, opened 2016-04-19
Add Graphics draw call with base vertex index (Offset Indexing)

svifylabs:
only  supported for opengl 3  and direct3d

### DIFF
--- a/Source/Urho3D/Graphics/Direct3D11/D3D11Graphics.cpp
+++ b/Source/Urho3D/Graphics/Direct3D11/D3D11Graphics.cpp
@@ -824,7 +824,7 @@ void Graphics::Draw(PrimitiveType type, unsigned indexStart, unsigned indexCount
     ++numBatches_;
 }
 
-void Graphics::Draw(PrimitiveType type, unsigned indexStart, unsigned indexCount, , unsigned baseVertexIndex, unsigned minVertex, unsigned vertexCount)
+void Graphics::Draw(PrimitiveType type, unsigned indexStart, unsigned indexCount, unsigned baseVertexIndex, unsigned minVertex, unsigned vertexCount)
 {
     if (!vertexCount || !shaderProgram_)
         return;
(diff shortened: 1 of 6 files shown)


cadaver:
Cool. This can be included as-is, but have you considered expanding the Geometry class to support defining base vertex index? This would allow it to be used in 3D scene rendering.

svifylabs:
hmm can look into expanding the Geometry. 
this was for implementing imgui. 

how can i implement a compile time check for  GLEW_ARB_draw_elements_base_vertex ? 

glDrawElementsBaseVertex is only implemented for opengl 3.2 and opengl es 3.2. how would i best test for that ? And should i add an assert if someone uses it for not supported versions or just remove it at compile time ?
 
for now added runtime check against gl3Support.

svifylabs:
can look into expanding the Geometry class.

cadaver:
In general we should check extensions. GLEW has a nasty bug however that 3.2 doesn't support checking them from an OpenGL 3.x context (at least Windows). Checking the function not being null is another way, but not guaranteed. In this case it's best to just not render anything if the check fails, we don't want to assert.

Thinking of Geometry support, it may be of limited usability so this PR may be best to merge as is. To be fully useful, the Geometries would have to come from Model resources, which would also need added support. But usually each model defines its own VB / IB and it's not useful. Instead, having to check for the base vertex index at draw time would actually slow down all draing code.

svifylabs:
ok. so it can be merged.
can that be used to improve the batch rendering ? dont know how the rendering is implemented.

cadaver:
In theory yes, by packing multiple models into the same buffer, but there should be a higher-level system coordinating this, and it's not obvious there would be benefits in each case, so it would likely be a case of application-specific tuning, while making the rendering more complex. For comparison Ogre 2.x goes quite heavily in this direction but as a consequence it's harder to use.

svifylabs:
ok thanks for the info !

PR #1332, opened 2016-04-19
extend Model for named geometries

svifylabs:
implemented the model class from atomic game engine ... without the version index .. 
because of the new model implementation, i thought of pushing this now.

todos:
scripting binding
editor extention

### DIFF
--- a/Source/Tools/AssetImporter/AssetImporter.cpp
+++ b/Source/Tools/AssetImporter/AssetImporter.cpp
@@ -974,7 +974,7 @@ void BuildAndSaveModel(OutModel& model)
 
         SharedPtr<Geometry> geom(new Geometry(context_));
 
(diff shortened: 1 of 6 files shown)


cadaver:
I would prefer not to add anything to StaticModel, as any additional memory access in there slows down all static geometry rendering. You are already able to essentially hide a submesh by setting a material with no techniques, or technique without passes.

cadaver:
Also, when I think of the idea of named geometries I rather think these should be their own scene nodes which you could simply enable / disable. Will need to check AssetImporter's prefab mode and modify if necessary so that it keeps the geometries in the same nodes as in the original source file, instead of collating them into one node. (Model mode naturally can't do anything else than to save a single model)

svifylabs:
that sounds like a better approach. 
i just needed this feature for manipulating my models and testing them but dont use this at runtime, only for editing. 
so it can be closed :-/ will create a gui assetimporter for that ...

PR #1333, opened 2016-04-20
Extend Vector and HashMap 

svifylabs:
added TryGetValue and Insert for the HashMap class
and added EraseSwap and RemoveSwap for the Vector class

i needed them for my game ...
i also have a SortedVector and SparseVector implementation that could be merged too ....

### DIFF
--- a/Source/Urho3D/Container/HashMap.h
+++ b/Source/Urho3D/Container/HashMap.h
@@ -385,6 +385,42 @@ public:
             InsertNode(*it++);
     }
 
+    /// Insert a key and value and return iterator to the value and if the value was already added.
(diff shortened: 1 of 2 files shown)


cadaver:
Nice features!
However I don't think findExisting should be exposed in a public function. findExisting = false is used in some of the member functions internally when it's certain that a duplicate insertion will not result (when the existing value has already been searched for unsuccessfully.) But the user could cause duplicate value insertion.

cadaver:
I can rework the function as I merge, so you don't necessarily have to change it yourself. The function signature may change as a result.

svifylabs:
yes thats ok. i think i never used the findExisting parameter and dont know why i added it ...

PR #1338, opened 2016-04-28
Fast check whether an unsigned integer is a power of two.

Y-way:
Fast check whether an unsigned integer is a power of two.


weitjong:
hmm... isn't that could be simplified further to
```
{
    return !(value & (value - 1));
}
```

cadaver:
Thanks for the initial work! Applied @weitjong 's further improved version.

Y-way:
My pleasure！

PR #1340, opened 2016-04-30
Editor: info about selected model (thanks MonkeyFirst for idea)

1vanK:


1vanK:
fixed unnecessary spaces

PR #1345, opened 2016-05-04
support resource drag to scene

299299:


cadaver:
Will have to check if there's already code for offsetting a model in the spawn editor, in that case GetNodeSize() would be redundant and can be replaced with the existing code.

299299:
found this variable : alignToAABBBottom

cadaver:
Thanks for the modification, will check now.

cadaver:
Merged. Did some further cleanup and support for raycasting the world position based on cursor pos.

PR #1348, opened 2016-05-04
Add substring comparison. Improve StartsWith, EndsWith functions.

reattiva:
This should make StartsWith, EndsWith faster on very long strings.


reattiva:
Ok, the default value for the last argument breaks a lot of things... I'll work on it.

reattiva:
Can't replicate the errors of the automatic builds, giving up.

PR #1353, opened 2016-05-06
Simple looped animation for UI

1vanK:
If all is corret, I will add bindings


cadaver:
I think that's an acceptable solution, it's naturally reimplementing things that also exist elsewhere (Urho2D) but short of extensive refactoring scene elements (drawables) cannot be rendered in the UI. Other opinions?

1vanK:
I wrote the code for the needs of my project, but perhaps it be useful for other peoples. This is not a universal solution, but better than nothing :) UI in any case requires significant changes, so later it can be rework.

1vanK:
Added bindings and squashed

1vanK:
How to force build lua bindings when I change *.pkg files?

weitjong:
It should be automatic as the build system knows the build depends on the pkg files too.

weitjong:
> I think that's an acceptable solution, it's naturally reimplementing things that also exist elsewhere (Urho2D) but short of extensive refactoring scene elements (drawables) cannot be rendered in the UI. Other opinions?

I just did a quick test. Out of the box it is possible to render a "small" RTT scene containing 2D/3D drawables into a texture's surface and then feed that texture object to current Sprite class. For example, I could render the 2D spriter animation in the sprite object without any modification. Granted that it is not as easy to setup compared to what have been proposed in this PR.

cadaver:
Yes, that is certainly usable too, with the caveat that if you need to render several separate RTT's, it can get costly, compared to an immediate draw call to the backbuffer.

weitjong:
I think in this case we could use a slightly bigger RTT instead of several and use the Sprite::SetImageRect(), assuming there are no interferences in the RTT scene. But I see your point.

cadaver:
After considering @weitjong 's remarks some more, and also remembering that you can animate a sprite's ImageRect attribute with a value animation, I'm starting to consider that this shouldn't be merged, as the functionality can already be implemented in a number of ways and is therefore not orthogonal.

Also there's nothing wrong with subclassing a Urho3D class in your application to extend the functionality, everything necessarily doesn't need to be part of the engine itself. In that case, sorry for the wasted work of writing the bindings!

1vanK:
No problems, but I do not understand how to use a value animation in this place, because it interpolates values smoothly, rather than fixed step; and how to pack undetermined amount of elements on the fly to single RTT

cadaver:
Ah, true. Either the IntRect interpolation should just be removed, or ValueAnimation interpolation should be turnable off. You see quite nicely in the LightAnimation example how a string is being "animated".

1vanK:
Also we can add example of sprite animating throught ValueAnimation. Perhaps someone will the need arise also

cadaver:
Done. Interpolation method IM_NONE added, and simple sprite animation example included in LightAnimation. Though that animates the texture ref, not image rect.

PR #1356, opened 2016-05-12
UIComponent

1vanK:
If this is acceptable, then I completed yet. Here is a modified example 02_HelloGUI

### DIFF
--- a/Source/Samples/02_HelloGUI/HelloGUI.cpp
+++ b/Source/Samples/02_HelloGUI/HelloGUI.cpp
@@ -36,6 +36,7 @@
 #include <Urho3D/UI/Window.h>
 
 #include "HelloGUI.h"
+#include "Mover.h"
 
 #include <Urho3D/DebugNew.h>
 
@@ -46,6 +47,7 @@ HelloGUI::HelloGUI(Context* context) :
     uiRoot_(GetSubsystem<UI>()->GetRoot()),
     dragBeginPosition_(IntVector2::ZERO)
 {
+    context->RegisterFactory<Mover>();
 }
 
 void HelloGUI::Start()
@@ -136,6 +138,10 @@ void HelloGUI::InitWindow()
 
     // Add the title bar to the Window
     window_->AddChild(titleBar);
+    
+    // Create ui component for the Window
+    Mover* mover = window_->CreateComponent<Mover>();
+    mover->Init(Vector2(0.0f, 0.0f), Vector2(40.0f, 40.0f));
 
     // Apply styles
     window_->SetStyleAuto();
--- a/Source/Samples/02_HelloGUI/Mover.cpp
+++ b/Source/Samples/02_HelloGUI/Mover.cpp
@@ -1,0 +1,61 @@
+//
+// Copyright (c) 2008-2016 the Urho3D project.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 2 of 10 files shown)


cadaver:
As an idea, this is good. I think probably the end state we would want would be to have UI elements in a scene hierarchy, where you can freely add logic components or scripts to them. But since the UI is not in the scene graph, and this is a separate implementation of components, this ends up duplicating a lot of code & functionality, and there would be even more duplication if you were to add scriptability and editor editing capability to the UI components. Due to the reason of duplication I likely wouldn't be merging this.

1vanK:
Do you mean that the UI will be instance of a scene and UIElements will be Components for nodes?
```
Engine::RegisterSubsystem(new UI());

class UI : Object
{
    Scene* scene_;

    Scene* GetRoot()
    {
        return scene_;
    }
}

CreateWindow()
{
    Scene* root = GetSubsustem<UI>()->GetRoot();
    Node* wndNode = root->CreateChild();
    Window* wnd = wndNode->CreateComponent<Window>();
    ...
}
```

cadaver:
If taken to the extreme, there doesn't need to be a "singleton" UI subsystem (except perhaps for convenience), but rather you would instantiate a "UI canvas" or similar component into the scene, that would be able to take input, and render UI elements in children beneath itself in the scene tree.

For example if you think of Doom3, the HUD could be one instance of an UI canvas, and the UI's you interact in the game world (on monitor screens) would be others.

svifylabs:
unity's ui system is open source https://bitbucket.org/Unity-Technologies/ui 
and i think it is a good example how urho3d can implement it.  i would add it if  someone can help me with the rendering part. 

i did add components for uielements, i split it up for drawable components ( has GetBatches() )  and interactable components (with OnHover,OnWheel ...) but did not find it good  enough ... 
my reference for that was https://github.com/ChilliWorks/ChilliSource/tree/master/Source/ChilliSource/UI 

i really would like to see urho3d with a scene ui system.

1vanK:
We could add new components to the engine one by one, but they need to have some new names (prefix?). Or the new system will replace the old UI completely?

cadaver:
If it would turn out good, then the old UI could be axed, but I'd imagine they would need to live side by side quite a while.

In Aster's branch in https://github.com/urho3d/Urho3D/tree/component-based-ui/ the new component based classes are named UIX. That branch is just probably quite a bit out of date since it was last worked in 2014.

1vanK:
> i would add it if someone can help me with the rendering part.

I little familiar with Unity but it seems to me that the UICanvas in Unity is equivalent to Renderer2D in Urho, so we already have rendering part :)

1vanK:
Can we add
```
SharedPtr<Object> userData_
```
to UIElement or to Object class to have attached object which will be automatically deleted with UIElement?

weitjong:
`UIElement:SetVar()` not good enough for you?

1vanK:
Can Variant store SharedPtr ?

weitjong:
Yes, it can.

eugeneko:
@1vanK Custom Variant types should be able to store SharedPtr without allocation.

1vanK:
```
#include <Urho3D/Urho3DAll.h>

class ElementLogic : public Object
{
    URHO3D_OBJECT(ElementLogic, Object);

public:
    ElementLogic(Context* context) : Object(context)
    {

    }

    ~ElementLogic()
    {
        URHO3D_LOGDEBUG("UI COMPONENT DELETED");
    }
};
```

```

    void CreateUI()
    {
        UIElement* testElem = UI_ROOT->CreateChild<Text>();
        SharedPtr<ElementLogic> testComp(new ElementLogic(context_));
        testElem->SetVar("comp", Variant(testComp));
```

```
[Wed Oct 25 03:55:39 2017] DEBUG: UI COMPONENT DELETED
```

How to use it correct?

cadaver:
Variant stores WeakPtrs of RefCounted (probably this choice was to avoid lifetime issues in scripting), so it would have to be kept alive elsewhere.

eugeneko:
> How to use it correct?

Smth like this must work. It'd be nice test for generic variants, btw.
```testElem->SetVar("comp", MakeCustomValue(testComp));```

1vanK:
> Smth like this must work.

Thanks!

1vanK:
> For example if you think of Doom3, the HUD could be one instance of an UI canvas (in a separate scene, rendered with separate camera), and the UI's you interact in-world would be other instances, rendered as part of the game scene.

How to do clipping of childred elements, when its is transformed in 3d-space (scissor does not work)? I thought about stensil, but number of elements will be limited 8bit buffer or will need to re-create whole stensil stack when rendering every element (also stencil is used already for optimization inside the engine).

Is there any other way than clip geometry on CPU side?

cadaver:
Shader uniform or extra vertex data to guide the clipping?

1vanK:
Good idea!

PR #1367, opened 2016-05-18
Added EventProfiler

### LINKED ISSUE #1360: EventProfiler
hey 
What do you think of an EventProfiler ? It just inherits from the profiler class (for now) to create a new class. 
The problem was that i had to somehow save the eventnames (if StringNames are used for the event IDs, that would be easier to implement), so i used a static HashMap ...

modified:
> // Object.h
> /// Register event names.
> struct EventNamesRegistrar
> {
>     static StringHash RegisterEventName(const char* eventName);
>     static HashMap<StringHash, String> eventNames_;
> };
> 
> /// Describe an event's hash ID and begin a namespace in which to define its parameters.
> #define URHO3D_EVENT(eventID, eventName) static const Urho3D::StringHash eventID(EventNamesRegistrar::RegisterEventName(#eventName)); namespace eventName

and added 
> // Object.cpp
> void Object::SendEvent(StringHash eventType, VariantMap& eventData)
> {
> ...
> #ifdef URHO3D_PROFILING
>     Urho3D::AutoEventProfileBlock autoEventprofileblock(GetSubsystem<Urho3D::EventProfiler>(), EventNamesRegistrar::eventNames_[eventType].CString());
> #endif
> ...

thats how it looks
http://imgur.com/NbPJfrY

svifylabs:
hey this resolves #1360 
It is my first try for an event profiler. 
It uses the same logic as the profiler. 
It collects events from Object::SendEvent(StringHash eventType, VariantMap& eventData).

Event names are collected in a global HashMap see EventNameRegistrar in Object.h.
Should EventNameRegistrar be in a URHO3D_PROFILING block as well ? 

I dont like the activation and deactivation process, if possible i want just use RegisterSubsystem and  RemoveSubsystem for that. What do you think?

### DIFF
--- a/Source/Urho3D/AngelScript/EngineAPI.cpp
+++ b/Source/Urho3D/AngelScript/EngineAPI.cpp
@@ -76,6 +76,7 @@ static void RegisterDebugHud(asIScriptEngine* engine)
     engine->RegisterGlobalProperty("const uint DEBUGHUD_SHOW_STATS", (void*)&DEBUGHUD_SHOW_STATS);
     engine->RegisterGlobalProperty("const uint DEBUGHUD_SHOW_MODE", (void*)&DEBUGHUD_SHOW_MODE);
     engine->RegisterGlobalProperty("const uint DEBUGHUD_SHOW_PROFILER", (void*)&DEBUGHUD_SHOW_PROFILER);
+    engine->RegisterGlobalProperty("const uint DEBUGHUD_SHOW_EVENTPROFILER", (void*)&DEBUGHUD_SHOW_EVENTPROFILER);
     engine->RegisterGlobalProperty("const uint DEBUGHUD_SHOW_MEMORY", (void*)&DEBUGHUD_SHOW_MEMORY);
     engine->RegisterGlobalProperty("const uint DEBUGHUD_SHOW_ALL", (void*)&DEBUGHUD_SHOW_ALL);
     engine->RegisterGlobalProperty("const uint DEBUGHUD_SHOW_ALL_MEMORY", (void*)&DEBUGHUD_SHOW_ALL_MEMORY);
--- a/Source/Urho3D/Core/Context.h
+++ b/Source/Urho3D/Core/Context.h
@@ -25,6 +25,7 @@
 #include "../Core/Attribute.h"
 #include "../Core/Object.h"
 #include "../Container/HashSet.h"
+#include "../Container/Ptr.h"
 
 namespace Urho3D
 {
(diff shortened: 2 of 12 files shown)


cadaver:
The initial performance hit comes already from trying to look up the EventProfiler subsystem. That's why I suggested a static bool flag in Object class (could also be a static in EventProfiler), to avoid the lookup if event profiling is not wanted. I agree that the mechanism to activate/deactivate EventProfiler as its instance member function is unnecessary.

cadaver:
There seems to be a misunderstanding. I don't think subsystems should be queried for statically. Rather I just suggested having a static bool be checked first for whether event logging should be done. If that bool is true, then get the subsystem normally.

svifylabs:
ah sry i just had to test something ... did forget to branch it out sry. will remove that.
but is that something usefull or really bad to do ?

cadaver:
I think it's better to require an access to Context for normal (non-static) subsystem query. Sending events requires a valid Object in any case, which means it will have Context instance access.

Logging is the only exception; it holds a Log instance pointer itself so that logging can be called statically without a valid Object.

1vanK:
Hm,
```
         screenshot.SavePNG(fileSystem.programDir + "Screenshot_" +
                            time.timeStamp.Replaced(':', '_').Replaced('.', '_').Replaced(' ', '_') + ".png");
```
does not works, when path contains cyrillic chars (and no any warnings in log), but
```
screenshot.SavePNG("Screenshot_" +
                            time.timeStamp.Replaced(':', '_').Replaced('.', '_').Replaced(' ', '_') + ".png");
```
works.

By the way, log.Error() is missing in doc http://urho3d.github.io/documentation/HEAD/_lua_script_a_p_i.html

cadaver:
WriteRaw is "Print" in AngelScript API. Could also expose at least the basic form of WriteRaw.

SavePNG is stbi_image_write territory, which may not support utf8 paths as expected. It would be preferable to get just a memory buffer from it and handling file IO with Urho files.

1vanK:
> WriteRaw is "Print" in AngelScript API. Could also expose at least the basic form of WriteRaw.

I understood my mistake, that I opened wrong doc page (LUA insted AS) XD

cadaver:
It seems we can write images to memory and save to file manually, which could work more reliably.

weitjong:
After merging and reviewing this commit, I feel there are too many code duplication in the EventProfiler and EventProfilerBlock classes. So, I decided to refactor them to subsclass from Profiler and ProfilerBlock classes, respectively, and have been successful in doing so. However, IMHO, what left behind after my refactoring do not really warrant for creation of new classes at all. I feel I can go further to eliminate them but still preserving the new "event profiling" functionality some how. Should I go further?

PR #1373, opened 2016-05-18
Update GLEW to 1.13.

cosmy1:
Its time to update GLEW as 1.9 is from 2012.


cosmy1:
@cadaver What happened to the stale freetype-update branch? Now is old but I would like to know if you encountered any issues as I want to update FreeType to 2.6.3 when I find a bit of more free time.

cadaver:
If I remember correctly a Freetype update was done in master after that branch. You just need to watch out for Freetype's compilation or configuration potentially changing, since we have our own CMakeLists for it, and check if the built-in example fonts still render OK. Particularly the UI default / editor font has been picky when Freetype's hinting implementation changes.

cadaver:
Yeah, compiles OK, hinting is ugly.

cosmy1:
Ok thanks.

cadaver:
GLEW itself seems OK to merge. It still has the bug of requiring glewExperimental to be set, as it doesn't know how to read extensions properly on a GL3 core context.

PR #1376, opened 2016-05-20
Civetweb update.

cosmy1:


cadaver:
Have you tested that basic http requests still work as expected?

cosmy1:
Yes I tested it and works.

cosmy1:
What about making a HTTP request sample?

cadaver:
It would be good for regression testing, yes. If you want to do one, feel free to.

cosmy1:
I just made a basic sample that request date and time but I had other ideas which require HTTPS, like printing Urho3D contributors by using GitHub APIs. If is ok for you I just PR that basic sample or if you have other ideas let me know :smiley:

cadaver:
A basic sample is fine.

PR #1379, opened 2016-05-22
Enhanced text effects.

bengtsts:
Adds drop shadow offset, stroke thickness and rounded stroke to be used in editor. Default stroke is the same as the previous 1 pixel stroke.


cadaver:
Good feature additions! Can you update also the script bindings, and add the attributes & "passthrough" accessors also for Text3D, which contains a Text object?

Minor nitpicks: the files don't need to be executable, and the new get accessors in Text are not const, which they should be.

cadaver:
Thanks! Will merge soon.

bengtsts:
I found a few things to do...

- Large stroke values on fonts made with small dots -- like fonts that have a dot in the number 0 -- can create gaps between the font and stroke, e.g. rather than a stroke, a halo is drawn around the dot. A stroke fill flag can be made to address this issue when it arises though it would increase the glyph count a lot.

- Make stroke thickness value unsigned. Its odd to have a negative stroke thickness especially since signed and unsigned values produce the same result.

- Push text effect values to Text shader.

- Generate custom stroke thickness in Text shader using an implementation better suited for distance field font rendering.

The enhancement should be "complete" with these.

PR #1382, opened 2016-05-22
Add Http Request demo sample.

cosmy1:
Commited just AngelScript version for now to avoid CI failure as require #1381 to be merged first. (Lua version have some issues with Deserializer API, I need to investigate because may be a bug in Urho3D/tolua++).

### DIFF
--- a/Source/Samples/43_HttpRequestDemo/CMakeLists.txt
+++ b/Source/Samples/43_HttpRequestDemo/CMakeLists.txt
@@ -1,0 +1,33 @@
+#
+# Copyright (c) 2008-2016 the Urho3D project.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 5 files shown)


cadaver:
#1381 was OK and merged.

cosmy1:
The lua sample have some issue with ```ReadLine()```.
http://puu.sh/p26Oj/ea45e0b8d3.png

I will look in depth later.

cadaver:
I recommend to just commit the offending code, that way more people can look into it.

cosmy1:
Ok commited.

cadaver:
Merged. The crash was related to multiple inheritance and was fixed by copying the Deserializer functions into HttpRequest .pkg file, similar to File.pkg, instead of letting tolua++ know that HttpRequest is a Deserializer subclass.

cosmy1:
:+1:

PR #1384, opened 2016-05-23
compile on arm with opengles2

proller:
Now almost all arm boards support opengl es2 with libs located in standard places (-lGLESv1_CM -lGLESv2)
except rpi - it have custom libs in /opt

cmake -DVIDEO_OPENGL=0 -DURHO3D_ANGELSCRIPT=0
need disable opengl because https://github.com/urho3d/Urho3D/blob/master/Source/ThirdParty/SDL/src/video/SDL_video.c#L48
and angelscript not compiles under arm

tested on odroid x2 (armv7)  and odroid c2 (aarch64)

Also RPI - strange flag, its just linux with arm arch and non standard libs path


weitjong:
Thanks for your PR. On one hand after integrating the original SDL's CMakeLists.txt (which performs all the heave duty target platform auto-detecting work for us), it seems like we have open ourselves to many more target platforms than the ones we currently officially support. On the other hand, however, there is a limit of how many platforms we could officially support due to the size of Urho3D core team. It is not just a one-time porting work but we need people to actually become the "maintainer" for that port, i.e. it does not become an additional burden later to the existing core team. Having said that, I suppose we can make an exception for "generic ARM" support. @cadaver, what's your view?

> Also RPI - strange flag, its just linux with arm arch and non standard libs path

The same can be said for ANDROID. They are both Linux on ARM in essence. :smile: I guess that is exactly the purpose of the flag, to differentiate them from one another.

cadaver:
If the ARM changes in this PR are all that's required it doesn't look like a big deal. In this case the additional platform is mostly an adaptation of existing platforms (Linux / RPI). Without a maintainer and CI tests, it's still an unoffical / unsupported platform that we don't guarantee that it stays in working order, but at this point it's fine by me to include.

weitjong:
Ok. I will handle this PR. Obviously it needs more work as it has broken the rest. I also don't believe AS doesn't work on 32-bit ARM, but I will not have any mean to verify it or fix it.

proller:
> >Also RPI - strange flag, its just linux with arm arch and non standard libs path

>The same can be said for ANDROID. They are both Linux on ARM in essence. ￼ I guess that is exactly the purpose of the flag, to differentiate them from one another.

ANDROID can be on x86 and mips archs and have much more differences (stdlibs, ...)

RPI is just linux with preferred opengles libs located at non standard path

now i see small problem with default lib selection - most arm boards gave gpu on soc and accelerated opengles but software mode opengl - for arm better to prefer opengles, and opengl for x86 


about failed tests: 
https://github.com/proller/Urho3D/blob/arm/CMake/Modules/Urho3D-CMake-common.cmake#L1560
at this line need to use already checked flag like HAVE_VIDEO_OPENGLES or better split to HAVE_VIDEO_OPENGLES_V1 HAVE_VIDEO_OPENGLES_V2
and sdl already run checks at this moment and set this flags
https://github.com/urho3d/Urho3D/blob/master/Source/ThirdParty/SDL/cmake/sdlchecks.cmake#L731
but all this flags HAVE_VIDEO_OPENGLES_V1 ... HAVE_VIDEO_OPENGLES not defined at Urho3D-CMake-common.cmake#L1560

angelscript: i will try look deeper and find problem (it was something with asm).

weitjong:
> about failed tests: 
https://github.com/proller/Urho3D/blob/arm/CMake/Modules/Urho3D-CMake-common.cmake#L1560
at this line need to use already checked flag like HAVE_VIDEO_OPENGLES or better split to HAVE_VIDEO_OPENGLES_V1 HAVE_VIDEO_OPENGLES_V2
and sdl already run checks at this moment and set this flags
https://github.com/urho3d/Urho3D/blob/master/Source/ThirdParty/SDL/cmake/sdlchecks.cmake#L731
but all this flags HAVE_VIDEO_OPENGLES_V1 ... HAVE_VIDEO_OPENGLES not defined at Urho3D-CMake-common.cmake#L1560

The highligted macro in the Urho3D-CMake-common module is already earmarked to be deprecated. It will be probably be removed in the refactor-buildsystem branch soon. The library dependency will be managed differently in the near future. Currently we sort of hard-coded in this macro, which is not ideal. The library dependency should be managed transitively, which is not the case now. For example, the GLESv2 dependency is coming from SDL and Urho3D depends on SDL. So, although Urho3D depends on GLESv2, it should get it transitively via SDL and not by hard-coding what we "think" is required by SDL in this macro.

Until the refactoring is done, I guess you can still modify this macro anyway you like to fix your problem.

proller:
now building ok, graphics ok, but some problems with running angelscript on arm happens.

weitjong:
I have temporarily merged your PR into a new branch so that it can be tested by the full CI matrix. Depend on the result, we may tweak the changes further. Can you tell me what is the specs of your arm device? Which compiler toolchain do you use?

proller:
i can test on three boards:
http://www.hardkernel.com/main/products/prdt_info.php?g_code=G145457216438&tab_idx=2
Linux odroid64 3.14.65-61 #1 SMP PREEMPT Wed May 25 03:16:39 BRT 2016 aarch64 aarch64 aarch64 GNU/Linux

gcc version 5.3.1 20160413 (Ubuntu/Linaro 5.3.1-14ubuntu2.1) 
Target: aarch64-linux-gnu

clang version 3.8.0-2ubuntu3 (tags/RELEASE_380/final)
Target: aarch64-unknown-linux-gnu



http://www.hardkernel.com/main/products/prdt_info.php?g_code=G135235611947&tab_idx=2
Linux odroid 3.8.13.30 #1 SMP PREEMPT Fri Feb 20 14:56:01 BRST 2015 armv7l armv7l armv7l GNU/Linux

gcc version 4.8.4 (Ubuntu/Linaro 4.8.4-2ubuntu1~14.04.3)
Target: arm-linux-gnueabihf

Ubuntu clang version 3.6.0-2ubuntu1~trusty1 (tags/RELEASE_360/final) (based on LLVM 3.6.0)
Target: arm-unknown-linux-gnueabihf




http://www.orangepi.org/orangepimini2/
Linux orangepipc 3.4.112-sun8i #30 SMP PREEMPT Sat Apr 30 19:49:10 CEST 2016 armv7l GNU/Linux

gcc version 4.9.2 (Debian 4.9.2-10)
Target: arm-linux-gnueabihf

Debian clang version 3.5.0-10 (tags/RELEASE_350/final) (based on LLVM 3.5.0)
Target: arm-unknown-linux-gnueabihf

weitjong:
I would like to squash your changes as their commit messages are not great before applying more changes on top. If I do the squashing for you then git commit history won't have your name. If you mind that then please do the squashing yourself in your branch with a readable commit message. We will credit you in the documentation regardless, of course.

weitjong:
Could you also test out the new commit in the ```proller-arm``` branch. For 32-bit build, you will need to pass in additional build option ```ARM_ABI_FLAGS```. For OrangePI, you could try "-mcpu=cortex-a7 -mfpu=neon-vfpv4". Keep in mind that I don't have any of these boards to verify my changes, so may be we may need to iterate a few times to get them work.

weitjong:
Thanks. I will your squashed commit in the next push.

proller:
for orangepi -DARM_ABI_FLAGS="-mcpu=cortex-a7 -mfpu=neon-vfpv4" works, 
but for odroid x2 need to find proper flags like -DARM_ABI_FLAGS="-mcpu=cortex-a7 -mfpu=vfpv3" by investigating /proc/cpuinfo and build fail with 'Illegal instruction"

Please dont make ARM_ABI_FLAGS mandatory. 
-DARM_ABI_FLAGS=" " works good on all boards.

weitjong:
I have made the new flag optional as per your request. I have made a few more commits in attempt to enable AS for 64-bit ARM. It compiles cleanly on my local repo for Android using arm64-v8a ABI. It would be great if you can also help to test out AS for both 32-bit and 64-bit odroid boards. Otherwise, I think this branch is ready to be merged.

proller:
still need disable angelscript on 32bit:
```
[ 76%] Linking CXX executable ../../../bin/Urho3DPlayer
../../../lib/libUrho3D.a(loslib.c.o): In function `os_tmpname':
loslib.c:(.text+0x17c): warning: the use of `tmpnam' is dangerous, better use `mkstemp'
../../../lib/libUrho3D.a(as_callfunc_arm.cpp.o): In function `CallSystemFunctionNative(asCContext*, asCScriptFunction*, void*, unsigned long*, void*, unsigned long long&, void*)':
as_callfunc_arm.cpp:(.text+0x4a8): undefined reference to `armFuncR0R1'
as_callfunc_arm.cpp:(.text+0x562): undefined reference to `armFuncR0'
as_callfunc_arm.cpp:(.text+0x584): undefined reference to `armFuncR0R1'
as_callfunc_arm.cpp:(.text+0x598): undefined reference to `armFuncObjLast'
as_callfunc_arm.cpp:(.text+0x5aa): undefined reference to `armFunc'
as_callfunc_arm.cpp:(.text+0x5be): undefined reference to `armFuncR0'
as_callfunc_arm.cpp:(.text+0x5d2): undefined reference to `armFuncR0'
as_callfunc_arm.cpp:(.text+0x5ea): undefined reference to `armFuncR0ObjLast'
```

weitjong:
Your problem with AS on 32-bit build should be fixed now. My apology that I am now using the new branch for my own experiment for Android build but it should not affect you. Please report back when everything is OK now for you.

proller:
On fresh branch proller-arm
cmake -DVIDEO_OPENGL=0 .. && make -j3

```
[ 16%] Building ASM object Source/ThirdParty/AngelScript/CMakeFiles/AngelScript.dir/source/as_callfunc_arm_gcc.S.o
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S: Assembler messages:
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:336: Error: thumb conditional instruction should be in IT block -- `ldrge r0,[r6]'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:338: Error: thumb conditional instruction should be in IT block -- `ldrge r1,[r6,#4]'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:340: Error: thumb conditional instruction should be in IT block -- `ldrge r2,[r6,#8]'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:342: Error: thumb conditional instruction should be in IT block -- `ldrge r3,[r6,#12]'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:413: Error: thumb conditional instruction should be in IT block -- `ldrge r0,[r6]'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:415: Error: thumb conditional instruction should be in IT block -- `ldrge r1,[r6,#4]'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:416: Error: thumb conditional instruction should be in IT block -- `movlt r1,r5'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:418: Error: thumb conditional instruction should be in IT block -- `ldrge r2,[r6,#8]'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:419: Error: thumb conditional instruction should be in IT block -- `movlt r2,r5'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:421: Error: thumb conditional instruction should be in IT block -- `ldrge r3,[r6,#12]'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:422: Error: thumb conditional instruction should be in IT block -- `movlt r3,r5'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:423: Error: thumb conditional instruction should be in IT block -- `movlt r5,#0'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:503: Error: thumb conditional instruction should be in IT block -- `ldrge r1,[r6]'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:505: Error: thumb conditional instruction should be in IT block -- `ldrge r2,[r6,#4]'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:506: Error: thumb conditional instruction should be in IT block -- `movlt r2,r5'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:508: Error: thumb conditional instruction should be in IT block -- `ldrge r3,[r6,#8]'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:509: Error: thumb conditional instruction should be in IT block -- `movlt r3,r5'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:510: Error: thumb conditional instruction should be in IT block -- `movlt r5,#0'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:514: Error: thumb conditional instruction should be in IT block -- `ldrge r7,[r6,#12]'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:515: Error: thumb conditional instruction should be in IT block -- `strge r7,[r6,#8]'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:520: Error: thumb conditional instruction should be in IT block -- `movge r5,#4'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:591: Error: thumb conditional instruction should be in IT block -- `ldrge r1,[r6]'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:593: Error: thumb conditional instruction should be in IT block -- `ldrge r2,[r6,#4]'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:595: Error: thumb conditional instruction should be in IT block -- `ldrge r3,[r6,#8]'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:597: Error: thumb conditional instruction should be in IT block -- `movge r11,#4'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:669: Error: thumb conditional instruction should be in IT block -- `ldrge r2,[r6]'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:671: Error: thumb conditional instruction should be in IT block -- `ldrge r3,[r6,#4]'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:673: Error: thumb conditional instruction should be in IT block -- `movge r11,#4'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:675: Error: thumb conditional instruction should be in IT block -- `movge r11,#8'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:676: Error: thumb conditional instruction should be in IT block -- `ldrlt r7,[r6,#8]'
/home/proller/games/Urho3D/Source/ThirdParty/AngelScript/source/as_callfunc_arm_gcc.S:677: Error: thumb conditional instruction should be in IT block -- `strlt r7,[r6,#12]'
make[2]: *** [Source/ThirdParty/AngelScript/CMakeFiles/AngelScript.dir/source/as_callfunc_arm_gcc.S.o] Error 1
make[1]: *** [Source/ThirdParty/AngelScript/CMakeFiles/AngelScript.dir/all] Error 2
```

weitjong:
I have never encountered those before so I am afraid you are on your own. If I were you then I will try to modify the ABI flags. Is that for the OrangePI or for the odroid?

proller:
for both 32 bit systems.

weitjong:
As the last resort, if none of the ABI compiler flags work for your 32-bit board with AS native calling convention then you could try to go to use generic calling convention route.

proller:
maybe make way to set -DAS_MAX_PORTABILITY=1 from cmake for 32bit arms or any other platforms?

weitjong:
Yes, that is basically what I suggested in my earlier comment. Having said that, I hope you would not give up so easily. There is a significant performance hit when using AS generic calling convention. As commented before, I actually believe AS should work on 32-bit ARM unless the odroid is really so much different than 32-bit Android and RPI. Android uses soft-float while RPI uses hard-float and yet AS supports them both just as well.

I am actually successfully figure out how to do a cross compilation using Linaro and sysroot from odroid C2. I think I should be able to cross compile using odroid X2 sysroot too. But still I have no way to tell whether the binaries I built work or not as I don't have the actual device to test them.

weitjong:
I plan to merge this branch soon when all the CIs cleared. Let me know if there are any objections.

proller:
ok. Everything works except AS on 32bit and it can be fixed later in other pulls.

weitjong:
You may want to know that I have figured out what went wrong with the 32-bit AngelScript build. It appears that the Angelscript ARM assembly code does not like to be generated using Thumb states. Forcing them to be generated in ARM states seems to work. I have committed my experiment with generic ARM build in the ```refactor-buildsystem``` branch, not in master branch. But basically you just need to supply '-marm' compiler flags. Again, I have no way to verify the generated target binaries. It would be great if you can help to verify that the same fix works for you and more importantly the generated target binaries run on your 32-bit ARM board.

proller:
refactor-buildsystem now builds ok without any flags.

But for enabling opengles still need to pass -DVIDEO_OPENGL=0 to cmake

weitjong:
That's good to hear. Thanks for your help.

About the flags - without them the compiler would not generate optimal codes for your specific ARM board. Of course generic codes would run on more boards, but in this case I believe you want to target specific board to get the best performance. 

I leave the VIDEO_OPENGL build option as it is being auto-detected. In my test sysroots I simply don't install OpenGL to avoid the problem. I think the SDL is correct here. Given both GL and GLES are available then it should prefer GL. And we don't hardcode to say ARM can only use GLES.

PR #1389, opened 2016-05-24
LZ4 Update

cosmy1:


cadaver:
Thanks, seemed to work troublefree.  Will push soon.

PR #1390, opened 2016-05-26
Prevent stroke thickness attribute from being negative.

bengtsts:


bengtsts:
Or would it be better to make the minimal thickness 1?

cadaver:
To prevent outright crashing, I already inserted code into the render phase to clamp the value to 0. I believe 0 minimum value is OK since it still renders at least one pass of the effect behind the text, possibly using a different color, and thus does something.

This is nice however to not allow illegal value to be shown in the editor. But to be precise you would also need to put similar validation to the setter functions.

cadaver:
Ok. That should be good. Then I can remove the clamping from the render side. And Text3D accesses the thickness through Text setter, so everything should work right.

PR #1391, opened 2016-05-27
Make escape key name consistent between keycode and scancode.

cosmy1:


cadaver:
Looks OK, since we don't have abbreviations used elsewhere (except KP for "keypad" which also SDL has). Other opinions? Naturally this is an API breakage but we seem to have a habit of those anyway :)

cosmy1:
What about creating a migration section between releases in the documentation? People will have a straight guide to port the code from old to new version without digging into the changelog which may be longer.
Just an idea.

cadaver:
It's a good idea, someone just has to spend the effort :)

PR #1399, opened 2016-05-30
Emscripten AngelScript support

SirNate0:
I believe it's in a working state, though I've only tested with the 12_PhysicsStressTest.as example. 

- Has been tested by running Urho3DPlayer with the 12_PhysicsStressTest script (the file copied in place of NinjaSnowWar.as, which I hard-coded into the player, as it didn't work for me with the CommandLine.txt)
- built with "-std=gnu++14 -I/usr/include/boost" for the CMAKE_CXX_FLAGS and "-I/usr/include/boost" for the CMAKE_C_FLAGS.
- /usr/include/boost contains a symlink pointing to itself so that boost/preprocessor files can be found, without interfering the different emscripten system includes
- Networking still doesn't work, so NinjaSnowWar shouldn't work
- Debug builds did not work for me: see http://urho3d.prophpbb.com/topic2151.html
- Defining FORCE_AS_PORTABLE can be used to force it to use the generic build with, for example, a gcc build (which should allow more apples-to-apples testing of the efficiency of the bindings) (I have not done such tests yet).
- Discussion of this work originally at http://urho3d.prophpbb.com/topic2044.html


cadaver:
Sweet!

cadaver:
Going in now, created a separate branch for CI testing. I added the Preprocessor library to ThirdParty for good out-of-the-box compilation experience. CharacterDemo & AngelScriptIntegration at least worked flawlessly on Emscripten.

cadaver:
Merged. Thanks for this excellent work!

PR #1408, opened 2016-06-07
fix a typo in UTF16 decoder

ArtemKulyk:
0xe00 => 0xe000

### DIFF
--- a/Source/Urho3D/Container/Str.cpp
+++ b/Source/Urho3D/Container/Str.cpp
@@ -1025,7 +1025,7 @@ unsigned String::DecodeUTF16(const wchar_t*& src)
         return '?';
     }
     
-    if (word1 < 0xd800 || word1 >= 0xe00)
+    if (word1 < 0xd800 || word1 >= 0xe000)
         return word1;
     else
     {


cadaver:
Thanks. This must be very little exercised code since the bug hadn't come up earlier.

ArtemKulyk:
Yes, looks like so. I found it after checking Urho3D code by static c++ code analyzer. I didn't find it in runtime.
No other major issues were found by cppcheck. Only this one, two false-positive errors in ParticleEffect2D.cpp and a lot of warning concerning lack of 'explicit' word before constructors with one argument and lack of initialization of class fields in constructors (false-positive for some math classes, which use SSE for faster initialization).
Later I'll create pull request with fix for 'explicit' constructors and uninitialized class members.

damu:
Initializing class members may have a relevant performance impact.
For example I recently made my own container class to replace a usage of std::vector as that does always initialize every element which caused a serious and unnecessary performance impact.
I still usually initialize class members outside of time critical code though to avoid potential problems (bugs).

ArtemKulyk:
OK, I agree with your point. I need to work with MISRA C restrictions at work, so I tried to use the same approach here. Yes, it is wrong, performance is more important here.

damu:
? It depends on the case. It may also fix or avoid future problems. I had several hard to find bugs due to not initializing stuff properly. I just wanted to throw in to watch out for time critical cases, to not cause performance issues. Would be still good in not performance critical cases, if you want.
Just out of curiosity: which tool do you use? I once used one static analyzer tool from Google long ago.

weitjong:
The problem with those static analyzers are their low signal to noise ratio. Their reports need devs that familiar with the code base to tell which are false positive and which are genuine defects. I have read a few big projects actually has the man power to run two or three of those to complement each others and then going through all those reports. For Urho3D though, we actually has setup Coverity-Scan sometime back but no one bother to go through its report. Before it can be of any use, someone needs to first classify the newly detected items to suppress those false positive. And if the ratio is too low (i.e. too many false positive) then it gets harder just to browse through the items, let alone to classify them. Which is a pity because once is a while when I glance through its report, I do see what looks like a genuine defects buried in them. This is one of them.

screenshot from 2016-06-09 11-43-44

PR #1409, opened 2016-06-08
Fix EventNameRegister typo.

cosmy1:


cadaver:
Here I suspect "Registrar" is the intended word (ie. an authority that handles event name registrations)

cosmy1:
Ops, my mistake, it sounded so Spanish :smile: but I checked and you are right.

PR #1419, opened 2016-06-11
Implement UI/ProgressBar

nerzhul:
Urho have scrollbars, sliders, but not ProgressBar for doing some loading screens, for example.
We have this progressbar working on our project and wanting it to be upstream, then enjoy

### DIFF
--- a/Source/Urho3D/UI/ProgressBar.cpp
+++ b/Source/Urho3D/UI/ProgressBar.cpp
@@ -1,0 +1,159 @@
+//
+// Copyright (c) 2008-2016 the Urho3D project.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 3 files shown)


eugeneko:
Maybe you need to fix codestyle first?
E.g. member names and convert tabs to spaces
http://urho3d.github.io/documentation/1.5/_coding_conventions.html.

nerzhul:
@eugeneko hmmm sorry i forget to see tabs should be spaces.
For members i looked at the slider code

nerzhul:
@eugeneko this is now fixed :)

cadaver:
Check http://urho3d.github.io/documentation/HEAD/_contribution_checklist.html. The brace placement is not the same as in existing code, and script bindings are missing. Otherwise looking good. However I'm a bit unsure if this is needed, since you could always implement a bar (with a style of your own choosing) with a BorderImage or two. Opinions?

nerzhul:
@cadaver i will fix the braces soon.
For the ProgressBar the idea is to have this component directly into core. Many users can have this need if we have a component to do this core side it permit everybody to don't reinvent the wheel :)
ProgressBar has the same goal as Slider, simplify the library end user coding :)
This permits to have a common interface for every body to do this work

nerzhul:
@cadaver indent is now fixed. For bidings you mean angelscript & lua ?

cadaver:
Yes.

Scellow:
I think slider is already a progress bar, since it just need a fill between begin and the handle
http://media02.hongkiat.com/jquery-ui-slider/anatomy.jpg
But having a specific control is i think a good idea, slider is for changing value, progress bar is for displaying a progress

cadaver:
Merged the version with added bindings.

Issue #1405, opened 2016-06-06
Smooth model LOD switch


eugeneko:
http://urho3d.prophpbb.com/topic2163.html

How to make it?
It could be fade through transparency for FR or dithering for DS renderer.

cadaver:
Likely you would need to subclass e.g. StaticModel and make it draw more batches (to draw both LODs) during the transition period.

EugeneKozlov:
To make smooth fade I need fade factor in the destination shader and I shouldn't use shader uniforms to reach this. So I need to pass this factor through SourceBatch into renderer.
Can I safely use e.g. [1] element of worldTransform_ for this purpose?
Do you think about implementing some kind of smooth lods or custom parameters at some point?

cadaver:
In my opinion the easiest/correct way is to use a special material (cloned from the original) for the transition time.This could have for example transparent blending set, and its DiffColor alpha modified for the transition, or a custom uniform used to control e.g. stippling / dithering.

cadaver:
You're right though that if you're after max performance and rendering hundreds of fading objects at the same time, using individually cloned materials, it will break instancing and generally cause slowdown in the render queue preparation stage. For that, you should perhaps look into expanding the object transform into a full 4x4 matrix so you can store extra data. This is a change I'd very likely not want into the core Urho3D, so you'd be diverging from it.

EugeneKozlov:
Extending matrix is easy and bad solution, I agree.. User shouldn't pay for what he don't use.
What about adding second instance array like 'user instance data'? It could be handy for many cases, on my opinion. To pass per-instance color or some other data.
It can look like
`#define NUM_PARAM 0/1/2/etc`
and new member in SourceBatch
`#if NUM_PARAM > 0
Vector4 params_[NUM_PARAM];
#endif`
and the same guards in renderer that allows anybody enable this buffer when needed.

EugeneKozlov:
One more question.
I found that dithering fade looks better when fading is performed during some small time interval instead of distance-based fade.
How can I get 'current time' in updateBatches()?
I don't want to pass it from external code.

cadaver:
In the shaders it's quite easy to add arbitrary per-instance data, just define it to e.g. an unused texcoord.

[...]

PR #1423, opened 2016-06-13
Add support of custom per-instance data in batches.

eugeneko:
Hi all.

These changes allow to set custom per-instance data in SourceBatch and implement things like smooth LOD switch.
https://github.com/urho3d/Urho3D/issues/1405

For now it it internal feature that can be used from Drawable derived-s.
So it is impossible to set these data from external code with public methods.
I just don't need such use case and have no ideas about good interface for it.

It could be settable property of `Drawable` like `Vector4 instancingData_[MAX]`, as suggested by **cadaver**, or something else.
I am not sure how much MAX should be. Now it is 4. I suppose it is enough for most cases. Is it enough? Is it okay to add extra 64 bytes for each drawable?
Where can I place this constant? Is `GraphicDefs.h` good place?

Custom per-instance data also does not work without instancing. I tried to implement it, but it requires extra uniform in all shaders and related changes, so I decided to do nothing.
I just added possibility to force instancing on for all render-ops. Could it break something?

Let's discuss..


eugeneko:
> Build execution time has reached the maximum allowed time for your plan (60 minutes).
I have no idea why it has failed.

weitjong:
Sometimes it is just a false alarm. I have retriggered the CI jobs and it looks good this time.

cadaver:
Ah, I didn't think of non-instanced case at all. For now I think it's sufficient that you allow to set minInstances to 1. Probably, if you're doing a high-end game where instancing is necessary, you aren't going to support old graphics cards (without instancing) at all, so it isn't that big of a problem.

PR #1436, opened 2016-06-21
SendEvent new renderpath command

OldSnake22:
A new "sendevent" render path command can now be added. This render path command will send an E_RENDERPATHEVENT event with the specified name. This can be used to execute some code at a specific point during the rendering.

_<command type="sendevent" name="MyCustomName"/>_

Feel free to suggest me some changes to the names that i have used. Since English is not my native language some names may not be nice :)


cadaver:
This looks solid. Thanks! Some very minor typo/punctuation issues only which I'll fix when merging.

PR #1442, opened 2016-06-29
08 Decals now draws debug geometry

define-private-public:
`false` was being passed into the call do `drawDebugGeometry`, now
it's actually using the variable `drawDebug_`.


weitjong:
I actually get worse result with the depth test on (using OpenGL API).

cadaver:
The "false" parameter is deliberately used in this example to draw without depth test and show the effect of occlusion. Decal bounding boxes are better drawn in this mode. Don't think this change should be merged.

Also, for future reference note that if you change the samples then similar changes should be made to the script versions.

PR #1443, opened 2016-06-29
MultiLineEdit

markomiz:


cadaver:
Sweet! Will check in detail a bit later. Since this supersedes the earlier multiline PR I'll close it now.

cadaver:
Ok, looked through this with more detail now, one thing that puzzles me is that MultiLineEdit subclasses from Text, yet it also creates an internal Text element. Why is this? Also, using the | character instead of a cursor sprite is a departure from the regular LineEdit and may not work well with all fonts.

danhambleton:
So, I've fixed some things here:
- Inherits from BorderImage now (like LineEdit)
- Uses a BorderImage for the cursor (like LineEdit)
- Fixed a bug with triggering OnKey multiple times per frame. Not sure I've got the most elegant solution, but it works.

Here are the files: 
- [MultiLineEdit.cpp](https://gist.github.com/danhambleton/7139fdad7007044bf38c5a2e18659d0b#file-multilineedit-cpp)
- [MultiLineEdit.h](https://gist.github.com/danhambleton/7139fdad7007044bf38c5a2e18659d0b#file-multilineedit-h)

cadaver:
Cool, thanks!

cadaver:
Hmm, tested the gist a bit, vertical scrolling and cursor up/down logic seemed somewhat odd. Don't think I'm going to merge the gist. Further improvements are welcome.

danhambleton:
> vertical scrolling 
What are you referring to here? Could explain what the desired behaviour would be?

> cursor up/down logic
Yes, this is off for sure. I think the right behaviour should be that on key up/down, the cursor goes to the character in the line at the same index as the one it came from. If that index is larger than the total characters in the line, go to the end of the line. Sound right?

cadaver:
The scrolling issue may have been related to the up/down behaving strangely. Basically, I had a short (about 48 pixels tall) MultiLineEdit, added a couple of rows to it so that the top scrolled away, and couldn't get it to show the start of the text any more.

LordGolias:
I suggest to close PR. It is stalled and I went through it and IMO it is not the way to go: MultiLineEdit must be a class subclassed from ScrollView and LineEdit, instead of implemented as a flag of LineEdit; It is way more difficult than what this PR proposes.

PR #1447, opened 2016-07-01
Graphics API agnostic headers

cadaver:
This refactors the Graphics subdirectory headers to not use different versions based on the graphics API. This should make it possible to swap Urho3D library using whatever API in a client application, since the ABI should be the same in any case. Some API-specific functions are present in the headers, but these are explicitly marked, and should be a safe no-op when called on the "wrong" API.

In addition code duplication was reduced by moving common functionality into a separate .cpp file in Graphics directory, away from the API-specific .cpp files.

Testing is welcome, since this is quite a large change.

Note that three explicitly API-specific classes remain, but these are for internal use only and not needed in applications: GraphicsImpl, ShaderProgram, and VertexDeclaration.


JoshEngebretson:
Fantastic!  This PR is a really good read and will set things up nicely for future work too :) 

+4,286 −11,331 <==== Wow!  

Great job! :+1:

cadaver:
I see a texture loading failure (black content) on GLES2. Will investigate.

cadaver:
Pushed a fix for that. However for some odd reason the branch is not updating in github web frontend.

cadaver:
I have now personally tested also Emscripten, Android and iOS and things appear to be working. Unless no objections, the next step is to just merge this and see whether then something gets uncovered :)

JoshEngebretson:
I'd say go for it, am sure we can all help figure it out in post! :)

PR #1452, opened 2016-07-04
Fixes some more serious warnings

TheComet93:
I compiled Urho3D with ```-Wall -Wextra -pedantic``` to fix some of the more serious warnings.
 * Removed a plaethora of commas at the end of enums
 * Removed a plaethora of superfluous semi-colons
 * Commented out unused variables
 * Fixed ordering of a initializer lists (-Wreorder)

During this, I also discovered some minor WTFs and even discovered a bug.

### Strangeness 1
What is this? As far as I can see, ```isASCII``` is always true no matter what. The if statement will always evaluate to being false because ```buffer[i]``` is a (signed) const char.
Source/ThirdParty/Assimp/code/STLLoader.cpp:103
```cpp
    bool isASCII( strncmp( buffer, "solid", 5 ) == 0 );
    if( isASCII ) {
        // A lot of importers are write solid even if the file is binary. So we have to check for ASCII-characters.
        if( fileSize >= 500 ) {
            isASCII = true;
            for( unsigned int i = 0; i < 500; i++ ) {
                if( buffer[ i ] > 127 ) { // warning: comparison is always false due to limited range of data type [-Wtype-limits]
                    isASCII = false;
                    break;
                }
            }
        }
    }
```

### Strangeness 2
This was an interesting warning, which I believe is actually an error.
Source/Tools/AssetImporter/AssetImporter.cpp:1191
```cpp
        // Find out the start time of animation from each channel's first keyframe for adjusting the keyframe times
        // to start from zero
        float startTime = duration;
        for (unsigned j = 0; j < anim->mNumChannels; ++j)
        {
            aiNodeAnim* channel = anim->mChannels[j];
            if (channel->mNumPositionKeys > 0)
                startTime = Min(startTime, (float)channel->mPositionKeys[0].mTime);
            if (channel->mNumRotationKeys > 0)
                startTime = Min(startTime, (float)channel->mRotationKeys[0].mTime);
            if (channel->mScalingKeys > 0) // warning: ordered comparison of pointer with integer zero [-Wextra]

                startTime = Min(startTime, (float)channel->mScalingKeys[0].mTime);
        }
```

I believe this should be this? This change is in the pull request.
```cpp
if (channel->mNumScalingKeys > 0)
```

### Strangeness 3
Why is it possible to copy ```ValueAnimation``` when the base class (```SharedPtr```) is non-copyable (i.e. ```RefCounted(const RefCounted&)``` is private).
Source/Urho3D/Scene/ValueAnimationInfo.cpp:55
```cpp
ValueAnimationInfo::ValueAnimationInfo(const ValueAnimationInfo& other) : // warning: base class ‘class Urho3D::RefCounted’ should be explicitly initialized in the copy constructor [-Wextra]
    target_(other.target_),
    animation_(other.animation_),
    wrapMode_(other.wrapMode_),
    speed_(other.speed_),
    currentTime_(0.0f),
    lastScaledTime_(0.0f)
{
}
```

### DIFF
--- a/Source/Urho3D/Graphics/Viewport.h
+++ b/Source/Urho3D/Graphics/Viewport.h
@@ -88,9 +88,9 @@ public:
 
     /// Return ray corresponding to normalized screen coordinates.
     Ray GetScreenRay(int x, int y) const;
-    // Convert a world space point to normalized screen coordinates.
+    /// Convert a world space point to normalized screen coordinates.
     IntVector2 WorldToScreenPoint(const Vector3& worldPos) const;
-    // Convert screen coordinates and depth to a world space point.
+    /// Convert screen coordinates and depth to a world space point.
     Vector3 ScreenToWorldPoint(int x, int y, float depth) const;
 
     /// Allocate the view structure. Called by Renderer.


cadaver:
Thanks for the mNumScalingKeys fix, that is at least a certain bug. Can you revert the whitespace / newline changes, especially third party libs, as otherwise it's quite painful to see what actually changed?

weitjong:
I am also not in favor of using dubious construct to suppress unused parameter warning.
Instead of
```
void LuaScript::HandlePostUpdate(StringHash eventType, VariantMap& eventData)
void eventType;
```
I would rather have
```
void LuaScript::HandlePostUpdate(StringHash /*eventType*/, VariantMap& eventData)
```

TheComet93:
Thanks for the swift feedback. Shall I revert all changes in thirdparty libraries in that case as well?

cadaver:
If you're changing a section of code that was already modified for Urho, that's at least OK to leave in (e.g. some variable init orders in Recast/Detour, that we had caused). If you revert the whitespace changes first then we'll see how much actually changed.

Issue #1457, opened 2016-07-05
Minor changes in Container


eugeneko:
I suggest some changes in Urho3D.Container that I found useful during writing some code that uses Urho3D as C++ library.

1. Add nullptr_t ctors to SharedPtr and WeakPtr (guarded by URHO3D_CXX11). This allows to write `foo(nullptr)` or `return nullptr` instead of `return SharedPtr<SomeClass>()`
2. Add aggregate ctor to PODVector (guarded by URHO3D_CXX11). It looks strange that Vector has aggregate ctor but PODVector doesn't.
3. Make [POD]Vector::Buffer() public. It allows to get rid of ugly `foo(&vec[0], vec.Size())` and _it prevents asserts if vector is empty_.

If you accept [some of] these changes I'll make PR.

cadaver:
I believe all of these should be acceptable.

cadaver:
PR merged, closing this.

PR #1458, opened 2016-07-05
Minor changes in Container

eugeneko:
#1457

### DIFF
--- a/Source/Urho3D/Container/Vector.h
+++ b/Source/Urho3D/Container/Vector.h
@@ -597,7 +597,16 @@ public:
     {
         *this = vector;
     }
-
+#if URHO3D_CXX11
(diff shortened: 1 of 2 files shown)


Source/Urho3D/Container/Vector.h
```diff
@@ -597,7 +597,16 @@ template <class T> class PODVector : public VectorBase
     {
         *this = vector;
     }
-
+#if URHO3D_CXX11
+    /// Aggregate initialization constructor.
+    PODVector(const std::initializer_list<T>& list) : PODVector()
+    {
+        for (auto it = list.begin(); it != list.end(); it++)
```
nerzhul: const auto is better than just auto here

PR #1462, opened 2016-07-08
get body properties when body exists

fredakilla:
to be consistent with set functions.


cadaver:
Is there actually a logic problem in the setters that they don't update the bodyDef if the body already exists? In case where we would need to later recreate the body from the bodyDef, and the values would then be out of date?

fredakilla:
On my side I didn't find the reason why bodyDef is not updated if body exists.
I made a test with always updating it, and didn't notice any difference.

PR #1468, opened 2016-07-13
Optimize direction billboard vertex buffer

ucupumar:
After developing RibbonTrail feature, I realized it's no longer need to passing camera position to direction billboard vertex buffer because Urho3D already has default uniform to do the job.


cadaver:
Thanks!

ucupumar:
@cadaver you're welcome! :smile:

PR #1474, opened 2016-07-16
CMAKE_SOURCE_DIR -> Urho3D_SOURCE_DIR

rokups:
This change makes Urho3D no longer assume it "owns the world". While `CMAKE_SOURCE_DIR` always points to directory of root project, `Urho3D_SOURCE_DIR` always points to directory of Urho3D project. Under normal circumstances these both values are equal so it has no impact on build at all. However it paves a way for using Urho3D with `add_subdirectory()` as a child of other project in the future.

### DIFF
--- a/CMake/Modules/Urho3D-CMake-common.cmake
+++ b/CMake/Modules/Urho3D-CMake-common.cmake
@@ -810,7 +810,7 @@ macro (create_symlink SOURCE DESTINATION)
     if (IS_ABSOLUTE ${SOURCE})
         set (ABS_SOURCE ${SOURCE})
     else ()
-        set (ABS_SOURCE ${CMAKE_SOURCE_DIR}/${SOURCE})
+        set (ABS_SOURCE ${Urho3D_SOURCE_DIR}/${SOURCE})
     endif ()
     if (IS_ABSOLUTE ${DESTINATION})
         set (ABS_DESTINATION ${DESTINATION})
@@ -1270,7 +1270,7 @@ macro (setup_main_executable)
     # Define resources
     if (NOT RESOURCE_DIRS)
         # If the macro caller has not defined the resource dirs then set them based on Urho3D project convention
-        foreach (DIR ${CMAKE_SOURCE_DIR}/bin/CoreData ${CMAKE_SOURCE_DIR}/bin/Data)
+        foreach (DIR ${Urho3D_SOURCE_DIR}/bin/CoreData ${Urho3D_SOURCE_DIR}/bin/Data)
             # Do not assume downstream project always follows Urho3D project convention, so double check if this directory exists before using it
             if (IS_DIRECTORY ${DIR})
                 list (APPEND RESOURCE_DIRS ${DIR})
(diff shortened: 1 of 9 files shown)


weitjong:
Thanks for your PR. Although I agree this is a step in the right direction (in fact someone has already suggested and done too), just applying these changes won't get you anywhere near that goal. i.e. there are other changes required so that all our modules would release its self-centered position. In view of that, I will probably not merge this. At least not until the other parts are ready.

rokups:
Is submitting general improvements against contributing rules? Somehow i missed that, sorry.

weitjong:
Of course there is no such rule. But there is also no rule that says we have to merge any PR being submitted too. Please do not get me wrong. I am just saying I will not merge this as it is (yet). It is anyway one of the plan item in the buildsytem-refactoring branch that I want to do when the time has come. However, I would certainly not just do a simple global search-and-replace operation without checking each occurrence personally. Have you just done that? At the very least I spot one occurrence that it should not be replaced in any use cases.

rokups:
> However, I would certainly not just do a simple global search-and-replace operation without checking each occurrence personally. Have you just done that?

I checked every one of them and at the time it seemed right. Turns out at least one in `Rakefile` should have not been replaced and this i already corrected. Is there something else too? I do not know build system as good so review by someone better knowledgeable would be great.

By the way any idea why AppVeyor build fails? Checking build logs i do not see any errors which is very strange.

weitjong:
As I said, I will keep this PR pending and will return to it when I have fixed the other issues that would make embedding Urho3D in other project possible. These include fixing the library dependency hard-coding issue,  making the dependencies propagate transitively for STATIC lib type, reducing the scope of the compiler defines to where they are needed, among other things.

As for the AppVeyor - it looks like the test run has hang up for the scaffolding test case. When it was one out of 4 hang up then it could be just a flux. But when all 4 did that then it means your commit broke some thing. Scaffolding test case is testing the use case where the newly built Urho3D library is being used in a new project as an external library.

rokups:
Indeed you are right. Due to be too densely concentrated on subdir use method i overlooked the fact of module sharing and that in turn totally broke using lib as SDK. I was looking more carefully through all the changes and short of introducing custom `URHO3D_IS_SDK` i cant see easy solution. Ill close this for now as patch just breaks things.

rokups:
@weitjong since build system refactoring was merged maybe you could comment on this? Is there anything still in the way of these changes? Should i reopen this PR? I tried building/installing android and x86_64 builds and it worked. Not sure how to test whatever was broken in the past without letting CI do it though.

weitjong:
The truth is my hand was kind of forced to merge that branch and that I have actually not even half completed what I plan to do in that branch originally! I kept getting side track by other things that interested me at different time. Also, as there are no clear milestones (that's intentional by the way because I am doing this just for fun and not as my day job), I do not do things in respect with their rightful priorities.

And as for your PR, my view remains the same as before.

rokups:
Would you help me to do it proper then? Also since there may be other work in progress on build system i figured i should coordinate with you in order to not interfere too much with your work. Be great if you could point me to the right direction.

weitjong:
There are a few things. The most important I guess is to do away with the define_dependency_lib() macro which has been marked as deprecated. We should let newer CMake to keep track of the dependencies of all the targets, from all the internal 3rd-party libs to ultimately the Urho3D lib itself. So, when downstream project depends on Urho, CMake knows when to apply the dependency list transitively, be it the Urho3D lib is configured as an internal target in an encompassing outer project or be it just as an external lib as it is known today.

rokups:
So every project should have `target_link_libraries()` after `setup_library()` right? I am asking because `target_link_libraries()` is not that new, it was there in 2.8, not sure why it was not used like that to begin with.

weitjong:
target_link_libraries() is used in the setup_target() macro which is in turn called by setup_library(). The signature of that command has changed from what we used to have in 2.8. I am not quite sure of the reason now, but most probably because I dislike the fact using its link interface forcing us to have IMPORTED library target of Urho3D, i.e Urho3D libary need to be installed somewhere first in the filesystem. Whereas with our macro approach, we could easily setup and use Urho3D as external library directly from its build tree, no installation is required.

PR #1477, opened 2016-07-16
PBR Major improvements

dragonCASTjosh:
A small commit that improves the quality of the current PBR implementation. Changes include Gamma correction to ensure the color and brightness of materials. I also made some changes to the demo scene that includes, pre-filtered skybox and zone texture along with a new floor texture that prefer.


cadaver:
Can you fix the typos? There's at least Propeties.jpg (should be properties.jpg I believe) and "lamda" in the shader code which looks it should be "lambda"

dragonCASTjosh:
@cadaver fixed :)

cadaver:
Thanks!

dragonCASTjosh:
@cadaver on a slightly side note PBR should be stable enough and possibly fast enough for promotion to the main shading model if you want it to be :) if it is not fast enough now then when i add the optional split sum model it will be fast enough.

dragonCASTjosh:
DO NOT MERGE, there is an issue with the example under directx i will investigate

cadaver:
At least for now I like to keep the default lighting model as is, considering also mobile use.

dragonCASTjosh:
It should be fast enough for mobile, i can put epics mobile ibl solution as an optional system

dragonCASTjosh:
Issue found a messed up the rename on lambda just have to rename the output. I'll push in the morning

hdunderscore:
I think for changes to the BRDFs, it would be better to raise an issue pointing to errors and a reference source or derivation for alternatives, approximations or optimisations. Alternatively, explain the changes in the PR. That way it's easier to follow the changes and reasoning, otherwise we could receive many PRs with corrections because of different approximations that were used. At a glance it's not always easy to spot issues.

There are a few small issues I see that are chained together in this PR, but maybe some simple explanations or references would help make it clear why you made the changes. Eg, the change to GGXDistribution and then its use appears strange at a glance. The random change to the BurleyDiffuse call too.

Correct me if I'm wrong, but deleting resources in a commit will leave the repo fat anyway, which isn't good. There is more benefit to simply adding a new example than replacing old ones when they include large resources.

dragonCASTjosh:
@hdunderscore  the changes to the GGX was to ensure its still physically correct after gamma correction. i used frostbites PBR paper for details on the change, as for the changes to the diffuse it was to ensure we have a roughness square curve to the materials, from reading i found that the calculation are not what is counted as the power curve for the roughness, therefore i made it roughness square as it has the best curve for materials.

dragonCASTjosh:
@cadaver it should be safe to merge

hdunderscore:
I think these kinds of changes are like playing whack-a-mole and will eventually lead to a situation where the PBR is hard to maintain and will probably need a re-write. Insisting on PRs to be merged when there are possible errors doesn't inspire confidence in me.

The simplest solution is to provide comparisons to other implementations and an understanding that tuning the PBR to one scene while there are major missing components (eg, area lights and inverse fall-off) is not ideal.

There are a few issues that need to be nailed down:
- Gamma correction -- what is correct and why. I believe the changes you made lead in the wrong direction. The lighting should all be done in linear space, and only at last should it be gamma-corrected. This can be done either via post-process or via graphics->SetSRGB(true). Eg, what if you want to make a post-process effect that relies on linear space lighting (which is the sane thing to do)? If you pass a gamma-corrected buffer and keep converting every pass, it's fragile.
- BRDFs -- I prefer we keep them clean and isolated, or implement combined optimisations as their own function. Pulling out the pi divide breaks our other functions and overall makes the code messier. I also believe you added an extra divide by pi.

cadaver:
Yeah I will definitely want to wait before @hdunderscore is happy before merging this.

dragonCASTjosh:
@hdunderscore i decided to do gamma correction within the shader using frostbites approach of both divide by PI and convert the final values to SRGB. This approach was taken to make the look of PBR in the editor match the look of PBR in the final game.

As for additional feature they are being worked on, Inverse fall-off need to have a small gl write up then is ready to make a PR. area lights are a much bigger beast to tackle.

I do test against Unreal 4 for quality reason. although there are some disparity in PBR it will require large changes to the rendering pipeline to fix the issues therefore im holding that for a later commit, alot of the issue is down to how IBL is applied and the IBL itself

dragonCASTjosh:
Ill extend this PR to the larger changes to everything PBR if that is ok with @cadaver and @hdunderscore

cadaver:
Yes I believe it's better that way instead of multiple PR's. I currently have planned for the next release to be after July and this either makes it in or not, but I wouldn't stress over that, it's better that whatever is merged, is well thought-out and doesn't have "magic" adjustments.

dragonCASTjosh:
I have rebuilt allot of the PBR from the ground up using the frostbite PBR paper.

changes include 
- Energy conserving Burley diffuse
- Smith geometry term changed to visibility turn 
- Mobile friendly IBL solution
- Broke the main part of PBR out into a separate shader ready for more PBR changes such as area lights (likely later commit)
- Inverse square light falloff under PBR
- Made the demo scene better demonstrate the feature

screenshot_mon_jul_18_21_41_29_2016

OldSnake22:
I had the chance to test it and it look much better and reliable than the old official PBR implementation. Good work !

dragonCASTjosh:
@OldSnake22  thanks

also i know about the few odd files that got reverted in the commit like the billboard stuff, ill fix that up now

dragonCASTjosh:
@cadaver @hdunderscore Open to feedback now

cadaver:
Don't have competence to comment at the math, but structurally the shader code looks good.

Can you put all PBR related resources under the Autoload/LargeData folder? The gun model wasn't there, only its source FBX was (which I don't think we necessarily require) There were also some material files written outside Autoload.

Is the gun model OK licensing-wise, does it need license information being included?

dragonCASTjosh:
Apparently i copied the wrong working directory the gun model was only supposed to be used for testing, i swapped it out just prior to committing but it looks like it was the wrong directory. I will change that right away.

dragonCASTjosh:
@cadaver  double checked everything and all licencing and file structures should be correct now with 0 misplaced files.

hdunderscore:
Could you keep the old code (eg, BRDFs and importance sampling), I don't see a reason to delete everything else. You can make the new code the defaults.

The IBL doesn't give a diffuse factor, which is noticeable especially on the vehicle or anything out of direct light.

Minor issues:
- the C++ sample doesn't enable gamma correction while angelscript does. 
- I believe the checker normal map is inverted.

dragonCASTjosh:
@hdunderscore some allot of the old BRDF code would not create correct results so i plan to rewrite them at a later date to work with the current structure, that said the distribution methods should work fine although i don't there there is any advantage to using blinn in PBR. I can add the importance sampling stuff back although its results are not to much different from the mobile friendly solution (assuming it has diffuse ibl). 

On epics blog post about there mobile IBL they did not do anything with diffuse ibl and left it out of there equations, i mostly followed there steps in this, although i will experiment with adding a diffuse ibl in.

The AS example should not have gamma correction enabled ill fix that.
Apparently the normal map is inverted good spot

dragonCASTjosh:
@hdunderscore does this IBl look correct to you

screenshot_tue_jul_19_13_53_05_2016

hdunderscore:
That IBL looks a lot better.

dragonCASTjosh:
@cadaver  part of the improvements should allow the PBR to work as high fps on mobile although i am unable to test. If this is true then it may solve the issue with not using PBR as a main renderer due to mobile :)

cadaver:
Will test, I'd honestly be surprised to see it even run, as mobile is limited in num. of available texture units , and anything that has to do with texcoord calculation in pixel shader has potential to run poorly. 

I'm aware of the potential impression of "poor graphics / shaders" if PBR is not used as default, however Urho is still mostly a coder's engine, to whom I'd give a bit more credit in not making snap judgements based on a superficial assessment. There is also merit in running as fast as possible by default (since we can also have usecases like cartoon or nonrealistic rendering, which necessarily do not have use for PBR) and then paying for more advanced lighting if you want to use it.

1vanK:
I'm not very well versed with PBR rendering, and have a question. It is limeted HDR rendring only? When I disable HDR, it looks darker.

dragonCASTjosh:
If you disable HDR you may want to enable the gamma correction it should solve the issue

1vanK:
Oh, its my mistake. When I disable hrd, I also remove postprocess effects (including gamma correction), therefore I get different brightness.

p.s. I tested OpenGL version and new realisation is really fast. Great works!

dragonCASTjosh:
@cadaver i will start documentation now but dont hold out on it as it may take a while to get it worded right and spelled correctly i end up with minor heat stroke daily at the moment making it hard to work.

@1vanK i designed PBR to be fast so it can be used on all Urho platforms

cadaver:
@dragonCASTjosh No hurry, it's something we can well include later, and/or keep amending.

dragonCASTjosh:
@cadaver is everything ready to merge then?

cadaver:
I'll wait @hdunderscore to answer if there's anything more to change. I'll also test for possible shader compile errors on GLES (Android) but that shouldn't affect you, as I'll fix them myself if there is something that needs (and can) be fixed.

dragonCASTjosh:
@cadaver i can work with you on fixing them if need be

cadaver:
One thing noticed, gun textures still there? Will not be a problem for me to remove, I'll squash the history anyway when merging.

dragonCASTjosh:
where did you find that i thought i removed it all

cadaver:
Check the "86 changed files" dropdown in the PR overview, Cerberus_A/N/P.jpg in bin/Autoload/LargeData/Textures

1vanK:
Why it's needed to hold RMB to move camera in example? Whe can make free flight without RMB. Or we can add to scene sphere with dynamic material and two sliders for change mettalic and roughness of this material.

dragonCASTjosh:
@cadaver ill look into it

@1vanK Its a feature of other Urho demos so i assumed it best to keep the functionality, if someone wants to add sliders for a extra sphere go ahead but im not well versed enough in either the UI or scripting to attempt this yet

1vanK:
Ok, I will modify example after merging

dragonCASTjosh:
@1vanK feel free to make a PR on my branch now if you like

cadaver:
Tested on Android, there were compile errors (mostly doing with certain functions like clamp or pow not accepting const int arguments) but I'm not going to spend time on them now, since we assume 8 texture units on mobiles, and that leaves out the zone cube map texture. So for now we can state that PBR requires desktop GPU.

dragonCASTjosh:
@cadaver can you not replace the sEnvMap under PBR for mobile, and i assumed i removed all that because gl shouldnt support const values

cadaver:
Envmap texture unit is commonly used by materials, so anything defined in the zone is liable to get overwritten by materials. We could use a rule where on mobiles the zone texture is assigned there, but it's not foolproof.

dragonCASTjosh:
i think performance wise it should be good enough its just getting the structure to work

hdunderscore:
Overall it looks good, I'm submitting a PR to your branch with minor tweaks. 

One issue I spotted is that the DX11 (maybe DX9 too, although I didn't test it yet) version doesn't have reflections on the glass of the vehicle.

yushli:
I would strongly suggest to make this feature running on mobiles. dragonCASTjosh  has already made it fast enough for mobiles I think. We should not let just zone to limit it only to desktop GPU.

amerkoleci:
Agree, need mobile GLES2 implementation before merging.

cadaver:
May not be that simple; most of the shader errors are indeed missing decimal points on constants, however texture2DLod in IBL lighting may not be widely supported. I tested on a LG G2 and it didn't have it. It's easy to do a fallback cube sampling without LOD selection, but naturally it will be wrong.

cadaver:
Got the scene running on Android (LG G2) without IBL LOD sampling. The new skybox displayed just black for some reason, so used the old one as a fallback. Got about 19fps without shadows, which is not super, but not a slideshow either.

dragonCASTjosh:
if there is any optimization needed for mobile it will likely be in the BRDF things like using Lambert diffuse

OldSnake22:
@hdunderscore Just looked at the changes in your branch. Having to put srgb=true in all textures xml files is not very practical in my opinion, is there any other simpler way to avoid that ?

dragonCASTjosh:
@OldSnake22 i originally had it built into PBR itself

OldSnake22:
Yeah what was wrong with this approach ?

dragonCASTjosh:
@hdunderscore feels it should be done in post processing as far as i can tell

hdunderscore:
@dragonCASTjosh , the issue @OldSnake22 raises is texture loading. The changes you had were in regards to outputting lighting as gamma-corrected.

It's probably safe to assume albedo and emissive textures are srgb encoded and decode them in the shader without much performance hit, although if that flag is set and the shader decodes additionally, things will get dicey fast. Similarly, if you happened to not have srgb encoded textures, then you would have to edit the shader.

If you follow a naming scheme, eg, Albedo.jpg, then it's simple to copy and paste the same Albedo.xml.

1vanK:
Actually, into PBR shader need send another one texture for ambient occlusion (finalColor *= ambientOcclusion) because for PBR we can not mix AO with diffuse texture

dragonCASTjosh:
@hdunderscore the pow(x, 1.0/2,2) was a short hand way of converting to SRGB color space. in the frostbite paper they recommend doing all calculation in linear color space then convert to SRGB before output. I admit they dont recommend the short hand as colors become slightly off but urho is a performance based engine so i decided it fits the needs best.

@1vanK you can use the Blue color channel of the PBR texture to load AO then apply it in the shader

1vanK:
> you can use the Blue color channel of the PBR texture to load AO then apply it in the shader

o, nice idea!

dragonCASTjosh:
@hdunderscore for reference page 87  [http://www.frostbite.com/wp-content/uploads/2014/11/course_notes_moving_frostbite_to_pbr.pdf](http://www.frostbite.com/wp-content/uploads/2014/11/course_notes_moving_frostbite_to_pbr.pdf)

cadaver:
Regarding performance and lighting spaces, on desktops a fullscreen postprocess pass should be nothing. On mobile it can have a quite significant impact. On mobiles sRGB pixel format also isn't supported. So it is rather hard to do something which is both correct, performant and workable on all platforms.

dragonCASTjosh:
then surly doing it as part of the PBR provides the best results? as we can then put a mobile work around to the color correction.

hdunderscore:
@dragonCASTjosh, right-- to do calculations in linear space you need to make sure you've loaded your textures into the right space. @OldSnake22 asked if there was an easier way to load textures with the srgb decoding, without making so many xml's. At the moment we are using the srgb=true flag in the texture xml which I believe does the decoding only the first time the texture is loaded, instead of every hit in the shader.

Using the xml's is good, but putting the decision into the artist hands is a source of difficulty and confusion. I personally get confused every time I have to decide whether I should enable that flag or not-- and it's also confusion because you shouldn't enable that flag for the normal map and pbr/properties map.

dragonCASTjosh:
from what i have experienced artists only develop textures in linear color space and that should be correct for loading it the shaders in there current space, this will mean there is not a need for the time consuming and sometimes confusing srgb flags

1vanK:
I sended PR with improved example to dragonCASTjosh repo and have some notes:

1) I have no ideas, why dynamic material does not works in LUA example. Some problems with dynamicMaterial:SetShaderParameter("RoughnessPS", Variant(newValue))

2) Sun does not rotating in LUA example. It use AS script.

3) Should we rename RoughnessPS and MetallicPS to Roughness and Metallic?
```
	<parameter name="MatDiffColor" value="1 1 1 1" />
	<parameter name="MatEmissiveColor" value="0 0 0" />
	<parameter name="MatEnvMapColor" value="1 1 1" />
	<parameter name="MatSpecColor" value="1 1 1 1" />
	<parameter name="RoughnessPS" value="0" />
	<parameter name="MetallicPS" value="1" />
```

cadaver:
If roughness and metallic are never used in vertex shaders, it's indeed better to drop the PS from the uniform name.

dragonCASTjosh:
i put it there origonally to make things clear when developing but the values are only used in the PS shader, i can work on dropping PS off the name

dragonCASTjosh:
Changes to the falloff re not correct, I'll change when I get on the computer in the morning

dragonCASTjosh:
Correction the math was right I was testing with VSM :) no changes needed

dragonCASTjosh:
some tweeks i want to look into before merging, i think i messed up some diffuse math

dragonCASTjosh:
@cadaver everything should be ready for a merge as @hdunderscore has had his changes merged to the branch

cadaver:
Thanks for the heads-up. I will take that latest revision and add the GLES fixes.

dragonCASTjosh:
for GLES it may be worth having it use lambert diffuse and blinn distribution for performance reasons

OldSnake22:
@cadaver Good that you are merging this now. This will let us test everything before the next official release of Urho

dragonCASTjosh:
Apparently the Siggraph talks on PBR are today so i may do a later PR on some of the changes they suggest. Talks include Frostbite, Epic and Unity so it could be some new and important changes

dragonCASTjosh:
@cadaver can you hold off on the merge i need to make some small changes as the merge with hdunderscore broke a few things.

cadaver:
No problem.

dragonCASTjosh:
@cadaver and @hdunderscore any idea what is causing this issue

screenshot_sun_jul_24_14_15_48_2016

dragonCASTjosh:
My guess is because it would only let me import 8-bit normals

OldSnake22:
In my opinion it is just a normal map compression artifact. Try to use xGxR normal map encoding (this has improved normal maps quality for me). 

Urho3D documentation say: 

Store as xGxR, ie. Y-component in the green channel, and X-component in the alpha. In this case use the DiffNormalPacked techniques: Z will be reconstructed in the pixel shader. This encoding lends itself well to DXT5 compression. To convert normal maps to this format, you can use AMD's The Compressonator utility, see http://developer.amd.com/Resources/archive/ArchivedTools/gpu/compressonator/Pages/default.aspx

cadaver:
This looks like an inaccurate specular map to me too (= nothing I'd worry about now.) If the specular map is jpg, there might be a block of same color values, which is OK when viewed as a photo, but causes artifacts when high power specular calculations are based on it.

dragonCASTjosh:
@cadaver not the main focus at the moment, the changes that need to be made is removing the multiplication of the diffuse color from output the pbr shader. im away at the moment but if you was planning to merge today you can still do it with that change, else ill commit it tomorrow morning

cadaver:
I can wait for tomorrow.

dragonCASTjosh:
@cadaver looks like hdunderscore beat me to it so you can merge :)

cadaver:
In process of merging. Removed the "PS" from Roughness & Metallic. Will do the GLES fixes tomorrow. 

I noticed two ways of calculating IBL mip selection, which results in different maximum sharpness for the cubemap. (GLSL is sharper)

GLSL does float mipSelect = exp2(GlossScale * roughness * roughness + GlossBias) - exp2(GlossBias);
HLSL does const float mipSelect = exp2(GlossScale * roughness + GlossBias) - exp2(GlossBias);

Which one is preferred?

dragonCASTjosh:
i just reverted that change although i didnt notice they where diffrent, give me a second to test

cadaver:
Now HLSL & GLSL produce same result.

dragonCASTjosh:
So here are some comparason shots. I will let you pick.

HLSL version:
hlsl

GLSL version:
glsl

Origonal:
prior

cadaver:
The new code (in your latest commit) appears simpler and seems to go smoothly from maximum sharpness to maximum blurriness so I'll leave that in.

dragonCASTjosh:
yea that was the origonal implementation i reverted it just prior to your comment on it :)

dragonCASTjosh:
@cadaver could you also turn down the brightness on the directional light, its casuing some over brightness on colored low rough materials. i recomend setting it to 1 or 2 on brightness

cadaver:
Yes, can do that.

cadaver:
Merged.

1vanK:
Hm, in latest version glare flicker with high frequency (like a Christmas tree :) ), and they are square о0. Before, everything was normal. (I tested OpenGL version)

dragonCASTjosh:
It's a side affect of bloom just dissabe the bloom shader. It was a commit that was missed

weitjong:
> It's a side affect of bloom just dissabe the bloom shader. It was a commit that was missed

Thanks. Commented out bloom post-processing effect does help. However, I think that is not the root cause of the "flickering". I can still see shadow cast on the bike appears to be stuttering as the light source moves.

1vanK:
I change ROUGHNESS_FLOOR to 0.05 and it resolved problem absolutly

dragonCASTjosh:
on my local build i had to disable bloom and set HDR to false, i then change the zone ambient color alpha to 1 in the editor. this solved all issues that i could see

Issue #1366, opened 2016-05-18
add anchorPoint to UIElement (feature request)


yushli:
Currently Sprite has hotspot property to set the center of the sprite. But UIElement doesn't have this property. further hotspot is using actual size instead of percentage. Suggestion: add anchorPoint to UIElement, using percentage to set the center of this element. E.g., anchorPoint(0.5,0.5) will set the anchor point at the center of the element. That will make positioning UI more conveniently.

cadaver:
If we add this, I believe we should remove the UI element alignments (left/right/center, top/bottom/center) because those calculate a fixed anchor point internally (eg. left if aligned left, center if aligned center). This could break existing UI's. Layouting code also assumes UI elements being anchored from the top-left.

If you're interested I recommend working on it on a PR, to see if you can get it compatible with the existing UI, and perhaps most importantly to not make it more confusing.

yushli:
I'd like to do but that sounds a lot of code to look at and a lot of corner cases to watch out. I may settle down just adding that to the top level text ui elements.

PR #1486, opened 2016-07-24
Add anchor and pivot to UIElement.

ninjastone:
I implemented new feature mentioned in #1366. New attribute mode AM_FILEREADONLY is used for keep compatibility with older layout files.

Please take a look at my changes. If it's not bad, I'll add script bindings as well.

### DIFF
--- a/Source/Urho3D/AngelScript/SceneAPI.cpp
+++ b/Source/Urho3D/AngelScript/SceneAPI.cpp
@@ -50,6 +50,8 @@ static void RegisterSerializable(asIScriptEngine* engine)
     engine->RegisterEnumValue("AutoRemoveMode", "REMOVE_COMPONENT", REMOVE_COMPONENT);
     engine->RegisterEnumValue("AutoRemoveMode", "REMOVE_NODE", REMOVE_NODE);
 
+    engine->RegisterGlobalProperty("const uint AM_FILEREADONLY", (void*)&AM_FILEREADONLY);
+    
     RegisterSerializable<Serializable>(engine, "Serializable");
 }
 
--- a/Source/Urho3D/Core/Attribute.h
+++ b/Source/Urho3D/Core/Attribute.h
@@ -46,6 +46,8 @@ static const unsigned AM_NODEID = 0x10;
 static const unsigned AM_COMPONENTID = 0x20;
 /// Attribute is a node ID vector where first element is the amount of nodes.
 static const unsigned AM_NODEIDVECTOR = 0x40;
+/// Attribute is readonly.
+static const unsigned AM_FILEREADONLY= 0x81;
 
 class Serializable;
 
--- a/Source/Urho3D/Core/Attribute.h
+++ b/Source/Urho3D/Core/Attribute.h
@@ -46,8 +46,8 @@ static const unsigned AM_NODEID = 0x10;
 static const unsigned AM_COMPONENTID = 0x20;
 /// Attribute is a node ID vector where first element is the amount of nodes.
 static const unsigned AM_NODEIDVECTOR = 0x40;
-/// Attribute is readonly.
-static const unsigned AM_FILEREADONLY= 0x81;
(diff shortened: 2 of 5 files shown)


cadaver:
Nice. And dangerous :) Will check in detail later. For my sanity I'd probably want to postpone this after the coming release.

This looks a bit suspicious (GetPosition get accessor and editor only), is this as intended? URHO3D_ACCESSOR_ATTRIBUTE("Min Offset", GetPosition, SetMinOffset, IntVector2, IntVector2::ZERO, AM_EDIT);

Read-only attributes are incompatible with binary serialization, since the structure being read and written is different. If they're only used for UI elements it's fine, but a note should be added of this.

cadaver:
Tested already since the PBR is still ongoing. It's excellent how you managed to keep this compatible to the existing UI features. However I couldn't quite understand how the max anchor parameter works. I think this needs better documentation in the class setters/getters, ie. explain briefly what they do.

ninjastone:
"Min Offset" line is intended, I wanted to add way to edit left/top positions of UI elements without touching their right/bottom positions.

As for read-only attribute mode, my bad, I didn't think about binary serialization. Is adding some notes to constant enough? Maybe I should change it's name also.

Thanks for testing. The max anchor is starting positions of right/bottom sides relative to their parents. The max offset is added afterwards. I'll add explanations how rectangles of UI elements are calculated.

cadaver:
Yes, it's enough if you add a note that this mode is not to be used with binary-serialized objects. Practically, anything that can live in the Scene.

yushli:
When will this be merged into main branch? I know it may be after the coming new release, but good to know if the merge can happen soon.

cadaver:
After release. I still have concerns that the new attributes may be dependant on setting order, on the other hand if you don't use them, then nothing breaks at least in old UI layouts.

cadaver:
@ninjastone One thing that could be good to add besides the script bindings, would be a brief paragraph to the UI documentation page (in Reference.dox).

cadaver:
Tested this a bit more, and found out that the new attributes interfere with saving and reloading some of the layouts that used to work. For example, try loading Data/UI/EditorHierarchyWindow.xml into the editor, set the size to for example 400x600 and observe that the content inside it looks correct. Now save that with another filename, then load the just saved layout. The sizes of elements within the layout aren't restored correctly. @ninjastone Can you look into that?

ninjastone:
Conditions I added to serialize function was wrong. Give me some days to fix it. I don't have much time to work on hobby projects this week.

cadaver:
Thanks for looking into it, no hurry!

cadaver:
The PR seems to now be in conflict, can you look into that still?

ninjastone:
sure

cadaver:
Thanks for the work. There's one thing that confused me, that having maxOffset zero doesn't let the anchor positioning work. I'll see if that can be cleared up. Maybe I'll set an explicit boolean enable for the anchoring, that way there is also no question that existing layouts would break, since it'd be disabled by default.

cadaver:
Also, I think the size not getting saved in layouts anymore is not desirable (for elements that are not anchor-positioned), so I'd rather remove the "readonly" mechanism and think for some other solution, if possible.

cadaver:
I believe I succeeded. The merge with modifications will be pushed soon to master.

cadaver:
Merged.

ninjastone:
thank you :)

1vanK:
How to use it right?

1vanK:
Ah, I understood. I had to use offsets for spaces.

1vanK:
Why not use Rect for anchor and offsets? At the moment top and left bounds store in one variable, and right and bottom in another. I think this is illogical.

cadaver:
Yes, you can use a mixture of anchor and offset. For an absolute sized but anchored element min/max anchor could be the same and offset would define the actual pixel size.

PR #1494, opened 2016-07-28
Add int64_t and uint64_t support for Serializer/Deserializer

nerzhul:
Because int are not always sufficient, support 64b values to Serialisation interface


cadaver:
These need to be long long and unsigned long long respectively to best ensure they're actually 64bit across platforms and build types.

nerzhul:
you are right, i'm fixing this issue

nerzhul:
@cadaver fixed

cadaver:
Thanks!

PR #1495, opened 2016-07-28
Fix minimum roughness

1vanK:


dragonCASTjosh:
I have just made a PR that should solve all issues with PBR #1496 

Increasing the minimum roughness makes some real world materials impossible within the engine, the cause is related to something other then minimum roughness issue if it is still a problem after trying the PR i submitted then we should probably look for the root cause :)

hdunderscore:
I think 1vanK's solution to raise the roughness makes more sense than simply turning off the bloom, although you can still choose to work with a more reasonable roughness floor when authoring materials.

On the flip side you could choose to implement a bloom shader that handles the spikes in roughness a bit better too (eg, UE4 has a 'physically based bloom'). There are other ways to address the issue too, google: "pbr specular aliasing".

dragonCASTjosh:
i dont think the issue lies within the roughness value. as mentioned making a physically based solution for the rest of the rendering is what is needed to correctly solve the issues.

the main 3 systems that need physically based solution:
- Camera
- Lights
- Bloom

weitjong:
I am not an expert in graphics department, but IMHO the bloom post effect was just simply amplifying the "problem" (if there is indeed one). To me, the specular reflection on the surface appears to be flickering is because there are "discontinuity" from one frame to another, i.e.at some places the patches that have reflection and the patches that don't appear to be varied frame by frame by too much. The problem can also be observed without bloom post effect and without light source' rotator logic in a "still" scene by looking at patches of tiles with refection and using the mouse to gently change the camera angle back and forth. In my observation, no matter how small the nudge is, I can still see flickering albeit at a much smaller scale. Raising the minimum roughness as proposed by 1vank seems to help to reduce this "problem" actually on my test (using OpenGL).

dragonCASTjosh:
After testing with bloom and looking around online, it seams post processing should occur after gamma correction for best results. After trying this it seams the results are more then acceptable, although the bloom technique is not physically correct

PR #1496, opened 2016-07-28
Fixes to PBR demo and HDR option in editor so PBR displays correct

dragonCASTjosh:
Changes include the following:

Removed Bloom from PBR demo,
Set zone texture ambleint to 1,
Added HDR option in editor settings (needed for PBR in editor)

### DIFF
--- a/bin/Data/LuaScripts/42_PBRMaterials.lua
+++ b/bin/Data/LuaScripts/42_PBRMaterials.lua
@@ -128,8 +128,8 @@ function SetupViewport()
 
     -- Add post-processing effects appropriate with the example scene
     local effectRenderPath = viewport:GetRenderPath():Clone()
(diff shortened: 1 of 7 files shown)


cadaver:
The AngelScript version of the PBR demo has the ambient slider, which defaults to 5 and overwrites the scene's value. Should it be removed, since there's no ambient slider in C++ & Lua versions?

dragonCASTjosh:
I think adding it into c++ and lua may be better as it shows such a feature exists

dragonCASTjosh:
Although may be best to default to 1

cadaver:
I can work those changes in before I merge, shouldn't be a large job. It's also better to allow Zone::SetAmbientColor() to set the alpha component directly, without having to go through SetAttribute.

dragonCASTjosh:
if your able to work on that then ill start work on area light.

PR #1499, opened 2016-07-30
CMake fixes for compiling on Ubuntu 16.04

feltech:
* rapidjson and STB CMake `install_header_files` call using `USE_FILE_SYMLINK` causes "Too many levels of symbolic links" error.
* Ensure necessary folder required for tolua++ to build.
* Link `gcc_s` and `gcc` to work around GCC 5.4 issue with STB in shared libraries (known issue https://github.com/nothings/stb/issues/280).

This is as much a bug report as pull request - there may be better ways to solve these issues.  Especially the symbolic links issue - the fix doesn't address the underlying cause, just disables symlinks.

### DIFF
--- a/Source/ThirdParty/STB/CMakeLists.txt
+++ b/Source/ThirdParty/STB/CMakeLists.txt
@@ -26,6 +26,5 @@ set (TARGET_NAME STB)
 # Setup target
 add_custom_target (${TARGET_NAME} ALL)   # Dummy target just so that its post-build step can install headers to the build tree
 set (STATIC_LIBRARY_TARGETS ${STATIC_LIBRARY_TARGETS} ${TARGET_NAME} PARENT_SCOPE)
-
 # Install headers for building the Urho3D library
-install_header_files (DIRECTORY ./ DESTINATION ${DEST_INCLUDE_DIR}/ThirdParty/STB FILES_MATCHING PATTERN *.h USE_FILE_SYMLINK BUILD_TREE_ONLY)  # Note: the trailing slash is significant
+install_header_files (DIRECTORY ./ DESTINATION ${DEST_INCLUDE_DIR}/ThirdParty/STB FILES_MATCHING PATTERN *.h BUILD_TREE_ONLY)  # Note: the trailing slash is significant
--- a/Source/ThirdParty/rapidjson/CMakeLists.txt
+++ b/Source/ThirdParty/rapidjson/CMakeLists.txt
@@ -28,4 +28,4 @@ add_custom_target (${TARGET_NAME} ALL)   # Dummy target just so that its post-bu
 set (STATIC_LIBRARY_TARGETS ${STATIC_LIBRARY_TARGETS} ${TARGET_NAME} PARENT_SCOPE)
 
 # Install headers for building the Urho3D library
-install_header_files (DIRECTORY include/rapidjson/ DESTINATION ${DEST_INCLUDE_DIR}/ThirdParty/rapidjson FILES_MATCHING PATTERN *.h USE_FILE_SYMLINK BUILD_TREE_ONLY)  # Note: the trailing slash is significant
+install_header_files (DIRECTORY include/rapidjson/ DESTINATION ${DEST_INCLUDE_DIR}/ThirdParty/rapidjson FILES_MATCHING PATTERN *.h BUILD_TREE_ONLY)  # Note: the trailing slash is significant
--- a/Source/ThirdParty/toluapp/src/bin/CMakeLists.txt
+++ b/Source/ThirdParty/toluapp/src/bin/CMakeLists.txt
@@ -64,6 +64,7 @@ if (URHO3D_UPDATE_SOURCE_TREE)
 endif ()
 file (GLOB TOLUA lua/*.lua)
 add_custom_command (OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/generated/toluabind.c
+    COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/generated
     COMMAND ${CMAKE_BINARY_DIR}/bin/tool/tolua++ -o ${CMAKE_CURRENT_BINARY_DIR}/generated/toluabind.c -H ${CMAKE_CURRENT_BINARY_DIR}/generated/toluabind.h -n tolua tolua_scons.pkg 2>${NULL_DEVICE} || ${CMAKE_COMMAND} -E copy_if_different ${CMAKE_CURRENT_SOURCE_DIR}/toluabind.c ${CMAKE_CURRENT_BINARY_DIR}/generated/toluabind.c
     ${UPDATE_COMMAND}
     DEPENDS ${TOLUA} tolua_scons.pkg
--- a/Source/Urho3D/CMakeLists.txt
+++ b/Source/Urho3D/CMakeLists.txt
@@ -419,6 +419,10 @@ endif ()
 # Use PIC on platforms that support it (shared library type has this property set to true by default, so we only have to deal with those static ones that the shared library links against)
 if (URHO3D_LIB_TYPE STREQUAL SHARED)
     set_target_properties (${STATIC_LIBRARY_TARGETS} PROPERTIES POSITION_INDEPENDENT_CODE true)
+    if (CMAKE_COMPILER_IS_GNUCXX)
+        # Required for Ubuntu 16.04 to use STB library: https://github.com/nothings/stb/issues/280
+        target_link_libraries(${TARGET_NAME} gcc_s gcc)
+    endif ()
     if (NOT MSVC AND NOT (MINGW AND CMAKE_CROSSCOMPILING) AND CMAKE_VERSION VERSION_LESS 2.8.9)  # todo: Remove this when CMake minimum version is 2.8.9
         set_property (TARGET ${STATIC_LIBRARY_TARGETS} APPEND PROPERTY COMPILE_FLAGS -fPIC)
     endif ()
--- a/Source/Urho3D/CMakeLists.txt
+++ b/Source/Urho3D/CMakeLists.txt
@@ -419,7 +419,7 @@ endif ()
 # Use PIC on platforms that support it (shared library type has this property set to true by default, so we only have to deal with those static ones that the shared library links against)
 if (URHO3D_LIB_TYPE STREQUAL SHARED)
     set_target_properties (${STATIC_LIBRARY_TARGETS} PROPERTIES POSITION_INDEPENDENT_CODE true)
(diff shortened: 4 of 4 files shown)


weitjong:
Thanks for your PR. It is interesting to see that Ubuntu 16.04 needs such tweaks to make the build works. Personally I don't have 16.04 yet to verify this but this is what I do know.

- The USE_FILE_SYMLINK option for the install_header_files() custom macro is there for a purpose, it allows Windows users without MKLINK privilege to fallback to hardcopying at individual file level using "cmake -E copy_if_different" command. Without this option, the hardcopying will be done at the directory level where there is no "if_different" counterpart. Since we are hardcopying source files (or headers to be more precise) in this case, we have to keep those files' timestamp in order not to retrigger the build unnecessarily on Windows platform. On Linux platforms where the symbolic link is always available, the create_symlink() macro will simply call "cmake -E create_symlink". So far I have not encountered first hand the "too many levels of symbolic links" in our build. Note that the macro in question has logic to convert all the relative paths to absolute ones to avoid the symlink recurses into itself. Perhaps, it is more useful if you could report back where the "too many levels" error occurred and see whether it is caused by our build scripts or something peculiar at your own build environment.

- We do not use command to create the "generate" directory explicitly because so far it seems that tolua++ is able to auto-generate it when necessary.

- As for the g++ bug, I have a little reservation to fix it at the application level. If it is indeed a g++ bug then I think it will be fixed rather sooner upstream by GCC devs, perhaps it is already fixed there but just not backported to 16.04 yet. If we really want to workaround this at our end then I suppose we have to tighten the GCC version check so that the workaround does not apply to all g++ compiler version indiscriminately.

NCrashed:
The same story is about STB issue on Fedora 24 x86_64, the PR successfully fixes it.

weitjong:
Thanks for the head up (had originally planned to upgrade my F23 system :smiley:).

feltech:
As soon as I get a chance, I will report back the console logs of the symlink and tolua errors, and I'll have a closer look at the STB issue to see if I can tighten the constraint on gcc version.

feltech:
OK, upon re-enabling `USE_FILE_SYMLINK` in `Urho3D/Source/ThirdParty/rapidjson/CMakeLists.txt`, doing a `make clean` and building again gives
```
<snip>
[ 64%] Building CXX object Source/Urho3D/CMakeFiles/Urho3D.dir/Resource/JSONFile.cpp.o
cd /home/dave/Dropbox/Workspace/Urho3D/Source/Urho3D && /usr/bin/c++   -DGLEW_NO_GLU -DGLEW_STATIC -DHAVE_SINCOSF -DHAVE_STDINT_H -DURHO3D_CXX11 -DURHO3D_IS_BUILDING -DURHO3D_LOGGING -DURHO3D_LUA -DURHO3D_LUA_RAW_SCRIPT_LOADER -DURHO3D_NAVIGATION -DURHO3D_PHYSICS -DURHO3D_THREADING -DUrho3D_EXPORTS -I/home/dave/Dropbox/Workspace/Urho3D/Source/Urho3D -I/home/dave/Dropbox/Workspace/Urho3D/include/Urho3D/ThirdParty -I/home/dave/Dropbox/Workspace/Urho3D/include/Urho3D/ThirdParty/Bullet -I/home/dave/Dropbox/Workspace/Urho3D/include/Urho3D/ThirdParty/Detour -I/home/dave/Dropbox/Workspace/Urho3D/include/Urho3D/ThirdParty/LuaJIT  -std=gnu++11 -Wno-invalid-offsetof -march=native -ffast-math -pthread -fdiagnostics-color=auto -include "/home/dave/Dropbox/Workspace/Urho3D/Source/Urho3D/Precompiled.h" -fvisibility=hidden -fvisibility-inlines-hidden -O2 -g -DNDEBUG -fPIC   -o CMakeFiles/Urho3D.dir/Resource/JSONFile.cpp.o -c /home/dave/Dropbox/Workspace/Urho3D/Source/Urho3D/Resource/JSONFile.cpp
/home/dave/Dropbox/Workspace/Urho3D/Source/Urho3D/Resource/JSONFile.cpp:34:32: fatal error: /home/dave/Dropbox/Workspace/Urho3D/include/Urho3D/ThirdParty/rapidjson/document.h: Too many levels of symbolic links
 #include <rapidjson/document.h>
                                ^
compilation terminated.
Source/Urho3D/CMakeFiles/Urho3D.dir/build.make:710: recipe for target 'Source/Urho3D/CMakeFiles/Urho3D.dir/Resource/JSONFile.cpp.o' failed
make[2]: *** [Source/Urho3D/CMakeFiles/Urho3D.dir/Resource/JSONFile.cpp.o] Error 1
<snip>
```
Let me know if you would like to see more build output, but I figured this is the relevant bit.

If I go into the generated include folder, then
```
dave@dave-desktop:~/Dropbox/Workspace/Urho3D/include/Urho3D/ThirdParty/rapidjson$ ls -la
total 72
drwxrwxrwx 3 dave dave 4096 Jul 31 20:36 .
drwxrwxrwx 3 dave dave 4096 Nov 12  2015 ..
lrwxrwxrwx 1 dave dave   94 Jul 31 20:36 allocators.h -> /home/dave/Dropbox/Workspace/Urho3D/Source/ThirdParty/rapidjson/include/rapidjson/allocators.h
lrwxrwxrwx 1 dave dave   92 Jul 31 20:36 document.h -> /home/dave/Dropbox/Workspace/Urho3D/Source/ThirdParty/rapidjson/include/rapidjson/document.h
lrwxrwxrwx 1 dave dave   97 Jul 31 20:36 encodedstream.h -> /home/dave/Dropbox/Workspace/Urho3D/Source/ThirdParty/rapidjson/include/rapidjson/encodedstream.h
lrwxrwxrwx 1 dave dave   93 Jul 31 20:36 encodings.h -> /home/dave/Dropbox/Workspace/Urho3D/Source/ThirdParty/rapidjson/include/rapidjson/encodings.h
lrwxrwxrwx 1 dave dave   98 Jul 31 20:36 filereadstream.h -> /home/dave/Dropbox/Workspace/Urho3D/Source/ThirdParty/rapidjson/include/rapidjson/filereadstream.h
lrwxrwxrwx 1 dave dave   94 Jul 31 20:36 filestream.h -> /home/dave/Dropbox/Workspace/Urho3D/Source/ThirdParty/rapidjson/include/rapidjson/filestream.h
lrwxrwxrwx 1 dave dave   99 Jul 31 20:36 filewritestream.h -> /home/dave/Dropbox/Workspace/Urho3D/Source/ThirdParty/rapidjson/include/rapidjson/filewritestream.h
drwxrwxrwx 2 dave dave 4096 Jul 31 20:36 internal
lrwxrwxrwx 1 dave dave   96 Jul 31 20:36 prettywriter.h -> /home/dave/Dropbox/Workspace/Urho3D/Source/ThirdParty/rapidjson/include/rapidjson/prettywriter.h
lrwxrwxrwx 1 dave dave   93 Jul 31 20:36 rapidjson.h -> /home/dave/Dropbox/Workspace/Urho3D/Source/ThirdParty/rapidjson/include/rapidjson/rapidjson.h
lrwxrwxrwx 1 dave dave   90 Jul 31 20:36 reader.h -> /home/dave/Dropbox/Workspace/Urho3D/Source/ThirdParty/rapidjson/include/rapidjson/reader.h
lrwxrwxrwx 1 dave dave   96 Jul 31 20:36 stringbuffer.h -> /home/dave/Dropbox/Workspace/Urho3D/Source/ThirdParty/rapidjson/include/rapidjson/stringbuffer.h
lrwxrwxrwx 1 dave dave   90 Jul 31 20:36 writer.h -> /home/dave/Dropbox/Workspace/Urho3D/Source/ThirdParty/rapidjson/include/rapidjson/writer.h
```
but
```
dave@dave-desktop:~/Dropbox/Workspace/Urho3D/include/Urho3D/ThirdParty/rapidjson$ pwd
/home/dave/Dropbox/Workspace/Urho3D/include/Urho3D/ThirdParty/rapidjson
```
so the symlinks are pointing to themselves!

Also, in case it's relevant
```
dave@dave-desktop:~/Dropbox/Workspace/Urho3D/include/Urho3D/ThirdParty/rapidjson$ cmake --version
cmake version 3.5.1
```
and
```
dave@dave-desktop:~/Dropbox/Workspace/Urho3D/include/Urho3D/ThirdParty/rapidjson$ gcc --version
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.1) 5.4.0 20160609
```

feltech:
I have tried to reproduce the tolua++ failing because a folder did not exist issue fixed in the PR, but I don't seem to be able to.  It was a while ago I experienced the issue, and it looks like some things have changed.  It looks like the necessary `generated` folder has moved from `/Urho3D/Source/ThirdParty/toluapp/src/bin` to `/Urho3D/Source/Urho3D/tolua++-prefix/src/tolua++-build`?

Regardless, it looks like that part of the PR can indeed be ignored.

weitjong:
I still don't fully understand why it happened to your build tree. Based on your logs, it is clear to me that you were using non-out-of-source build tree. We recommend to use out-of-source build tree. Still, our build system actually supports both. I have quickly generated a non-out-of-source build tree to test and I did not manage to reproduce your issue. In my test the symlinks in the "include" directory point to the original headers and those original headers remain the same as in the repo, i.e. not turned into a symlink pointing to itself like in your case. I am using F23 with CMake 3.4.3 against the latest master branch. So, I believe the culprit is something else. If I have to guess then perhaps it is your source tree cum build tree. Here is one example where it could gone horribly wrong. Each time one change the option of the ```install_header_files()``` macro from directory mode to file mode later, the macro does not clean up after itself. The remnant symlinks from the previous setup in the build tree most likely will interfere with the new changes and the end result could be exactly as you have observed. Since the remnant symlinks were pointing to the original source tree, further modification to them will then modify the original source tree. If this happens, doing a git reset may or may not fix this problem due to .gitignore file. Probably may need to manually delete the whole ```Thirdparty/rapidjson``` directory and then ```git checkout -- Thirdparty/rapidjson``` to get it back to its pristine state.

weitjong:
I can almost sure now that your symlink issue is caused by mode switching when calling ```install_header_files()``` macro. I have added a new logic to perform a clean up process to prevent this kind of issue to happen in future. However, the changes will be checked in our ```refactor-buildsystem``` branch first, so you will have to wait a little while before you can test this.

feltech:
Great!  So
* symlink issue - to be fixed in another branch.
* tolua++ missing folder issue - can't replicate, probably fixed some time ago.
* STB gcc 5.3/5.4 issue - awaiting upstream fix.

So I'll close this PR.

weitjong:
I bite the bullet to upgrade my system to F24, knowing that it would break my development environment for Urho3D. I just want to add a comment that whatever it is, the problem persists until GCC version 6.1.1 (current version in F24), so not just GCC 5.4.x. Probably we should open another issue to track this.

PR #1506, opened 2016-08-06
Input events

rokups:
Added three input-related events:

* E_INPUTBEGIN : input handling starts.
* E_INPUTEND : input handling ends.
* E_SDLRAWINPUT : raw SDL event is sent for customized event processing.

`E_SDLRAWINPUT` additionally checks `P_CONSUMED` value (which is `false` by default) and if it is set to true event is discarded. This is for libraries that would like to take over input handling completely.

Need for other two events arose from handling nuklear UI events. It requires to do some things before and after input handling. Without these events doing those tasks took bit of improper hackery abusing other events and their order which is not good at all.


cadaver:
This looks acceptable to include even just before release, since it's just sending events and checking a boolean.

PR #1508, opened 2016-08-07
Minor Typo correction

BlueMagnificent:
The current history (v1.6 right before release) has the below line
> AngelScript support on also on Emscripten and 64-bit iOS, based on generic bindings

corrected to 
> AngelScript support on Emscripten and 64-bit iOS, based on generic bindings


cadaver:
Thanks!

BlueMagnificent:
Wow,
That was fast

PR #1517, opened 2016-08-12
Added light temperature

dragonCASTjosh:
screenshot_tue_aug__9_18_14_10_2016

The image should explain things well enough, if not then i have added physically based light temperature to lights, it changes the hue based on the kelvin scale.

Reference scale:
http://www.jozifolk.com/resources/kelvin-chart.jpg

### DIFF
--- a/Source/Urho3D/AngelScript/GraphicsAPI.cpp
+++ b/Source/Urho3D/AngelScript/GraphicsAPI.cpp
@@ -1201,7 +1201,9 @@ static void RegisterLight(asIScriptEngine* engine)
     engine->RegisterObjectMethod("Light", "void set_perVertex(bool)", asMETHOD(Light, SetPerVertex), asCALL_THISCALL);
     engine->RegisterObjectMethod("Light", "bool get_perVertex() const", asMETHOD(Light, GetPerVertex), asCALL_THISCALL);
     engine->RegisterObjectMethod("Light", "void set_color(const Color&in)", asMETHOD(Light, SetColor), asCALL_THISCALL);
-    engine->RegisterObjectMethod("Light", "const Color& get_color() const", asMETHOD(Light, GetColor), asCALL_THISCALL);
+	engine->RegisterObjectMethod("Light", "const Color& get_color() const", asMETHOD(Light, GetColor), asCALL_THISCALL);
+	engine->RegisterObjectMethod("Light", "void set_temperature(float)", asMETHOD(Light, SetTemperature), asCALL_THISCALL);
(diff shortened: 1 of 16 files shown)


dragonCASTjosh:
no idea how to handle lua

cadaver:
Does this do anything else than modulate the light RGB color with the temperature formula? In that case it seems unnecessary to repeat the calculation per pixel in the shader.

I'd rather recommend something like this:
- Add bool attribute "Use Temperature" to Light, can default to false to not cause unintended effects. When enabled, the temperature modulation is calculated on the CPU in Light's GetEffectiveColor() function.
- No shader changes should be necessary if I understand right.

dragonCASTjosh:
The default value should be white light although I was considering adding a bool. I also tried moving the calculation to light.cpp and apply it at the same place as brightness although Color * Color caused some issues I don't have time to look at this week.

dragonCASTjosh:
@cadaver committed a version with no shader changes, and made temperature values optional

cadaver:
Thanks, implementation looks good now. However there's tabs being used in Light.h / Light.cpp (please use spaces) and you should revert the shader changes completely, if you check the changed files list there's still shader changes.

dragonCASTjosh:
i changed visual studio to use spaces under c++ so im not sure whats up there

cadaver:
Use Edit -> Advanced -> View White Space in VS to be sure. Maybe you touched the files with another editor that uses tabs?

dragonCASTjosh:
@cadaver fixed spacing and removed shaders should all be good now

cadaver:
Thanks. I'll check it and add Lua bindings, likely today.

dragonCASTjosh:
thanks, and where would i look for doing lua bindings for future reference

cadaver:
Look at the .pkg files in Source/Urho3D/LuaScript/pkgs . In easy cases this means just copy-pasting the new function declarations from header files. Note that we may change Lua binding system, but that's still upcoming until the build problems are resolved for all supported platforms.

dragonCASTjosh:
@cadaver im am also working on implementing lumans for optional physical based light brightness, if you want to hold of on the merge i can push it here and change the use temperature option to use physical values and add lua

cadaver:
That's fine, I'm in no hurry to merge.

dragonCASTjosh:
light color is in kelvin and is now multiplied with with the standard color to allow for artist tweeking. 
light brightness is in lumens power

all physical values are only used if the use physical value check box is used.

cadaver:
Merged.

PR #1520, opened 2016-08-15
replaced jack model w/ mutant model in 18 character demo

Lumak:


1vanK:
I agree that it is necessary to replace the content in the demos to more beautiful. Engine code is very high quality. Examples perfectly demonstrate its capabilities. But they look unattractive. But the demos - the face of the engine.

cadaver:
Looks excellent! Almost too good considering the simple background assets in the demo :)

Some minor nitpicks. Lua changes are missing, and there seems to be some tabs instead of spaces in the AngelScript code. These are not a big thing to fix, but naturally it'd be welcome if you fixed them.

yushli:
When will this be merged into main branch? I can't wait to replace the naked Jack with this new model.

cadaver:
Yushli, please stop with all variations of "when will this.." type of questions, the answer is always "when it's done" or "when it's ready" and you don't expedite anything by asking, or add any value to the topic at hand. It's a bit harsh, but it has to be said.

yushli:
Ok, will follow your advice.

Lumak:
ok, done. also corrected the spacing for .as file.

cadaver:
Thanks!

cadaver:
Merged. Converted textures to jpg to save repository space. Admittedly .dds format would be faster to load, but it would have needed converting the normal map to the green/alpha channel format.

1vanK:
I think "Mutant_Jump.ani" looks better than "Mutant_Jump1.ani"

Lumak:
The Jump is a full animation: starts with a wind up (near squat), jump, fall and land. And the Jump1 is a cut segment of that animation.  While the Jump animation might look better if standing still, it would fall apart from a run, and it would require changing the timing when to apply the impulse and also a change in the timing in a fall when walking off blocks.  
I used Jump1 to simplify this submission and w/o having to rewrite the existing timing logic. However, I've included 10 animations to encourage modding the code.

Lumak:
I'd like to add a few more animation files: 
-layer animations: block and wave
-jumpfall, kick, and hiphop1

Just assets, no code change.

cadaver:
That's fine.

BlueMagnificent:
Sorry for resurrecting this... I noticed that there is no reference to this Mutant model in the "credits" list like we have Jack credited to realXtend project

cadaver:
Neither do we have to the PBR models and materials. The credits section is somewhat outdated and could even be cropped; what's important are the license texts for the assets that are used.

PR #1521, opened 2016-08-16
Fixed the possibility of node collision event normalB not matching nodeB

Enhex:
This problem always existed as far as I can remember, and I used to work around it by using the direction between the nodes instead of the contact normal (which aren't the same).

When I reload a level (first load is consistent AFAIR) sometimes the normals of contacts flip. It only happens after reloading a level, possibly because the order of the bodies changes.

In `Urho3D::PhysicsWorld::SendCollisionEvents()` the order of the bodies A and B may be flipped to ensure unique pairs, so original A becomes new B and original B becomes new A, but the contactManifold's points' `m_normalWorldOnB` stays the same, and later on when collision event's contacts_ is populated it handles the normals according to the new A & B while the normals are according to the original A & B.

The contact points' normals can't be flipped directly because it seems Bullet still uses them (tried that and it caused bodies to wiggle).
A possible solution is to track if the order of the bodies was flipped, and take that into account when populating the event data's contacts_ by flipping the normals too (tried it and it worked).

Also, does Bullet generate duplicate contactManifolds? If it doesn't it may be possible to avoid using a HashMap for the purpose of iteration, thus needing to reorder the bodies, and that would be a more efficient solution.
I do see that the HashMap is used for lookup to check if a previousCollisions_ ended.

A simple alternative way to do that is to use both a vector for iteration, and a HashMap for checking against previous collisions, which would be possibly faster too.

A better is to send the events from inside the contactManifolds loop and using the original A & B bodies, removing the need to iterate over the HashMap all together, which is also faster.
I tried it with my project and some of the Urho samples and it seems to work fine, suggesting Bullet doesn't generate duplicate manifolds.
This is the solution in this pull request (this post was originally written as an issue).

### DIFF
--- a/Source/Urho3D/Physics/PhysicsWorld.cpp
+++ b/Source/Urho3D/Physics/PhysicsWorld.cpp
@@ -896,25 +896,19 @@ void PhysicsWorld::SendCollisionEvents()
             // First only store the collision pair as weak pointers and the manifold pointer, so user code can safely destroy
             // objects during collision event handling
             currentCollisions_[bodyPair] = contactManifold;
-        }
 
-        for (HashMap<Pair<WeakPtr<RigidBody>, WeakPtr<RigidBody> >, btPersistentManifold*>::Iterator i = currentCollisions_.Begin();
-             i != currentCollisions_.End(); ++i)
-        {
-            RigidBody* bodyA = i->first_.first_;
-            RigidBody* bodyB = i->first_.second_;
+            // Send start/ongoing collision events
             if (!bodyA || !bodyB)
                 continue;
 
-            btPersistentManifold* contactManifold = i->second_;
-
+            // Using the original body A & B order so m_normalWorldOnB matches them
             Node* nodeA = bodyA->GetNode();
             Node* nodeB = bodyB->GetNode();
             WeakPtr<Node> nodeWeakA(nodeA);
             WeakPtr<Node> nodeWeakB(nodeB);
 
             bool trigger = bodyA->IsTrigger() || bodyB->IsTrigger();
-            bool newCollision = !previousCollisions_.Contains(i->first_);
+            bool newCollision = !previousCollisions_.Contains(bodyPair);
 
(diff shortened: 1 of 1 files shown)


cadaver:
Thanks for raising the issue. This fix can't be applied directly, because we can't take the risk of bodies getting destroyed by event handlers while the manifold is being accessed. I'll try to make a fix that retains this PR's "spirit", e.g. bodyA & bodyB shouldn't be swapped.

Lumak:
I'm not sure if this needs to be fixed. I learned about the this when I first started learning about the engine and come to understand that the normals are "collision normal" acting on the object.

Enhex:
What was the reason for the re-ordering in the first place? I assume Bullet doesn't create duplicates body pairs, so is it because the order of the bodies may change between current and previous collisions?

cadaver:
Actually I do find that NinjaSnowWar generates duplicate manifolds / pairs, at least at the moment of snowball hitting a ninja.

However, what happens with the current code is definitely incorrect too, because if there is a duplicate, then the second manifold will get discarded from collision event processing.

Easiest would be to just ignore the matter and signal the collision for both manifolds in case a duplicate happens, however this could have issues with some game logic code, e.g. damage being dealt twice. A certainly proper solution would be to collect the contact points from both pairs into our own structure before sending collisions.

Enhex:
Collecting the contact points before sending them means the copied normals can be corrected if there was an order flip.

cadaver:
Yes. It's just a possible heap memory allocation strain, if there's a hashmap, which contains contact vectors.

Lumak:
When I looked at the code, I made my interpretation of what was happening and understood it as "collision normal" acting on the object, but from the looks of what cadaver is finding, maybe that wasn't a good idea.

cadaver:
Committed a fix that saves both manifold pointers (if necessary) in a pair-like structure. Contact handling became just slightly slower, but no extra data copies are being made. Removed Abs() from CharacterDemo & NinjaSnowWar ground collision handling to prove that it should work properly now.

PR #1526, opened 2016-08-19
more mutant anims

Lumak:
-renamed Jump1 anim to JumpStop and replaced Jump1 with jump-to-fall animation
-added more anims: block layer, wave layer, kick, hip hop


cadaver:
Merged!

PR #1529, opened 2016-08-21
Make LineEdit text and cursor to be always verticaly centered

Scellow:
I believe that it is how it should be by default, it is not a MultiLineEdit so it should be centered Vertically

http://i.imgur.com/qjAYO0f.png


cadaver:
This looks to be the default behavior also in e.g. operating system edit fields, so makes sense.

PR #1532, opened 2016-08-22
added a feature to suppress $fbx nodes in assetImporter

Lumak:
If an fbx model contains non-reset bone transforms (non-zero or having pivot point) it will generate multiple $AssimpFbx$ nodes per bone.  Because of this some Mixamo characters will generate near 200 bones per model which will fail to animate the characters correctly, not to mention the over head it causes to interpolate all these bones at run time.


Lumak:
A good test case would be the Y_Bot model from Mixamo.

cadaver:
Thanks! I don't have time to go through the code in intimate detail but I trust your implementation. Tested a non-FBX model briefly and nothing appeared broken. In case there's any trouble reported, this can be revised later.

PR #1550, opened 2016-08-30
fixed a bug where crash is caused by geometry having key animation

Lumak:
In addition, refactored the code a bit.  Two additional layered animations are carried over from my canceled submission.


cadaver:
It's ok this time to merge both the animation & fixes, but using feature branches would be more pleasant to work with for PR's. A master branch sort of implies it could get more work appended to it any time.

Lumak:
Ok, thanks, and I'll look into how to create a branch and learn more about github features.

cadaver:
Learn git, not github :)

PR #1553, opened 2016-08-30
IntVector2::ToHash()

henu:
Hello

I think it's common in games that IntVector2 is used as a key in set or map. For example, I use it to keep track of tiles that need updating. I think it needs to be hashable, so I can use it in HashSet. Hope I understood the HashSet correctly and hope you agree with me :)

About the hashing function. Why does it have two random numbers? Honestly, I don't know. I just made a wild guess that if I put two big prime numbers there, it's hard for hash numbers to make a collision.

### DIFF
--- a/Source/Urho3D/Math/Vector2.h
+++ b/Source/Urho3D/Math/Vector2.h
@@ -323,6 +323,9 @@ public:
     /// Return as string.
     String ToString() const;
 
+    /// Return hash value for HashSet & HashMap.
+    unsigned ToHash() const { return x_ * 15485863 + y_ * 32452843; }
+
     /// X coordinate.
     int x_;
     /// Y coordinate.
--- a/Source/Urho3D/Math/Vector2.h
+++ b/Source/Urho3D/Math/Vector2.h
@@ -324,7 +324,7 @@ public:
     String ToString() const;
 
     /// Return hash value for HashSet & HashMap.
-    unsigned ToHash() const { return x_ * 15485863 + y_ * 32452843; }
+    unsigned ToHash() const { return (unsigned)x_ * 31 + (unsigned)y_; }
 
     /// X coordinate.
     int x_;


OvermindDL1:
If you are doing it for 'tiles' that are centered around 0, then striping it would be far more efficient for a hash function:
```cpp
(n << 1) ^ (n >> 31) // For 32-bit integers

(n << 1) ^ (n >> 63) // For 64-bit integers
```
And then you can has that to a prime but even then not needed at all as they encode to unique values, it is an almost perfect (though not 'noised') hash space just as it is above.

henu:
The tile thing was just one example. I've been using integer vectors as keys in maps and sets in other cases as well.

Anyway, I didn't quite understood this striping method.. Isn't that just circular bit shifting?

eugeneko:
Signed integer overflow is Undefined Behavior, do you know?

2016-08-31 2:41 GMT+03:00 Henrik Heino <notifications@github.com>:

henu:
>Signed integer overflow is Undefined Behavior, do you know?

Hmm, didn't know that. After quick googling I found that [unsigned overflow is defined](http://stackoverflow.com/questions/18195715/why-is-unsigned-integer-overflow-defined-behavior-but-signed-integer-overflow-is) and [casting to unsigned should also be safe](http://stackoverflow.com/a/4975363/323352). So I guess I should convert them to unsigned first...

But do you think the whole "multiply by big prime" is stupid approach, because if it is, I would like to ditch it and change to some other method instead. I know there was already a suggestion in second comment, but I didn't understood it.

One thing that came to my mind too: Would it be a good idea if Hash.h would contain a function for this kind of combinations of two or multiple hashes?

weitjong:
How about just ```x_ * 31 + y_```? 31 is a prime number. There is no reason to choose a big prime for this case IMHO as it is not for cryptography. BTW, what I just proposed is considered a good hash in Java world for a class having two integer field members.

1vanK:
from http://stackoverflow.com/questions/892618/create-a-hashcode-of-two-numbers
```
int hash = 23;
hash = hash * 31 + item1Hash;
hash = hash * 31 + item2Hash;
hash = hash * 31 + item3Hash;
hash = hash * 31 + item4Hash;
hash = hash * 31 + item5Hash;
// etc
```

cadaver:
Looks good to merge now.

cadaver:
Merged.

PR #1560, opened 2016-09-02
Draw navigation debug geometry in Editor

1vanK:


1vanK:
If I understand correctly, NavigationMesh and DynamicNavigationMesh always connect to root node.
But SceneRebuildNavigation() search components recursive.

1vanK:
Also NavMesh does not write log message
```
URHO3D_LOGERROR("CrowdManager is a scene component and should only be attached to the scene node");
```
like void CrowdManager::OnSceneSet() when attached to not root node

1vanK:
fixed

cadaver:
Thanks! I was indeed missing something like this when debugging the Obstacle OnMarkedDirty() implementation.  Are the UI XML changes intentional?

1vanK:
> Are the UI XML changes intentional?

yes, just style fix, it is not directly related to the patch

cadaver:
Merged.

Issue #1556, opened 2016-08-31
Adding elements from [POD]Vector to itself crashes applictaion.


eugeneko:
Any insertions that consume elements from the same vector may crash applictaion because of wrong resize implementation.

```
    Vector<String> vec = { "One", "Two", "Three" };
    vec.Push(vec.Front()); // Dies here
```

eugeneko:
I disagree.
Huh... I mean... Urho need better solution.
Even with these new checks this vector is vulnerable to such problems. And it is also not-self-assignable because of similar reason. I suppose it is better to implement Vector in STL style when old buffers are not removed before new ones are completely filled. It does not look important because 
nobody noticed that bug until now. However, I think it should be completely fixed at some point.
When I have propper solution, I'll create a PR.

Here is some synthetic crash example:

```
struct Иode
{
    Vector<Иode> children;
    String tag;
};

    Vector<Иode> nodes(1);
    nodes[0].tag = "One";
    nodes[0].children.Push(Иode());
    nodes.Push(nodes[0].children[0]);
```

cadaver:
Yes, better solution is welcome. Looking at the fix I was making I knew it was a band-aid, because there are compound operations where the resize happens first (deleting the buffer), followed by move / insert operations.

PR #1561, opened 2016-09-03
Refactor Vector and PODVector implementation.

eugeneko:
Related to #1556.
I re-organized vector code to prevent old buffer destruction in case of reallocation until new buffer is constructed. I also removed this hacky macro and added one new ctor for both vectors.
**Upd:** I smoke-tested Editor and all demos.

### DIFF
--- a/Source/Urho3D/Container/Vector.h
+++ b/Source/Urho3D/Container/Vector.h
@@ -36,8 +36,6 @@
 #pragma warning(disable:6293)
 #endif
 
-#define URHO3D_VECTOR_ISINSIDE(val) (&val >= Buffer() && &val < Buffer() + size_)
-
 namespace Urho3D
 {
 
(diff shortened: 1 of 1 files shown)


cadaver:
Appreciated very much! The code became much less repeated, too.
I believe List-like containers also have a potential problem with the next/prev links when self-inserting, but I can look into that separately.

eugeneko:
What kind of problems do you expect from List?
As far as I see, List always copies passed element and then initializes its links.

cadaver:
Yes, I meant range self-insertion, which currently works one-by-one. I think it could be made workable by first constructing copies of all the elements in the range, and simply chaining them to each other, then fixing the links into the rest of the list as the final step.

1vanK:
I am sorry for offtopic, why not use std containers?

cadaver:
It's explained in the documentation, there are a few reasons, such as guaranteed binary size for constructing inside a Variant.

1vanK:
Thanks  for the clarification, I found docs article

cadaver:
Briefly tested self-range insertion with List and didn't see problems, at least in a simple case. Added just a self-assignment check to containers to prevent mistaken clearing.

eugeneko:
PS: I notced that I've written an assert in Vector::InsertElements but not in PODVector::InsertElements. Please add this forgotten assert when you commit something next time.

PR #1568, opened 2016-09-06
Add new face camera mode to BillboardSet and Text3D.

eugeneko:
This PR is a small improvement of Urho's billboard system.

**LookAt Y** and **Rotate Y** modes are useful for rendering some grounded objects like grass or trees that shouldn't be freely rotated. Such billboards degrade when camera looks downside and this case is often unwanted.

This PR introduces new mode called **LookAt Mixed** that is a hybrid of **LookAt XYZ** and **LookAt Y**. New attribute **minAngle** controls minimum angle between billboard normal and look-at direction. 0 is **LookAt Y** mode, 90 is **LookAt XYZ**. Small values of **minAngle** allow user to have almost-lookat-Y billboards that don't degrade when looking downside. See screenshot.

Unfortunatelly, implementation for **Rotate Y**/**Rotate XYZ** looks much more difficult. But maybe it will be implemented at some point.

image


cadaver:
Looks good. Will look in more detail later today.

PR #1577, opened 2016-09-11
Add NavigationPushiness::NAVIGATIONPUSHINESS_NONE

1vanK:


1vanK:
I think good practice is set NAVIGATIONPUSHINESS_LOW for moving agents, but when they arrive to target point - set pushiness to NAVIGATIONPUSHINESS_NONE

PR #1581, opened 2016-09-13
Fix LookAt Mixed rotation mode.

eugeneko:
Mixed mode needs true look-at vector instead of XZ projection.

Recent refectoring changed this: https://github.com/urho3d/Urho3D/commit/d19d8d3dbfd962cce273bf43dadaa32bb08bcab8

PS: Yes, I _do_ like const-s and I will write'em wherever possible. Because they avoid errors like this one.


weitjong:
My apology if my recent change break it. Personally I dislke code bloat due to similar looking code, especially when even the comments are evidently the same which was in fact misleading.

eugeneko:
@weitjong Anyway, refactoring was a good idea. I wanted to do the same but I feel uncomfortable to refactor owner's code without discussion.

PR #1587, opened 2016-09-14
Fix memory leak in Dictionary.

### LINKED ISSUE #1582: MSVC detects a lot of memory leaks in Editor
MSVC 2015 Update 3 detects a lot of small memory leaks if I try to run Editor in Urho3D player. Demos are leak-free both executables and scripts, even script integration demos. Most leaks are located in Str.cpp(431).

Have you tested Urho for memory leaks? Are these leaks real or just side effect of some static variables?

eugeneko:
RefCastObject adds one ref by itself. So extra AddRefScriptObject call cause memory leak when Dictionary.Get is called from script.

I'm not 100% sure in this solution. However, I've smoke-tested Editor and my project - no crashes, no leaks.

Closes #1582


cadaver:
This looks OK since the Array At() doesn't increment ref count either (rather, when storing the refcount is increased)

PR #1604, opened 2016-09-19
Expose mouse position manipulation api in Input class

rokups:
`Input::SetMousePosition()` and `Input::CenterMousePosition()` are moved to public access scope to allow access to these APIs. Sometimes there is need to manually position cursor and there are no other means.

One of such usecases is  having cursor hover UI button when application is opened should button be positioned right in center of the screen. This is more of a problem on mobile platforms where cursor is not visible and it is not clear why one (hovered) button looks differently than the others.

### DIFF
--- a/Source/Urho3D/Input/Input.h
+++ b/Source/Urho3D/Input/Input.h
@@ -297,6 +297,11 @@ public:
     /// Return whether application window is minimized.
     bool IsMinimized() const;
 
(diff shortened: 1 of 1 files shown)


cadaver:
These may cause unwanted interaction in hidden mouse mode, but should be OK in visible mouse mode when used responsibly. Will merge.

PR #1612, opened 2016-09-23
Minor changes

### LINKED ISSUE #1605: Can SetModel and SetMaterial be virtual?
Subject.
I've found that it could be useful for writing my own Drawable's.
If I confirm this, can I make'em virtual?
It shouldn't affect performance because Urho already has tons of virtual calls.

### LINKED ISSUE #1608: Cast from U to int in EnumAttributeAccessorImpl::Get
EnumAttributeAccessorImpl::Set picks int from Variant and casts it to actual enum type.
If EnumAttributeAccessorImpl::Get would do the same, can it have some unexpected side effects?
I sure no, but I want to confirm.

I want to make this change to allow EnumAttributeAccessorImpl support enum classes that are not implicitly convertible to integer types.

eugeneko:
Closes #1608 
Closes #1605 
And also add Resize overload that I've found useful.

### DIFF
--- a/Source/Urho3D/Container/Vector.h
+++ b/Source/Urho3D/Container/Vector.h
@@ -353,6 +353,16 @@ public:
     /// Resize the vector.
     void Resize(unsigned newSize) { Vector<T> tempBuffer; Resize(newSize, 0, tempBuffer); }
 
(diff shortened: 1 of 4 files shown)


cadaver:
Thanks!

PR #1617, opened 2016-09-26
Multisample textures

cadaver:
This implements ability for Texture2D and TextureCube rendertargets to be multisampled. On OpenGL / Direct3D11 Texture2D can operate with or without automatic resolve to 1-sample; when operating without, a shader can instead access the individual samples. Also adds multisample option for VSM shadows, which makes them smoother but costs some performance.

Multisampling is not (at least currently) supported on OpenGL ES.


cadaver:
I believe this is ready to go in.

JoshEngebretson:
Fantastic!  This will really help scene view aliasing issues 👍

PR #1630, opened 2016-10-04
Move Texture2D::GetImage and TextureCube::GetImage implementations.

eugeneko:
I needed Texture2D::GetImage for last months and today I've found out that Urho already has such function in AS binding. So I've moved it to C++ class and updated Lua pkgs. I've also added some error messages and simplified some Lua bindings around Detach.


cadaver:
Nice cleanup, thanks!

PR #1632, opened 2016-10-06
Add AnimationController::IsPlaying(layer) method

1vanK:


1vanK:
A player can have different animations for attacks ("Hit", "Slash", etc). So I use different layers for different TYPES of anymations (layer1 - attack animations, layer 2 - walk animations). So I always know whether playing at the moment any of the animations of attack.

But may be correct way is add types of animations:
```
animCtrl->Play("PlayerKick.ani", "Attack");
if (animCtrl->isPlayingType("Attack"))
{
    ...
```

cadaver:
Thanks, this is a good idea. Layer controls blending order but if your type also corresponds to that, it's perfectly valid. Even more lightweight than tags would be to add a function with string Contains() check to the animation name, if you follow a convention with the animation names.

1vanK:
I like idea with specific names. But what if I wont to use String::StartsWith() instead? Also I want to check fade state of played "attack" animation (when I want to start another animation without waiting fading of ended animation). So I think best solution is direct acces to "animations_" member. Currently is GetAnimationsAttr() but it copy data to another vector ant it can affect to perfomance if it is used often. C++ still hard for my, so I am not sure, is it possible to make readonly acces to "animations_" member with deprecation of changing it?

cadaver:
Exposing the internal animation control struct to script will be laborious, and it will also be poor performance if a copy needs to be made. Would this be better served by just iterating animationstates first and then checking additional information from AnimationController if necessary?

cadaver:
Exposed AnimationControl. At least on Lua you're probably able to wreak some damage if using it incorrectly, and on AngelScript I used, the slightly nasty fake refcounting mechanism, but as long as you're aware of what you're doing, it should work OK.

1vanK:
I have another idea, but it seems I'm late xD
We can just move "animations_" to protected section to allow acces in inheritanced class. Ofc it works only for C++ apps.

PR #1639, opened 2016-10-10
Fixed an issue for SQLite queries where String::Trimmed was being corrupted

victorholt:
This should allow sql queries to run properly when DbConnection::Execute is provided a string. This fix is only for SQLite queries.


iSLC:
The second variable `cSqlStr` is unnecessary.

victorholt:
Agreed, cSqlStr has been removed.

Issue #1637, opened 2016-10-09
Narrowing conversion error trying to compile Urho3D under C++11.


hsnabn:
I'm encountering an error with narrowing conversions, while trying to compile Urho3D under C++11 (with `URHO3D_C++11` build option). The error occurs in the Bullet code:
```
C:\Projects\Game Development\Engines\Urho3D\Source\ThirdParty\Bullet\src\LinearMath\btScalar.h:339:54: error: narrowing conversion of '+Inf' from 'double' to 'float' inside { }
 #define btAssign128(r0,r1,r2,r3) (__m128){r0,r1,r2,r3}
                                                      ^
C:\Projects\Game Development\Engines\Urho3D\Source\ThirdParty\Bullet\src\LinearMath\btVector3.cpp:442:21: note: in expansion of macro 'btAssign128'
     float4 dotmin = btAssign128( BT_INFINITY,  BT_INFINITY,  BT_INFINITY,  BT_INFINITY );
                     ^
```
That's the basic error (it is repeated a couple of times, in different places, sometimes with `-Inf` instead of `+Inf`).

My compiler is TDM-GCC 5.1.0, under Windows. This error cropped up this time with commit `9b78bbacb60ea3a37faa3001ee60cd87c3ec31f1`, but it has occurred to me earlier as well (cannot recall the commit hash).

I managed to work around it by changing the Bullet source code which throws the error (mentioned above in the log):
```
 #define btAssign128(r0,r1,r2,r3) (__m128){r0,r1,r2,r3}
```
to:
```
#define btAssign128(r0,r1,r2,r3) (__m128){static_cast<float>(r0),static_cast<float>(r1),static_cast<float>(r2),static_cast<float>(r3)}
```
I feel like this is a time bomb that's ticking away, though.

Any help would be appreciated. Please ask if you need more information.

**EDIT:** Adding `Wno-narrowing` and/or `Wno-error=narrowing` to `CMAKE_CXX_FLAGS` did not help with the issue.

cadaver:
Can you check how the BT_INFINITY gets defined on your compiler? I understood that the INFINITY constant (which should be used for win32 code path) should be a float.

hsnabn:
It's defined like so:
``
`#define BT_INFINITY INFINITY
```
But I have no idea what `INFINITY` refers to... maybe it's in the `math.h` header? In which case it's being defined:
```
#define INFINITY	__builtin_inf()
```
[...]

PR #1640, opened 2016-10-10
Fixed compile error about narrowing conversions.

hsnabn:
See #1637.

The compile error was apparently being caused by cases where `BT_INFINITY` was defined as a double, and the `btAssign128` macro was being used, causing a narrowing conversion error. This commit fixes it by casting the values passed to `btAssign128` to floats.


cadaver:
Used a different localized fix in the btVector3.cpp, but thanks for your work anyway!

PR #1649, opened 2016-10-14
UIElement::CreateChildXML() returns child instead of bool

TheComet93:
As discussed on the forum: http://urho3d.prophpbb.com/topic2437.html

 + The return value of LoadChildXML() is now a pointer to the new child (or NULL if it fails)
 + The child is destroyed if loading fails, instead of keeping an empty child around


cadaver:
Merged.

PR #1652, opened 2016-10-15
Mouse visible by default

rokups:
~~Currently mouse is hidden by default. We can make mouse visible in `Application::Start()` but at this point engine has already centered mouse cursor. If game is started in a windowed mode it artificially changes mouse cursor position to the center of a window which is undesired behavior. With mouse visible by default this does not happen. Games that do not require mouse to be visible or need mouse centered can explicitly do this in `Application::Start()` method.~~

New engine parameters:

* MouseVisible: controls initial OS cursor visibility.
* MousseMode: controls initial OS cursor mode.

Use of engine parameters allows to avoid undesired mouse cursor centering in windowed applications that require visible cursor.

Notable changes:

* `Engine::HasParameter()` is no longer static.
* `Engine::GetParameter()` is no longer static.

These methods were used only in `Engine::Initialize()` method so impact is minimal.

Setting mouse parameters before window initialization did nothing and setting them after window initialization centered mouse anyway. The only sensible way was to set these parameters upon `Input` subsystem initialization when window is available. This conveniently happens on first `E_SCREENMODE` event. I had to store parsed parameters in `Engine::engineParameters_` member to allow `Input` subsystem to access parameters at correct time.

### DIFF
--- a/Docs/Reference.dox
+++ b/Docs/Reference.dox
@@ -224,6 +224,8 @@ The full list of supported parameters, their datatypes and default values:
 - TouchEmulation (bool) %Touch emulation on desktop platform. Default false.
 - ShaderCacheDir (string) Shader binary cache directory for Direct3D. Default "urho3d/shadercache" within the user's application preferences directory.
 - PackageCacheDir (string) Package cache directory for Network subsystem. Not specified by default.
+- MouseVisible (bool) Visible operating system mouse cursor. Default false.
+- MouseMode (MouseMode) Operating system mouse cursor mode. Default MM_ABSOLUTE.
 
 \section MainLoop_Frame Main loop iteration
 
--- a/Source/Urho3D/Engine/Engine.cpp
+++ b/Source/Urho3D/Engine/Engine.cpp
@@ -160,8 +160,10 @@ bool Engine::Initialize(const VariantMap& parameters)
 
     URHO3D_PROFILE(InitEngine);
 
+    engineParameters_ = parameters;
+
     // Set headless mode
-    headless_ = GetParameter(parameters, "Headless", false).GetBool();
+    headless_ = GetParameter("Headless", false).GetBool();
 
     // Register the rest of the subsystems
     if (!headless_)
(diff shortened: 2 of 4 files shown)


weitjong:
IMHO, the default cannot be changed without also changing all the documentation regarding this flag. Also, note that not all the platforms that Urho supports have OS mouse, take Raspberry Pi for example.

rokups:
What effect would this change have on raspberry pi?

weitjong:
The Input::SetMouseVisible() is a no-ops for RPI, it expects the visible flag to be always false. Having said that, I have no idea what impact it will have on RPI when you initialize the flag to true by default without looking at the code in detail. It may still work without any negative effect.

rokups:
Hmm, either way it sounds like no good. Mouse could be made hidden by default on operating systems not providing hardware cursor, however it would make engine behave differently in different contexts. That is bad.

Anyhow [this](https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/Input/Input.cpp#L2352) is the culprit of the problem and there is no workaround for it. Maybe better fix would be adding new engine parameter that sets mouse input mode to `MM_FREE` before rendering starts? In that case unwanted centering would be avoided and mouse can be made visible later if needed. Or while at it i could add option to control mouse visibility as well. Think that is a good as a solution?

weitjong:
The Input subsystem was easier to understood back then before the time we have those changes for Emscripten. At that time I think the mouse visible flag can be easily flipped by overriding Application::Setup() method, i.e. we could call ```Input::SetMouseVisible(true)``` even when the Input subsystem has not been initialized yet. This is the last commit when this was still possible. https://github.com/urho3d/Urho3D/blob/54565005e2719f1e2d94f43716b8a3f451d72761/Source/Urho3D/Input/Input.cpp#L540

Why I point this out is because I also think there is something deeply wrong with what we have currently but I am not sure how best to fix it at the moment. Personally I feel what you proposed is a bit clunky although it should fix your issue. Ideally when done correctly, we should just use the Application::Start() method or script's Start() function,

cadaver:
It's not much less clunky, but less changes overall would be needed by having something like "initialMouseMode" in Input class which the Engine could set, like it sets other parameters to other subsystems. It's preferable that subsystems don't depend on Engine if possible.

weitjong:
I have created a patch to attempt for a quick fix for this issue. I have also removed a Linux-specific workaround for fixing initial input focus problem. It does not appear to be needed anymore with our current code base. So far so good in my brief tests on Linux platform. With this patch, the OS mouse visible can be set just in the Application::Start() method or Script Start() function without the unwanted mouse centering. [patch.txt](https://github.com/urho3d/Urho3D/files/531644/patch.txt)

In the patch I have removed the offending mouse centering logic from the screen mode change event handler. Going back through the change history one can see the logic was simpler at first and it was still compatible with the Input::Update(). But that logic in the Input::Update() is so convoluted now, I figure it would be best to remove the centering logic call from the event handler than to replicate all its complex condition check before calling Input::CenterMousePosition(). In effect deferring the call to Input::Update() if and only if the logic there requires it to be centered.

rokups:
Sounds great to me. User can always center cursor manually should it be needed.

weitjong:
If there are no objections then I will commit it.

cadaver:
No objections, appears to work fine on Windows. On Windows I observe the mouse to be limited to the window area if it was outside on startup (e.g. when starting editor)

weitjong:
This is strange, I don't observe that in Linux.

cadaver:
I can test some more, supposedly it could be from hiding the mouse initially and if on Windows SDL does the "capture to window" at the same time.

PR #1653, opened 2016-10-15
Utility functions

rokups:
First of all - im sorry, i know it should be one PR per feature, but i did not want to flood project with 5-line PRs. If anything is not agreeable i will overwrite these commits based on discussion for a clean merge in the end.

For my own needs i added bunch of utility functions that ease life. 

`UIElement::FindChildrenWithTag()` and `Node::GetChildren*()` functions that return vectors with results. This is very convenient as it allows to use said functions in modern c++ loops like `for (auto child : node->GetChildren())`. It also is not a performance issue if modern compiler is used since compilers rewrite calls that return objects by value to calls that receive variable by reference for storing return value.

`IntVector2::Length()` - self explanatory.

`Node::SetTransform(const Matrix3x4& matrix)` - a convenience function expanding to `SetTransform(matrix.Translation(), matrix.Rotation(), matrix.Scale());`.

~~Log writing functions `Log::Write(int level, const char* messageFormat, ...)` and `Log::WriteArgs(int level, const char* messageFormat, va_list ap)` - write formatted log messages using `String::AppendWithFormatArgs()`. Now debug message logging is way more convenient.~~

### DIFF
--- a/Source/Urho3D/AngelScript/APITemplates.h
+++ b/Source/Urho3D/AngelScript/APITemplates.h
@@ -668,6 +668,7 @@ template <class T> void RegisterNode(asIScriptEngine* engine, const char* classN
     engine->RegisterObjectMethod(className, "void SetTransform(const Vector3&in, const Quaternion&in)", asMETHODPR(T, SetTransform, (const Vector3&, const Quaternion&), void), asCALL_THISCALL);
     engine->RegisterObjectMethod(className, "void SetTransform(const Vector3&in, const Quaternion&in, const Vector3&in)", asMETHODPR(T, SetTransform, (const Vector3&, const Quaternion&, const Vector3&), void), asCALL_THISCALL);
     engine->RegisterObjectMethod(className, "void SetTransform(const Vector3&in, const Quaternion&in, float)", asMETHODPR(T, SetTransform, (const Vector3&, const Quaternion&, float), void), asCALL_THISCALL);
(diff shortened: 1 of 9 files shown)


eugeneko:
Unsure about `Log::Write(int level, const char* messageFormat, ...)`
If I now have code like Log::Write(0, "100%"), what will happen after your changes?

rokups:
Good catch. It goes through formatting function and discards last %. In case of % in the middle of string it would result in a warning. Best i can think of right now is renaming functiosn to `Log::WriteFormat()`. I would be happy to do it differently if anyone has better idea.

weitjong:
It looks like a duplicate of what URHO3D_LOGDEBUGF(), URHO3D_LOGINFOF(), etc, macros currently do.

rokups:
These macros totally slipped past my radar. Removed that commit as these functions are not needed.

cadaver:
Good work, can you add script bindings where appropriate?

rokups:
Added `IntVector2` stuff. I could not find any of `GetChildren*()` or `Node::SetTransform()` functions exposed. So maybe they should not be exposed?

cadaver:
You should find overloads of Node::SetTransform being exposed in APITemplates.h & LuaScript's Node.pkg.

rokups:
Cleaned up commits, squashed appropriate ones, rebased on to master, exposed `Node::SetTransform()` to scripts. `GetChildren*()` calls already were exposed to scripts in a manner where return value is array with children so nothing to do there. Please review.

cadaver:
Looks OK. I found one tab in the Lua pkg file, will change to space before merging.

PR #1661, opened 2016-10-21
Add an emplace back method to the vector class when C++11 is enabled.

iSLC:
However, following the style of C++17 by returning a reference to the created instance.

### DIFF
--- a/Source/Urho3D/Container/Vector.h
+++ b/Source/Urho3D/Container/Vector.h
@@ -218,6 +218,7 @@ public:
     void Push(const Vector<T>& vector) { InsertElements(size_, vector.Begin(), vector.End()); }
 
 #if URHO3D_CXX11
+    /// Create another element at the end.
     template <typename... Args> T & EmplaceBack(Args&&... args)
     {
         if (size_ < capacity_)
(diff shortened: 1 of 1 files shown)


eugeneko:
I dislike these changes.
First, you sometimes call constructor twice (inside Resize(n+1) and right below), this is errorneous and dangerous.
Second, your EmplaceBack discards vector growth logic, so it will perform reallocation on each insertion if don't have enough capacity, this is confusting and may cause misuse of container and performace hit.
Third, empalace back looks pretty useless until the whole container support move-semantic and storing non-copyable-but-movable objects. If you want to omit one copy-op, it would be better to add move-push.

iSLC:
Ah, I see what you mean now. I should've used `Reserve` instead. Should I go ahead and implement move semantics? Although there's one thing I'm not sure how to approach. When the engine is built as a DLL, and it allocates memory which I can move back in my application and application then needs to free the memory. From what I could gather, some issues could appear. Unless the vector also stores the initial allocator alongside the buffer.

iSLC:
Actually hold on. `Reserve` wouldn't work either. The code to deal with buffer growth was moved in the private `Resize` which also would construct the instance upon resize. Yeah, looking back at it now, it might take a little more work to achieve this. It would've worked before but I see that the vector went through some refactoring. Might as well just remove this pull request.

eugeneko:
Actually, my second phrase was wrong because I mixed up Resize and Reserve.
Urho containers are not designed as STL ones and most probably they are not going to be re-designed. So it will be hard to implament some guarantees or features from STL.

Issue #1664, opened 2016-10-27
Bottleneck in hierarchy ListView


eugeneko:
I've noticed that adding new nodes in Editor costs much more than I expected.
E.g. when I have several thousands of nodes, adding new node takes about 15ms and this time grows with scene complexity. This is quite important for procedurally spawned nodes (even temporary), because spawning several hundreds of nodes take more than second.
The biggest part of time is spent inside _nodeadded_ event.
I suppose it is caused by rebuilding and reallocating of items list even if hierarchy is hidden or collapsed.
I am trying to somehow optimise it... E.g. create tree data structure instead of plain vector inside HierarchyContainer. Or maybe just disable all these updates if hierarchy is hidden.
Does anybody have good ideas?

MonkeyFirst:
>Does anybody have good ideas?
Catch only one nodeadded event per second or two and update hierarchy ListView, other incoming events ignore.

cadaver:
The UpdateHierarchyItem() function in the editor is quite generic, it could optimize for new node added case. For one thing it looks through the existing items in the list to check if it's already there (and find list index in that case), for a new node we know that's false. Furthermore it's doing the loop in script and not C++.

eugeneko:
@cadaver Then I'll try to optimize it locally, thank you.

@MonkeyFirst It's not so easy because these updates are accumulative and we can't just throttle 'em.

eugeneko:
I tried. It looks like the main problem is in ListView and single vector for whole hierarchy.
I want to at least try to write some other control like ListView but without such indexes.
Is is possible to avoid global indexes at all and use only UIElement ptrs in all ops?

And what's the meaning of overlay checkboxes inside HierarchyContainer?

eugeneko:
I optimized Editor work with invisible temporary nodes. Now it has O(1) complexity instead of O(N), so my ~1s lags just entirely disappeared.
However, it is just a half of work. Spawinig objects throught _Run Script..._ has O(N^2) complexity and this is.. bad.
I suppose that turning off scene updated during _Run Script..._ and `UpdateHierarchyItem(editorScene, true)` call after will significantly improve performance for heavy cases. However, it will probably slow down lightweight scripts.
@cadaver What do you think about proposed solution and adding some flag to Settings or Preferences dialog that will control this behavior?

cadaver:
[...]

PR #1669, opened 2016-10-29
Optimize Editor work with temporary objects

eugeneko:
Related to #1664 
I'ven't managed to fix bottleneck in Hierarchy View. However, I've optimized temorary nodes and how Editor deal with them. Since procedural nodes like terrain pathces are temporary, these changes are pretty good workaround for my issue.
Please, review changes in **EditorHierarchyWindow.as** for possible unwanted side effects.


cadaver:
Shouldn't be anything wrong with this. Feel free to merge to master.

cadaver:
One thing, did you check toggling the "show temporary" checbox in the editor at runtime, does it work as before? If yes then there shouldn't be any other possibilities for trouble.

eugeneko:
Seemed to work OK for me too.
BTW I added the same checks to HandleNodeNameChanged, HandleNodeEnabledChanged and HandleComponentEnabledChanged. It shouldn't be a problem too. Then I'll build it and merge.

cadaver:
Yes, makes sense.

PR #1672, opened 2016-11-03
Added SetDrawRect and SetTextureRect in StaticSprite2D

gogoprog:
This PR adds function to override the draw and texture rectangles of a StaticSprite2D


cadaver:
Thanks for the PR. I notice AngelScript bindings missing, and there's no attributes for the overrides, so they wouldn't be serialized. There's also no mechanism to "stop" the overrides. 

If you do all of those, then there shouldn't be any problem to merge this PR. If you don't add the attributes, then I suppose this would still be useful for programmatic use.

gogoprog:
Of course I'll do it. Thanks for the feedback

gogoprog:
I think I should be fine now.

cadaver:
Thanks! I'll get to checking this in very near future.

cadaver:
Merged. Needed some additional bindings and possibility to edit Rect attributes. I also set the default values to zero so that it wouldn't show infinity (undefined rect) in the editor when the sprite had not rendered itself yet.

Issue #1646, opened 2016-10-14
pbr problems,how to solve env-map edge line like unity or unreal


xiongyaya:
![qq 20161014231916](https://cloud.githubusercontent.com/assets/13036261/19392909/516bd74a-9265-11e6-8a9a-e39a56177784.png)
![qq 20161014232118](https://cloud.githubusercontent.com/assets/13036261/19392910/52bfe9e2-9265-11e6-80d8-9fcb890b7715.png)
cubemap cut-line is obvous on the fronter photo in pbr demo, unity or unreal  have a smooth surface even with a full-roughness material. how to improve this like the two engines?

xiongyaya:
the second parameter has some error, metallic should be 1
![qq 20161014233349](https://cloud.githubusercontent.com/assets/13036261/19393304/b8f49e64-9266-11e6-87ab-f8311ab62445.png)

cadaver:
I don't have a solution, but it's fairly clear where this comes from: it's sampling the roughest mip level of each environment cubemap face, however in that case (at least with the default sampling operation) it doesn't blend across the face edges. If all faces of the cubemap at the roughest mip had same color, the problem wouldn't show.

dragonCASTjosh:
I have been planning to improve the mip level sampling, but the seams themselves are less of an implementation issue and more because the cubemap is not that well filtered, in addition to this we really should be filtering cubemap in engine for best quality and usability but that is not set up and would be a large addition to the engine.

xiongyaya:
good aticle, I'm interesting with pbr and will try to find some solutions on the internet.

dragonCASTjosh:
The error mostly lies in the fact that seamless cubemaps are not supported out the box in D3D9 and is possibly not enabled in GL so we will have to correct it manually

xiongyaya:
thank you for your explains， filtered-method results this error,right? I find another good pbr filtered example at
https://github.com/derkreature/IBLBaker
it do the cube-filter with shader, I do't kown whether his method solve this problem or not and will have a test.

dragonCASTjosh:
Baking the IBL in engine will make the results less apparent but the main causes of the issue is DX9 itself thats there the first link i posted comes in as that is a work around to that issue

dragonCASTjosh:
@xiongyaya if you are running in DirectX feel free to try my latest commit on my personal branch

dragonCASTjosh:
Can someone confirm these issues are fixed

[...]

PR #1673, opened 2016-11-05
Area lighting and Quality improvements to IBL

dragonCASTjosh:
This commit includes the following:

-Sphere Area Lighting
-Tube Area Lighting
-Mip selection based on epics approach for higher quality
-Physically accurate attenuation
-Improve Cubemap seams on d3d9 and opengl, as mentioned in #1646 
-Update scene to demonstrate area lighting
-Re-import all textures to fix normal map artifacts

Screenshots:
screenshot_sat_nov__5_05_52_15_2016
screenshot_sat_nov__5_05_52_21_2016
screenshot_sat_nov__5_05_52_35_2016
screenshot_sat_nov__5_05_53_02_2016
screenshot_sat_nov__5_05_53_22_2016
screenshot_sat_nov__5_05_56_13_2016


eugeneko:
There are some tabs and probably garbage lines in Batch.cpp and Light.cpp..

dragonCASTjosh:
ill take a look, although its a bit odd the only some files have the issue.

dragonCASTjosh:
hopefully i didnt miss anything else

cadaver:
I'll test in detail hopefully soon. Meanwhile, the Light's "radius" & "length" only have an effect when it's an area light, right? Please explain this in the setters in Light.h, because otherwise looking at the Light class is going to get confusing for people who don't know how it's supposed to work. As there's the "range" parameter already from before.

dragonCASTjosh:
It's currently set up so that if the radius is above 0 then it will run area lights, this is handled by an if in the shaders. I understand this has a slight overhead but it's nowhere close to the area lighting code. Epic also take this same approach in their shaders. At a later date I'd like to understand the define system and set it to only run on a define the same way as shadows but i was unable to work it out.

As for what the values mean. The radius is the size of the light source and the length is the length of the light source. The existing range parameter is used for the range the light effects, in other words, the attenuation.

Hopefully that makes sense.

cadaver:
Yes, makes sense, but assume a user hasn't read the relevant forum threads + discussion of this PR, so that same info should be either in the class API doxygen comments, or in an added documentation section :)

dragonCASTjosh:
That makes sense im just not good with writing learning resources and am not even sure how to do docs

cadaver:
The least work will be to cram all the most vital information in the one-liner header comments.

dragonCASTjosh:
hopefully that covers what you was asking for

cadaver:
Yeah that should part be good now. 

Since the area light calculation is kept inside the PBR paths it's acceptable to me to do the branching in pixel shader, if you have personally tested the performance of ordinary lights under PBR to not suffer much. Though using a shader compile flag would still be preferable.

As isAreaLight_ is not actually used, it should not be defined as an attribute, since that's unnecessary operations being done in serialization for every lights, and it increases the memory payload of the Light component. Easiest is to just remove. You could still have IsAreaLight() which checks for the radius / length parameter values.

Also, in IBL shader code, it would be preferable to fix spelling mistake + function name not following convention of the rest: GetMipFromRougness() -> Roughness and fix_cube_lookup -> FixCubeLookup.

dragonCASTjosh:
Should be good now.

cadaver:
Thanks. Think I'll perform additional cleanup before merge, because I don't think there should be commented out lines in the engine classes (doesn't inspire confidence), or unused variables. But that should be a minor change, you don't have to touch it.

dragonCASTjosh:
Any update on when it will be merged :) im exited to get started on more goodies

cadaver:
Sorry, I've been away a bit due to extreme finish line stretch in http://www.nanowrimo.org but I hope it should be imminently soon. Will also need to check how much the repo will grow.

cadaver:
Ok, back. Checked that the Autoload/LargeData folder is now 700MB, I think that's a *bit* too much considering the old one is about 40MB. Due to the extremely large .png's the demo also takes a long time to start up. Or did I make a mistake in merging? Got a shader compile error (complaining of finalColor) on non-deferred PBR, OpenGL 3. The engine code changes themselves are looking fine.

dragonCASTjosh:
i felt the need to change the textues as the old one has a lot of artifacts from being exported as 8-bit jpg files. most of the textures are to the same resolution as a modern game outside of the 1 4k textures because i could see alot of pixels in the 2k version. If you are unable to merge the textures id recomend making the samples as a sepereate repo or download posible. As for the shader error im not sure what would cause that nothing in the None-PBR shaders should of changed.

OldSnake22:
You can just use DDS DXT1 format for albedo and PBR maps and for normal maps just save them in DDS DXT5 'XGXR' format. This will remove normal maps artifacts and you will save a huge file size amount. This is the workflow that i am using in my project and it is working fine.

Infos for xGxR format conversion here: https://urho3d.github.io/documentation/1.4/_materials.html

cadaver:
The shader error can be a result of the minor changes that were applied for soft particle support, can recheck that.

cadaver:
I could merge this without the texture changes first. If you want to convert them to DXT for smaller size you could do that later in a separate PR. Ok?

dragonCASTjosh:
that sounds fair

cadaver:
Merged.

1vanK:
Lighting looks different with different distances (I tested DX9 version only)

1vanK:
same problem with GL3

dragonCASTjosh:
@1vanK can you make the floor surface of that area to use the roughness 0.5 material it will make it easier to see whats happening

1vanK:
bug

1vanK:
hm, I removed "cast shadows" from sunlight and problem was solved xD

dragonCASTjosh:
that seams very odd as it only affects the tube light and it uses the same shadow system as every other light

PR #1675, opened 2016-11-05
Added namespace prefix for various calls to GetVariantType

matanshukry:
Various calls to GetVariantType are missing a using namespace.

As such, when a project doesn't have the line ```using namespace Urho3D``` (such as my own ^^), a compile error will be generated regarding the use of GetVariantType.

This fixes that.


cadaver:
Thanks.

PR #1676, opened 2016-11-06
Solid debug geometry for NavArea

1vanK:


Source/Urho3D/Graphics/DebugRenderer.cpp
```diff
@@ -505,7 +541,7 @@ void DebugRenderer::Render()
     graphics->SetBlendMode(lineAntiAlias_ ? BLEND_ALPHA : BLEND_REPLACE);
     graphics->SetColorWrite(true);
     graphics->SetCullMode(CULL_NONE);
-    graphics->SetDepthWrite(true);
```
eugeneko: This change looks... Widespread.
Does it have noticeable side-effects?

1vanK: As far as I understand, after it is rendered another viewport or UI. And every set own settings

MonkeyFirst: mb similar "fill" for navMesh ? lines on floor looks very simple and do not so solid?

eugeneko: Since this is DebugRenderer-wise change, what noticeable changes in rendering of debug geometry could appear? For example..
If depth write is disabled, degug geometry will be rendered in draw order without Z-test with another debug geometry. Am I right? Is it ok?

cadaver: If there's a lot of overlapping lines, it's true the depth writing is a rather important part of debug rendering. Some applications will choose to render debug without depth *testing* in which case it doesn't matter, but it's best to not assume that. Probably the ideal would be to disable depth write during rendering of the filled geometry, and enable otherwise.

eugeneko: > If there's a lot of overlapping lines, it's true the depth writing is a rather important part of debug rendering.

I see AddTriangle method. So somebody may use DebugRenderer for drawing some solid stuff. So it may be important to draw such solid triangles with propper depth sorting.

cadaver:
Merged. I left depth write for lines and turn it off for triangles.

PR #1683, opened 2016-11-12
Remove unnecessary copy in random access iterator.

iSLC:
The temporary copy should not be necessary when there's only one member which already supports post increment/decrement operations.


eugeneko:
Although new version is shorter, old version was almost canonical implementation of post++-- ops and easier for understanding (for me).
Unsure what is better since you didn't _removed_ unnecessary copy, but just moved it after `return`.

Just wondering: how do you faced the necessity to change this place of code?

PR #1686, opened 2016-11-14
Added a check before deleting stuff in hashmap

matanshukry:
When a hashmap is being declared and initialized (```m_map()``` in constructor), but not actually used in any way, an NPE will be thrown when closing the app.

This is due to the last line in the destructor, ```delete[] ptrs_;```, which deletes a NULL variable.

Since all other calls inside the destructor are relaying on some data to exists,
and we're checking ptrs_ anyway (assuming the PR is good, that is), there's no reason
not to wrap the other calls as well; At least IMO, correct me if I'm wrong.

### DIFF
--- a/Source/Urho3D/Container/HashMap.h
+++ b/Source/Urho3D/Container/HashMap.h
@@ -253,10 +253,13 @@ public:
     /// Destruct.
     ~HashMap()
     {
-        Clear();
-        FreeNode(Tail());
-        AllocatorUninitialize(allocator_);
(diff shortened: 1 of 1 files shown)


iSLC:
Doesn't `delete` already check for null?

eugeneko:
> When a hashmap is being declared and initialized (m_map() in constructor), but not actually used in any way, an NPE will be thrown when closing the app.

I cannot reproduce this. Can you please provide more detailed steps?

PR #1696, opened 2016-11-25
Crash due to call of StaticSprite2D::SetSprite if param is a null pointer.

Y-way:
If mouse left button hit a sprite tile in 36_Urho2DTileMap sample, it make a crash on windows. Forgive me poor English.


eugeneko:
Wow... What does this bucket of code do?

weitjong:
Thanks for your PR, however, you need probably reapply your changes on top of a recent commit before it will get reviewed. As it is now it even contains code from a commit that has been reverted since (you can blame me for that). Also, you may want to consider to leave out the refactoring on the math class. For sure, shorter code does not necessary mean faster code, unless you have profiled your code to be faster, which I doubt.

Y-way:
I am so sorry!I am not good at Git.

Y-way:
I only want to fixed a bug which would make crash in sometime.But I did not sync my fork, so, ... So sorry! @weitjong @eugeneko

cadaver:
Cherrypicked the commit related to StaticSprite2D.

PR #1702, opened 2016-12-04
Material preview model

gleblebedev:
A special Urho-themed model for material preview
2016-12-04 23_00_46-urho3d sample


cadaver:
It's very nice looking and not particularly large in filesize.
Is it wholly your own work so that there isn't any copyright / license problems?

gleblebedev:
Thanks! I made it from a scratch so there should not be any license issues. The font used for Urho3D text at the base of the object is Impact. I hope it won't cause any issues. If you think it may - I can remove the text.

gleblebedev:
By the way I have 1 more thing I want to add to content of Urho3D: https://github.com/urho3d/Urho3D/commit/43b579feb6a006a71d8490ac0e8994f61d5611f2

These are MDL files I've generated that aren't trangle-based models but models made of lines
lineprimitives

PR #1705, opened 2016-12-07
Optimize shader code and loader 

Y-way:
Faster loading and less 1/2 memory used than befor for shader code.


eugeneko:
What is optimized? How does it work if I don't want to re-write my shaders?

PS: Actually I want to immediatelly close PR that has so many garbage commits.

Y-way:
Fast loading and less memory used for shader code.
BTW: how can i remove the grabage commits? @eugeneko

cadaver:
This is something that changes especially GLSL shaders quite fundamentally (as they have just main() now) and that should be documented in the .dox files. Considering that the Shader class loads the code just once and then it's reused for all subsequent variation compilations, and the shader compile itself should be the main time occupation, I wouldn't take this in lightly.

You should have a clean version of Urho's master branch (reset it if necessary) then apply your changes on top, then make a PR out of that state. It's not really in our scope to teach use of git, though.

@eugeneko If there's a strongly justified reason I'd have no problem with core contributors closing PR on their own. You're basically one of the rulers already, in that respect. :)

In this case it looks @Y-way is still learning use of git and I'd show some leniency, but there's no reason why the reasoning for the changes couldn't be described in the pull request description already. Having to "interrogate" them afterward from the author is less than ideal. Basically, just try to make our job as easy as possible to either take, or refuse the PR.

1vanK:
You need use tabs instead spaces

I like idea with COMPILEVS COMPILEPS for main() functions as part of the overall Ubershader approach without "internal magic", but I'm not sure, it is so important to break backward compatibility :)

eugeneko:
> Fast loading and less memory used for shader code.

Probably you don't have any measurements, so preformance/memory benefit is questionable.
Furthermore, I dislike an idea to break compatibility even if it give some noticeable benefit (I am sure it don't).

Y-way:
I test all samples on windows, everything is ok. In this case, it will not modify the Shader code original logic, just replace the main () and use macro condition judgment.

eugeneko:
Ok, just imagine that I have bucket of _my_ shaders and I know nothing about your changes.
What will happen to me?
I suppose that all my shaders will be silently and suddenly broken with some strange errors.

Y-way:
Yes,your right. I have considered that case,but I didn't realize the problem severity. I have a idea about how to avoid it.Define a new maco for shader code and make sure backcompatibility in c++ by version macro.What do you think?

Y-way:
Ok,I know how should I do.Thanks for your help and generosity.@cadaver

eugeneko:
Finally, it is not very good to expect that somebody will fix code in your PR:
Please, follow all codestyle guidelines from Urho documentation and try to keep things consistent even if it is not mentioned in codestyle.

Y-way:
OK.Thanks.

PR #1709, opened 2016-12-12
Nuklear

fire:
Initial work on Nuklear UI:

I took `https://github.com/rokups/Urho3D-nuklear-ui`'s work on Nuklear UI and moved it into the engine code.

Having problems with migrating to pure urho3d graphics pipeline though. There's some commented code that shows where the progress is. The Nuklear UI code uses OpenGL 3 and only works on OpenGL 3.

It is currently missing:

- [x] urho3d graphics pipeline integration
- [x] DirectX 11 integration.
- [x] fonts
- [x] maybe urho3d resource system for images

If someone could help me with the graphics pipeline. It'll unstick me. Please look at 074e7b0699fd158f63bc6bd6070877d3aa0005d9.

@yushli There's two samples here for Nuklear UI. One is the overview and the other is a node editor.


eugeneko:
It would be great to have second UI in Urho for Editor...
Probably separate project branch is better.

cadaver:
Will have to look at this a bit, at least the graphics integration shouldn't be anything impossible, and the shader looks very basic, so Urho's already used UI / basic shader should be enough.

yushli:
That it only works with opengl 3 concerns me a little bit. If it is used in game, it'd better support opengl es2 as well. I don't think Urho3D core members will rewrite the editor with this new UI anytime soon, so the integration of this new UI would be more used in game, either as game tool or actual game UI. Anyway, it will be nice to see it integrated into Urho3D.

cadaver:
I don't think we will ever merge something that bypasses Urho's rendering. It has to be made to work using the Graphics subsystem to ensure working on all API's.

fire:
@cadaver Got your reply, I appreciate if you could spent some time on 074e7b0699fd158f63bc6bd6070877d3aa0005d9 to use the api. It's around 100 lines but it goes deep into Urho3d internals. I don't know that much about urho3d internals.

The Nuklear PS looks like the Basic shader, but the VS is slightly different.

nk_draw_foreach is the loop construct in Nuklear. In Urho3d code there's a batch submit.

cadaver:
I hope to look at this when I have time. In the meanwhile you could look at how Urho's UI draws itself. The Graphics class should map quite straightforwardly to GL core profile (VBOs, shaders used always) type rendering.

cadaver:
Ok. Looked into this just a bit. Basically you would need to convert *all* of nuklear_sdl_urho3d_gl3.h to use Urho Graphics subsystem operations (eg. creating shaders, textures and rendering) and not attempt to do anything directly through OpenGL.

The usual applies: if you want to make sure this happens, do it yourself. That's the default assumption for PRs, that they're already in a ready state. I may eventually get inspired to do that myself, but please do not count on it. Of course anyone else is welcome too.

cadaver:
Ok, this is what I've so far. It appears to be rendering the geometry, but the textures are not set correctly (comment out the SetTexture to just use what was last bound)

https://gist.github.com/cadaver/eaf784926ea394351731e319e8c6cef9

Feel free to work forward from this, it possibly isn't much missing at this point.

yushli:
Nice! I will check this out.
But i cannot access gist from my place, which is blocked. Can you send it to my by email(liyusheng@gmail.com), or put it somewhere on github.com?( github.com is not blocked, but gist is blocked...)

cadaver:
@yushli: It is not user-usable and I don't recommend you trying it out. Wait for @fire to finalize the PR.

yushli:
OK. will wait then. Thank you for your effort. 
BTW, I heard your song, not expected that from a programmer... now I look forward to seeing your picture holding a gituar...

fire:
I'll investigate later tonight. Thank you for your help.

cadaver:
Ok. Rather stupid issue in the gist. Just put SetNumLevels before SetSize in nk_sdl_font_stash_end() and it should render with the textures correctly. Note the commented out sharedptr vector for remembering the created atlases; I got refcount asserts with it on and didn't investigate further.

fire:
Updated, but Urho3d still complains.

```

Detected memory leaks!
Dumping objects ->
E:\Mine\Urho3d\Urho3D\Source\Urho3D\Container\VectorBase.cpp(34) : {26303} normal block at 0x000001A1122FBC30, 8 bytes long.
 Data: <        > 00 00 00 00 00 00 00 00 
E:\Mine\Urho3d\Urho3D\Source\Urho3D\Container\RefCounted.cpp(33) : {25194} normal block at 0x000001A1122FC040, 8 bytes long.
 Data: <        > 00 00 00 00 01 00 00 00 
DataDeserializer.cpp(1) : {25193} normal block at 0x000001A1122ECAC0, 256 bytes long.
 Data: <0 @     @ /     > 30 B5 40 B3 F6 7F 00 00 40 C0 2F 12 A1 01 00 00 
E:\Mine\Urho3d\Urho3D\Source\Urho3D\Container\RefCounted.cpp(33) : {6339} normal block at 0x000001A10CD856A0, 8 bytes long.
 Data: <        > FF FF FF FF 01 00 00 00 
Object dump complete.
The program '[21072] 46_NuklearUIOverview_d.exe' has exited with code 0 (0x0).
```

fire:
46_nuklearuioverview_d_2016-12-20_17-08-41

What do you think of putting Nuklear in a namespace? It has a lot of bits. I like Urho3d::Nuklear.

cadaver:
About the namespace, for me wrapping inside Nuklear would be enough, instead of Urho3D::Nuklear. Practically you're just exposing direct access to a 3rd party library through a Urho subsystem. However this matter isn't super important to me at all.

fire:
@cadaver What should I work on next for this integration?

cadaver:
The resource integration is probably the most critical to follow. D3D9/11 shader should be easy to write and you don't even necessarily have to do it.

fire:
@cadaver Would a image file browser demonstrate the resource integration? I'm looking for an example.

cadaver:
Just loading even one image in a GUI layout is sufficient (as long as it's done totally through the resource system without hacks), doesn't have to be a browser.

fire:
How do I convert a texture2d to a raw pointer with n size and probably height and width info?

1vanK:
Texture2D::GetImage() ?

fire:
Loaded an image with the Urho3d image resource and began a start of a file manager. 

@cadaver How does it look?

fire:
47_nuklearuibrowser_d_2016-12-30_20-15-19

cadaver:
I'll have to study it in more detail, is it running GetImage()->GetData() every frame? That's not good for performance, since that's an unnecessary GPU->CPU transfer. But rather, it's OK to just use Image (no GPU representation at all) type resources, if all Nuklear needs is the image pixels in CPU memory.

Also font loading through resources (if it isn't there already) would be kind of necessary, as otherwise e.g. Android file operations are going to fail if they just assume to load fonts through a filesystem path, whereas Urho's resource system is able to handle files within the apk. Note that ResourceCache::GetFile() is already sufficient for that, it doesn't have to be an actual resource subclass.

cadaver:
I would suggest not having so many Nuklear samples, for example proper font loading should be demonstrated from the very first Nulkear sample, if it draws any text.

1vanK:
When I comment 
```
static void node_editor_init(struct node_editor* editor)
{
    ...
    //editor->show_grid = nk_true;
}
```
Artifacts is missing

1vanK:
Is in nuklear some option to round float coords to int like https://docs.unity3d.com/ScriptReference/Canvas-pixelPerfect.html ?

1vanK:
May be cause of noise is  vertex buffer is ulocked when used Draw call:
```
nk_sdl_render(enum nk_anti_aliasing AA, int max_vertex_buffer, int max_element_buffer)
{
    ...

            sdl.graphics->Draw(Urho3D::TRIANGLE_LIST, offset, cmd->elem_count, 0, 0, dev->vbuf->GetVertexCount());

            offset += cmd->elem_count;
        }
        nk_clear(&sdl.ctx);
    }

    dev->vbuf->Unlock();
    dev->ebuf->Unlock();

    sdl.graphics->SetScissorTest(false);
}
```

cadaver:
Thanks for fix. On OpenGL the buffer isn't actually locked / mapped, so old data would be used in that case. On D3D consequences could be something more severe.

1vanK:
I correctly understand that there is a limit of elements on screen?

fire:
Only have time to hobbyist code on weekends. Progress will be slow, but I have an idea of recreating a node graph editor.

fire:
Having a bit of problem with this code:

```

				struct nk_image logo = nk_image_ptr(logoTexture->GetImage()->GetData());
				logo.h = logoTexture->GetHeight();
				logo.w = logoTexture->GetWidth();
				nk_button_image_label(ctx, logo, "logo", NK_TEXT_CENTERED);
```

https://github.com/fire/Urho3D-1/blob/6e5c6b4cd2e3a234153ee09ccb8dea3775b4bafc/Source/Samples/46_NuklearUIOverview/NuklearSample.cpp#L238

The goal is to have image show up in the list, but the load fails.

fire:
@cadaver I'll change the includes later when I figure out why it breaks the build.
@cadaver Also can you help me with this?

```
static struct nk_image
icon_load(const char *filename)
{
    int x,y,n;
    GLuint tex;
    unsigned char *data = stbi_load(filename, &x, &y, &n, 0);
    if (!data) die("[SDL]: failed to load image: %s", filename);

    glGenTextures(1, &tex);
    glBindTexture(GL_TEXTURE_2D, tex);
    glTexParameterf(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR_MIPMAP_NEAREST);
    glTexParameterf(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR_MIPMAP_NEAREST);
    glTexParameterf(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_CLAMP_TO_EDGE);
    glTexParameterf(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE);
    glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA8, x, y, 0, GL_RGBA, GL_UNSIGNED_BYTE, data);
    glGenerateMipmap(GL_TEXTURE_2D);
    stbi_image_free(data);
    return nk_image_id((int)tex);
}
```

Logo in this function call is the output of the above function.  It converts   glGenTextures(1, &tex) 's tex to a int and returns is as a nk_image_id. In Urho3d this is https://github.com/urho3d/Urho3D/blob/ee054a1507cb3518c57d4ebc43cfd6dc93de9a27/Source/Urho3D/Graphics/GPUObject.cpp#L39 object._name. Is there any way to access that?
 
`nk_button_image_label(ctx, logo, "logo", NK_TEXT_CENTERED);`

cadaver:
When I fixed the font atlas code, I used a pointer to a Texture2D object as the image ID. I think you should be able to do the similar, with the idea that nuklear requests that texture to be bound when rendering. You'd just have to keep the created textures alive somewhere.

yushli:
Any updates on this feature?

fire:
Sorry, work is ramping up. I haven't spent much time on this work :(

cadaver:
I think this is the case where lack of script bindings can be accepted. It's OK to me as long as Nuklear is not mandatorily required, and it can't be the "default UI", but an optional addition.

1vanK:
My second attempt to win this library... It is absolutely ugly :( Default Urho's UI 1000 times better
Just look it: https://github.com/vurtun/nuklear/blob/master/demo/overview.c
No any logic here, just output window. I should write 60kb strange code to output window? It is absolutely unreadable. Ok, this windows has a lot controls, but in Urho I can use editor and save it to xml. But for nuklear it is impossible. If after a month I want to add a button, I just do not understand my own code. Maybe visual it looks nice? Yes. But it is just skin. No any reason, why default Urho's UI may not look the same.

fire:
Well the way Unreal Engine solved this problem was creating a graphical editor. Their code for creating ui code is just as annoying.

1vanK:
Ok, we write own code for rendering. We write own code for correct moving and reordering controls. We write own code to interpetate C functions and serialize/deserialize. Why we neet this lib?

rokups:
@1vanK this library solves some specific problems. You may not like it but it is useful for some. For example consider a tree of nodes in a scene. With normal UI you would have to hook some event and traverse a node graph and update UI displaying this tree. With nuklear/ImGui you traverse a node graph when UI is rendered and render required controls in-place. Result of that is **any** change to scene graph is immediately reflected in the UI. Immediate mode UIs are not best tool for all the jobs. Actually i dont even think nuklear support should be incorporated into urho3d itself, however this could be a very cool **external** project for urho. Just like blender exporter. @fire did a great job here cleaning up my quick and dirty opengl integration code and using engine APIs instead and i am very likely to use this code for my own purposes. Thing is usecases of this UI library are pretty specific and it will likely not be used in games much if at all.

rokups:
Hey @fire i incorporated your work to [Urho3D-nuklear-ui](https://github.com/rokups/Urho3D-nuklear-ui). You may want to use those cleanups as well. Thanks!

fire:
@rokups I'll look into when I have time. Currently busy. Sounds great though.

fire:
@rokups Are you familiar with how to load a image? I'm stuck.

rokups:
Got dx9 shader to work, but still stuck with what seems to be that 0.5px problem as everything is blurry.

```
#include "Uniforms.hlsl"
#include "Samplers.hlsl"
#include "Transform.hlsl"

void VS(float4 iPos : POSITION,
    float2 iTexCoord : TEXCOORD0,
    float4 iColor : COLOR0,
    out float4 oColor : COLOR0,
    out float2 oTexCoord : TEXCOORD0,
    out float4 oPos : OUTPOSITION)
{
    oPos = mul(iPos, cViewProj);

    oColor = iColor;
    oTexCoord = iTexCoord;
}

void PS(float4 iColor : COLOR0,
        float2 iTexCoord : TEXCOORD0,
        out float4 oColor : OUTCOLOR0)
{
    oColor = iColor * Sample2D(DiffMap, iTexCoord);
}
```

fire:
@rokups updated my fork. Still waiting for the build results. 

Also updated the logo image to be a square one.

slapin:
Thanks, looks so good! I hope this GUI will be easier for me to use than stock Urho GUI
if it is more predictable/easier to understand.

fire:
Updated to Nuklear 1.37.0.

fire:
@cadaver Can you code review? It would be helpful. Particularly the NuklearUI.cpp is fairly rough.

cadaver:
Like said in the Raycastvehicle PR, my free time is limited for the time being. Anyone else should also be able to review, but I will eventually get to it.

cadaver:
Ok. Had an initial very short look. Just a couple quick notes which should be easy to fix:

- Samples 45 & 46 are already taken, so you will have to renumber.
- For NuklearUI member variables, prefer underscore in the end instead of beginning, to follow the code convention.

cadaver:
The URHO3D_NUKLEAR build define also seems ineffectual in the end, since it controls only the inclusion of the thirdparty lib, which is header-only. The Nuklear samples and NuklearUI engine directory get compiled nevertheless if I understood right.

Probably would also need a brief documentation page in the Doxygen docs (Reference.dox) just so that people know there's a Nuklear UI option. Doesn't have to tell much more than how to enable the build option and point to the samples.

I don't have anything bad to say of the code itself. The FinalizeFonts() function seems inflexible after being used to being able to load any resources any time, but I suppose there isn't much of a choice (or that you can always choose to not use Nuklear..)

slapin:
Is AS binding available yet?

cadaver:
I don't think script binding is possible with a sensible amount of work, and I certainly won't be requiring it in this case, since the build inclusion is optional.

fire:
I'll address these items over the next few weekends.

cadaver:
Thanks, it's good to know the timescale.

fire:
This weekend I spent some time making a lobby screen mockup to implement in NuklearUI. Still feeling raw.

The goal is to have a demo that approximates a real gui. (Also I didn't feel like coding this weekend.)

This is not nuklear ui yet.

new mockup 4 copy

slapin:
What is new about this? I can't wait...

fire:
* Renumbered Samples 45 & 46
* Prefer underscore in the end instead of beginning, to follow the code convention.
* Made more effective URHO3D_NUKLEAR
* Wrote a very brief documentation
* Having trouble setting the font to anything else.

@rokups Any suggestions on how to set the font? It should be possible to set the font to any resource, but I wasn't able to get it working. It crashes.

```
void NuklearUI::FinalizeFonts() // This could have an argument of a filepath to a font.
{
	ResourceCache* cache = GetSubsystem<ResourceCache>();
	SharedPtr<File> defaultFont = cache->GetFile("Fonts/Anonymous Pro.ttf", true);
	unsigned int bufSize = defaultFont->GetSize();
	unsigned char* buf = new unsigned char[bufSize];
	defaultFont->Read((void*)buf, bufSize);
	
	nkDefaultFont_ = nk_font_atlas_add_from_memory(&atlas_, buf, 14, 0, NULL);

	const void* image;
	int w, h;
	image = nk_font_atlas_bake(&atlas_, &w, &h, NK_FONT_ATLAS_RGBA32);

	SharedPtr<Texture2D> fontTex(new Texture2D(graphics_->GetContext()));
	fontTex->SetNumLevels(1);
	fontTex->SetSize(w, h, Graphics::GetRGBAFormat());
	fontTex->SetData(0, 0, 0, w, h, image);

	// Remember the created texture for cleanup
	font_textures_.Push(fontTex);

	nk_font_atlas_end(&atlas_, nk_handle_ptr(fontTex), &null_texture_);

        nk_init_default(&nk_ctx_, &nkDefaultFont_->handle);
}
```

Lumak:
why not just add a setfont function, instead of hardcoding in FinalizeFonts() func.:

`
bool NuklearUI::SetFont(const Urho3D::String &fontFile, float fontSize)
{
    Urho3D::File file(context_, fontFile);
    bool success = false;

    if (file.IsOpen())
    {
        unsigned fileLen = file.Seek(M_MAX_UNSIGNED);

        if (fileLen != 0)
        {
            file.Seek(0);

            SharedArrayPtr<unsigned char> data(new unsigned char[fileLen]);

            if (file.Read(data, fileLen) == fileLen)
            {
                struct nk_font* nkFont = nk_font_atlas_add_from_memory(&atlas_, data.Get(), (nk_size)fileLen, fontSize, 0);
                nk_style_set_font(&nk_ctx_, &nkFont->handle);
                success = (nkFont != 0);
            }
        }

        file.Close();
    }

    return success;
}

`

Lumak:
got the font working.  48_NuklearUIOverview looks pretty sweet 👍 
Awesome work, fire. Thank you.

fire:
Have some code formatting changes, but added the SetFont function.

fire:
Haven't had time, but there's Unicode support including CJK (Chinese, Japanese, Korean) if anyone wants a task.

https://github.com/vurtun/nuklear/wiki/UTF-8-Support

xiliu98:
Compile failed on vs2017:
1>*********\Urho3D\Source\Urho3D\NuklearUI\NuklearUI.cpp(357): error C2668: “_wassert”: ambiguous call to overloaded function
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.15063.0\ucrt\assert.h(26): note: may be “void _wassert(const wchar_t *,const wchar_t *,unsigned int)”
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.15063.0\ucrt\assert.h(26): note: or    “void Nuklear::_wassert(const wchar_t *,const wchar_t *,unsigned int)”

Lumak:
I got the same error in vs2015. The only way I found to fix it was to remove Nuklear namespace, i.e.:
```
namespace Nuklear
{
//#include "../ThirdParty/Nuklear/nuklear.h"
}
#include "../ThirdParty/Nuklear/nuklear.h"
```
and remove all `Nuklear::` from code.

Lumak:
replaced Seek() with GetSize() in my previous post - just discovered GetSize() function is in the Deserializer class, and this whole time I was thinking it's weird that File class doesn't have that function...  More you know.

fire:
I'll review the changes this weekend.

fire:
Removed NK_ASSERT. It should not be erroring anymore and replaced Seek(M_MAX_UNSIGNED) with GetSize().

eugeneko:
TL;DR
May you give any ETA on this PR?

fire:
Please try this branch and tell me if it works.

eugeneko:
Tested it a bit. Works fine for DX9 and DX11.

These issues shall probably be re-directed to Nuklear itself.

1. Properties sometimes are broken.

Here I was just clicking on field with cursor, no more actions.
When one property is 'active' I can't edit others.
47_nuklearuinodeeditor_d-2017-06-18-23-42-32-452

2. Selection and cursor are the same

3. Property editor < > arrows pixel hunting is annoying. Active area shall be bigger.

fire:
Summer cleanup.

These are the patches for the code I currently have. 

I am currently working on Visual Script for Godot Engine. https://github.com/godotengine/godot/issues/18831 It is not as good as Unreal Engine but there's momentum.

[patches.zip](https://github.com/urho3d/Urho3D/files/2066442/patches.zip)

Issue #1252, opened 2016-03-17
Mouse scaling issue on retina display


NightBlaze:
There is an issue on Macbook with Retina. Actually cursor position is calculated wrong. Video http://www.youtube.com/watch?v=UiGhXOoWKfM

On commit 28fb9e7b88045df218098f3259473d660d9ce1a5 and before all is good (except retina support :) ). Video http://www.youtube.com/watch?v=_vgCS6It464

weitjong:
Thanks for reporting this. It is good to know that now we have retina support, thanks to SDL 2.0.4. The issue should be caused lack of mouse scaling when on retina display.

weitjong:
I don't actually have a retina display to test any of this. We have enabled the SDL_WINDOW_ALLOW_HIGHDPI on all Apple platforms now in our engine code. @NightBlaze, in your opinion, is it the mouse move too fast and need to be scaled down? Or the mouse movement feels just right and it should be the UI response to be scaled up?

NightBlaze:
@weitjong The mouse movement feels just right and it should be the UI response to be scaled up.

cadaver:
I don't have a retina display Mac either. So is the issue that SDL is giving us mouse coordinates that don't match the window's pixel size?

weitjong:
I haven't got time to look into this yet, but the release notes in WhatsNew.txt may give us some clues.

```
Mac OS X:
* Added support for retina displays:
  Create your window with the SDL_WINDOW_ALLOW_HIGHDPI flag, and then use SDL_GL_GetDrawableSize() to find the actual drawable size. You are responsible for scaling mouse and drawing coordinates appropriately.
```

Actually this has been made available since 2.0.1 but our engine has never utilize this flag before and I took the chance :smile:  while I was merging the 2.0.4. After setting the flag, I remember I have to replace all the SDL_GetWindowSize() call with SDL_DrawableSize(). I suppose this is not enough. Somewhere in our UI or Input subsystems we still need to let the code how to differentiate between screen coordinates and pixels.

weitjong:
I think we need to fix the UI subsystem to properly support the high DPI mode before enabling this by default. IMHO, just scaling the mouse movement does not really help much if the overall UI is too small to use.

cadaver:
Should be fixed now, so closing. Can be reopened if necessary.

PR #1725, opened 2016-12-23
Proposed fix for Issue #1252: mouse scaling on Retina displays

elissa-ross:
I've added some code to address the mouse position issues on Retina displays with HighDPI enabled (issue #1252). This fix works for our purposes but there may be other places where the issue comes up that I haven't encountered. 

Regarding the UI scaling, we find that with HighDPI enabled on OSX the UI should be scaled at about 1.5X. This is possible using the Scale() functionality.


weitjong:
Thanks for your PR. IMHO, the scaling ratio should not be computed at every mouse move event. Probably the ratio should be precalculated in the screen mode setting event. Please use the same code style for brace placement as ours. Also avoid adding project-specific settings into ```.gitignore``` file unless they are commonly used by other projects, or otherwise.our ```.gitignore``` will be hard to maintain.

eugeneko:
Some time ago I had code like `aspect=width/height`
It worked fine untill I minimized window. Then it exploded.
Is it possible to get zeros from these nice SDL getters?

cadaver:
Don't think I've personally seen zero size from SDL getters, but it could be possible in a situation where the window isn't visible yet, so I'd probably put a zero check before divide anyway.

Operating directly on SDL_GL_DrawableSize is a bit unsecure feeling in the case we're not running on GL. However Graphics->GetWidth() / Graphics->GetHeight() should already be returning the correct GPU backbuffer size, so these should be preferable to use.

elissa-ross:
Thanks for your comments, and sorry for the delay in my response. I have incorporated some of the minor suggestions above, which were very helpful -- thanks. However, it sounds like the fix may not be what you are looking for. As I mentioned, it works for our purposes, but please feel free to close the PR.

I will take a look at the screen mode setting event, however I'm probably not in the best position to modify this. I am happy to help with testing if anyone else is interesting in working on this though, as I am actively developing on a Retina display Mac.

cadaver:
I think it should be easy enough to adapt/expand as necessary. The preferable is if Input subsystem only hands out mouse coordinates and moves that have been scaled to the backbuffer. Only problem is that I at least will be doing every modification blind, with no possibility of testing on real retina HW.

PR #1729, opened 2016-12-26
Fix minor Terrain bugs.

henu:
- Missing include added.
- Terrain UV coordinates fixed.

The image below shows how UV coordinates currently go in the place where one Terrain object changes to another. This PR fixes it.

terrain fix


cadaver:
Confirmed the bug with a simple checkerboard texture. Thanks!

PR #1731, opened 2016-12-28
replaced Jack with Jill in 06_SkeletalAnimation sample

Lumak:


Lumak:
screenshot_wed_dec_28_11_15_31_2016

1vanK:
I think need add "anisotropic" for ground like
https://github.com/1vanK/Urho3DHabrahabr06/blob/master/MyData/Textures/Ground.xml

Lumak:
I like the submission like it is.

cadaver:
Would you mind if I'd merge this without the plane material? Using the existing stone floor seems to me to make the "black spot" problem less severe.

Lumak:
Nope, don't care as long as I don't have to come up with another texture that may or may not work.

cadaver:
Merged, thanks!

1vanK:
Is some problem



Also I suggest to change zone and sun colors:
```
zone.ambientColor = Color(0.4f, 0.4f, 0.4f);
light.color = Color(0.6f, 0.6f, 0.6f);
```

1vanK:
Ok, if you do not mind, I would also like to make model more large

cadaver:
By scaling? Fine by me.

1vanK:
What about this wariant? (scale = 2)

1vanK:
just seemed, that new model is much less than jack

Lumak:
Asian girls are "smaller" :smile:

cadaver:
Looks OK. I think it's enough if each sample makes sense in isolation. Though viewed as static models in editor Kachujin appears a bit larger than Jack. Jack should be about 2 world units high.

1vanK:
Ah, It is an optical illusion :) Jack walks, but gilr crouched  :)

PR #1733, opened 2016-12-29
06_SkeletalAnimatin: fix running out bounds, lighting correction

1vanK:


1vanK:
What about default walk animation?

1vanK:
It seems models fly over ground (both Walk and Run animations)

1vanK:
I reimport walk animation from https://www.mixamo.com/store/#/search?page=1&type=Character (Blender sucefully open it), fix animation curves, correct direction (+ remove localSpin node) and also fix material

cadaver:
Cool, will check out shortly!

cadaver:
It does look more visually pleasing now. The performance characteristics naturally change with less models. clb was using this for Emscripten perf-tests so the benchmark is no longer the same, but I don't think that needs to be the deciding factor. He can make his own custom torture version :)

1vanK:
> The performance characteristics naturally change with less models. 

models also has another count of polygons, so anyway  benchmark is no longer the same :)

cadaver:
Yeah, I think he was mostly interested in the animation + skin matrix CPU update part and optimizing it with SIMD, bone structure is similar to Jack I think, which would make the CPU hit per model stay about the same.

Lumak:
@1vanK , in the course of 24 hrs you made several commits to your branch, right? Yet, it only shows 1 commit on your PR.  How's that done? Is it some kind of git command?

1vanK:
Yes. For example I sent 3 commits to server, and after it I wont to merge 3 commits to one
```
git rebase -i HEAD~3
```
This will open text editor with this content:
```
pick 50b6eda Comment commint 1
pick 9d7dceb Comment commint 2
pick 4d7dceb Comment commint 3
```
In this file a replece all "pick" to "f" except the first, save and close text editor.
```
pick 50b6eda Comment commint 1
f 9d7dceb Comment commint 2
f 4d7dceb Comment commint 3
```
Local repo will be updated. In a result I have one commit with with firs comment ("Comment commint 1").

After it for sending changes to server need use --force option:
```
git push origin master --force
```

Lumak:
Cool, good to know. Thx.

PR #1734, opened 2016-12-30
Console auto complete

Enhex:
Console auto complete feature.

Didn't include binding for Lua.

### DIFF
--- a/Source/Samples/26_ConsoleInput/ConsoleInput.cpp
+++ b/Source/Samples/26_ConsoleInput/ConsoleInput.cpp
@@ -83,6 +83,12 @@ void ConsoleInput::Start()
     console->SetCommandInterpreter(GetTypeName());
     console->SetVisible(true);
     console->GetCloseButton()->SetVisible(false);
+    console->AddAutoComplete("help");
+    console->AddAutoComplete("eat");
+    console->AddAutoComplete("hide");
+    console->AddAutoComplete("wait");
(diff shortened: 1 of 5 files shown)


cadaver:
Great idea. However Urho itself is not yet mandatorily C++11 so please avoid using auto in engine code. Also it would be preferable to use Urho containers and not leak std containers to Urho headers. There's also a minor compile issue due to a class member being initialized in header, which should be easily fixable.

Enhex:
Working on fixing these issues.
BTW Urho doesn't provide an alternative for std::set, which is ordered (needed for searching results in order).
I'll just replace it with a sorted vector of unique elements, which should increase complexity and insertion removal cost, while reducing search cost.

darwikey:
Or you can use HashSet instead of std::set

Enhex:
HashSet isn't ordered

Enhex:
If you're going to manually order you're better off using vector for better search performance.

Since the vector is sorted search can be optimized with `std::lower_bound` and `std::lower_bound`, but I'm getting compiler errors that seems to be related to Urho3D::Vector's Iterator type.

cadaver:
Merged slightly changed. I got a crash when pressing up in sample 26 as the first thing, and since there were potential problems with iterator invalidation, I changed the system to use a position index. Feel free to make a new PR in case I messed something up.

PR #1735, opened 2016-12-30
Rename Resource::Load and Resource::Save filename overloads to Resour…

eugeneko:
…ce::LoadFile and Resource::SaveFile to prevent Resource::Save shadowing.

I noticed this annoying behaviour with my new Resource functions.
When some resource overrides Resource::Save(Serializer&), Resource::Save(String) became shadowed and inaccessible. So I renamed these functions in C++ API. Script APIs remain untouched because such functions are quite old and scripts don't have shadowing problems.


cadaver:
Should be no problem for you to merge this. I think those overloads were mostly made for scripting convenience, since it's a straightforward Urho File class access.

PR #1739, opened 2017-01-01
Avoid special chars in paths in MRI script passed to ar utility

koja:
I have run into issue with MRI script special characters when building inside my "projects/c++/" directory.

Seems like not only spaces are problematic and no escaping is available.
https://sourceware.org/binutils/docs/binutils/ar-scripts.html

> The syntax for the ar command language is straightforward:
...
>     comments are allowed; text after either of the characters `*' or `;' is ignored.
...
>     Whenever you use a list of names as part of the argument to an ar command, you can separate the individual names with either commas or blanks. Commas are shown in the explanations below, for clarity.
...
>     `+' is used as a line continuation character; if `+' appears at the end of a line, the text on the following line is considered part of the current command. 

In case I understand it correctly no space, '+', '*', ';' or ',' should be allowed in script passed to ar.


weitjong:
Thanks for your PR. Can you try whether this regex pattern also works for you: ```"[ +*;,]"```
The regex metacharacters loses their meaning in a character class, so there is no need to escape them there.

koja:
Right you are, that's more readable. Thanks.

Issue #1741, opened 2017-01-01
[Feature Request] Allow to define vector structs for editor for external components.


KonstantinTomashevich:
In Urho3DEditor, editing of structures such as animation states (in AnimatedModel) and billboards (in BillboardSet) implemented using special vector structs (See lines 1490-1579 in Scripts/Editor/AttributeEditor.as). This mechanism works fine for engine components, but it can't be used for external components (special game components, Unit for example). In my opinion, it will be good to move this structures definitions into AttributeInfos (for easy defining them in external components).

cadaver:
In general, you should not take the editor as some immutable piece of code; quite natural approach would be to just be running your customized editor with your customized attribute structs.

However you're right it would be preferable to do this at the component definition, if we find a good way to do it. Possibly it doesn't need to be anything else than a vector of strings that is optionally filled.

KonstantinTomashevich:
Yes, Now I'm trying to implement it in my fork.

cadaver:
With the AngelScript API functions GetObjectCategories() and GetObjectsByCategory(), defined in SceneAPI.cpp.

KonstantinTomashevich:
Thanks!

PR #1742, opened 2017-01-01
Allow define of variant vector structs attributes

KonstantinTomashevich:
See issue #1741 .
Now variant vector structs elements names can be defined from AttributeInfos. Now AnimatedModel, CrowdManager, SplinePath, BilboardSet and StaticModelGroup use this way to do it.

### DIFF
--- a/Source/Urho3D/AngelScript/CoreAPI.cpp
+++ b/Source/Urho3D/AngelScript/CoreAPI.cpp
@@ -824,6 +824,15 @@ static CScriptArray* AttributeInfoGetEnumNames(AttributeInfo* ptr)
     return VectorToArray<String>(enumNames, "Array<String>");
 }
 
+static CScriptArray* AttributeInfoGetVariantStructureElementsNames(AttributeInfo* ptr)
+{
+    Vector<String> variantStructureElementsNames;
+    const char** variantStructureElementsNamesPtrs = ptr->variantStructureElementsNames_;
+    while (variantStructureElementsNamesPtrs && *variantStructureElementsNamesPtrs)
+        variantStructureElementsNames.Push(*variantStructureElementsNamesPtrs++);
+    return VectorToArray<String>(variantStructureElementsNames, "Array<String>");
+}
+
 static void SendEvent(const String& eventType, VariantMap& eventData)
 {
     Object* sender = GetScriptContextEventListenerObject();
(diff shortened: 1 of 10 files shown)


cadaver:
Thanks!

I'll look in detail shortly. Meanwhile, there are some minor stylistic concerns, as well as misspelling of Attribute in the added function. Typically, we'd do a loop iterating an Urho container like this, using an unsigned counter variable to avoid warnings, and no extra spaces before (), and preincrement.

for (unsigned index = 0; index < attributes->Size(); ++index)

instead of 

for (int index = 0; index < attributes->Size (); index++)

KonstantinTomashevich:
Codestyle fixed.
And I added "***" in start of name of any first property because editor doesn't add any white spaces after end of structure (see BillboardSet for example).

cadaver:
No problem. I could check if a small spacer could be added.

eugeneko:
I wanted some kind of this for a long time, because hard-coded structs in Editor are ugly.
However, won't it be redundant to have this functionality in _Urho3D core_ while it is actially needed _only for editor_? It brings some API heaviness.

KonstantinTomashevich:
@eugeneko AttributeInfo has enumTypes_ var only for editor too and it doesn't lead to API heaviness for now. In my implementation structure elements names contained in pointer (char**) and will not require additional memory.

eugeneko:
Enums are used for human-readable and portable serialization attributes into XML and Json.
Stuctures are not. This is the difference.

We already have many things that are hardcoded in the Editor: non-continuously-edited attributes, bit fields, these vector structs, probably something else.

And I can't easily answer whether it is good that such non-functional stuff don't bloat attribute system or it is bad that such things can't be specified at code level.

PS: What about crowdManagerAreaCostVariables?

KonstantinTomashevich:
AreaCost in CrowdManager is not an attribute and because of it can't be handled this way.

KonstantinTomashevich:
Tested this feature in my game (which is downstream c++ project). All works successfully.
[Variant Vector Structure in external projectpng](https://postimg.org/image/kskxwhr2r/)

cadaver:
This sort of metadata of how a vector of variants is being used for data storage (by the engine code) could theoretically be useful to any deserialization or inspection utility, not just the editor. So I don't have a problem of it being here.

eugeneko:
Okay, go to plan B.
I have several small questions about this code.

1) What't the point of copying things in GetObjectsAttributeInfos?
2) Is it possible to rename all 'variantStructureElements' to something more brief, e.g. just 'vector'. I mean, like 'enum', but 'vector'.
3) I still don't understand why does *** needed in new code but not needed in old.
4*) It would be nice to have an ability to display any vector in Editor. Have you tried to implement it? I haven't managed to, shame on me.

KonstantinTomashevich:
1. To save them from any changes.
2. Yes, I think it will be good.
3. "***" was needed in old code too. For example try to create BillboardSet with 2 or more billboards in editor on 1.6 release (or see image below). There isn't any separator between variables of different billboards. "***" marks that this separator will be added.
[](https://postimg.org/image/xhy2io7o7/)
4. For now it can be implemented by variant vector structure with 1 element. Small example from my project:
```c++
// Register attribute.
URHO3D_MIXED_ACCESSOR_VARIANT_VECTOR_STRUCTURE_ATTRIBUTE ("Polygon Points",
 GetPolygonPointsAttribute, SetPolygonPointsAttribute, Urho3D::VariantVector,
 Urho3D::Variant::emptyVariantVector, polygonPointsStructureElementsNames, Urho3D::AM_DEFAULT);

// Structure elements names.
const char *polygonPointsStructureElementsNames [] =
{
    "Polygon Points Count",
    "***Point",
    0
};

// Getter.
Urho3D::VariantVector District::GetPolygonPointsAttribute () const
{
    Urho3D::VariantVector variantVector;
    variantVector.Push (polygonPoints_.Size ());
    for (int index = 0; index < polygonPoints_.Size (); index++)
        variantVector.Push (Urho3D::Variant (polygonPoints_.At (index)));
    return variantVector;
}

// Setter.
void District::SetPolygonPointsAttribute (const Urho3D::VariantVector &polygonPoints)
{
    polygonPoints_.Clear ();
    if (!polygonPoints.Empty ())
    {
        int requestedSize = polygonPoints.At (0).GetInt ();
        if (requestedSize > 0)
            for (int index = 0; index < requestedSize; index++)
            {
                if (index + 1 < polygonPoints.Size ())
                    polygonPoints_.Push (polygonPoints.At (index + 1).GetVector3 ());
                else
                    polygonPoints_.Push (Urho3D::Vector3 ());
            }
    }
}
```

eugeneko:
I still don't see how something can be changed if you write `return VectorToArray<AttributeInfo>(*attributes)`

Then, if @cadaver has no objections, I'll rename these long names.

KonstantinTomashevich:
1. Yes, maybe copying is excessive and it will be deleted. I understood now, thanks.

cadaver:
Changes / renames are fine to me.

PR #1743, opened 2017-01-02
Switched to new icon

Modanung:
Like this?


1vanK:
I think the smaller the image, the less there should be details. For example gradient looks like diagonal line for me

cadaver:
Looks fine to me, including the gradient.

PR #1745, opened 2017-01-03
log message colour in console displayed based on it's type

ArnisLielturks:


eugeneko:
_offtop_
Why every second contributor ignores codestyle?..

ArnisLielturks:
@eugeneko Thanks for pointing this out!

weitjong:
Is there any special reason why you remove the "hover" attribute in the base console text style? Note the previous highlighted style was based on it so the attributre got inherited. Probably you can use an array of styles indexed by log level instead of using if-else blocks.

ArnisLielturks:
@weitjong Can you take a look at the latest changes? Changed the code according to your suggestion. And relating to "hover" effect, it was removed by mistake since I used the styles from one of the projects that I'm working on.

ArnisLielturks:
Console appearance after these changes: http://imgur.com/a/Eb3AG

weitjong:
It looks ok by me.

cadaver:
Yes, the changes look rather clean/minimal now, and also screenshot is fine.

PR #1748, opened 2017-01-05
Input scaling between backbuffer / window

cadaver:
Scales mouse position and deltas if backbuffer (Graphics) has different size than SDL window. Inspired by work of Elissa Ross in PR #1725. Related to issue #1252, might resolve it, but not tested on actual Retina hardware.

For code simplicity, this now applies scale always, even if it's (1,1). Involves somewhat horrible int-float-int conversions. If anyone has time, please review so that I didn't commit too bad sins. I tested in D3D9 by intentionally creating half sized or double sized backbuffer, and it seemed to work right.


elissa-ross:
This is great! I just tested on my Retina display and it works as expected. Thank you for taking this up!

cadaver:
Thanks for testing!

cadaver:
Will merge as no especial further comments. We'll see afterward then :)

weitjong:
> Will merge as no especial further comments. We'll see afterward then :)

Submit a patch to make this happen sooner rather than later.

PR #1752, opened 2017-01-08
SDF supersampling

1vanK:
Only GL version currently. Supersampling can be disabled by comment define.
SDF rendering improved even without using supersampling (dynamic calculation of border width + smoothstep for outline). But outlines and shadows still dirty for far distance. Any notes?

### DIFF
--- a/bin/CoreData/Shaders/GLSL/Text.glsl
+++ b/bin/CoreData/Shaders/GLSL/Text.glsl
@@ -19,40 +19,85 @@ void VS()
     mat4 modelMatrix = iModelMatrix;
     vec3 worldPos = GetWorldPos(modelMatrix);
     gl_Position = GetClipPos(worldPos);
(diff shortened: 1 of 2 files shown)


1vanK:
Some artefacts are seen on borders of glyphs at a certain angle at a certain distance



Source it is incorrectly generated SDF fonts (its just cut):

1vanK:
Added DX vesrion

cadaver:
Cool. Will look in detail shortly.

cadaver:
Seems to work good on both GLSL & HLSL. Stroke effect in the sample 35 seems less noticeable / thinner with supersampling. But IMO can be merged as is if you don't have further changes in mind.

1vanK:
> But IMO can be merged as is if you don't have further changes in mind.

no ideas currently

PR #1755, opened 2017-01-11
WebGL Module.arguments not work on Urho3DPlayer.

abdllhbyrktr:
Hi,

When I attempt to pass arguments to Urho3DPlayer web sample, It fires unexpected application failure. So I changed and tested the related source file. Now it works.


cadaver:
Thanks for noticing the issue. It can perhaps be solved more elegantly by removing the earlier ifdef which prevents Web build from initially reading the arguments, and only making it so that Web build doesn't attempt to load CommandLine.txt from filesystem.

cadaver:
Check if it works now in the master branch.

abdllhbyrktr:
Yes it works, Thank you!

cadaver:
Thanks for checking. Will close this now.

PR #1760, opened 2017-01-14
Minor documentation expansion on use of std::bind() for events.

carnalis:


cadaver:
Can someone who has used std::bind for events comment? (or merge directly, if good)

carnalis:
I changed my events to this syntax and it seems to work (I'm using GCC 6.3.1) .. I made the PR because the former syntax (without placeholders) errored with incorrect number of arguments also on MinGW/GCC 6, but I am curious what others experience.

rokups:
Yes this is correct.

carnalis:
I think now it should look a bit less scary.

cadaver:
Thanks, merging would still add several merge commits to the history, but can do a squash merge.

cadaver:
Merged.

carnalis:
I see what happened, will avoid that in the future.

PR #1767, opened 2017-01-17
Paint selection and origins

MonkeyFirst:
Hi folks! 
I took feature from my fork for editor and tried to put it on new rails of new version of the engine. 
if this needed for editor, checkout this please

- PaintSelection tool on KEY_C (use origins)
- Origins for show and select nodes (empty, light), button in toolbar + KEY_SPACE for show/hide names

* currently range visibility of origins is 32 units, but may increased to 64 or 128...


cadaver:
Is there new key shortcuts? Does the editor documentation need updates? If not I think this could go in as is, if yes it shouldn't be a large work.

MonkeyFirst:
little question: the proc DeselectNode() was removed for some reason from new editor, but I bring it back. but it's seems not working properly now in new behavior. Is there exist another proc for deselecting node from EditorHierarchyWindow ?

cadaver:
I have no memory of that function.

MonkeyFirst:
Deselecting work in std.mode with Ctrl, so I'm change this key for both modes.
I think for now it's ready. Checkout.

cadaver:
Thanks. Will check soon.

cadaver:
Merged.

PR #1770, opened 2017-01-19
Remake 'Anonymous Pro.sdf' font

1vanK:


1vanK:
Some chars (w for example) slightly elevated. It is problem of "Anonymous Pro" font when font size is big.
(SDF font geterator used huge font size for more quality).

1vanK:
there are some problems.

Issue #1728, opened 2016-12-23
Update AngelScript


alex-rass-88:
AngelScript 2.31.2 released.
Fixed bug with saved byte code not being portable between 32bit and 64bit platforms.
And more...
http://angelcode.com/angelscript/changes.php

cadaver:
Yes, this is long overdue. It will bring some API changes and require going over the arrays / dictionarys again, but nothing in Urho API should change.

cadaver:
Being worked on in update-angelscript branch.

cadaver:
AngelScript branch (which updates to 2.31.2) merged.

PR #1771, opened 2017-01-19
Update AngelScript

cadaver:
Updates AngelScript to version 2.31.2. Customizations regarding platform support (e.g. Emscripten generic call convention) should be preserved. Whoever has time, please review. Addons (array & dictionary) were only fixed where necessary without taking completely new code from the SDK, for easier work of preserving customizations. Closes issue #1728


cadaver:
Will go in shortly, if no objections in the meanwhile.

PR #1772, opened 2017-01-19
Remake 'Anonymous Pro.sdf' font

1vanK:
Old version https://github.com/urho3d/Urho3D/pull/1770 had problems with some glyphs.

This is caused by too small image size

test1_0

So I recreate it with another font size (anyway user can scale SDF font to any size). It also solved problem with elevated glyphs.

But now Anonimous Pro used 2 pages instead 1.


1vanK:
I use Image::Resize() for downscale image
```
    int originalFontSize = 50;
    int scale = 16;
    FT_Set_Char_Size(face_, 0, originalFontSize * 64 * scale, 96, 96);
    ...
    FT_Glyph_To_Bitmap(&glyph, FT_RENDER_MODE_MONO, nullptr, true);
    ...
    int radius = 8;
    glyphManipulator.CalculateSDF(radius * scale);
    SharedPtr<Image> resultImage = ConvertToImage(glyphManipulator, Color::WHITE);
    resultImage->Resize(resultImage->GetWidth() / scale, resultImage->GetHeight() / scale);
    PackGlyph(resultImage);  
```

Could this be the cause of damage sdf ?
```
 bool Image::Resize(int width, int height)
{
    ...
    /// \todo Reducing image size does not sample all needed pixels
    ...
}
```

cadaver:
Image class is not a high-quality image processing package, an actual paint program is preferred. You could also possibly use the GPU for size changes (e.g. render to texture, then read back the pixels). To me it's fine if it takes more glyph pages, that way the multiple page support also gets exercised more.

1vanK:
> You could also possibly use the GPU for size changes (e.g. render to texture, then read back the pixels).

Hm, I also can calculate sdf on gpu by shaders for speedup.

1vanK:
Is it possible to force rendering only viewport with render-to-texture scene? 600 glyphs required 600 rendering

cadaver:
I think you should use the Graphics class directly for this kind of task, so you don't have to wait for a frame to complete.

1vanK:
I found Graphics::ResolveToTexture(), that's what I need?

cadaver:
No, that is for resolving a multisample texture.
Rather, create a rendertarget texture, Graphics::SetRenderTarget() to it, perform rendering with Graphics calls, then you should be able to read pixels with Texture2D::GetData().

1vanK:
Thanks!
```
    SharedPtr<Texture2D> renderTarget(new Texture2D(context_));
    renderTarget->SetSize(width, height, graphics->GetRGBAFormat(), TEXTURE_RENDERTARGET);
    graphics->SetRenderTarget(0, renderTarget);
    graphics->SetViewport(IntRect(0, 0, width, height));
    graphics->Clear(CLEAR_COLOR, Color::RED);
    SharedPtr<Image> result = renderTarget->GetImage();
```
works nice

PR #1780, opened 2017-01-26
Fix crash related to TileMap2D when no scene is set

gogoprog:
Hello, here is a fix for the following case :

 * Create a node without a scene
 * Add TileMap2D component
 * call SetTmxFile(...)

This was crashing because 'renderer_' was null in StaticSprite2D.

Now it won't crash and StaticSprite2D will update its material when added to a scene.


cadaver:
Thanks. Will test briefly.

PR #1782, opened 2017-01-26
fixes #1781 - Moved SDL_Init() into Context class

### LINKED ISSUE #1781: Move SDL_Init() into Context
As discussed here: http://discourse.urho3d.io/t/sdl-init-doesnt-get-called-in-headless-mode/2746

Will submit a PR.

TheComet93:
```SDL_Init()``` and ```SDL_Quit()``` are now located in the ```Context``` class. Urho3D subsystems can use ```context_->RequireSDL(SDL_INIT_SUBSYSTEM);``` and ```context_->ReleaseSDL()``` in their constructor/destructors if they require the use of SDL functions. The ```Context``` class globally keeps track of the number of calls to ```RequireSDL()``` so it knows when to call ```SDL_Quit()```

### DIFF
--- a/Source/Urho3D/Audio/Audio.cpp
+++ b/Source/Urho3D/Audio/Audio.cpp
@@ -58,6 +58,8 @@ Audio::Audio(Context* context) :
     sampleSize_(0),
     playing_(false)
 {
+    context_->RequireSDL(SDL_INIT_AUDIO);
+
     // Set the master to the default value
     masterGain_[SOUND_MASTER_HASH] = 1.0f;
 
@@ -70,6 +72,7 @@ Audio::Audio(Context* context) :
 Audio::~Audio()
 {
     Release();
+    context_->ReleaseSDL();
 }
 
 bool Audio::SetMode(int bufferLengthMSec, int mixRate, bool stereo, bool interpolation)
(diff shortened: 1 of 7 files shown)


cadaver:
Cool! It occurs we don't init Timer at all, but we do our own timing anyway so it's probably not going to hurt. Note some code convention issues, mostly spaces between parentheses, and we don't prefix globals with g_, it's fine to call the variable just sdlInitCounter, especially since it's static and can't be seen anywhere else.

TheComet93:
I fixed the coding conventions. Is there a reason for SDL_INIT_TRAMPOLINE?

cadaver:
Thanks. Do you mean SDL_INIT_NOPARACHUTE? That is to avoid SDL messing with signal handlers.

TheComet93:
Whoops, yeah that's what I mean. That flag does nothing in SDL 2.0

cadaver:
Ah, then it can be removed :) Must have been copied directly from old SDL-interfacing engine code that was using SDL 1.x.

cadaver:
There appears to be a conflict with Input.cpp, can you rebase with latest master to fix it?

cadaver:
Thanks, merged now.

PR #1784, opened 2017-01-27
IntVector3

henu:
Hello.

I quickly created this code for one of my projects, and I haven't actually tested it at all. At least it compiles. I copied the code from IntVector2, so it might miss some functionality that is needed with three dimensions.

If you want, you can merge it, but I'm also happy to make any improvements you might suggest.

### DIFF
--- a/Source/Urho3D/LuaScript/LuaScriptInstance.cpp
+++ b/Source/Urho3D/LuaScript/LuaScriptInstance.cpp
@@ -190,6 +190,13 @@ void LuaScriptInstance::OnSetAttribute(const AttributeInfo& attr, const Variant&
                 tolua_register_gc(luaState_, lua_gettop(luaState_));
             }
             break;
+        case VAR_INTVECTOR3:
+            {
+                IntVector3* value = new IntVector3(src.GetIntVector3());
+                tolua_pushusertype(luaState_, value, "IntVector3");
+                tolua_register_gc(luaState_, lua_gettop(luaState_));
+            }
+            break;
         default:
             URHO3D_LOGERROR("Unsupported data type");
             lua_settop(luaState_, top);
@@ -266,6 +273,9 @@ void LuaScriptInstance::OnGetAttribute(const AttributeInfo& attr, Variant& dest)
         break;
     case VAR_INTVECTOR2:
         dest = *((IntVector2*)tolua_tousertype(luaState_, -1, 0));
+        break;
+    case VAR_INTVECTOR3:
+        dest = *((IntVector3*)tolua_tousertype(luaState_, -1, 0));
         break;
     default:
         URHO3D_LOGERROR("Unsupported data type");
@@ -582,10 +592,12 @@ void LuaScriptInstance::GetScriptAttributes()
                     info.type_ = VAR_QUATERNION;
                 else if (typeName == "Color")
                     info.type_ = VAR_COLOR;
-                else if (typeName == "Intrect")
+                else if (typeName == "IntRect")
                     info.type_ = VAR_INTRECT;
-                else if (typeName == "Intvector2")
+                else if (typeName == "IntVector2")
                     info.type_ = VAR_INTVECTOR2;
+                else if (typeName == "IntVector3")
+                    info.type_ = VAR_INTVECTOR3;
             }
             break;
         default:
--- a/Source/Urho3D/AngelScript/CoreAPI.cpp
+++ b/Source/Urho3D/AngelScript/CoreAPI.cpp
@@ -277,6 +277,11 @@ static void ConstructVariantIntVector2(const IntVector2& value, Variant* ptr)
     new(ptr) Variant(value);
 }
 
+static void ConstructVariantIntVector3(const IntVector3& value, Variant* ptr)
+{
+    new(ptr) Variant(value);
+}
+
 static void ConstructVariantPtr(RefCounted* value, Variant* ptr)
 {
     new(ptr) Variant(value);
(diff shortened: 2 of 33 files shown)


cadaver:
Thanks for the work, however I wouldn't be in favor of merging a class that the engine itself does not have direct need for, as that way it doesn't get regularly exercised / tested.  Even though it would be useful for someone.

henu:
Hmm okay, I didn't though about that... I just felt it would be natural addition for of IntVector2.

So should I just close it? Or would it be a good idea to let it wait here for a while in case other people think it's a good idea?

cadaver:
I'd wait for comments for other core devs, but if there's no comments for a few days, then you can close it.

Modanung:
An `IntVector3` is ideal for 3D grid coordinates. Which is not an unusual use case.

eugeneko:
I've never needed it. Hovewer, I agree that it may be useful for some tasks.
If it will be decided to merge it to Urho, please write appropriate script bindings.

henu:
@eugeneko I have no experience about scripts in Urho3D. Are you talking about similar function as RegisterIntVector2 in MathAPI.cpp? Looks rather easy to make a copy from that function for IntVector3.

henu:
Okay, there are all the bindings, conversions, serialization, etc. I could find.I have no idea how to test them, so I just tried to compile everything and then I tried Ninja Snow War, Editor and I also tried to use the library in my code.

I made copies of every IntVector2-thingy I could find, and I tried to put them nicely after IntVector2 things, except with enum VariantType. Since enum values are generated automatically, I put my VAR_INTVECTOR3 to the end, so it would not mess numbering of other enum constants.

Source/Urho3D/LuaScript/LuaScriptInstance.cpp
```diff
@@ -586,6 +596,8 @@ void LuaScriptInstance::GetScriptAttributes()
                     info.type_ = VAR_INTRECT;
                 else if (typeName == "Intvector2")
                     info.type_ = VAR_INTVECTOR2;
+                else if (typeName == "Intvector3")
```
eugeneko: Does anybody know why lowercase here?

eugeneko:
You missed Lua bindings in *.pkg files.
Finally, I think it's ok to merge it.

Source/Urho3D/LuaScript/LuaScriptInstance.cpp
henu: Looks like a bug to me. The string "Intvector2" cannot be found anywhere else in the code. Also other Lua stuff that deal with user types use "IntVector2". I'll fix it.

henu:
@eugeneko Hmm, You mean these files?

- Source/Urho3D/LuaScript/pkgs/IO/Deserializer.pkg
- Source/Urho3D/LuaScript/pkgs/IO/File.pkg
- Source/Urho3D/LuaScript/pkgs/IO/NamedPipe.pkg
- Source/Urho3D/LuaScript/pkgs/IO/Serializer.pkg
- Source/Urho3D/LuaScript/pkgs/IO/VectorBuffer.pkg
- Source/Urho3D/LuaScript/pkgs/Math/Vector3.pkg
- Source/Urho3D/LuaScript/pkgs/Network/HttpRequest.pkg

I'll modify them, and also include the typo fix, and then I'll amend all changes to the latest commit, so commits look clean.

henu:
If I understood you right, the Lua bindings in *.pkg files are now done. I also fixed the "Intvector2" typo. If you merge this, please make sure it really was a typo.

1vanK:
```
else if (typeName == "Intrect")
```

henu:
@1vanK Damn, I must be blind... Fixed :)

1vanK:
It seems that no one is using Lua, if exists such bugs :)

eugeneko:
> It seems that no one is using Lua, if exists such bugs :)

I've just thought the same. Is it really broken now? I'm too lazy to test.

1vanK:
> I've just thought the same. Is it really broken now? I'm too lazy to test.

If you tell me how to test it xD

eugeneko:
If I understood this code corectly, Lua scripts shall not have IntVector2 and IntRect attributes in Attribute Inspector.

Upd: Checked. Seems that it's really broken.

cadaver:
It's quite a bit of added code. But kudos for the effort, I don't oppose this going in now. Thanks to the new VAR_INTVECTOR3 being last it will not break existing binary-serialized scenes, which is good.

henu:
Yeah, these kind of big changes are always ugly. If you give me any hints how to test it, I can try to look for bugs. If others can test it too and check the code, that would be great.

cadaver:
Running all samples is what I'd do.

henu:
The examples seems to work.

KonstantinTomashevich:
Running samples is good, but, in my opinion, big projects such as Urho3D should be automatically tested, not manually. Graphics can be tested manually, but script bindings, math and etc should use automatic testing.

eugeneko:
@KonstantinTomashevich There are autotests with samples. But nobody wants to write thousands of UTs on simple functionality.

cadaver:
Yes, comprehensive tests would require them to be some developer's specific passion, otherwise it will not happen. Having them would have a likely negative impact to the other developers. In this regard, Urho is not a super-professional project and it's unlikely to change.

henu:
This conversation is starting to go way off topic, but I'll still ask: Is there any way to implement tests in case you would like to do so?

eugeneko:
Since we don't have tests now you shan't start to write 'em.
However, I recommend you to write few test scripts in AS and Lua with IntVector3 and check that this class is usable.
Then, I suggest to merge and fix defects as they occur.

henu:
I tested some simple operations and printing with both script languages and it seems to work.

cadaver:
Thanks for confirmation. Will review (& merge if OK) soon.

henu:
Yeah it's ok.

cadaver:
A couple of AngelScript global functions were missing. Otherwise looked OK.

PR #1791, opened 2017-02-01
Improve Box2D joints handling, fixes #1787

### LINKED ISSUE #1787: Box2D joints handling bug
The way Box2D joints are handled by Urho is not valid. For now every time some joint parameter is changed, the joint gets recreated. It involves destruction of an old object, change of joint definition and then creation of a new joint with altered parameters. For most cases it might be a good way to handle this, despite the fact this is not so performant.

But some joints have internal state which gets lost during the process. For example the prismatic joint with its lower and upper translation. Limits are computed for rigid bodies during the creation of a prismatic joint. If bodies move away during the simulation (due to gravity or a motor) and then joint gets recreated, the limits are computed again but for bodies in other positions, which stretches (or narrows) the limits creating strange behaviors.

I tried to change the joint parameter directly (for example a motor speed) without joint recreation and then it worked correctly. Are there any other reasons this is not implemented this way?

barograf:
This PR contains improved handling for several constraints. I checked it in Urho2DConstraints example and it seems to work fine. Moreover I have added SetLimit and GetLimit for distance constraint.


eugeneko:
Code standard was written to be followed, not ignored, do u 'now?
Fix tabs please.

barograf:
@eugeneko Sure thing. I'm going to fix that.

eugeneko:
Thanks!

cadaver:
New attribute in the distance constraint is missing script bindings. Note this for future (it's listed in the contributor checklist) but I can add it this time while reviewing.

barograf:
@cadaver Thanks. I'll keep it in mind.

PR #1794, opened 2017-02-04
Compact viewport mode

Liichi:


eugeneko:
Sooo... what's it?

Liichi:
It is a layout that puts the viewport area in the center of the screen (not fullscreen), the hierarchy window on the left and the attribute inspector window on the right.
This layout also deactivates the ability to move or scale the hierarchy window and the attribute inspector.
PS: sorry for my english.

preview

cadaver:
Looks useful. However, can you edit the code to match the engine code convention (e.g. braces)? The idea is to use the same convention in both C++ code and scripts.

Liichi:
Yes, of course, i will also comment on the code.

Liichi:
I have a problem with the grid. I do not know if it's a problem with my code or an editor issue.
grid

Liichi:
I think everything is ready.
Preview:
preview2

eugeneko:
Let's squash commits when you finish. Dozen of non-functional commits like codestyle fixup IMO decrease readability of repo tree.

Liichi:
It's ready.

weitjong:
Just want to chip in on Eugene's point. I understand this could be a personal habit to commit as frequent as possible. However, one don't have to push to remote repository on each commit, especially when a PR has been submitted to upstream. Each push would not only send a notification to all of repo watchers' email accounts, but also trigger a CI build unnecessarily. It increases our carbon footprint :)

Liichi:
I agree, I had not thought of that :( . So should I edit the files locally?
PS: sorry for email and CI flood. C:

1vanK:
How to squash https://github.com/urho3d/Urho3D/pull/1733#issuecomment-269802300

cadaver:
I would have one suggestion, when switching from Compact mode back to the others, could you reset inspector & attribute editor window sizes to the defaults (same which is done on editor startup)? It's a bit of a pain to resize them manually after they've been expanded to the window edges.

cadaver:
Thanks, will test again!

cadaver:
Merged as squashed.

PR #1797, opened 2017-02-05
Update Renderer.h

Inoc4:
Fixed a typo.


darwikey:
Actually it's not an error of typo, it's an interval. It means that _LightBleeding_ can be between 0 and 1 but with 1 exclude.
https://en.wikipedia.org/wiki/Interval_(mathematics)#Including_or_excluding_endpoints

Inoc4:
ahh, ok.
sorry for that and thanks for the explanation

PR #1805, opened 2017-02-09
JS: Add IES cookie generation to RampGenerator

JSandusky:
Adds Gobo/Cookie baking from IES files to RampGenerator.

Part of a series of PRs I'll submit for old functionality I'd like to see merged in.

**Note:** I couldn't figure out how to filter these based purely on the data, the blur could be removed if someone else can sort out this lat/lon ray thing better, I never found a method that dealt with it robustly enough (bollard lights break everything). Blurring is generally sufficient.

I'll be opening a series of issues, they won't be actual issues but more matters of serious discussion regarding merging such as OpenCL, Texture cube arrays, geom/hull/dom shaders, etc.

### DIFF
--- a/Source/Tools/RampGenerator/RampGenerator.cpp
+++ b/Source/Tools/RampGenerator/RampGenerator.cpp
@@ -37,7 +37,7 @@
 using namespace Urho3D;
 
 // Kernel used for blurring IES lights
-static const float Sigma3Kernel9x9[9 * 9] = {
+static const float sigma3Kernel9x9[9 * 9] = {
     0.00401f, 0.005895f, 0.007763f, 0.009157f, 0.009675f, 0.009157f, 0.007763f, 0.005895f, 0.00401f,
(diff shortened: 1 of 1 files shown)


eugeneko:
_offtop_ I suggest you to PR other issues. It'd be easier to understand and discuss about.

cadaver:
Very minor code convention issue: sigma3Kernel9x9 (start with lowercase) would be preferable.

JSandusky:
@eugeneko negative, only a few of these are small enough to occur without discussion.

@cadaver is the correction sufficient, would it be preferable to base off of *Image* and include the convolution filter routines (arbitrary gauss calc. is part of that) with this?

cadaver:
No need, this can be taken in now as-is. 

Is it OK to include your name in the contributor list again?

JSandusky:
@cadaver, it's fine with me.

cadaver:
Merged.

JSandusky:
Okay, so the AppVeyor thing was just a build system choke or something of the sort then?

cadaver:
AppVeyor builds are being worked on. I saw the errors had nothing to do with your code so I went ahead.

PR #1818, opened 2017-02-12
Update Assimp

Florastamine:
Unfortunately, as of Assimp 3.3 the developers have switched to C++11 and does not provide any pre-C++11 fallback paths. Reversing that would take some time and effort, so I've only updated to the latest release which still sticks with C++03 and here's a PR for that.

### DIFF
--- a/Source/ThirdParty/Assimp/CMakeLists.txt
+++ b/Source/ThirdParty/Assimp/CMakeLists.txt
@@ -675,6 +675,8 @@ SET( zlib_SRCS
 SOURCE_GROUP( zlib FILES ${zlib_SRCS})
 
 SET ( openddl_parser_SRCS
+  contrib/openddlparser/code/OpenDDLExport.cpp 
+  contrib/openddlparser/code/OpenDDLCommon.cpp
   contrib/openddlparser/code/OpenDDLParser.cpp
   contrib/openddlparser/code/DDLNode.cpp
   contrib/openddlparser/code/Value.cpp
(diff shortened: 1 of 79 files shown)


benrob0329:
Why is C++11 still a problem? Its been out for 6 years now.

eugeneko:
@benrob0329 Because Urho3D was promised to be compilable on e.g. MSVC 2008 that doesn't have C++11?

weitjong:
There is a build option to enable C++11 but it is not set by default.

cadaver:
If I'm honest I rarely test on VS2008 / VS2010 nowadays, so for all I know it could be broken on those for other reasons. However, how would the GCC on our CI servers fare if C++11 would be mandatorily enabled?

weitjong:
According to this [page](https://gcc.gnu.org/projects/cxx-status.html#cxx11), GCC 4.8.1 was the first feature-complete implementation of the C++11 standard. By default Travis (Ubuntu 12.04 LTS) comes with GCC 4.6.3, however, in the "refactor-buildsystem" branch I have successfully bump the version up to 4.9.4 across the build matrix, so it should be no problem there.

cadaver:
Good to know. We can return to the issue of updating to newer assimp when that branch is merged.

Florastamine:
Agree. On a side note, VS2010 and older platforms need to be dropped at some point, imo. Even the RPI can be configured to use C++11 and newer standards.

cadaver:
Pushed to a branch for full CI testing.

cadaver:
CI testing OK, merged. Thanks for the work!

PR #1820, opened 2017-02-12
Added Android build script to support building for multiple ABIs at once

IDWMaster:
Added a nice helper shell script to assist with building for multiple Android ABIs.


weitjong:
Thanks for your PR. Personally, I see a few problems with your approach like: it has hard-coded build option, only work for *NIX host system, and multiple ```make``` invocation using a same build tree location (although the script nuke the build tree on each make and should not cause any build issue, still it means it will be slower than it should be).

If there is really such a need, perhaps, we should do it using the same approach we have done for iOS with so-called "universal" build. One call to ```make``` and it takes care of all the ABIs internally.

IDWMaster:
A "universal" build would be much appreciated, although I'm not sure how I'd go about implementing that when the CPU architectures vary substantially (I haven't done much iOS development, but was under the impression it was all ARM?). Should I close this pull request and create a new feature request?

weitjong:
The iOS universal library build contains mix of ARM and Intel archs (for simulator) as well.

IDWMaster:
To my knowledge; Android has no equivalent to universal libraries (or fatelf files), and each architecture must be built separately.

weitjong:
I was just talking about the approach. Surely, the end result won't be a Mach-O universal binary.

IDWMaster:
I've updated my build script. It now uses separate build directories for each ABI and doesn't delete them.

weitjong:
After thinking more about this, I feel there is really no need nor it is possible to actually automatically going through "all" the ABIs when targeting Android platform. There are quite a number of permutations besides the one you just listed in your range loop, especially for the 32-bit ARM arch. See [here](https://github.com/urho3d/Urho3D/blob/refactor-buildsystem/CMake/Toolchains/Android.cmake#L39-L69) for the complete list of the supported ABIs. Although it is nice to have a convenience script which loop through one's desired ABIs, the list of the ABIs itself is in fact another hard-coded value which may not be universal, i.e. it  depends on each intended target audience. Since the proposed script contains hard-coded values, I am not in favor of it to be resided in the master branch. Having said that, this is just my personal opinion.

cadaver:
I would agree, particularly as it's just a multiple make invocation, and furthermore with hardcoded build options. I wouldn't have anything against a documentation page giving this kind of script as an example of how to automate building multiple ABIs.

PR #1823, opened 2017-02-13
GetDerivedComponent for Physics2D to allow using custom PhysicsWorld2D

1vanK:


cadaver:
Nice, it was only in those two places. I thought of making a function but the duplication isn't yet bad. Can merge this as-is.

1vanK:
May be GetDerivedOrCreate is useful elsewhere

cadaver:
Yes, on the other hand this makes you think that you're doing a relatively expensive lookup first.

PR #1832, opened 2017-02-24
Extract some code to increase reusability.

eugeneko:
Main change here is in 4th commit of this PR.
I wanted to allow change paths on-the-fly with the same semantic as they are initialized.
I intentionally didn't add force ResourceCache force cleanup because it can be easlily done outside.

@cadaver, please check for important things that I may have missed.

**Note for reviewers**: it's easier to review changes in Engine.cpp via https://github.com/urho3d/Urho3D/pull/1832/commits/f386678667396b342bc40a9ba9a52154914452e3


cadaver:
Looks good. Can you add IsChildOf() to script bindings too?

PR #1836, opened 2017-02-28
AbstractFile class

rokups:
Added AbstractFile class. It is used as parent class for classes that implement both Serializer and Deserialized.

While doing profiler work i found a need to interact with objects that implement both `Serializer` and `Deserializer`. Urho3D has several such objects but all of them inherit both classes separately. That means there is no way to have type system require an object implementing both. For this reason `AbstractFile` class was created and objects that inherited both `Serializer` and `Deserializer` now inherit `AbstractFile` instead.

### DIFF
--- a/Source/Urho3D/IO/AbstractFile.h
+++ b/Source/Urho3D/IO/AbstractFile.h
@@ -1,0 +1,43 @@
+//
+// Copyright (c) 2008-2017 the Urho3D project.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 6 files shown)


cadaver:
Seems to be missing include in NamedPipe.h?

eugeneko:
Nice interface, but I somewhy dislike its name. I'd prefer somethink like `Stream` or `IOStream` or something-else-feel-free-to-suggest... Let's think more.

1vanK:
Also (this is just my opinion) InputStream and OutputStream is more correct names than Deserializer and Serializer (or StreamReader and StreamWriter), but it was too late to change

rokups:
`IOStream` is probably better than `Stream` since it clearly states that stream handles input and output. It certainly is better than `AbstractFile`. @1vanK is also right. Names can always be changed and aliases to old names can be created. We can even deprecate those aliases if needed. gcc has `__attribute__((deprecated))`, msvc has `__declspec(deprecated)`, and anything else emulates these two compilers. And finally c++11 has [[[deprecated]]](http://en.cppreference.com/w/cpp/language/attributes) attribute.

rokups:
@cadaver final word on `AbstractFile` class name?

cadaver:
I used Stream in Turso3D codebase for a similar case. Though, it doesn't necessarily work as a stream, as also seeking is involved, so I think AbstractFile is actually closer to what it does and is fine for me. Could merge this as is.

eugeneko:
:(
But `Stream` is nicer.

cadaver:
It's a bit imperfect in any case. Deserializer says it's an "abstract stream" for reading. Serializer similarly for writing. To me it feels a bit funny if they combine into a class called "Stream" which is still abstract. In Turso there were no base classes for Stream, it already handled both writing & reading.

PR #1838, opened 2017-02-28
64bit integer variant

rokups:
Title ^. Please make sure i did not forget anything here.

### DIFF
--- a/Source/Urho3D/AngelScript/CoreAPI.cpp
+++ b/Source/Urho3D/AngelScript/CoreAPI.cpp
@@ -183,6 +183,16 @@ static void ConstructVariantUInt(unsigned value, Variant* ptr)
     new(ptr) Variant(value);
 }
 
+static void ConstructVariantInt64(long long int value, Variant* ptr)
+{
+    new(ptr) Variant(value);
+}
+
+static void ConstructVariantUInt64(unsigned long long int value, Variant* ptr)
+{
+    new(ptr) Variant(value);
+}
+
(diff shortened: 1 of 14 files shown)


eugeneko:
I think that GetInt/GetUInt shall also work with 64-bit ones.

_offtop_
And this Variant.h is a bit too dirty, IMO. So many copy-pasted stuff and dirty casts.
Do we really have no better choise than this?

cadaver:
It's legacy, what do you think? You're free to refactor (later) but the upcoming release should still use the old ways, so this PR is IMO proper in respect to it.

eugeneko:
It wasn't an issue to this PR (so I marked it as _offtop_)
Just a random thought. Variant and offsetof-attributes are the dirtiest places in Urho, IMO.

rokups:
These script bindings are... Really annoying :\ We should really steal binding generator from Atomic.

Could anyone explain what exactly is going on in [this struct](https://github.com/rokups/Urho3D/blob/f90dde08fec0a07f8bf36ec95ef270f4d1b0bedf/Source/Urho3D/Core/Variant.h#L70-L100)? Why multiple unions? I wanted to add my int64 value but didnt because i do not quite understand why this is made so. Thats why i copied code from double (`reinterpret_cast`).

cadaver:
Multiple unions are for many-component objects, for example Vector3 uses the first 3 floats, and IntVector3 would use 3 ints. If the data doesn't fit into the union, then you have to heap allocate.

cadaver:
Going to merge with minor modifications. Thanks!

PR #1842, opened 2017-03-02
Add info about shapes to PhysicsEvents2D

1vanK:
Sometimes it is necessary to identify collided shapes: http://www.iforce2d.net/b2dtut/jumpability

Can we also add to components name_, tags or vars? (currently I compare pointers to identify shape)

EDIT: May be move "name", "vars" and "tags" to Animatable (or Serializable) class?

### DIFF
--- a/Source/Urho3D/Scene/Animatable.cpp
+++ b/Source/Urho3D/Scene/Animatable.cpp
@@ -80,6 +80,7 @@ void Animatable::RegisterObject(Context* context)
 {
     URHO3D_MIXED_ACCESSOR_ATTRIBUTE("Object Animation", GetObjectAnimationAttr, SetObjectAnimationAttr, ResourceRef,
         ResourceRef(ObjectAnimation::GetTypeStatic()), AM_DEFAULT);
+    URHO3D_ACCESSOR_ATTRIBUTE("Name", GetName, SetName, String, String::EMPTY, AM_FILE);
 }
 
 bool Animatable::LoadXML(const XMLElement& source, bool setInstanceDefault)
@@ -184,6 +185,11 @@ bool Animatable::LoadJSON(const JSONValue& source, bool setInstanceDefault)
     }
 
     return true;
+}
+
+void Animatable::SetName(const String& name)
+{
+    name_ = name;
 }
 
 bool Animatable::SaveXML(XMLElement& dest) const
--- a/Source/Urho3D/Scene/Animatable.h
+++ b/Source/Urho3D/Scene/Animatable.h
@@ -121,6 +121,11 @@ public:
     /// Return object animation attribute.
     ResourceRef GetObjectAnimationAttr() const;
 
+    /// Set name.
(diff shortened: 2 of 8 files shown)


1vanK:
What for is needed NodeImpl?

eugeneko:
> What for is needed NodeImpl?

To make Node class more light-weight for some performance stuff.

1vanK:
> To make Node class more light-weight for some performance stuff.

For example?

eugeneko:
> For example?

Check related commits, there were some performance degradation.

BTW, why have you removed Name attribute from ::RegisterObject? Have you checked e.g. Editor?

1vanK:
Thanks, I think it copys base attributes, but it not

1vanK:
> UPD: fee755e

I still do not understand the circumstances under which performance is improved

1vanK:
Hm, I should add
```
URHO3D_ACCESSOR_ATTRIBUTE("Name", GetName, SetName, String, String::EMPTY, AM_DEFAULT);
```
to every derived Component
or create Component::RegisterObject() and URHO3D_COPY_BASE_ATTRIBUTES() to every derived Component

1vanK:
Should I add
```
URHO3D_ACCESSOR_ATTRIBUTE("Name", GetName, SetName, String, String::EMPTY, AM_DEFAULT);
```
to every derived components

or create Component::RegisterObject(Context* context) and add URHO3D_COPY_BASE_ATTRIBUTES to every derived components

eugeneko:
> I still do not understand the circumstances under which performance is improved

I can't find an issue, but you can check some heavy benchmarks like Demo 20

cadaver:
The point with NodeImpl is to reduce the binary size of the Node class to make it fit into cache better for "hot" access like a Drawable asking its world transform. 

I can cherrypick the first commit from this PR but don't think will be taking the Node / Animatable change. That's btw. also why I find master branch PR's a bit scary, because master usually represents author's continuing development. Separate feature branch for PR would tell more clearly that it's this specific change (mentioned in the PR) you want to be pulled in now, and it won't be changing in the meanwhile.

1vanK:
Any another way to identify components?

1vanK:
Box2D uses UserData for identification, but curently UserData is pointer to Component

cadaver:
Currently there's not much else than the index order within the parent node's component list. When you know e.g. specific shape components are for specific parts of the object, you could hold them in pointers / handles in logic code.

1vanK:
> you could hold them in pointers / handles in logic code.

Yes, I am doing so, but it is not very convenient (sensors placed all around main collision shape and is needed compare with every)

1vanK:
It is possible to accept only first commit or I should modify PR?

cadaver:
I will cherrypick and merge, you don't have to do anything. The hash changes so it will not show the exact same commit you made, but I don't think that's an issue.

cadaver:
First commit merged.

PR #1849, opened 2017-03-07
Remove duplicate macro in SDL windows joystick implementation.

iSLC:


weitjong:
Thanks for spotting this. It is really weird. The merging of upstream SDL changes with our local changes was mainly done automatically via git without human intervention, except when there was a conflict, of course. Not sure the mistake was simply caused by faulty conflict resolution or wrong auto merge result, or worse wrong git subtree operation. Did you just come across this by chance or did you review all the SDL code and only found this particular anomaly?

iSLC:
I reviewed most of the windows/unix related code in SDL to port the changes to the svn version od SDL (_for experimentation purposes, nothing else_). I've only noticed this one. But I can't speak for mac/android/ios platforms.

It was mostly straight forward copy/paste and the only other difference I've noticed was the fixed leak from `SDL\src\video\x11\SDL_x11opengl.c`.

weitjong:
If it is one-off then probably it was just faulty conflict resolution. Will double check later.

weitjong:
Did a quick scan and indeed only found this one anomaly.

PR #1856, opened 2017-03-10
Improve Box2D integration

eugeneko:
- Added serialized contact info for each physical2D event;
  + Temporary pre-allocated buffer inside PhysicsWorld2D is used, so it shan't be heavy;
  + New parameter is called `ContactPoints` and similar to `Contacts` parameter of physical events;
  + I wanted to call new parameter `Contacts` too, but it can be easily mixed up with already existing `Contact` parameter. May be discussed.

- Added event for contact update.
  + Similar to `NodeCollision` event, called from Box2D internals while contact is on between pre-step and post-step;
  + Can be used for simple check of ongoing contact in script logic;
  + Contact may be disabled in event handler;
  + Can be used for creating simple one-way obstacles and so on.

- By default Box2D re-enable contact before each check. I disabled this behavior because I find it disturbing. One can easily re-enable any contact on his own.

### DIFF
--- a/Source/Urho3D/Urho2D/PhysicsEvents2D.h
+++ b/Source/Urho3D/Urho2D/PhysicsEvents2D.h
@@ -31,6 +31,21 @@
 namespace Urho3D
 {
 
+/// Physics update contact. Global event sent by PhysicsWorld2D.
+URHO3D_EVENT(E_PHYSICSUPDATECONTACT2D, PhysicsUpdateContact2D)
+{
+    URHO3D_PARAM(P_WORLD, World);                  // PhysicsWorld2D pointer
+    URHO3D_PARAM(P_BODYA, BodyA);                  // RigidBody2D pointer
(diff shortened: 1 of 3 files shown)


cadaver:
Looks good to me.

1vanK:
I do not fully understand. Does Box2D send the same contact several times?

1vanK:
disabledContacts_ never cleared?

cadaver:
@eugeneko Are you still going to do additional work to this?

eugeneko:
@cadaver Yes. I encountered that my way of storing contacts is bad design from Box2D PoV.
I was going to roll back few lines from this PR when I ensure that my tasks are still solvable.
If you want to e.g. label new version, I can roll back these lines ASAP and then merge.

cadaver:
I don't think there's hurry right now, since there's also the IK PR ongoing, and since our releases are not *that* frequent, it would be nicer to get that in now rather than waiting for another half year or more :)

eugeneko:
Ok, I reverted. I'll wait for build and then merge.

Issue #1412, opened 2016-06-09
fixed size slider bug


Lumak:
When the slider's knob is set as fixed size, it hops when slid because of the 0.5f value added in the computation.

code below fixes it, tested both horizontal and vertical settings.

void UpdateSlider()
{
    const IntRect& border = knob_->GetBorder();

    if (range_ > 0.0f)
    {
        if (orientation_ == O_HORIZONTAL)
        {
            int sliderLength = (int)Max((float)GetWidth() / (range_ + 1.0f), (float)(border.left_ + border.right_));

            if ( knob_->IsFixedSize() )
            {
                sliderLength = knob_->GetWidth();
            }

            float sliderPos = (float)(GetWidth() - sliderLength) * value_ / range_;

            if ( !knob_->IsFixedSize() )
            {
                knob_->SetSize(sliderLength, GetHeight());
                knob_->SetPosition(Clamp((int)(sliderPos + 0.5f), 0, GetWidth() - knob_->GetWidth()), 0);
            }
            else
            {
                knob_->SetPosition(Clamp((int)(sliderPos), 0, GetWidth() - knob_->GetWidth()), 0);
            }
        }
        else
        {
            int sliderLength = (int)Max((float)GetHeight() / (range_ + 1.0f), (float)(border.top_ + border.bottom_));

            if ( knob_->IsFixedSize() )
            {
                sliderLength = knob_->GetHeight();
            }

            float sliderPos = (float)(GetHeight() - sliderLength) * value_ / range_;

            if ( !knob_->IsFixedSize())
            {
                knob_->SetSize(GetWidth(), sliderLength);
                knob_->SetPosition(0, Clamp((int)(sliderPos + 0.5f), 0, GetHeight() - knob_->GetHeight()));
            }
            else
            {
                knob_->SetPosition(0, Clamp((int)(sliderPos), 0, GetHeight() - knob_->GetHeight()));
            }
        }
    }
    else
    {
        if ( !knob_->IsFixedSize())
        {
            knob_->SetSize(GetSize());
        }
        knob_->SetPosition(0, 0);
    }
}

cadaver:
Thanks for the fix! Will test shortly.

PR #1857, opened 2017-03-11
Fix jumping slider knob when it's size is fixed.

rokups:
Slider has a feature where setting min and max size of knob allows us to control it's size. However when knob size is enlarged moving know causes it to jump forward and back. This change fixes the issue.


Lumak:
There was an issue with this in both orientations, https://github.com/urho3d/Urho3D/issues/1412.  Did you text vertical orientation?

rokups:
Yes, vertical slider works as well.

Lumak:
ok, cool.

PR #1865, opened 2017-03-14
Updated sound effects sample to use SetAutoRemoveMode() instead of nodes

Enhex:


1vanK:
What about lua and as examples?

cadaver:
Nice code simplification!

Enhex:
There's a bug in the Lua bindings (`AudioLuaAPI.cpp` line `2335`):
```C++
AutoRemoveMode mode = *((AutoRemoveMode*)  tolua_tousertype(tolua_S,2,0));
```
should be
```C++
AutoRemoveMode mode = *((AutoRemoveMode*)  tolua_tousertype(tolua_S,1,0));
```

Lua bindings seems to be automatically generated, and I don't know how it works in order to fix it (in case it isn't just something messed up on my side).

eugeneko:
> should be

I don't understand why.

cadaver:
Looks possibly related to this, and it's an ancient bug: http://lua-users.org/lists/lua-l/2004-01/msg00193.html

eugeneko:
> Looks possibly related to this, and it's an ancient bug

Huh... The @Enhex 's question is about index, but your link is about syntax... I doubt that it is related.

weitjong:
> Huh... The @Enhex 's question is about index, but your link is about syntax... I doubt that it is related.

No, I think I agree with @cadaver that it is related.

weitjong:
Adding these lines in the ```SoundSource.pkg``` seem to correct the auto-generated binding.

```
enum AutoRemoveMode
{
    REMOVE_DISABLED = 0,
    REMOVE_COMPONENT,
    REMOVE_NODE
};
```

cadaver:
I remember we've done that elsewhere too. So that is the reason :)

eugeneko:
May somebody explain what's wrong in generated binding?
I mean in this case, not at posted link.
I don't understand :(

weitjong:
The way I understand it the problem is caused by tolua++ makes a wrong assumption when it first encounters the "AutoRemoveMode" symbol. It didn't know what it is and erroneously assumed it to be one of a user type. By explicitly declaring "AutoRemoveMode" is an enum then tolua++ knows what it is correctly and generates the binding accordingly.

eugeneko:
[offtop] *confused* I haven't noticed any changes in generated files after rebuild.

PR #1867, opened 2017-03-16
Disable Visual Studio warnings for third party

Sasha7b9:


weitjong:
Thanks for your PR. Personally I don't think we should suppress those warnings so that we, the devs, know they exist. I think the only time we are forced to suppress them is when we are scripting to target Apple platforms on Travis CI in order to not exceeding its log entries limit. https://github.com/urho3d/Urho3D/blob/513f9592fa0b616e6a09b8ec86740558b31ca858/CMake/Modules/UrhoCommon.cmake#L1381-L1385
If the warnings trouble you then perhaps you can add the MSVC-specific compiler flag in the same macro instead of pepper it all over the places.

Sasha7b9:
Okay. Thanks.

PR #1868, opened 2017-03-18
Add fullscreen monitor and refresh rate options

PredatorMF:
Adding essential options for fullscreen in multimonitor environment. 
Tested on Windows 10 with DX9, DX11 and OpenGL renderers.


cadaver:
This is excellent, thanks! Some minor inconsistency in the scripting API for Graphics::GetResolutions(), the monitor number parameter could be exposed to script too.

PredatorMF:
Thanks for pointing that out, gonna push in a minute.

PR #1871, opened 2017-03-21
Inverse kinematics v1.0

TheComet93:
For the past few weeks I've been pushing development on a FABRIK based ik library of mine and I think it's reached a point where it's stable enough to be usable. I'd like to make it more accessible to the Urho3D community so I can get better feedback for future improvements.

The relevant forum thread can be found here: http://discourse.urho3d.io/t/inverse-kinematics/1819/43

An overview of the changes:
  + Included a new thirdparty library for solving IK
  + Added an URHO3D_IK build option, which is enabled by default
  + Added 3 components exposing the ik library to Urho3D
  + Added a new editor icon for the IKEffector component
  + Wrote AngelScript and Lua bindings (lua is currently untested)
  + Wrote a new sample (45_InverseKinematics) and ported it to AngelScript

Let me know if there is anything to be improved upon or changed.


iSLC:
I could never really understand why Lua was necessary when AngelScript was more than just fine. So far I've only seen Lua to be a hindrance to new features and maintenance.

cadaver:
Looks like a very formidable PR with almost everything required in place, thanks! The example seems to be segfaulting on Linux, and there's some VS project files mixed in the PR.

About Lua: if there would be overwhelming consensus to drop it I would have nothing against it. Though I imagine there would be quite a few users who would be disappointed.

eugeneko:
Lua has nice syntax. But I don't use it too.

@TheComet93 The only style issue/question: are multiline doxygen comments used in Urho?
Probably they should be converted into single lince comments, ast least in cases when there is only one line.

TheComet93:
Can you give me more information on the segfault? I developed this on Linux but the C++ and AngelScript samples run fine. The Lua sample currently crashes and I'm not sure why, I'll make a post on the forums about it.

As to the int32_t/uint32_t redefinition, I'm not sure how to fix this. The error occurs in Context.cpp and it seems SDL is defining int32_t before I include "ik/pstdint.h", which promptly redefines it. I would have assumed pstdint.h checks whether int32_t is defined or not before defining it?

iSLC:
Would it be a bad practice to manually define `_PSTDINT_H_INCLUDED` before you include the IK library and  if SDL has `HAVE_STDINT_H` defined as well then they'll both include `stdint.h` so that'll keep them both happy.

TheComet93:
Well, that didn't quite work. :/

```cpp
#ifdef URHO3D_IK
// Hopefully fixes issue on windows where ik/pstdint.h redefines int32_t
#if defined(HAVE_STDINT_H)
#define _PSTDINT_H_INCLUDED
#endif
#include <ik/memory.h>
#include <ik/log.h>
#endif
```

iSLC:
Uhm, maybe a pull request to the IK library author to modify `effector.h`, `node.h` and `solver.h` to use the following:
```
#ifdef HAVE_STDINT_H
	#include <stdint.h>
#else
	#include "ik/pstdint.h"
#endif
```

And to be honest, it doesn't make sense to force your own `stdint.h` header on the user. Therefore the IK library should've had something similar in the first place.

cadaver:
About multiline comments, they're sort of discouraged, but if you need to tell a lot to the function's user it's acceptable.

I hope to look at the Linux crash at some point, during the weekend by latest if it's not solved by then.

TheComet93:
I have since been able to reproduce both the sample crash and the compile error on my Windows machine. I expect to have it solved by tomorrow.

As to the list bindings, @cadaver could you please take a look at the of files in LuaScript/pkgs and tell me if you see anything obviously wrong? The engine segfaults whenever I try to create any IK component, e.g. `node:CreateComponent ("IKEffector")`. I believe to have done everything correctly but it seems I was wrong.

cadaver:
Yes, will take a look.

cadaver:
You are not actually initializing the IK bindings into the Lua state yet. Insert the following (guarded by URHO3D_IK define) to LuaScript.cpp, similar to the other subsystem bindings:

extern int tolua_IKLuaAPI_open(lua_State*);

and later in LuaScript constructor:

    tolua_IKLuaAPI_open(luaState_);

TheComet93:
Fixed the segfault and the stdint.h issues.

The Windows build just failed with:
```
Installing Urho3D SDK to default system-wide location...
Failed to install Urho3D SDK
Command exited with code 1
```

I'm not sure what that means...

weitjong:
It means something is probably wrong with the newly added CMakeLists.txt file which caused the "install" built-in target to fail. At a glance I see the new script does not call our common macro to setup the new IK target, but that may or may not be the contributing factor for this failure.

Mike3D:
Very robust and packed with many useful features & options.

When testing in sample 18, I noticed a couple of tiny details:

- In IKSolver.cpp:
Auto solve need to be exposed, otherwise it won't be restored when reloading a saved scene:
```c++
URHO3D_ACCESSOR_ATTRIBUTE("Auto Solve", AutoSolveEnabled, EnableAutoSolve, bool, true, AM_DEFAULT);
```
- A typo in IKEffector.cpp (default value set to true instead of 0):
```c++
URHO3D_ACCESSOR_ATTRIBUTE("Chain Length", GetChainLength, SetChainLength, unsigned, true, AM_DEFAULT);
```
- In the samples:
Foot offset is the delta between foot and ground, foot being above the ground:
```angelscript
float footOffset = rightFoot_.worldPosition.y - jackNode_.worldPosition.y ;
```
Eager to see constraints and full body integrated

TheComet93:
Here's what I'm doing. This can be found in the file `Source/ThirdParty/ik/cmake/Urho3D/CMakeLists.txt`
```
set (TARGET_NAME IK)
set (SOURCE_FILES ${IK_SOURCES} ${IK_HEADERS})
set (INCLUDE_DIRS ../../ik/include ${CMAKE_CURRENT_BINARY_DIR}/../../include)

setup_library ()

# Install headers for building and using the Urho3D library
install_header_files (DIRECTORY ../../ik/include/ik/ DESTINATION ${DEST_INCLUDE_DIR}/ThirdParty/ik)
install_header_files (DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/../../include/ik/gen/ DESTINATION ${DEST_INCLUDE_DIR}/ThirdParty/ik/gen)
```

Mike3D:
Many thanks.

In lua & as samples, the IKSolver is attached to spine (matching with the comment above), while in C++ sample it is attached to jackNode.

1vanK:
Great works! I have a question. When chain == 3 shoulder excessively flexibly.

Is there a way to reduce weight of some bones?

TheComet93:
@1vanK As of now, it is difficult. There is very basic support for constraints but it's not exposed to Urho3D yet. This is a feature I will be working on next.

TheComet93:
So what's the deal with this build issue? How can we get it fixed? Because I'm out of options.

weitjong:
I will have a closer look this weekend but no promise.

weitjong:
Just to let you know that I will probably need to rewrite your CMake build script to make it fit with our build system and coding convention. I hope you don't mind.

TheComet93:
@weitjong No problem, do whatever is necessary.

slapin:
being offtopic, please don't drop Lua support, I have some plans for it.

weitjong:
I have done my part. I have pushed my local merged branch to Urho3D repo for full CI build tests. It is up to Lasse now where it goes from there.

cadaver:
Thanks! Will test it, hopefully in very near future.

cadaver:
Merged @weitjong's branch. Thanks to everyone involved!

PR #1872, opened 2017-03-22
Add Animation::GetTrack(unsigned)

Florastamine:
This variant of Animation::GetTrack() is listed in the documentation but I couldn't see it implemented anywhere in the source code.

### DIFF
--- a/Source/Urho3D/Graphics/Animation.cpp
+++ b/Source/Urho3D/Graphics/Animation.cpp
@@ -379,8 +379,9 @@ SharedPtr<Animation> Animation::Clone(const String& cloneName) const
 } 
 
 AnimationTrack* Animation::GetTrack(unsigned index) 
-{
-    index = (index < 0) || (index > tracks_.Size() - 1) ? 0 : index; 
+{ 
+    if (index >= GetNumTracks()) 
+        return (AnimationTrack*) 0;
 
     int j = 0; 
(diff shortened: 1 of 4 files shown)


Source/Urho3D/Graphics/Animation.cpp
```diff
@@ -376,8 +376,25 @@ SharedPtr<Animation> Animation::Clone(const String& cloneName) const
     ret->SetMemoryUse(GetMemoryUse());
     
     return ret;
+} 
+
+AnimationTrack* Animation::GetTrack(unsigned index) 
+{
+    index = (index < 0) || (index > tracks_.Size() - 1) ? 0 : index;
```
eugeneko: Ewww... Let's just return 0, as everywhere.

cadaver:
AngelScript binding for the new overload missing? It's OK to be just a GetTrack function, since the property access happens with a string. 

And yes, agree with @eugeneko, if index is illegal, just early-out to returning a null pointer.

Florastamine:
Done. Apologies for the obvious mistakes.

Mike3D:
Wasn't this already exposed by 'tracks' array ?

cadaver:
Thanks, will wait for the CI and merge then.

It's OK to have a convenience overload that accesses by index directly.

PR #1874, opened 2017-03-23
Allow root UI element to be moved.

rokups:
Root UI element can be resized, however setting it's position renders UI with bottom and right portions cut-off. Change fixes the issue and resized/positioned root UI element is rendered properly.

Before anyone asks - tested `GetElementAt()` and it works properly ;)


cadaver:
The root element was supposed to represent the whole screen canvas and not be moved, but you've managed to allow that quite benignly. Since there is no actual "locking" of the root element pos/size to the default (though I'd still discourage changing them), could as well take this in.

rokups:
Yes, i understand. I came to needing this when making UI editor with nuklear. Then it is very useful to scale and move root UI window so that it does not cover sidebars provided by nuklear.

PR #1885, opened 2017-03-30
Update Transform.glsl

cheece:
let the user define the number of outputs for advanced rendering effects such as subsurface scattering


cadaver:
Cool! Should be no problem to take this.

PR #1891, opened 2017-04-02
Fix Thread::GetCurrentThreadID() if threading is disabled.

ombre5733:
When URHO3D_THREADING is not defined, Thread::GetCurrentThreadID() was missing a return statement. Also the helpers for invoking a thread should be static, which will hide their symbols.


cadaver:
Thanks! Will be merged.

PR #1892, opened 2017-04-03
Improve AngelScript API dump, object attribute animations, 2D-particles

orefkov:
In AngelScript API dump fixed template classes dump, add ctors/dtor dump in classes.
Also add some defines.
In object attribute animations add the ability to specify a child node by name when setting attribute animation path.
In 2D particles fix particle emitting duration, add event for ending particles emitting and for all particles over.


Source/Urho3D/Urho2D/ParticleEmitter2D.cpp
```diff
@@ -262,11 +262,21 @@ void ParticleEmitter2D::UpdateMaterial()
...
 
+URHO3D_EVENT(E_PARTICLESEND, ParticlesEnd) {
+}
+URHO3D_EVENT(E_PARTICLESDURATION, ParticlesDuration) {
+}
+
 void ParticleEmitter2D::HandleScenePostUpdate(StringHash eventType, VariantMap& eventData)
 {
     using namespace ScenePostUpdate;
+    unsigned tests = (numParticles_ > 0 ? 1 : 0) | (emissionTime_ > 0.0f ? 2 : 0);
```
eugeneko: Please avoid this unneeded esoteric bitmagic, e.g. in this way:

```
bool wasEmitting = emissionTime_ > 0.0f;
bool hasParticles = numParticles_ > 0;
```

PS: And check for trailing spaces

Source/Urho3D/AngelScript/ScriptAPIDump.cpp
```diff
@@ -386,11 +396,26 @@ void Script::DumpAPI(DumpMode mode, const String& sourceTree)
...
+                    String str = "\ntemplate <";
+                    for (asUINT tt = 0, ttm = type->GetSubTypeCount(); tt < ttm; tt++) {
+                        asITypeInfo* pSubType = type->GetSubType(tt);
+                        str += String("class ") + pSubType->GetName() + (tt < ttm - 1 ? ", " : ">");
+                    }
+                    Log::WriteRaw(str);
+                }
+                Log::WriteRaw("\nclass " + typeName + "\n{\npublic:\n");
+            }
+            if (mode == C_HEADER)
```
eugeneko: I don't really understand this code, but `else if (mode == C_HEADER) { ... }; if (mode == C_HEADER) { ... }` looks suspicious

Source/Urho3D/Urho2D/ParticleEmitter2D.cpp
orefkov: Yes you are right. Initially, there were several other tests, so it remained so. I will change it.

Source/Urho3D/AngelScript/ScriptAPIDump.cpp
orefkov: This is the result of several consecutive code changes in the search for a solution. Thank you for noticing this. I'll fix it now.

cadaver:
Thanks! While merging, I'll add a similar mechanism for name lookup to UI element attribute animations.

orefkov:
May be add ability to choice for search childs by name recursive or not?
For example, if after '#' follow '.' - search child not recursive.
In some my cases Im animating many objects in big scene, and will be more optimised set direct path without searching in all nodes.

cadaver:
The search string format is already rather arcane, so I'd prefer to not add complexity to it.

However there's room for optimization in Node::GetChild(), it could first check immediate children before starting recursion. In that case you wouldn't need the option. Failed queries will be expensive regardless but a finished project shouldn't have failing node animations anyway.

PR #1896, opened 2017-04-07
Fix crash in AngelScript, improve Editor.

orefkov:
Fixing critical error in AngelScript, cause segfault on my script.
Add color wheel in Editor for Text3D component.


cadaver:
Merged, thanks!

PR #1900, opened 2017-04-09
Fixed blend modes for 2D particles

ombre5733:
The 2D particles had missing blend mode functions for
BlendMode::BLEND_SUBTRACT and BLEND_SUBTRACTALPHA. The values for the
functions have been copied from OGLGraphics.cpp.

If C++11 is enabled, the size of the struct is checked with a static
assertion, now.


cadaver:
Thanks!

ombre5733:
@cadaver Wouldn't it be better to serialize the blend mode directly instead of the blend functions? This would also avoid invalid combinations of src/dst functions. Am I missing something?

eugeneko:
> Am I missing something?

What if one has custom blend mode?

cadaver:
It's not a Urho3D native format, but .pex files produced by an external particle editor, which the ParticleEffect class is interpreting.

PR #1903, opened 2017-04-11
Bullet RaycastVehicle support

slapin:
Hi!

This is initial release of RaycastVehicle component, which is wrapping of Bullet btRaycastVehicle.

Many features are missing, which will be added later.
The commit contains C++ code and AngelScript + Lua binding. But there is no Lua demo yet.
The component is not able to save/load yet (don't know how to do it).

The purpose of this PR is to recieve feedback and review. Please remember that:
1. It is my first submission and many things can be wrong. Please have patience and explain to me
what is wrong, I'll be more than happy to fix all things. Especially coding style errors might happen.
2. I'm from C/Linux kernel/MCU scenery and have very little (VERY little) C++ experience.
So please comment on general design errors, if I'm doing something wrong in general.
3. I have very little experience with gamedev, so I might do things not efficiently because of lack
of experience. Please pint me on these, I want to learn.

The motivation behind this PR is current vehicle example is not that effective, and also very rigid.
I can't spawn 100 cars and have no frame drop.
However with this component I can run 100 cars without slowdowns. Also the flexibility is very good.
And it is easier than making everything from scratch.

The code is based on forum example by @Lumak edited and simplified by me to be minimal.
I plan to implement all features provided by btRaycastVehicle and provide Lua demo too. Also I plan to do general cleanup.

This video demonstrates abilities: https://youtu.be/LKjzcN3s65Y

### DIFF
--- a/CMake/Modules/UrhoCommon.cmake
+++ b/CMake/Modules/UrhoCommon.cmake
@@ -105,6 +105,7 @@ cmake_dependent_option (IOS "Setup build for iOS platform" FALSE "XCODE" FALSE)
 cmake_dependent_option (URHO3D_64BIT "Enable 64-bit build, the default is set based on the native ABI of the chosen compiler toolchain" "${NATIVE_64BIT}" "NOT MSVC AND NOT ANDROID AND NOT (ARM AND NOT IOS) AND NOT WEB AND NOT POWERPC" "${NATIVE_64BIT}")     # Intentionally only enable the option for iOS but not for tvOS as the latter is 64-bit only
 option (URHO3D_ANGELSCRIPT "Enable AngelScript scripting support" TRUE)
 option (URHO3D_IK "Enable inverse kinematics support" TRUE)
+option (URHO3D_RAYCASTVEHICLE "Enable raycast vehicle component support" TRUE)
 option (URHO3D_LUA "Enable additional Lua scripting support" TRUE)
 option (URHO3D_NAVIGATION "Enable navigation support" TRUE)
 # Urho's Network subsystem depends on kNet library which uses C++ exceptions feature
@@ -354,6 +355,7 @@ if (URHO3D_CLANG_TOOLS)
             URHO3D_DATABASE_SQLITE
             URHO3D_FILEWATCHER
             URHO3D_IK
+	    URHO3D_RAYCASTVEHICLE
             URHO3D_LOGGING
             URHO3D_LUA
             URHO3D_NAVIGATION
@@ -405,6 +407,7 @@ foreach (OPT
         URHO3D_DATABASE
         URHO3D_FILEWATCHER
         URHO3D_IK
+	URHO3D_RAYCASTVEHICLE
         URHO3D_LOGGING
         URHO3D_LUA
         URHO3D_MINIDUMPS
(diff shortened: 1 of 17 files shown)


slapin:
I wonder why all win32 shared library builds failed... What do I need to change to fix it?

weitjong:
Didn't look at the code yet but from the build  log it seems due to missing new symbols in the DLL that your sample referencing to.

slapin:
Strange, I did not do any external references, probably
some files were not linked into the library. I wonder why, as Linux builds fine for both
static and shared.

iSLC:
If I remember correctly, this depends on some new (_and modified_ ) Bullet files. And the Bullet library wasn't updated in a long while. Which is strange, because I tried it locally and I only had to drop the new Bullet sources straight from repository and worked without a hitch.

But back to the main issue, please include the Bullet changes for this to work.

slapin:
Now, I did not change Bullet in any way.

slapin:
Well, all AppVeyor  builds succeeded, I wonder what is going on with Travis win32 shared builds...

weitjong:
PR only goes through limited CI build test. On AppVeyor it only uses VS to test and not the complete VS and MinGW-W64 builds, or otherwise I believe you would get the same error on the latter.

cadaver:
Thanks for the work, will see how this develops! As long as the build is sorted out, I would be fine with raycast vehicle being compiled in whenever URHO3D_PHYSICS is specified, so I don't see the separate build option as absolutely necessary.

slapin:
Removed Bullet include file from sample header, will see what outcome will be. Too bad I never set up
mingw builds...

slapin:
Well, I know what is going on, but I wonder what to do - if bullet construct is accessed from
.h file the undefined reference is produced. If inside .cpp file - no error is produced.
I don't really like having all these tiny accessors in .cpp, hoped them to be inlined...

cadaver:
In this kind of case, where otherwise a large(r) part of the third party library would be exposed because of inlining, the accessors should be in .cpp. Compare to e.g. RigidBody or Constraint class. The header will look cleaner too as a result.

slapin:
The remaining errors are these:
CMakeFiles/46_RaycastVehicle.dir/objects.a(RaycastVehicleDemo.cpp.obj):RaycastVehicleDemo.cpp:(.text$_ZN17btTypedConstraintD0Ev[btTypedConstraint::~btTypedConstraint()]+0xb): undefined reference to `btAlignedFreeInternal(void*)'
CMakeFiles/46_RaycastVehicle.dir/objects.a(RaycastVehicleDemo.cpp.obj):RaycastVehicleDemo.cpp:(.data$_ZTV17btTypedConstraint[vtable for btTypedConstraint]+0x60): undefined reference to `btTypedConstraint::serialize(void*, btSerializer*) const'

And I'm really out of options now - I see that these are use all over bullet even in btRigidBody but
for mingw win32 build these are omitted somehow. And why these are triggered by header inclusion?

weitjong:
The builds are queueing up on the AppVeyor. In the interest of time, I will cancel those jobs coming from your PR as I need the computing resource for other tests. Please do not be alarmed.

slapin:
It loos like I will not be able to resolve this any time soon, so I close this PR now
for better times and not consuming anyone's resources nor wasting time.

weitjong:
But why? The CI jobs queuing up was just a temporary situation. In any case, I respect your decision. Feel free to reopen and submit more commits later if you decide to reopen it.

slapin:
Well, I don't want it to hang there for no reason.

1. I don't know any way to fix this. Absolutely no idea.
And there's no indication I will have any idea in near future.
My knowledge of C++ and Urho build system is a not on level where
I can do any educated guesses. I don't know what Urho build system does
differently
in case of WIN32.
2. I will not have time to set up mingw build for at least a month to find
out.
3. I will not get any feedback until it builds.

So there's no reason for PR to just hang there.

So for now I will slowly improve things locally in hope that
some other time I will be able to resolve the build issue somehow.
As soon as I will be able to test mingw locally I will give it another try.

slapin:
Well, a friend of mine helped me to setup mingw find the solution (actually did all the work)
so the reason is that just inclusion of btRaycastVehicle.h produces unresolved references.
I reimplemented RaycastVehicle function using opaque data, so no bullet stuff is required in header,
which builds and works fine.

PR #1905, opened 2017-04-12
Added missing includes.

ombre5733:
The header ValueInformationInfo.h did not include the Variant.h and so its usage was sensitive to
the include oder.

### DIFF
--- a/Source/Urho3D/Scene/ValueAnimationInfo.h
+++ b/Source/Urho3D/Scene/ValueAnimationInfo.h
@@ -22,7 +22,9 @@
 
 #pragma once
 
+#include "../Container/Ptr.h"
 #include "../Container/RefCounted.h"
+#include "../Core/Variant.h"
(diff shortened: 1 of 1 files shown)


ombre5733:
At least Ptr.h must be included because SharedPtr cannot be fwd declared. For Variant a fwd declaration could be added I guess.
Currently application code does not compile if ValueInformationInfo.h is the first include.

cadaver:
It's true inclusion of an include file shouldn't ever cause compile error, however the class API doesn't necessarily have to be fully usable after that without additional includes (if you're creating Variants to pass to this class, it makes sense you need to include Variant.h yourself) 

Ptr.h and Vector.h are necessary, Variant can be forward declared like you said. Can make the changes and close this then. Thanks for bringing the issue to attention!

ombre5733:
@cadaver Thanks for taking care of this! :-)

PR #1906, opened 2017-04-12
Fixed: Erasing sibling nodes from component destructor can result in segfault.

ssinai1:


cadaver:
Thanks for noticing! Will add a note describing why this is necessary.

ssinai1:
It is necessary to keep the reference of the node, because it must not be destructed in the Vector::Moverange() function. Destructing here, and calling a "siblingNode->Remove()" in one of its (user defined) component destructor can corrupt the array..(sorry for my english)

PR #1908, opened 2017-04-12
fixes #1907 - Compiler error when URHO3D_IK=OFF

### LINKED ISSUE #1907: Compile fail with -DURHO3D_IK=0
When i try to compile the engine without the IK module. Visual studio fail to compile Context::RequireIK() and void Context::ReleaseIK() functions. 

It seems that those function are not encapsulated inside any define like #ifndef URHO_IK

TheComet93:
Is the sample 45_InverseKinematics supposed to fail compilation if you set URHO3D_IK=OFF? I don't see any header guards around the physics samples (though I haven't tried compiling the physics samples with URHO3D_PHYSICS=OFF).

### DIFF
--- a/Source/Urho3D/AngelScript/IKAPI.cpp
+++ b/Source/Urho3D/AngelScript/IKAPI.cpp
@@ -20,7 +20,7 @@
 // THE SOFTWARE.
 //
 
-#ifdef URHO3D_PHYSICS
+#ifdef URHO3D_IK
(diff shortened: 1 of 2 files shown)


TheComet93:
I see that disabling physics also causes the samples that use physics to be disabled. How this happens, however, appears to be pure magic. I just don't understand this build system.

PR #1919, opened 2017-04-25
IKSolver.spp: PVS-Studio: Non-void function should return a value

SvyatoslavRazmyslov:
We have found and fixed a bug using PVS-Studio tool. PVS-Studio is a static code analyzer for C, C++ and C#: https://www.viva64.com/en/pvs-studio/

We suggests having a look at the emails, sent from @pvs-studio.com.

Analyzer warning: [V591](https://www.viva64.com/en/w/V591/) Non-void function should return a value. IKSolver.cpp 57


cadaver:
Thanks!

PR #1921, opened 2017-04-26
Libs update

cosmy1:


iSLC:
These would've looked better if there was one per pull request. But I guess that' fine.

cadaver:
Thanks for the work. These are fairly small and isolated libraries in their use, so as long as nothing appears to break it can actually be convenient to test and merge as bulk.

weitjong:
Merged to a new remote branch for full CI build tests.

PR #1927, opened 2017-05-03
Minor fix and helper function

eugeneko:
Added helper function to get an axis/angle from Quaternion.
Also fixed a bug when Urho crashes if AS ScriptObject throw an exception from ctor.


cadaver:
Looks good.

eugeneko:
Any ideas why it's failed?

BTW, @weitjong, may you somehow grant me an ability to re-start jobs in AppVeyor?

weitjong:
Most probably it was just a technical error. The CI jobs come in pair of group. A genuine error would usually trip in pair. So when you see a lone CI Job error for no obvious reason then don't think too much about it. Of course, if you want to be sure then the job(s) can be re-triggered. I have just done that.

Which come to your next question. Currently no one else can retrigger beside myself. The whole setup in Urho3D is in fact using my personal AppVeyor account, i.e. I have cloned my personal setup for Urho3D. I think it is good idea to setup a separate account for Urho3D as an org or team so we could have multiple users or collaborators. Will do that when I have time.

eugeneko:
Thank you!

PR #1928, opened 2017-05-04
Update rapidjson lib

1vanK:
Lib now supports comments in JSON file: http://rapidjson.org/md__c_h_a_n_g_e_l_o_g.html

p.s. I'm not sure that correct change JSONFile::Save(Serializer& dest, const String& indendation) function - it requires some allocator instead poolallocator


cadaver:
It is using the document's allocator (whatever subclass / template class that is) so looks correct.

cadaver:
Will merge. Thanks for the work!

PR #1930, opened 2017-05-05
40_Localization: Center mouse on start

1vanK:


1vanK:
trailing commas also is part of json5 as well as comments

PR #1931, opened 2017-05-08
Query the amount of installed memory

Florastamine:
Only works on Linux and Windows atm. (I don't have access to other platforms for proper testing).


cadaver:
Can you make it dynamic load the symbol on Windows? Otherwise this will bump Urho's Windows version requirement to Vista SP1, while in theory otherwise nothing prevents it running on very ancient systems (XP, possibly even 2000)

Stinkfist0:
`GlobalMemoryStatusEx()` could be used instead for XP support.

Florastamine:
Done. I switched to use `GlobalMemoryStatusEx()` instead of `GetPhysicallyInstalledSystemMemory()` as suggested. Also there's no need to do dynamic loading as the API is useable on XP systems and newer.

Florastamine:
I'll close the PR for now as I'm still having a few more functions to contribute to.

PR #1932, opened 2017-05-08
Add Object::IsTypeOf to AngelScript API.

KonstantinTomashevich:
It could be useful in some cases. For example in my game I have abstract `Unit` component and it's implementations: `FleetUnit`, `TradersUnit`, `ColonizatorsUnit` and `ArmyUnit`. And if have an array of `Unit`'s, this function is only one way to determinate which unit type is it.
P.S. I don't know, how to make same lua binding without changing C++ `Object` code.

### DIFF
--- a/Source/Urho3D/AngelScript/APITemplates.h
+++ b/Source/Urho3D/AngelScript/APITemplates.h
@@ -363,6 +363,11 @@ template <class T> bool ObjectHasSubscribedToSenderEvent(Object* sender, const S
     return ptr->HasSubscribedToEvent(sender, StringHash(eventType));
 }
 
+template <class T> bool ObjectIsTypeOf(const String& typeName, T* ptr)
+{
+    return ptr->GetTypeInfo()->IsTypeOf(StringHash(typeName));
+}
+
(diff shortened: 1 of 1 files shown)


eugeneko:
I doubt that it is good solution.
`IsTypeOf` is strict type check that is the fastest but not the best soluton for hierarchies.
Why not to use plain cast?

KonstantinTomashevich:
@eugeneko So, maybe plain cast is better idea. My general idea to add `IsTypeOf` to AngelScript API is to use the same functionality as in C++.

eugeneko:
Even in C++ code dynamic_cast is the best way unless you are working with performance-critical code. Urho's type info system is a kind of low-level high-performance casts mostly for internal usage, hidden under ResourceCache or Node interfaces.

AngelScript don't need it in most cases, for other cases something like `Object.type == "ClassName"` shall be enough. So, `Object.IsTypeOf("ClassName")` seems duplicate.

> Should I close PR?

I don't enforce you, but if you agree with my reasonings, then close.

KonstantinTomashevich:
I close the PR. Plain cast works well in my situation.

PR #1934, opened 2017-05-09
Platform-specific system query functions

Florastamine:
Continuation of PR #1931 

Looks good on my machine. Only supports Windows and Linux platform at the moment. The Windows API used requires Windows 2000 Pro/XP or higher.

### DIFF
--- a/CMake/Modules/UrhoCommon.cmake
+++ b/CMake/Modules/UrhoCommon.cmake
@@ -857,7 +857,7 @@ macro (define_dependency_libs TARGET)
             if (IOS OR TVOS)
                 list (APPEND LIBS "-framework AudioToolbox" "-framework AVFoundation" "-framework CoreAudio" "-framework CoreGraphics" "-framework CoreMotion" "-framework Foundation" "-framework GameController" "-framework OpenGLES" "-framework QuartzCore" "-framework UIKit")
             else ()
-                list (APPEND LIBS "-framework AudioToolbox" "-framework Carbon" "-framework Cocoa" "-framework CoreAudio" "-framework CoreServices" "-framework CoreVideo" "-framework ForceFeedback" "-framework IOKit" "-framework OpenGL")
+                list (APPEND LIBS "-framework AudioToolbox" "-framework Carbon" "-framework Cocoa" "-framework CoreFoundation" "-framework SystemConfiguration" "-framework CoreAudio" "-framework CoreServices" "-framework CoreVideo" "-framework ForceFeedback" "-framework IOKit" "-framework OpenGL")
             endif ()
         endif ()
 
--- a/Docs/AngelScriptAPI.h
+++ b/Docs/AngelScriptAPI.h
@@ -15945,6 +15945,7 @@ String GetParentPath(const String&);
 String GetPath(const String&);
 String GetPlatform();
 uint64 GetTotalMemory();
+String GetLoginName(); 
 uint GetRG16Format();
 uint GetRGBA16Format();
 uint GetRGBAFloat16Format();
--- a/Docs/AngelScriptAPI.h
+++ b/Docs/AngelScriptAPI.h
@@ -15946,6 +15946,7 @@ String GetPath(const String&);
 String GetPlatform();
 uint64 GetTotalMemory();
 String GetLoginName(); 
+String GetHostName();
 uint GetRG16Format();
 uint GetRGBA16Format();
 uint GetRGBAFloat16Format();
--- a/Docs/AngelScriptAPI.h
+++ b/Docs/AngelScriptAPI.h
@@ -15944,6 +15944,7 @@ Array<String> GetObjectsByCategory(const String&);
 String GetParentPath(const String&);
 String GetPath(const String&);
 String GetPlatform();
+uint64 GetTotalMemory();
 uint GetRG16Format();
 uint GetRGBA16Format();
 uint GetRGBAFloat16Format();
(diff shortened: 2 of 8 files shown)


Florastamine:
It builds and runs fine on my local machine, as well as on Travis g++. Don't know why Travis MinGW rejected it. :(

cadaver:
You could try downgrading to the exact same compiler version as used by CI. If it barfs on the typedef specifically, could you write the function pointer the hard way into the function, instead of using a typedef at all?

Florastamine:
Ok. I have pushed out a patch which fixed the `NTSTATUS` typedef issue. I took a quick look at the CI MinGW headers (`x86_64-w64-mingw32-gcc-4.6.3-2`, same version) and found out that they also defined it. So it wasn't about `NTSTATUS` being left out in earlier MinGW releases but I suspect that they changed the definition/location of `NTSTATUS` (it's in `<ntdef.h>`, but my MinGW 6.3.0 and 5.3.0 went fine without including that header, whereas the CI version which is 4.6.3 fails if the header wasn't explicitly included).

cadaver:
Good! Could push this then into a feature branch for full CI tests.

Florastamine:
So... is there anything left I should do?

cadaver:
The branch is pushed now, we'll see how it compiles in the full CI battery.

cadaver:
You can look at the branch build status at https://travis-ci.org/urho3d/Urho3D/branches . OSX failed due to missing include. Android failed too apparently, but I didn't look in detail.

Florastamine:
Corrected. I forgot to exclude other platforms, so when building on OSX the Linux headers got accidentally included.

cadaver:
Android fails due to HOST_NAME_MAX being undefined. I did a fast search through one version of NDK and didn't find it in any of the include files, which was surprising.

Florastamine:
Strange. These functions are only meant to support Windows and Linux. So I isolated the appropriate headers (in `ProcessUtils.cpp`, at the beginning). But the headers weren't somehow isolated, which would cause `HOST_NAME_MAX` being "leaked out" in the Android build and thus caused the error. Same as with OSX, the missing header was intended for Linux. Would you mind checking the most recent commit (specifically the `#ifdef`-header include to see if there's anything I'm missing?

Florastamine:
```
#if defined(_WIN32)
	#include <windows.h>
	#include <io.h>
	#if defined(_MSC_VER)
		#include <float.h>
		#include <Lmcons.h> // For UNLEN. 
	#elif defined(__MINGW32__)
		#include <lmcons.h> // For UNLEN. 
		#include <ntdef.h> 
	#else 
	#endif 
#elif defined(__linux__)
	#include <pwd.h> 
	#include <unistd.h>
	#include <sys/sysinfo.h>
	#include <sys/utsname.h>
	#include <limits.h> // For HOST_NAME_MAX. 
#else
#endif
```

This is what I've got in the lastest commit, don't know if there's something wrong. Linux headers are not supposed to "leak" into the OSX/Android build.

cadaver:
Linux would also be defined on Android I believe.

weitjong:
> Linux would also be defined on Android I believe.

I can definitely confirm that.

Florastamine:
Thanks. The build should be smooth now.

cadaver:
You also need to exclude the functionality, if it does not work without the includes.

Florastamine:
Problem is I don't have the appropriate platforms for proper implementation (e.g. OS X). Might need some time for all the functions to have full support for whatever systems Urho can be compiled.

cadaver:
Yes, in that sense Urho has a significant barrier to contribution, when it's not something that is e.g. covered by SDL for all platforms automatically. I'll see if I get inspired / if I have time / if it's easy to add OSX corresponding functionality, but no hard promises.

cadaver:
Fixed enough of the OSX functions so that I'm satisfied. I didn't bother with the OS version for OSX at least for now. Pushed the changes to the feature branch and we'll see how the build of all platforms goes now.

Florastamine:
Great, thanks! By the way, two things: 

I took a quick glance at your code and noticed you were using `getenv()`, which is quite unreliable. A user could alter the `USER` variable (for whatever purpose) during the current session and `GetLoginName()` would return the wrong result. 

```
NDMKs-Mac:Desktop ndmkhoi$ ./Urho3DPlayer PlatformAPI.lua
Logon:     ndmkhoi
Computer name:     NDMKs-Mac.local
NDMKs-Mac:Desktop ndmkhoi$ export USER="NewName"
NDMKs-Mac:Desktop ndmkhoi$ ./Urho3DPlayer PlatformAPI.lua
Logon:     NewName
Computer name:     NDMKs-Mac.local
NDMKs-Mac:Desktop ndmkhoi$ 
``` 

- I have access to a borrowed Mac now, which means that I'll be able to inspect and fill in the missing details for `GetOSVersion()` and also correct `GetLoginName()`.

cadaver:
Ok, thanks. Please use the platform-query-functions branch as a base for further modifications.

Florastamine:
Done. Please review. In `OSGetVersion()` I have added detection support for Mac OS X Lion (10.7) and higher, but older versions can be added with very minimal effort. 

Result from my device: 
```
Login name:     ndmkhoi
Computer name:     NDMKs-Mac.local
Total useable memory:     4096     (MiB)
OS version:     macOS/Mac OS X El Capitan 10.11.6  (Darwin kernel 15.6.0)
```

cadaver:
You could use ToInt() from StringUtils.h instead of calling atoi() manually to be more "Urho-like". Will review in detail later, hopefully this day.

cadaver:
Pushed into the Urho feature branch and fixed compile on iOS.

cadaver:
CI is down for the moment, but did manual builds to verify, and should be OK now (had to just fix iOS to not attempt calling unsupported functions related to the login name.) Thanks for the work!

PR #1939, opened 2017-05-11
Add support of arbitrary data in Sound, Animation and Texture resources.

eugeneko:
I found it useful.
Please, review. Discussion about naming is open.

It would be great to have such ability for all binary resources, but I don't know nice way to implement it.

Example:
```
Animation@ anim = cache.GetResource("Animation", "Animations/Kachujin_Idle.ani");
Print(anim.metadata["speed"].GetFloat());
anim.metadata["dir"] = Variant(Vector3(1, 1, 1));
Array<Variant> vec;
vec.Push(Variant("String"));
vec.Push(Variant(10));
vec.Push(Variant(15.0));
anim.metadata["vec"] = Variant(vec);
anim.Save(cache.resourceDirs[0] + anim.name);
```

### DIFF
--- a/Source/Urho3D/AngelScript/APITemplates.h
+++ b/Source/Urho3D/AngelScript/APITemplates.h
@@ -834,6 +834,33 @@ template <class T> void RegisterResource(asIScriptEngine* engine, const char* cl
     engine->RegisterObjectMethod(className, "uint get_useTimer()" ,asMETHODPR(T, GetUseTimer, (), unsigned), asCALL_THISCALL);
 }
 
+static void ResourceAddMetadata(const String& name, const Variant& value, ResourceWithMetadata* ptr)
+{
+    ptr->AddMetadata(name, value);
+}
+
+static const Variant& ResourceGetMetadata(const String& name, ResourceWithMetadata* ptr)
+{
+    return ptr->GetMetadata(name);
+}
+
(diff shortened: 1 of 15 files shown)


cadaver:
It's a good idea. Would probably be worth it to use this mechanism to replace all the existing resource "side-by-side" xml's, e.g. textures and sounds too, but that would be a larger work.

cadaver:
If you want to merge this before release, it's OK by me, as it's still useful, and expanding it to encompass more resources would be a large work.

eugeneko:
I'll look into this PR this day, then decide.

eugeneko:
Ok, I implemented this that way.

I dislike names like `ObjectWithSomething`, but I don't know better name. I decided to integrate this feature into main hierarchy to avoid copy-paste and keep hierarchy simple.

It is also possible to make metadata container class outside of hierarchy and use multiple inheritance to attach it to resources. However, this makes hierarchy more complicated and probably causes problems e.g. in C# facade.

cadaver:
Looks fine. Should also be fairly safe as nothing in the existing side-by-side xml parsing changes. Can go in to master by me.

eugeneko:
Ok, I tested it a bit. I'll merge it now, I hope I won't break anything.

PR #1940, opened 2017-05-11
Help the compiler eliminate calls to strlen at compile time, wherever possible.

iSLC:
Also Fix add-assign operator of arbitrary types to return a reference and not a copy.

Normally you would see this using something like `constexpr` but since there's heap memory involved, might have undesired results. However, any decent compiler is able to fold `LiteralStrLen` at compile-time.

I know these kinds of additions are likely to be discarded but I thought I should make a pull request anyway.

Performance wise, I opened the Character demo and left it on idle after I pressed F2 to bring the debug text (so there's some actual text being used). Without this, the CPU was averaging around 19-21% and with it, around 8-10%. Might have to run some further tests to see how much (if any) performance is gained.

I've tried this with MinGW 7.1 x64. Although, theoretically, the code should be C++98 compatible.

### DIFF
--- a/Source/Urho3D/Container/Str.h
+++ b/Source/Urho3D/Container/Str.h
@@ -36,6 +36,50 @@ static const int MATRIX_CONVERSION_BUFFER_LENGTH = 256;
 
 class WString;
 
+/// Helper structure used as a type trait to differentiate between character arrays and a character pointers.
+template <typename T, typename U> struct IsCharPtr
+{
+};
+/// Specialization for constant character pointers.
+template <typename T> struct IsCharPtr<const char*, T>
+{
+    typedef T type;
+};
+/// Specialization for character pointers.
+template <typename T> struct IsCharPtr<char*, T>
+{
+    typedef T type;
+};
+/// Helper structure used as a type trait to make a distinction between other types and character pointers.
+template <typename T, typename U> struct IsNotCharPtr
+{
+    typedef U type;
+};
+/// Specialization for constant character pointers.
+template <typename T> struct IsNotCharPtr<const char*, T>
+{
+};
+/// Specialization for character pointers.
+template <typename T> struct IsNotCharPtr<char*, T>
+{
+};
+
+/// Exclude (at compile-time, if possible) the null character from the actual size of string literals and/or array of caharacters.
+template < unsigned N > inline unsigned LiteralStrLen(const char(&a)[N])
+{
+    return a[N-1] == '\0' ? N-1 : N;
+}
+/// Exclude (at compile-time, if possible) the null character from the actual size of string literals and/or array of caharacters.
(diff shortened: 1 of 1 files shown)


eugeneko:
Why do we need `char*` and `const char*` overloads? (I definetly dislike SFINAE in C++)

`LiteralStrLen` shall fallback to `strlen` if last symbol is not zero.

iSLC:
I agree with the `strlen` suggestion.

As for why there's `char` and `const char`. Because `char` should be treated as a regular string pointer since situations like these are more common:

```
    char b[32];
    sprintf(b, "test");
    // dumb example but shows what I mean
    Urho3D::String s(b);
```

And if I remove it then it'll be treated as a `const char` and there's no point in trying to identify the length at compile time.

eugeneko:
Huh... Now I see. Actually, you have just illustrated the problem with this code.

This code is a kind of hack. You assume that `const chat[]` is probably literal and `char[]` is probably buffer, so you get compile-time length for first and compute length for second.

However, these assumptions are not stated in code. That's the way to surprises.

If I had such problems, I would pre-allocate strings (that gives much more performance) or use special keyword like `Literal("some string")` to explicitly flag places where it is possible to get string length in compile-time (if I had tons of small strings).

PS: Neither STL nor EASTL have such optimization, BTW.

iSLC:
I don't really assume anything. I simple let the compiler decide since it can do that on compile time. There is exactly one overload, the `const char` one. The other `char` overload is simple so that it won't be covered by the `const char` overload.

The behavior should be exactly the same as before with a single rare (_at least I'd like to think so_) exception:

```
    const char a[] = {'a', 'b', 'b', '\0', 'x', 'y', 'z', '\0'};
    Urho3D::String s1(a);
    const char * p = a;
    Urho3D::String s2(p);
    printf("S1 = %u __ S2 = %u", s1.Length(), s2.Length());
```

Which outputs `S1 = 7 __ S2 = 3`. Everything else works exactly the same as you'd expect it to work.

Even though I believe this to be a rare case. For someone to manually create a constant array with a series of characters that may contain multiple null terminators and one at the end. And then simply passing that onto a string.

But this isn't my decision to make. So I think I'll just close this for now. Sorry for the late reply.

eugeneko:
That case is actually synthetic and not a problem.

My concern was mostly about another. You cannot check on the language level if `const char[]` is literal. In some cases `char[]` may be treated as `const char[]`. No one can swear that `const char[]` have valid symbols at the end instead of garbage.

So, I'd like to avoid any checks that just rely on probably garbage memory. Check this example and how it would work with your ctors.
https://ideone.com/Z8TrS4

iSLC:
I see now. I guess I'll have to wait until the c++11 switch and I can make a user defined literal to take care o this.

PR #1941, opened 2017-05-11
Random seek while playing audio file

meshonline:
Hi,

My project uses a huge audio file to play sounds, and I want to play only certain pieces of the audio file.

So I achieved a random seek function to jump to any time while playing, and it supports all audio file types except buffered stream, for buffered stream usually feeds data on the fly.

Usage:
// Jump to 12.5 seconds from beginning.
musicSource->Seek(12.5f);

Regards,

Art

### DIFF
--- a/Source/Urho3D/Audio/SoundStream.h
+++ b/Source/Urho3D/Audio/SoundStream.h
@@ -36,6 +36,9 @@ public:
     /// Destruct.
     ~SoundStream();
 
+    /// Seek to sample number
+    virtual int Seek(unsigned int sample_number) = 0;
+    
     /// Produce sound data into destination. Return number of bytes produced. Called by SoundSource from the mixing thread.
     virtual unsigned GetData(signed char* dest, unsigned numBytes) = 0;
 
--- a/Source/Urho3D/Audio/SoundStream.h
+++ b/Source/Urho3D/Audio/SoundStream.h
@@ -37,7 +37,7 @@ public:
     ~SoundStream();
 
     /// Seek to sample number
-    virtual int Seek(unsigned int sample_number) const { return 1; }
+    virtual int Seek(unsigned int sample_number) = 0;
     
     /// Produce sound data into destination. Return number of bytes produced. Called by SoundSource from the mixing thread.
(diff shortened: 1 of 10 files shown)


cadaver:
Two things, please mind the coding convention regarding on which lines braces are inserted, and script bindings. Otherwise looking good.

Source/Urho3D/Audio/SoundStream.h
```diff
@@ -36,6 +36,9 @@ class URHO3D_API SoundStream : public RefCounted
     /// Destruct.
     ~SoundStream();
 
+    /// Seek to sample number
+    virtual int Seek(unsigned int sample_number) const { return 1; }
```
eugeneko: Please comment return value.

Why 1?

meshonline: Fixed it, should work now.

eugeneko:
I still don't understand meaning of Seek return value.
BTW, why do you need `#include "../Audio/OggVorbisSoundStream.h"`?

meshonline:
It does not need to include "../Audio/OggVorbisSoundStream.h", I will remove the line later.
The return value is C style like that in the STB module, I will change to C++ style later too.

meshonline:
I have never tried script binding before, please help.

cadaver:
The short of it is that for Angelscript, you add the function definitions to AudioAPI.cpp to the relevant place (there is a function per class that registers bindings) and for Lua, you add the new function to SoundSource.pkg in Source/Urho3D/LuaScript/pkgs/Audio. You pretty much learn by looking at the existing functions' bindings.

meshonline:
AudioAPI.cpp has not export useful SoundSource interface, so I will not touch the file, which means that we can not use the method in Angel script, I only added the method to Lua bindings.

cadaver:
In this case SoundSource AS registration uses helper functions in APITemplates.h to register both classes' base interface, look there and you'll find it does expose things, and Seek would already go there.

meshonline:
Yes, I see, finally I add the method into AudioAPI.cpp.
Thanks you, I have learned a lot today.

cadaver:
Merged as squashed. Thanks!

PR #1962, opened 2017-05-27
Fix RigidBody2D failure to apply transforms that would return it to o…

### LINKED ISSUE #1960: RigidBody2D fails to apply transforms that would return it to original configuration
[Here](https://github.com/urho3d/Urho3D/blob/84c8be164b5187c10edf9a1ad04315d094194b34/Source/Urho3D/Urho2D/RigidBody2D.cpp#L600-L603). The code does not agree with the comment when `body_` is not null.

ricab:
…riginal position and rotation; fixes #1960

### DIFF
--- a/Source/Urho3D/Urho2D/RigidBody2D.cpp
+++ b/Source/Urho3D/Urho2D/RigidBody2D.cpp
@@ -104,7 +104,7 @@ void RigidBody2D::OnSetEnabled()
 
 void RigidBody2D::SetBodyType(BodyType2D type)
 {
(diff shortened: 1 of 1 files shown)


ricab:
Hi, could I get some feedback on this please? (other than seemingly unjustified thumbs down by strange person :stuck_out_tongue: )

cadaver:
I will still have to review properly, have had little time. Should be certainly this weekend.

cadaver:
Looked OK, so merged. Thanks!

ricab:
My pleasure

PR #1966, opened 2017-06-01
Re-factored PhysicsWorld2D::ContactInfo. Removed NodeEndContact2D::P_…

### LINKED ISSUE #1958: Crash: PhysicsWorld2D uses destroyed contact when overlap ceases in broadphase
When `b2ContactManager` finds that a former contact ceases to overlap, it [destroys the contact](https://github.com/urho3d/Urho3D/blob/84c8be164b5187c10edf9a1ad04315d094194b34/Source/ThirdParty/Box2D/Box2D/Dynamics/b2ContactManager.cpp#L162). Note that [`EndContact` is called](https://github.com/urho3d/Urho3D/blob/84c8be164b5187c10edf9a1ad04315d094194b34/Source/ThirdParty/Box2D/Box2D/Dynamics/b2ContactManager.cpp#L46) on `PhysicsWorld2D`, which saves a pointer to the expiring contact before it is actually destroyed. Later on (but still in the same `PhysicsWorld2D::Update` call) [`PhysicsWorld2D::SendEndContactEvents` is called](https://github.com/urho3d/Urho3D/blob/84c8be164b5187c10edf9a1ad04315d094194b34/Source/Urho3D/Urho2D/PhysicsWorld2D.cpp#L350). This method [tries to serialize the recently destroyed contact into the event](https://github.com/urho3d/Urho3D/blob/84c8be164b5187c10edf9a1ad04315d094194b34/Source/Urho3D/Urho2D/PhysicsWorld2D.cpp#L815), dereferencing rubbish and causing a crash.

### LINKED ISSUE #1965: [retirement request] Remove b2Contact pointer from EndContact event
According to #1958, b2Contact pointer in EndContact event may be dangling and end user is unable to check it. In fact, this is the root cause of the bug. 
I don't see a nice way to fix it, so I suggest to completely remove the pointer from the event.
To compenaste this I suggest to keep contact info in `ContactInfo`.

Any objections, @cadaver ?

eugeneko:
…CONTACT and PhysicsEndContact2D::P_CONTACT event parameters. Fixed #1958. Fixed #1965.

BTW, there is one more step possible that I am unsure about.
I can remove P_CONTACT parameter from other events too.

**Cons:**
- Broken compatibility.

**Pros:**
+ Encapsulation of b2 internals;
+ Consistency of event parameters;
+ Avoid misuse of b2Contact;
+ This stuff is unusable from scripts anyway;

Or (at least) consider this parameter as deprecated and remove it in the future.

I also want to rename recently added P_CONTACTPOINTS to P_CONTACTS (like in bullet physics events).
I've chosen first name to avoid ambigity with P_CONTACT. However, if P_CONTACT is deprecated and almost useless, ambigity doesn't matter very much.

### DIFF
--- a/Docs/Urho3D.dox
+++ b/Docs/Urho3D.dox
@@ -1089,6 +1089,7 @@ From 1.6 to master:
 - AnimatedModel allows unclamped morph weights, also negative.
 - StaticModel::SetModel() will no longer warn and redirect to AnimatedModel::SetModel() if the wrong function is called from C++ code. The AngelScript API will instead redirect properly without producing a warning.
 - Graphics::SetMode(), Graphics::GetDesktopResolution() and Graphics::GetResolutions() have gained an extra parameter to specify the monitor to use. Use Graphics::GetMonitorCount() to get the number of attached monitors. Monitor number 0 means the default primary monitor.
+- P_CONTACT parameter was removed from 2D physics collision events. Use P_CONTACTS buffer instead.
 - Build system - the "Urho3D-CMake-common.cmake" file is now renamed to "UrhoCommon.cmake".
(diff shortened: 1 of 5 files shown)


cadaver:
It's fine by me to change the event parameters in a breaking manner (in the other events too), you can just mark the changes in the porting notes. If I got it right the contact parameters initially appeared in 1.6.

Issue #1970, opened 2017-06-05
Couldn't set position for borderless window


pengfei666:
Not sure it is a bug or an intended behavior.

When I want to create a fixed size borderless window, but it seems to ignore the window position I set.

The engine parameters are below
```
    engineParameters_[EP_BORDERLESS] = true;
    engineParameters_[EP_WINDOW_POSITION_X] = 100;
    engineParameters_[EP_WINDOW_POSITION_Y] = 100;
    engineParameters_[EP_WINDOW_WIDTH] = 500;
    engineParameters_[EP_WINDOW_HEIGHT] = 500;
```
I get the below result:
![4_h 9z kzvt0n_ gxtjj](https://cloud.githubusercontent.com/assets/25144819/26776667/b1b00c70-4a0c-11e7-966f-4d99208d21a4.png)

pengfei666:
UPDATE: I find this issue in OpenGL mode and have a fix for it. I am seeing whether it works in DX mode.

I will open a PR for that later.

pengfei666:
After my PR:

![pho95i owyi_0mzasmdj 4](https://cloud.githubusercontent.com/assets/25144819/26777043/53976ce4-4a0e-11e7-838a-0f4a701030e2.png)

cadaver:
Merged PR which fixes this.

PR #1971, opened 2017-06-05
Pengfei666/bordless window without position

pengfei666:
For issue https://github.com/urho3d/Urho3D/issues/1970

pho95i owyi_0mzasmdj 4

### DIFF
--- a/Source/Samples/Sample.inl
+++ b/Source/Samples/Sample.inl
@@ -63,6 +63,11 @@ void Sample::Setup()
     engineParameters_[EP_HEADLESS]     = false;
     engineParameters_[EP_SOUND]        = false;
 
+    engineParameters_[EP_BORDERLESS] = true;
+    engineParameters_[EP_WINDOW_POSITION_X] = 100;
+    engineParameters_[EP_WINDOW_POSITION_Y] = 100;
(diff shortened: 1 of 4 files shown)


cadaver:
Thanks. Will need to review. I believe the original intention of borderless has been to be a substitute for fullscreen, in which case the position wouldn't matter.

eugeneko:
If borderless window is used without specifying position/size, it is fullscreen in version _before_.
How does it behave in version _after_?

pengfei666:
Okay. Wait a moment. I will paste the picture.

pengfei666:
z681fhjf w3i z 1vp l k

cadaver:
Since we got just some time ago the ability to position the window on any monitor in multi-monitor setups, that might get broken now for borderless windows? So what it possibly needs to do is to still do the auto-positioning for both borderless & fullscreen, if a position was not explicitly specified. How to detect that situation, is another thing.

castiel520:
yeah, I think window positioning is fixed for borderless if a position was specidied. 
but what if we want to drag a borderless window?

cadaver:
I believe I managed to make a change that should enable borderless monitor select to still work properly. May not be perfect but can be improved later. Will merge / push shortly.

cadaver:
Merged, with changes.

PR #1972, opened 2017-06-05
Static VertexBuffer::UpdateOffsets()

henu:
I noticed VertexBuffer has many static functions to handle `PODVector<VertexElement> elements` that are outside VertexBuffer. However, it is missing a static function to update offsets of elements, so static `VertexBuffer::GetElementOffset()` does not work in certain situations. This PR adds a new static function that user can manually call to calculate offsets.

I tested it in one of my projects and it seems to work.

Another option would be to not use precalculated values, and do more manual work in static `VertexBuffer::GetElementOffset()`.


cadaver:
Thanks. Shouldn't be a problem to merge. It's a simplified version, since this doesn't concern with a vertex buffer's element hash / mask, but that's OK.

PR #1977, opened 2017-06-08
Fix bug in FontFaceFreeType::CanLoadAllGlyphs that can make text vanish

### LINKED ISSUE #1976: Text elements occasionally don't render
This is hard to reproduce but it occurs with certain fonts at certain sizes. I reproduced it with the BlueHighway font at 20 points, but only after forcing the load mode to FT_LOAD_NO_HINTING (I'm experimenting with FreeType settings).

iainmerrick:
Closes #1976

FontFaceFreeType checks at runtime whether the whole font can be loaded
into a single texture, by preflighting all the necessary texture atlas
placements in CanLoadAllGlyphs.

However, this preflight operation was slightly different from the real
texture atlas generation, so sometimes CanLoadAllGlyphs was incorrectly
returning true. The LoadCharGlyph would then switch to a new texture
atlas (via SetupNextTexture). This would usually make Text elements
vanish, as the rendered glyphs were in the wrong texture.

We fix this bug with two changes:
- Make CanLoadAllGlyphs process the same char codes in the same order
- Make CanLoadAllGlyphs skip zero-size glyphs (as LoadCharGlyph does)

We also log an error if SetupNextTexture is called unexpectedly, to
help catch future occurrences.

### DIFF
--- a/Source/Urho3D/UI/FontFaceFreeType.cpp
+++ b/Source/Urho3D/UI/FontFaceFreeType.cpp
@@ -160,25 +160,24 @@ bool FontFaceFreeType::Load(const unsigned char* fontData, unsigned fontDataSize
     // Load each of the glyphs to see the sizes & store other information
     loadMode_ = (int)(ui->GetForceAutoHint() ? FT_LOAD_FORCE_AUTOHINT : FT_LOAD_DEFAULT);
     ascender_ = RoundToPixels(face->size->metrics.ascender);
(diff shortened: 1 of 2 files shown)


cadaver:
Thanks! Should review shortly, likely during weekend.

iainmerrick:
Great, thanks! I just added a second commit that tweaks the code some more, to fix a related bug.

After this I plan to try changing some of these integer fields to floats, and adding an option for sub-pixel character positioning, as discussed on the forum.

PR #1980, opened 2017-06-12
Add a 'Typography' sample program

iainmerrick:
Displays text in various sizes, with checkboxes to play with settings that affect the way text is rendered:
- Color (white-on-black versus black-on-white)
- Autohinter
- SRGB output conversion (not available on all platforms)

### DIFF
--- a/Source/Samples/47_Typography/CMakeLists.txt
+++ b/Source/Samples/47_Typography/CMakeLists.txt
@@ -1,0 +1,33 @@
+#
+# Copyright (c) 2008-2017 the Urho3D project.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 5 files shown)


iainmerrick:
Not sure if this is the correct style for sample programs (for example, do I need to add a Lua version?)

cadaver:
Thanks! It was not stated in the contribution checklist (added just now), but yes, the idea would be to implement the sample in C++ and both scripting languages. 

In this case the pedagogic need isn't that great, since creating text elements in all APIs has already been demonstrated in earlier samples, and checking the text rendering quality & correctness with different options is the most important point of this sample. But to keep consistency, it would still be good to have the script versions too.

iainmerrick:
Fixed Travis build errors and added Lua and AngelScript versions. Ready for another look now.

cadaver:
Thanks, will see when have time, shouldn't take long.

Issue #1916, opened 2017-04-24
PBR Artifacts


rasteron:
Not really using PBR at the moment, but checking the demos I noticed some pixel artifacts being rendered. Maybe someone can clarify if this can be tweaked or fixed by setup, but you can also clearly see it on some submitted high resolution screenshots like this one here: http://i.imgur.com/4PdYkGA.png

1vanK:
try it https://github.com/urho3d/Urho3D/pull/1495/commits/4effd6f60693862ce618414c5262efce42b904bf

dragonCASTjosh:
Not sure what is causing the issue without finding time to look into. My initial guess would be the fresnel term within the BRDF as that's is the area the fresnel affects. It could also be an issue with IBL but I will have to look into it fully to find the problem,

@1vanK 
I dont think it has anything to do with the minimum roughness because it is affecting 0.5 roughness materials.

@rasteron If you want a quick resolution then applying AA may hide the issue. Alternatively, you could try changing the Fresnel model in use https://github.com/urho3d/Urho3D/blob/master/bin/CoreData/Shaders/HLSL/BRDF.hlsl#L29

rasteron:
Thanks guys, just pointing this one out. Actually not just me as I have said I'm not doing any PBR stuff atm. I sure hope this is just a minor issue, but it would be nice to update the setup or demos as others might stumble and encounter the same problem. 

Please discard or close this issue if needed, thanks.

dragonCASTjosh:
@rasteron what API's did you test on and is there any easy way to reproduce the issues as i dont have it on my local D3D11 build

rasteron:
It's on both D3D9 and OpenGL, have not tested with D3D11 though but will try this next.

rasteron:
Same result with D3D11.

![pbr](https://cloud.githubusercontent.com/assets/3676827/25736898/540bb9ac-31a7-11e7-9c70-2db992332c93.png)

dragonCASTjosh:
Interesting as I can't replicate on desktop or laptop and I'm not sure why

rasteron:
Thanks Josh. As I have stated, maybe this is just an isolated case so just throwing this out there for reference as others might find the same issue.

BTW, you may have noticed that the first screenshot there was from you on one of your forum PBR threads.

KonstantinTomashevich:
[...]

PR #1981, opened 2017-06-13
Fixed PBR artifacts

dragonCASTjosh:
Fixed the issues noted in https://github.com/urho3d/Urho3D/issues/1916

This was done by rebuilding the sphere lights, changing the visibility model and expanding the diffuse and fresnel models.

The new diffuse model should be more performant due to much less and much more simple math in the shader.

In addition to fixing the issue noted i also resolved another issue i didnt realise existed until tackling this fix, the old PBR implementation didnt correctly match the physical light values, now the light values should map to real world equivalents.


slapin:
Are these changes the same for OpenGL and DirectX?

cadaver:
Merged, thanks!

Issue #1952, opened 2017-05-24
FontFace should allow fractional font sizes


iainmerrick:
Currently the FontFace class only allows integral sizes (11pt, 12pt, etc).

Internally, FontFaceFreeType multiplies the size by 64, so it looks like FreeType is capable of rendering fonts in steps of 1/64th of a point. It would be good to make that functionality available in Urho. That would make it easier to render crisp text at arbitrary scales.

I'm happy to try implementing this! But I could use some feedback on the best approach:

- Just change pointSize from an int to a float? That changes the API, but I think almost all client code would work unchanged.

- Add new parallel methods that take floats, alongside the existing int methods. Safer but makes the API more complicated.

cadaver:
We make API-breaking changes from time to time, so I'd be fine with the first option. Like you say most code would still continue to work, just making an (ugly) implicit conversion where necessary.

iainmerrick:
Great, I'll give it a try. Thanks for the feedback, I appreciate it!

iainmerrick:
It looks like this change doesn't work right now, because of a bug in FreeType. Here's a long discussion: http://lists.nongnu.org/archive/html/freetype-devel/2017-04/msg00042.html

Quick version: glyphs are being incorrectly rounded to pixel boundaries.

It's fixed in 2.8, so updating our copy of FreeType should allow fractional sizes to work correctly.

iainmerrick:
Yep, confirmed that fractional sizes work in 2.8. (I'm working on the Urho API change for that)

PR #1984, opened 2017-06-15
Update FreeType to 2.8.

iainmerrick:
This is needed for correct handling of fractional sizes (issue #1952).

I downloaded http://download.savannah.gnu.org/releases/freetype/freetype-2.8.tar.gz and copied over the src, include and docs directories. I kept the Urho3D patch in src/gzip. Compiles and seems to work. :)

If you have a more formal process for updating third-party libraries, let me know!


cadaver:
For 3rd party libraries that aren't yet as own repos and used as subtrees, this procedure is fine. Note that documentation refers to lib versions (README.md and Urho3d.dox). Can fix in this case. Will now make a branch for complete CI-testing first.

cadaver:
CI builds looked ok, merged now!

Florastamine:
Just a very minor thing: IMO the `docs/` folder should be removed as there's no real need for it.

cadaver:
True, at least the reference folder seems extraneous, since we don't keep full API documentations to other 3rdparty libs either.

PR #1987, opened 2017-06-16
Added "emitting" flag for ParticleEmitter2D

kostik1337:
Flag that enables/disables emitting particles in ParticleEmitter2D, just the same as in ParticleEmitter


Source/Urho3D/Urho2D/ParticleEmitter2D.cpp
```diff
@@ -325,23 +336,26 @@ void ParticleEmitter2D::Update(float timeStep)
         }
     }
 
-    if (emissionTime_ > 0.0f)
+    if (emitting_)
```
eugeneko: Any chance to use 'logical AND' here?

Source/Urho3D/AngelScript/Urho2DAPI.cpp
```diff
@@ -198,6 +198,8 @@ static void RegisterParticleEmitter2D(asIScriptEngine* engine)
     engine->RegisterObjectMethod("ParticleEmitter2D", "Sprite2D@+ get_sprite() const", asMETHOD(ParticleEmitter2D, GetSprite), asCALL_THISCALL);
     engine->RegisterObjectMethod("ParticleEmitter2D", "void set_blendMode(BlendMode)", asMETHOD(ParticleEmitter2D, SetBlendMode), asCALL_THISCALL);
     engine->RegisterObjectMethod("ParticleEmitter2D", "BlendMode get_blendMode() const", asMETHOD(ParticleEmitter2D, GetBlendMode), asCALL_THISCALL);
+    engine->RegisterObjectMethod("ParticleEmitter2D", "void set_emitting(bool)", asMETHOD(ParticleEmitter2D, SetEmitting), asCALL_THISCALL);
+    engine->RegisterObjectMethod("ParticleEmitter2D", "bool is_emitting() const", asMETHOD(ParticleEmitter2D, IsEmitting), asCALL_THISCALL);
```
eugeneko: Use `get_` prefix here

cadaver:
Thanks, believe I will change the attribute to accessor type, so that side-effects are respected, and make it a default type attribute, since it's only 1 bool more and would possibly be useful also for net synchronization.

Issue #1953, opened 2017-05-24
Text class should allow per-character subpixel positioning


iainmerrick:
This is another feature suggestion, a followup to #1952 (allow fractional font sizes).

Inspecting both the code and the rendered output, it looks like Text objects align each character to a pixel boundary. That's OK at large sizes but leads to awkward letter spacing at small sizes.

It would be great to align characters horizontally at subpixel offsets, say 1/4 pixel offsets. I think that could be done by extending FontFace to cache up to 4 renderings of each glyph.

That would obviously use more memory, but it's only needed for small fonts, which don't take up much memory in the first place! So we could use an adaptive strategy -- single pixel alignment for large fonts, half-pixel for medium fonts, quarter-pixel for small fonts, etc.

This will become less important as high-resolution screens get more common, but many users still have low-res screens. As Urho aims to be a lightweight framework, it would be good to have high-quality text rendering by default on those low-end devices.

cadaver:
You're welcome to work on this. Possibly would still be preferable if the multiple copies mechanism can be disabled completely if necessary. There are already functions in the UI subsystem which configure behavior for all fonts globally. See e.g. SetUseMutableGlyphs, SetForceAutoHint.

iainmerrick:
Just to confirm that I'm working on this. Now that I understand the code better, I plan to do it in two steps:
1. Add a "subpixel character positioning" option that positions chars using floats
2. Add a "horizontal oversampling" option to improve rendering quality

I belatedly realized that we don't need N separate copies of each glyph for item 2, we just need a single stretched and filtered image. That makes things a lot simpler! The code changes shouldn't be too big.

iainmerrick:
Forgot about item 0 :). We need to be able to disable hinting, otherwise subpixel positioning makes no sense!

iainmerrick:
To show the current state of play, here's what you get with the normal autohinter:

![autohinted_text](https://user-images.githubusercontent.com/252111/27597610-7ade2506-5b28-11e7-911f-1a5eb9f8fbd9.png)

And here's unhinted text with subpixel alignment:

![unhinted_subpixel_text](https://user-images.githubusercontent.com/252111/27597619-7f3ffd4a-5b28-11e7-989b-b304c4b973a5.png)

Pretty big difference, and it's up to game authors to decide what settings are most appropriate for them.

[...]

PR #1991, opened 2017-06-20
Add a 'subpixel glyph positions' option to UI (#1953)

iainmerrick:
When this option is enabled, text will be formatted with subpixel
(fractional) positions on the x-axis. Positions on the y-axis are
still pixel-aligned.

Note that this option has no effect if the hinting level is set
to FONT_HINT_LEVEL_NORMAL, as each glyph is rounded to an integral
pixel size by the hinter. It only makes a different if the hinting
mode is LIGHT (vertical hinting only) or NONE.

With subpixel positioning, the output will look blurrier due to
texture filtering. TODO: Add horizontal oversampling to improve
sharpness. That needs extra memory so it should be configurable.

### DIFF
--- a/Source/Samples/47_Typography/Typography.cpp
+++ b/Source/Samples/47_Typography/Typography.cpp
@@ -81,16 +81,20 @@ void Typography::Start()
     CreateCheckbox("White background", URHO3D_HANDLER(Typography, HandleWhiteBackground))
         ->SetChecked(false);
 
-    // Add a checkbox for the global ForceAutoHint setting. This affects character spacing.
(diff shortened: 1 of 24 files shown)


iainmerrick:
I've added a boolean UI::SubpixelGlyphPositions flag, but I'm not sure that that's the best approach -- there are an awful lot of text configuration options now!

We're going to need at least one more option, some kind of threshold for oversampling. But maybe that can be combined with the subpixel flag, as you probably wouldn't want subpixel positions without oversampling in practice.

cadaver:
I think that's OK, that they're "global" in UI instead of being set for each Text element individually, at least.

cadaver:
Will merge, nothing at least in the examples and editor appears to break.

PR #1997, opened 2017-06-25
Add Vector3::Orthogonalize helper function.

eugeneko:


cadaver:
Should be OK to go in.

PR #1999, opened 2017-06-30
Explicitly call RefCounted constructor.

iSLC:
ValueAnimationInfo was not calling the RefCounted constructor.


weitjong:
I don't think it is needed.

cadaver:
Yes, lots of other places where the default constructor of base class isn't called, for example for Object.

PR #2000, opened 2017-06-30
Add configurable oversampling to improve subpixel font rendering (#1953)

iainmerrick:
Subpixel-positioned text looks blurry, due to bilinear filtering of the
underlying texture. The basic idea here is to stretch the font glyphs
horizontally to increase the sharpness at subpixel positions. Note that
the stretched images need to be smoothed to avoid aliasing artifacts;
this is done in the FontFaceFreeType::BoxFilter() method.

Glyphs are always pixel-aligned vertically, so no vertical oversampling
is needed.

To make this feature comprehensible (I hope!) I've removed the
'subpixelGlyphPositions' flag and replaced it with a couple of values:
'fontSubpixelThreshold' sets the point at which subpixel positioning
kicks on, and 'fontOversampleLimit' caps the maximum amount of stretching.

The default values are 12pt text and max 3x oversampling. These are fairly
conservative settings, which should improve small text without wasting
a lot of memory.

Note that when the font hint level is NORMAL (the default), subpixel
positioning and oversampling are both disabled. So, this feature doesn't
change any default behavior, and applies some hopefully sensible values
if the hint level is set to LIGHT or NONE.

### DIFF
--- a/Docs/Reference.dox
+++ b/Docs/Reference.dox
@@ -2234,6 +2234,18 @@ For FreeType fonts, it is possible to adjust the positioning of the font glyphs.
 </font>
 \endcode
 
+The \ref UI "UI" class has various global configuration options for font rendering. The default settings are similar to Windows-style rendering, with crisp characters but uneven spacing. For macOS-style rendering, with accurate spacing but slightly blurrier outlines, call \ref UI::SetFontHintLevel "UI::SetFontHintLevel(FONT_HINT_LEVEL_NONE)". Use the Typography sample to explore these options and find the best configuration for your game.
+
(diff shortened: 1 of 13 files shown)


cadaver:
Thanks, again! Since the features related to font hinting /  rendering are already complex, it'd be appreciated if you would also write a bit onto the "User interface" Doxygen page in Docs/Reference.dox. The Typography sample is a very good live test, but it's still good to also have textual reference of the concepts.

iainmerrick:
No problem, will do!

cadaver:
Thanks again! Will do minor cleanup of comments / docs and merge. Nothing at least appeared to explode and the changes to font rendering are quite contained despite the potential for a huge change.

PR #2002, opened 2017-07-01
Minor incoherence fix

Florastamine:
Different platform-specific functions return different values upon failure, for example one returns `String::EMPTY` while another returns `(?)`. From now on everything will return `(?)` if the function doesn't succeed.

### DIFF
--- a/Source/Urho3D/Core/ProcessUtils.cpp
+++ b/Source/Urho3D/Core/ProcessUtils.cpp
@@ -540,7 +540,7 @@ String GetLoginName()
 {
 #if defined(__linux__) && !defined(__ANDROID__)
     struct passwd *p = getpwuid(getuid());
-    if (p) 
+    if (p != NULL) 
         return p->pw_name;
 #elif defined(_WIN32)
     char name[UNLEN + 1];
@@ -580,7 +580,7 @@ String GetHostName()
     if (GetComputerName(buffer, &len))
         return buffer;
 #endif
-    return String::EMPTY; 
+    return "(?)"; 
 }
 
 // Disable Windows OS version functionality when compiling mini version for Web, see https://github.com/urho3d/Urho3D/issues/1998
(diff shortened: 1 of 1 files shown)


cadaver:
Thanks. I'll further change GetPlatform() to do the same, though it should be unlikely to hit that path.

Source/Urho3D/Core/ProcessUtils.cpp
```diff
@@ -540,7 +540,7 @@ String GetLoginName()
 {
 #if defined(__linux__) && !defined(__ANDROID__)
     struct passwd *p = getpwuid(getuid());
-    if (p) 
+    if (p != NULL)
```
weitjong: I cannot help but notice this inconsistency from our code convention. The former form is the desired convention (https://urho3d.github.io/documentation/HEAD/_coding_conventions.html).

Doing a quick scan using ```git grep``` between current master and 1.6 shows that the IK subsystem is the biggest offender.

```
git grep -w NULL
git grep -w NULL origin/1.6
```

Florastamine: While this is indeed violated the coding convention, I saw that comparing to NULL (`!= NULL`) would make it easier for the later plan of porting the code to C++11 (I saw a thread discussing about the move to C++11 somewhere if I remember right), you can just do a simple find & replace (`!= NULL` with `!= nullptr`). With only `if (p)` instead of `if (p != NULL)`, it's a little harder.

weitjong:
Although that is a good and valid point, it could probably be months away before the C++11 ball will start rolling. Also, when that time comes we could use Clang modernizer tool to automate the migration instead of doing everything by hand.

Source/Urho3D/Core/ProcessUtils.cpp
eugeneko: I suppose even in C++11 epoch pointer-to-null comparison will never be used.

Florastamine:
I haven't thought of using automated tools at all. Nice catch.

PR #2005, opened 2017-07-04
Add CSV and Base64 encodings for TmxFile (#1325)

### LINKED ISSUE #1325: TmxFile using deprecated file format
The TmxFile loading implementation is using uncompressed XML and was [officially deprecated](http://forum.mapeditor.org/t/tiled-0-15-0-released/1001) by @bjorn in January. This means any maps saved with version 0.15.0+ will not load. 

This was also [mentioned in a PR](https://github.com/urho3d/Urho3D/pull/641#issuecomment-106905685) and suggested to support an alternate format, preferably the default of the latest version as Tiled is getting closer to 1.0.

kostik1337:
TmxFile currently supports only deprecated XML layer format. This PR adds support for CSV and Base64 formats. I'm not sure it fixes #1325, since compression is still not supported (zlib nor gzip). Base64 decompression snippet taken from https://stackoverflow.com/a/13935718 with some minor modifications to work with Urho's containers instead of std

### DIFF
--- a/Source/Urho3D/Core/StringUtils.cpp
+++ b/Source/Urho3D/Core/StringUtils.cpp
@@ -31,6 +31,11 @@
 namespace Urho3D
 {
 
+static const String base64_chars =
+        "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+        "abcdefghijklmnopqrstuvwxyz"
+        "0123456789+/";
+
 unsigned CountElements(const char* buffer, char separator)
 {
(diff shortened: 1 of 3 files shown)


cadaver:
Thanks. Will have to include the base64 license since otherwise the codebase would be in violation, also since the base64 code is in-line in the Urho StringUtils.cpp file, I'll modify to Urho's code convention.

cadaver:
Merged, with changes.

PR #2008, opened 2017-07-05
Scale kerning values correctly (closes #2007)

### LINKED ISSUE #2007: Kerning values are not scaled correctly
When reading TrueType kerning tables, we don't currently scale the kerning values to match the size of the font. This means the kerning is too tight at small font sizes, as seen in this screenshot (the smallest size actually ends up with negative width):

kerning_bug

iainmerrick:


1vanK:
after some patches example 35_SignedDistanceFieldText does not works correctly

iainmerrick:
Ack! I'll take a look.

iainmerrick:
The C++ version of 35_SignedDistanceFieldText works, but the AngelScript version doesn't. Looks like the text geometry is wrong -- it's squished into thin strips. Most likely some bug in the API bindings.

iainmerrick:
Filed issue #2035

1vanK:
Thanks!

PR #2010, opened 2017-07-05
Fix Graphics::GetHighDPI() on OS X (closes #2009)

### LINKED ISSUE #2009: Graphics::GetHighDPI() always returns true on OS X
The docs for GetHighDPI() say "Return whether window is high DPI", but in fact it just returns the value of the EP_HIGH_DPI engine parameter, which defaults to true.

GetHighDPI() should check whether the window is really high DPI.

iainmerrick:


cadaver:
Thanks! Will likely change D3D9/11 to behave similarly, e.g. also send the highdpi flag on resize event, though it's always false.

iainmerrick:
Yeah, I left D3D alone because it looks like it always just ignored highdpi. Good idea to make it send the flag on the event anyway!

PR #2012, opened 2017-07-08
Background loading: add resource into cache before sending E_RESOURCE…

Mezozoysky:
Hello! I found that E_BACKGROUNDLOADED listeners are not able to use just loaded resources becouse of resources are not exist in the cache when event is raised. So, where is a fix, as i tough. Hope, it will be merged. Thanx.


cadaver:
Thanks! Makes sense.

PR #2020, opened 2017-07-13
Fixed ordered comparsion between pointer and zero in AssetImporter.cpp

ArXen42:
Tried to build Urho3D using clang, got error on that line. Seems like error: `mNumRotationKeys` and `mNumPositionKeys` are `unsigned int`'s while `mScalingKeys` is `C_STRUCT aiVectorKey*`, so comparing this pointer to the inequality with NULL seems more logical.

Hope I did not break anything with that change.


Lumak:
That looks more like a typo and probably meant to use the 
`mNumScalingKeys`
variable consistent with the other two comparators instead of a pointer.

eugeneko:
@Lumak ++

cadaver:
Agreed, very much looks like an oversight / typo.

cadaver:
Corrected the variable that was being compared. But thanks for bringing to attention!

PR #2024, opened 2017-07-18
Reimport Mutant with correct forward direction

1vanK:
Src blend files: https://github.com/1vanK/UrhoDataSrc/tree/master/Mutant


eugeneko:
(Offtop) Huh. I definetely hate these binary commits... They make repo grow so fast.

Lumak:
Is this a bug or just your personal preference? As you might be aware, most Unity users will have their fbx models facing -z axis and this sample demonstrates how to correct that in code without having to convert all their models.

Are you simply mandating that Urho3D cannot accept models facing -z axis?

1vanK:
I think in standart pack sould be correct models, that can be used withoud additional manipulations

Lumak:
I don't know what this "standard pack" is or who actually came up with the standard. Was there a committee that decided that all assets must face +z axis?

I see this PR as counter productive and will not answer the questions for people with models facing -z axis how to correct the orientation in game. I had to ask the question on the forum myself to figure out the best way to do this.

PR #2026, opened 2017-07-19
Initial commit of WebP image library

carnalis:
https://developers.google.com/speed/webp/

Urho3D native loading of WebP lossless and lossy formats (*.webp).
Enabled with URHO3D_WEBP cmake build option.
Did basic checks with Valgrind.
SaveWEBP() to be added.


cadaver:
The idea would be to include the webp library in Urho in the ThirdParty directory (like the others), since we can't assume the user has to have it installed. Note that you should modify or recreate the library's build system in CMake so that it builds along with the rest of the libraries.

carnalis:
@cadaver I had set it up as you describe and it correctly installs headers and source to the build tree, but I committed it as a submodule. Remaking branch...

PR #2033, opened 2017-07-20
Add DebugRenderer::AddSphereSector. Expose Sphere helper funcions.

eugeneko:
I also want to re-use AddSphereSector for drawing SoundSource3D.
IMO it's more fair because sector is circle, not pyramidal.
Any objections?

Before:
image
After:
image


cadaver:
Shouldn't be any objections from me.

PR #2034, opened 2017-07-20
Several tilemap improvements

kostik1337:
This PR contains several tilemaps improvements and bugfixes:

- Added support for tilesets, based on collections of images
- Added support for tile collision object groups
- Fixed debug drawing of isometric rectangle and circle objects
- Fixed tilemap debug drawing when node is transformed

Actually, this does not seems completed, I wanted to ask your advice on how to make these right:
1. I wrote function TransformNode2D(Vector2), it is simple helper function, which transforms Vector2 from node's local space to global space. It is used in two classes, and now I just declare it in one class and define in another. Where I should put it? I'm not sure, it doesn't looks so generic to put it in Node class.
2. To load tile collision object group, I copy TileMapInfo2D structure into local var and set there custom width and height, based on current tile width and height. Besides it looks hackish, it will work wrong, if tile width is not multiple map's tile width. I see at least two ways how to fix this: a) Create another class, some kind of "abstract objectgroup info", which will delegate position conversion somewhere else, and b) just make width_ and height_ in TileMapInfo2D some floating-point type instead of int. What do you think?

### DIFF
--- a/Source/Urho3D/AngelScript/Urho2DAPI.cpp
+++ b/Source/Urho3D/AngelScript/Urho2DAPI.cpp
@@ -299,6 +299,12 @@ static void RegisterTileMapLayer2D(asIScriptEngine* engine)
     engine->RegisterObjectMethod("TileMapLayer2D", "Node@+ get_imageNode() const", asMETHOD(TileMapLayer2D, GetImageNode), asCALL_THISCALL);
 }
 
+static CScriptArray* TileMap2DGetTileCollisionShapes(int gid, TileMap2D* tileMap)
+{
+    Vector<SharedPtr<TileMapObject2D> > result = tileMap->GetTileCollisionShapes(gid);
+    return VectorToArray<SharedPtr<TileMapObject2D> >(result, "Array<TileMapObject2D@>@");
+}
+
 static void RegisterTileMap2D(asIScriptEngine* engine)
 {
     engine->RegisterObjectMethod("TileMap2D", "void set_tmxFile(TmxFile2D@+)", asMETHOD(TileMap2D, SetTmxFile), asCALL_THISCALL);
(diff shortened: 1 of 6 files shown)


Mike3D:
I already fixed all the bugs and added all missing tmx features, and also added some experimental features (animations, physics and constraints from tmx, collision shapes using Bayazit decomposition (so that we don't need external tools like PhysicsBodyEditor-Box2DEditor or PhysicsEditor, we create our shapes in Tiled) and a Recast experiment for navigation).

Unfortunatley I never got time to push these (almost 1 year ago). If you have time to work on it, I can push a fork on my github account.

kostik1337:
Sure, I'd like to see it and push into engine, this sounds like exactly what I need! Can't say I will do it immediately, I don't have much time too, but I can find at least some time to push some features

Mike3D:
Thanks!

Archive put here:
https://github.com/Mike3D/Urho3D/commit/24c562755b9df39d3696c8fb506d04ce50bb64d5

After building, run the 5 AS samples to test some of the new features.

I've also included many test tmx files I've used for testing purpose.

Hope this can be helpful to you.

cadaver:
@kostik1337: Should we wait while you evaluate Mike3D's code and possibly update this PR?

kostik1337:
@cadaver Yes, I'm working on it, will complete in the nearest future

cadaver:
Thanks! Will wait then.

kostik1337:
Alright, I have updated PR. I'm sorry it took so long.
I left tileset loading as in my first version, not like in Mike3D's version - I think, loading tileset as one texture is better, then creating a texture for each tile.
Unfortunately, I don't know, when I'll have time to push other Mike3D's changes, so @Mike3D, if you decide to push them by yourself - please let me know

cadaver:
No worries. This looked OK to me. If no other comments can merge to master as is, and further improvements can be done later then.

PR #2038, opened 2017-07-23
Enable multi-thread compilation in Visual Studio. Fixed #1407.

### LINKED ISSUE #1407: Multithreaded compilation for Visual Studio
Visual Studio is able to compile code in several threads.
Unfortunately, VS by default uses only one thread.
Unfortunately x2, it is needed to manually enable this flag for each project.
To enable this feature it is needed to add compiler option '/MP' for each source file.
Can you enable multithreaded compilation for Visual Studio?
Or can you at least tell me where should I add this option to set it globally?

eugeneko:
@weitjong ?


weitjong:
I have commented before that I prefers more how `make` does it, where the option to spawn parallel process is passed at the actual build time and not baked into the solution itself with `/MP` flag. But since I am not a VS user, I don't have strong objection either. If you think it has more benefits than the potential issue it may arise (see https://blog.kitware.com/cmake-building-with-all-your-cores/) then you can go ahead to commit it.

eugeneko:
I've read these old comments, yep.

Unfortunatelly, Urho library is on the critical path during compilation. And Urho.lib actually compiles with single core used. Using MP makes Visual Studio always utilize all cores.

Thanks for link. I actually agree with it:
> However, the speed up is significant so it is usually worth the extra trouble it causes

PR #2041, opened 2017-07-24
Copy bat-files in win32 instead make symlinks

1vanK:
Tested for both console and gui cmake


weitjong:
I have my reservation for this. It is not the size it takes or the speed that lost, but having a hard copy lying around when you know the fact that MKLINK flag is set to true. With soft symlink you can edit the file in the build tree and the master copy in source tree automatically get the update.  I am preventing myself to become a victim in the future here. 

And also because I am not able to reproduce the original issue.  I think you need to get others to vouch for this first. So far I haven't heard other to have this issue of yours.

1vanK:
or we can change bat-files from
```
start "" "%~dp0Urho3DPlayer%DEBUG%" Scripts/Editor.as %OPT1% %*
```
to
```
start "" "%~dp0Urho3DPlayer%DEBUG%" Scripts/Editor.as %OPT1% %*
```

>>> And also because I am not able to reproduce the original issue.

You have repeatedly said that windows 10 is not your platform and when I asked to show your %~dp0 to identify the source of the problem, you responded that you BELIVE that everything is ok.

>>> I think you need to get others to vouch for this first.

I do not know why others are silent. Do not confirm and do not deny. I myself did not have such a problem before, because resources were simply copied and there were no symlinks

1vanK:
> With soft symlink you can edit the file in the build tree and the master copy in source tree automatically get the update.

How often do you edit bat-files to run the editor?

1vanK:
Another variant https://github.com/urho3d/Urho3D/pull/2044

weitjong:
Do not get me wrong here. I have scripted those batch files with same care I have done the shell files a few years ago. Reasonably tests have been done before I make all those commits. Plus the recent tests I did just for your issue. I have actually no obligation to do nor that I am trying to defend my work.

1vanK:
I saw warning in cmake-gui from the very beginning of the use of the engine but did not pay any attention to him (several years already), because everything worked, and resources just copied to build/bin. But I install Windows 10 and now I have symlinks instead copy of resources

weitjong:
My last test was using Windows10.

weitjong:
I am using cmake CLI naturally. Not the cmake-gui. A few years ago, we didn't even support cmake-gui use case.

1vanK:
Change your editor.bat
```
:: @echo off


echo %cd%
echo %~dp0%
pause

if exist "%~dp0Urho3DPlayer.exe" (set "DEBUG=") else (set "DEBUG=_d")
if [%1] == [] (set "OPT1=-w -s") else (set "OPT1=")
start "" "%~dp0Urho3DPlayer%DEBUG%" Scripts/Editor.as %OPT1% %*
```

and show your %~dp0

If %~dp0 is sources/bin then you need to find the reason why in your case everything works. (I guess you have path to SDK in your enveronment variable)

weitjong:
wait until I am home :)

cadaver:
On my Windows 7, I see the symlinked bat files starting e.g. Editor correctly from the build directory. Could test Win10 too.

1vanK:
May be in Windows 10 with some update were changes and in current version %~dp0 for symlink to bat file return target folder (which is logical) but not path to symlink like before

cadaver:
Win10 also tested OK for me from a build directory (no exes were present in source directory).

1vanK:
can you show %~dp0 ?

1vanK:
when I run build\bin\symlink
%~dp0 is sources\bin\ (not build\bin\)
 I do not think there is a way to incorrect use Windows, to get this behavior. May be it be changed with some update

1vanK:
I tested "mklink" command on not my computer with Win 7 x64. Mklink on Win7 required administrative rights. This is the reason why I've never seen a link before. And yes, in Win 7 %~dp0 shows path to symlink unlike my version of Windows

cadaver:
When running a symlinked batch file, for me %~dp0 shows the directory of the symlink, not the original, both Win7 and Win10. Because I'm evil, I run the command prompt as administrator though, but shouldn't make a difference..

1vanK:
Run console as admin not make a difference to me also

weitjong:
First of all, my apology that I forgotten to follow up with the %~dp0% test result on my Window10 earlier. Anyway here is mine now.

```
C:\Users\weitjong\native-Build\bin>Editor.bat
C:\Users\weitjong\native-Build\bin
C:\Users\weitjong\native-Build\bin\

C:\Users\weitjong\native-Build\bin>cd ..\..

C:\Users\weitjong>native-Build\bin\Editor.bat
C:\Users\weitjong
C:\Users\weitjong\native-Build\bin\

C:\Users\weitjong>cd \

C:\>%HOMEPATH%\native-Build\bin\Editor.bat
C:\
C:\Users\weitjong\native-Build\bin\

C:\>set PATH=%PATH%;C:\Users\weitjong\native-Build\bin\

C:\>Editor.bat
C:\
C:\Users\weitjong\native-Build\bin\

```

As you can see it does not really matter where my current directory is, passing a fully qualified path to the symlinked batch file always return the same result for the %~dp0% and it is the directory where the batch file is being executed, in my case it is the "bin" directory of the build tree.

> Because I'm evil, I run the command prompt as administrator though, but shouldn't make a difference..

Ahem, you should say that out loud! BTW, I am using the proper way, MKLINK privilege is given to my normal user account. See the [howto](https://urho3d.github.io/documentation/HEAD/_misc__how_tos.html#symlinking_on_windows). Disclaimer, be sure to check with your Windows Admins first if you are not.

1vanK:
All ignore my question about version of windows :) (command "winver")

weitjong:
Already shutoff my PC and ready for bed. IMHO, it is highly unlikely Windows engineers would break such simple thing as I can guarantee you if they do then it may hit more users than just you. Last time when I was still Windows fans, I actually accustomed to reinstall my OS every now and then to make it shipshape. 😂

1vanK:
Maybe not broken, but repaired)

weitjong:
More plausible cause is this one, you have umlaut or other kind of funny characters in the path. My first fews hit in Google search result shows how the %~dp0 breaks in a some situations. Try to test by changing your build tree location. HTH.

1vanK:
I test "mklink" command manually:
I create 2 files:

c:\test\test.bat
```
echo %~dp0%
pause
```

c:\run\create_link.bat
```
mklink c:\run\link.bat c:\test\test.bat
```

Next I launch c:\run\create_link.bat to create a file c:\run\link.bat

After it is 2 variants:
1) I launch cmd.exe and use command
```
c:\run\link.bat
```

%~dp0 is path to simlink

2) But when I launch link.bat from Explorer or Total Commander



%~dp0 is path to target

1vanK:
When I launch link from another bat-file, %~dp0 is path to parent bat-file

cadaver:
Could replicate the problem when launching bat from explorer, not command prompt.

weitjong:
I could reproduce the problem as well on my Win7 VM when using Explorer. I have totally forgotten of such use case and quite surprise by such inferior behavior to exist. But since the issue here is the variable expansion and not the symlink itself, I think the more proper fix to this issue is to replace the usage of %~dp0 with something else that could give consistent result.

Just for comparison. In Linux host system, double clicking any shell script usually invokes the default action for the item being clicked and in this case the default simply "display" the script for editing. And even when I choose the action to become "run" the script, the symlinked script runs as it should.

weitjong:
Could you raise this as another new issue in the Issue Tracker so that we can close this PR and also the other PR of yours. The action item is to replace the `%~dp0` variable in the batch file.

PR #2042, opened 2017-07-24
Implemented SaveWEBP().

carnalis:


cadaver:
Thanks! Is there some work you haven't pushed yet, as AngelScriptAPI.h has changed, yet there appears no script bindings changes (for the compression parameter) in this PR?

Issue #1957, opened 2017-05-26
IK crash


slapin:
I wonder how this part can work at all:
```c++
    while (ikNode == NULL)
    {
        missingNodes.Push(iterNode);
        iterNode = iterNode->GetParent();
        ikNode = ik_node_find_child(solver_->tree, iterNode->GetID());
    }
```
It always crashes at iterNode->GetID() as iterNode is eventually NULL after
```c++
iterNode = iterNode->GetParent();
```
if ik_node_find_child returns NULL.

And no check is there.
Any fixes?

slapin:
Well, it looks like IK always crashes regardless of the order I create components in. Looks like
something non-obvious is missing. By understanding the code I see that tree was not
created for some reason or was not containing any node it looks for. So I wonder how I'm supposed to use
it.

cadaver:
IK is waiting for a fix PR from @TheComet93, though I don't know if that specific issue is under work. I agree it looks suspect.

slapin:
The problem is that IKSolver::IKSolver is not called on second and following instances of component.
Is there some magic I don't know?

I call 
node_->CreateComponent<IKSolver>();
node_->CreateComponent<IKEffector>();

in all object instances, but I see that IKSolver constructor is called only on first one.
The reason why this happens is beyond my understanding.

Or probably it is related to REPLICATED mode.

0  0x0000000000976aa4 in Urho3D::Node::GetID (this=0x0) at /home/slapin/Urho3D/Source/Urho3D/Scene/../Scene/../Scene/Node.h:338
#1  0x0000000000b86e2e in Urho3D::IKSolver::BuildTreeToEffector (this=0x1b7837c0, node=0xe9467a0) at /home/slapin/Urho3D/Source/Urho3D/IK/IKSolver.cpp:391
#2  0x0000000000b87140 in Urho3D::IKSolver::HandleComponentAdded (this=0x1b7837c0, eventType=..., eventData=...) at /home/slapin/Urho3D/Source/Urho3D/IK/IKSolver.cpp:425
#3  0x0000000000b89b2c in Urho3D::EventHandlerImpl<Urho3D::IKSolver>::Invoke (this=0x1b7c6670, eventData=...) at /home/slapin/Urho3D/Source/Urho3D/IK/../IK/../Scene/../Scene/../Scene/../Core/Object.h:307
[...]

PR #2048, opened 2017-07-28
IK crash fixes

TheComet93:
This should hopefully fix the crashes reported in #1957. @slapin can you verify that this is the case?


cadaver:
Thanks! I should also be able to get into testing this shortly.

Source/ThirdParty/ik/CMakeLists.txt
```diff
@@ -19,11 +19,16 @@
...
 #
 
+include (CheckIncludeFiles)
+
 # Define target name
 set (TARGET_NAME ik)
 
-include (CheckIncludeFiles)
-check_include_files (stdint.h HAVE_STDINT_H)
+check_include_files (stdint.h IK_HAVE_STDINT_H)
```
weitjong: There are two ways to pass the variable from build system to the C++ compiler:

1. def/undef a preprocessor variable either using `cmakedefine` command or  `add_definitions()` conditionally in CMake build script, then using `ifdef / ifndef` in C/C++ code.

2. always define the variable but set its value to 0/1 conditionally, then using `if` directive in code.

I believe Urho prefers the first way.

Source/ThirdParty/ik/CMakeLists.txt
```diff
@@ -32,6 +37,18 @@ endif ()
 option (IK_MEMORY_DEBUGGING "Global switch for memory options. Keep track of the number of allocations and de-allocations and prints a report when the program shuts down" ${DEFAULT_MEMORY_DEBUGGING})
 cmake_dependent_option (IK_MEMORY_BACKTRACE "Generate backtraces for every malloc(), making it easy to track down memory leaks" "${DEFAULT_MEMORY_DEBUGGING}" "IK_MEMORY_DEBUGGING AND NOT WEB" FALSE)
 
+# Enable restrict keyword in quaternion and vector operations if not in debug
+set (IK_RESTRICT "")
+if (NOT CMAKE_BUILD_TYPE STREQUAL Debug)
```
weitjong: This construct only works for a single-config CMake-generator like Makefile or Ninja. This condition always fails for multi-config generator, such as Xcode and VS.

The following loop may also be potentially slow as although CMake by itself caches the c source compile test result, the way you have written makes the loop to always take place. Perhaps should read to see the variable has already been cached and only loop once in the initial configuration.

Source/ThirdParty/ik/CMakeLists.txt
```diff
@@ -38,19 +34,20 @@ option (IK_MEMORY_DEBUGGING "Global switch for memory options. Keep track of the
 cmake_dependent_option (IK_MEMORY_BACKTRACE "Generate backtraces for every malloc(), making it easy to track down memory leaks" "${DEFAULT_MEMORY_DEBUGGING}" "IK_MEMORY_DEBUGGING AND NOT WEB" FALSE)
 
 # Enable restrict keyword in quaternion and vector operations if not in debug
-set (IK_RESTRICT "")
-if (NOT CMAKE_BUILD_TYPE STREQUAL Debug)
+if (NOT IK_RESTRICT)
+    set (IK_RESTRICT "")
```
weitjong: I believe this line is redundant. In Urho we generally does not initialize variable to "".

TheComet93: Is there a way to check if a cmake variable is cached or not? Ideally that's what should be there
```cmake
if (IK_RESTRICT ISNT CACHED)
    ... do the stuff ...
endif ()
```

weitjong: You already do it the right way.

```
if (NOT IK_RESTRICT)
 ... # ONLY DO ONCE
 set (IK_RESTRICT .. CACHE ...)
endif ()
```

I just don't think we need the following line to init to "".

TheComet93: If IK_RESTRICT has an empty value it will always do the check, regardless of whether it's been cached or not (I don't know of any compilers that don't support the restrict keyword however, but this would still be useful to know for other CMake stuff).

This prints "lol" every time you run it:
```cmake
cmake_minimum_required (VERSION 3.3)
project (lalala)
if (NOT lol)
    set (lol ${lol} CACHE STRING "")
    message (STATUS "lol")
endif ()
```

weitjong: Of course, the "lol" snippet of yours won't work. The cache variable value must be "truthy" in order to not get into the protected block again. My previous snippet just want to give you the idea. I should probably use another more proper variable name instead of IK_RESTRICT. So let me try that again.

```
if (NOT IK_RESTRICT_CHECKED)
 message ("LOL once")
 set (IK_RESTRICT_CHECKED 1 CACHE INTERNAL "IK restrict check")
endif ()
```

TheComet93:
This is ready for merging, if there are no further issues.

cadaver:
Ok, thanks, had less time than expected, but will certainly look at it in beginning of next week!

Mike3D:
Do you plan to update your great wiki ?

TheComet93:
@Mike3D Yes certainly, I'll be working on it today and tomorrow

cadaver:
Builds came out OK and samples + editor OK too, will merge.

Issue #2058, opened 2017-08-02
IKSolver should ignore child IKSolver components


TheComet93:
As discussed here: https://discourse.urho3d.io/t/inverse-kinematics-mass-instances/3405/23

It should be possible to have multiple solvers in a skeleton, where one solver can potentially be in a parent solver's subtree. In this case, the parent solver's tree should end at the node the child solver is attached to.

cadaver:
What is your feeling, should the release wait for this and the 0,0,0 effector fix?

TheComet93:
I feel this issue should be fairly trivial to fix, and should definitely be included if the 1.7 release can wait for it.

I've yet to reproduce the 0,0,0 effector bug a second time, but my gut tells me once I find out what's causing it, it's going to be something really stupid and easy to fix.

cadaver:
No problem for me at least to wait a little more. It's good to have IK in a good condition.

Issue #2054, opened 2017-08-02
IK fails to set/adhere to target position setting with Makehuman skeleton


slapin:
With yet unknown circumstances (will debug and add) IK ignores targetPosition setting
using Makehuman skeleton. While it works with stock Urho provided skeletons this
might be by accident. Something is definitely lurking there, will look into this this weekend (unless @TheComet93 provides some solution faster). Now the behavior is like with target set to (0, 0, 0)

Normal scene (IK enabled but not solved):

https://youtu.be/4q72eSFKFiA

Full IK enabled for NPCs (not for player character)

https://www.youtube.com/watch?v=Ca4TuZpX7A4

slapin:
Cc: @TheComet93

TheComet93:
Hopefully we can get to the bottom of how to reproduce this. I would appreciate a minimal example demonstrating effector targets pointing to 0,0,0.

PR #2060, opened 2017-08-06
Iss2058

TheComet93:
This PR contains the following changes:
  - It's now possible to have IKSolver components that share the same subtrees ([#2058](https://github.com/urho3d/Urho3D/issues/2058)).
  - Urho avoids the call to abort() by not calling ik_solver_solve() if the tree is invalid.
  - Fixed a bug in solver_FABRIK_solve() where the root position was incorrectly being set to the original root node position instead of the active root node position. This bug made it impossible to move the root node if USE_ORIGINAL_POSE was disabled.
  - Added a chain tree iterator, which allows Urho to iterate over only the nodes that are affected by the solver instead of having to iterate over all nodes in the tree.
  - The scripting API for SetFeature()/GetFeature() was changed, such that now you can write ```solver.FEATURE_NAME = true``` directly, instead of having to type ```solver.SetFeature(IKFeature.FEATURE_NAME, true)```
  - Updating/Adding docstrings.

There are still two changes I'd like to get in before the 1.7 release. I believe I am onto what's causing the 0,0,0 effector bug, see my thread on more details here: https://discourse.urho3d.io/t/inverse-kinematics/1819/90


Mike3D:
Awesome!
I confirm the fix for #2058 

I don't know if this relates to #2054 but when commenting the PlayAnimation() line in the samples, the solver's behavior becomes weird.

Also noticed that we cannot toggle the IKSolver & IKEffector components (missing virtual void OnSetEnabled();) If it is trivial to add, then it certainly should be in 1.7

Great to see that you are still working on constraints

cadaver:
Thanks! Will look today or soon.

Source/Urho3D/AngelScript/IKAPI.cpp
```diff
@@ -32,32 +32,38 @@
...
+    engine->RegisterObjectMethod("IKSolver", "bool get_UPDATE_ACTIVE_POSE() const", asMETHOD(IKSolver, GetUPDATE_ACTIVE_POSE), asCALL_THISCALL);
+    engine->RegisterObjectMethod("IKSolver", "void set_UPDATE_ACTIVE_POSE(bool)", asMETHOD(IKSolver, SetUPDATE_ACTIVE_POSE), asCALL_THISCALL);
+    engine->RegisterObjectMethod("IKSolver", "bool get_USE_ORIGINAL_POSE() const", asMETHOD(IKSolver, GetUSE_ORIGINAL_POSE), asCALL_THISCALL);
+    engine->RegisterObjectMethod("IKSolver", "void set_USE_ORIGINAL_POSE(bool)", asMETHOD(IKSolver, SetUSE_ORIGINAL_POSE), asCALL_THISCALL);
+    engine->RegisterObjectMethod("IKSolver", "bool get_CONSTRAINTS() const", asMETHOD(IKSolver, GetCONSTRAINTS), asCALL_THISCALL);
+    engine->RegisterObjectMethod("IKSolver", "void set_CONSTRAINTS(bool)", asMETHOD(IKSolver, SetCONSTRAINTS), asCALL_THISCALL);
+    engine->RegisterObjectMethod("IKSolver", "bool get_AUTO_SOLVE() const", asMETHOD(IKSolver, GetAUTO_SOLVE), asCALL_THISCALL);
+    engine->RegisterObjectMethod("IKSolver", "void set_AUTO_SOLVE(bool)", asMETHOD(IKSolver, SetAUTO_SOLVE), asCALL_THISCALL);
```
cadaver: Using uppercase / underscore in object properties is unlike the rest of the Urho AngelScript API. I can see the reasoning, as these are boolean feature flags and not usual properties like "size", "position", so I'm not necessarily asking you to change it. Other opinions are welcome though.

Source/Urho3D/IK/IKEffector.h
```diff
@@ -113,12 +140,13 @@ class URHO3D_API IKEffector : public Component
...
     /// How strongly the target node's rotation influences the solution
     float GetRotationWeight() const;
+
     /*!
      * @brief Sets how much influence the target rotation should have on the
      * solution. A value of 1 means to match the target rotation exactly, if
      * possible. A value of 0 means to not match it at all.
      * @note The solver must have target rotation enabled for this to have
-     * any effect. See IKSolver::EnableTargetRotation().
+     * any effect. See IKSolver::Featire::TARGET_ROTATIONS.
```
cadaver: Typo.

Source/Urho3D/LuaScript/pkgs/IK/IKEffector.pkg
```diff
@@ -37,4 +40,7 @@ class IKEffector : public Component
     tolua_property__get_set float weight;
     tolua_property__get_set float rotationWeight;
     tolua_property__get_set float rotationDecay;
+
+    tolua_property__get_set Feature WEIGHT_NLERP;
+    tolua_property__get_set Feature INHERIT_PARENT_ROTATION;
```
cadaver: The property type should be bool, right?

cadaver:
Made some minor comments.

Source/Urho3D/AngelScript/IKAPI.cpp
TheComet93: Yeah, I'm still not too happy with this. Writing ```solver.SetFeature(IKFeature.FEATURE)``` seemed rather ghastly. Other possible ways to wrap it might be:
  - Add the flags as lowercase under a new structure? ```solver.feature.feature_name = true/false```
  - Turn them into setters/getters? ```solver.SetFeatureName(true/false)```
  - Expose the features flag directly so you can can OR things together? ```solver.features = IKFeature.AUTO_SOLVE | IKFeature.CONSTRAINTS;```

cadaver:
Thanks for the update, going in now.

PR #2067, opened 2017-08-13
IKEffector target bug fix

TheComet93:
This should fix the "(0,0,0) effector" bug.

IKEffector can either explicitly be given a target position with ```SetTargetPosition()```, or it can be given a node name to "follow", in which case IKEffector will search the scene for a matching node name right before solving.

The problem was that it would still perform this search, even with an empty node name. This behaviour is clearly wrong.


cadaver:
Cool, thanks! Does it now look that all the IK work you want in, is in?

PR #2071, opened 2017-08-15
IK documentation

TheComet93:
As the title says.


cadaver:
Thanks!

cadaver:
Think I'll remove the tentacle joke, as there should be no jokes elsewhere either, but otherwise fine, will go in shortishly.

PR #2074, opened 2017-08-18
3D UI

rokups:
This change adds support for rendering GUI into a texture. This is achieved by adding `UIComponent` component to `Node`. `UIComponent` overrides material of `Node`'s `StaticModel`. UI subsystem is changed so that batches belonging to UI elements owned by `UIComponent` are rendered into a texture that is also contained in `UIComponent`. Input is handled by UI subsystem, so anything that worked on normal UI will work on UI rendered on to 3D object. This required a minor change for input handling. Previously "screen" was considered surface of entire window. Now screen is considered area of `root_element_position` + `root_element_size`. This is fully backwards-compatible.

There is one issue i am unable to solve: since on OpenGL texture coordinates start from bottom-left corner a workaround is needed. Texture is flipped by projection matrix, however it breaks scissor test. I do not know how to properly fix it therefore scissor test is disabled when rendering into texture on OpenGL. Issue is clearly visible when scrolling a `ListView` in `HelloGUI` sample when UI is rendered on to a cube. Any help regarding OpenGL issue would be greatly appreciated.

Other changes: lua and angelscript bindings, HelloGUI sample can toggle between rendering on screen or rendering on cube, added a ListView to HelloGUI sample window.

screen

### DIFF
--- a/Source/Samples/02_HelloGUI/HelloGUI.cpp
+++ b/Source/Samples/02_HelloGUI/HelloGUI.cpp
@@ -53,7 +53,8 @@ HelloGUI::HelloGUI(Context* context) :
     uiRoot_(GetSubsystem<UI>()->GetRoot()),
     dragBeginPosition_(IntVector2::ZERO),
     animateCube_(true),
-    renderOnCube_(false)
+    renderOnCube_(false),
+    drawDebug_(false)
 {
 }
 
@@ -119,7 +120,8 @@ void HelloGUI::InitControls()
 
     instructions_ = new Text(context_);
     instructions_->SetStyleAuto();
-    instructions_->SetText("[TAB]   - toggle between rendering on screen or cube.\n[Space] - toggle cube rotation.");
+    instructions_->SetText("[TAB]   - toggle between rendering on screen or cube.\n"
+                           "[Space] - toggle cube rotation.");
     uiRoot_->AddChild(instructions_);
 }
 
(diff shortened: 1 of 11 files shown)


cadaver:
Sweet! Can investigate post 1.7 -release. I'm surprised at the small amount of changes. The OpenGL flip problem shouldn't be insurmountable, will just need a bit of meditation on how to solve the best way.

cadaver:
One thing to check, does this affect the renderpath RENDERUI command, which could already render UI to texture as part of renderpath execution? It works so that View class first setups rendertarget, then calls UI to render.

rokups:
Hmm possibly. It also sounds like a more proper approach too but.. probably 
that does not handle input right? Maybe it would be better to make new 
example rendering UI to texture using your mentioned method and rig input 
handling? I am just not sure how that all is done.

cadaver:
It's a bit different usecase; your method is good for inworld UIs (on object), while RENDERUI would be something fixed like rendering UI to VR dual buffers.

rokups:
@cadaver i took a second look at my changes as yesterday it was pretty late already.

`RENDERUI` command would call rendering to texture as usual, then UI subsystem, when rendering to texture, would do `graphics_->SetRenderTarget(0, surface);` where `surface` is a texture. 
`RENDERUI` command requests to *not reset* render targets. Given that i modified `UI::Render()` method to restore previous render surface after rendering to texture if `resetRenderTargets == false`. I think this should prevent any interference with renderpath `RENDERUI` command.

cadaver:
Ok, will check in more detail when I actually review the code. It sounds a bit like it shouldn't even be UI's responsibility to restore existing target, as usually the caller imperatively does that before, or after. For now you don't have to be concerned of that, I'll make changes if necessary.

cadaver:
Merged with changes. Created a new example, 48. The UI::Render() methods and how they are called in conjunction with the RenderUI renderpath command got tidied up a bit. The flipped projection for OpenGL was good, just finalized the scissor reversing as well.

1vanK:
It's amazing!

stark7:
Thank you both for your hard work!

PR #2084, opened 2017-08-20
Added Clip, Merge, and contructor from pair of IntVector2 to IntRect.

rainerdeyke:
I added a few utility functions to IntRect for better parity with Rect.


cadaver:
Thanks. Can you also update the Lua .pkg?

PR #2085, opened 2017-08-20
Selectable UI element

Enhex:
I ran into a problem where I wanted to have ListView that uses for items an UIElement that contains other elements, but only Text has proper item selection & hover coloring.

Selectable is a class that only implements selection & hover coloring. I refactored Text to inherit from it.

Still missing scripting and editor support.

Also I took the code from my own implementation which uses modern C++, hopefully it isn't a problem.


Enhex:
I sent the PR before the CI was updated to C++11. Please re-run it.

eugeneko:
Restarted.

Enhex:
Error is only with Clang, caused by inheriting constructor.

It also gives warnings about C++11 usage:
`
warning: defaulted function definitions are a C++11 extension [-Wc++11-extensions]
`

Perhaps Clang wasn't properly updated to allow C++11?

weitjong:
No, rerunning CI job is not what you want in this case. Instead, Enhex needs to rebase the PR to take the commit in your own PR.

CI server is just a VM running our build script. It is the build script itself that pulling the string. The VM is just a "blank".

PR #2088, opened 2017-08-21
Update coding conventions for C++11.

eugeneko:
Okay, I suppose someone should try to do it at some point. I updated code standard with C++11 items.

They are more advisory than mandatory but I hope they will prevent style mess in the Urho codebase.

I tried to not cutoff useful C++11 features on the one hand, and keep old code more or less consistent with the new one on the other hand.

I understand that some items are potentially disputable, but I really tried to balance opinions mentioned on the Forum. So, this PR is the last chance to discuss freely.

I still heavily rely on the common sence of contributors. At least, let's document the basic guideline for this common sence.

### DIFF
--- a/Docs/Reference.dox
+++ b/Docs/Reference.dox
@@ -3775,15 +3775,15 @@ byte[]     Bytecode, produced by AngelScript serializer
 
 - Indents use 4 spaces instead of tabs. Indents on empty lines should not be kept.
 
-- Class and struct names are in camelcase beginning with an uppercase letter. They should be nouns. For example %DebugRenderer, FreeTypeLibrary, %Graphics.
+- Class and struct names are in camelcase beginning with an uppercase letter. They should be nouns. For example \c %DebugRenderer, \c FreeTypeLibrary, \c %Graphics.
 
-- Functions are likewise in upper-camelcase. For example CreateComponent, SetLinearRestThreshold.
+- Functions are likewise in upper-camelcase. For example \c CreateComponent, \c SetLinearRestThreshold.
 
-- Variables are in lower-camelcase. Member variables have an underscore appended. For example numContacts, randomSeed_.
+- Variables are in lower-camelcase. Member variables have an underscore appended. For example \c numContacts, \c randomSeed_.
 
-- Constants and enumerations are in uppercase. For example %Vector3::ZERO or PASS_SHADOW.
+- Constants and enumerations are in uppercase. For example \c %Vector3::ZERO or \c PASS_SHADOW.
 
-- Pointers and references append the * or & symbol to the type without a space in between. For example Drawable* drawable, %Serializer& dest.
+- Pointers and references append the * or & symbol to the type without a space in between. For example \c Drawable* drawable, \c %Serializer& dest.
 
 - The macro \c NULL and 0 should not be used for null poitners, \c nullptr is used instead.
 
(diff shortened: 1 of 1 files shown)


Docs/Reference.dox
```diff
@@ -3800,6 +3800,26 @@ byte[]     Bytecode, produced by AngelScript serializer
...
+
+- Avoid \c enum \c class to keep Urho API consistent. TODO: This isn't the best decision.
+
+- Avoid \c auto variables unless it increases code readability. For example, \c auto is tolerable for iterators, pairs and long template types.
+  Inspired by https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-auto.html
+
+- Prefer \c auto to manual type deduction via \c decltype and \c typename.
+
+- Prefer old style constuctor syntax `Vector3(x, y, z)` to new constructor syntax `Vector3{x, y, z}` unless new syntax increases code readability.
+  Example: `Vector3 vec(x, y, z)` but `node.SetPosition({x, y, z})`
```
gogoprog: Would be clearer to say :
Not `node.SetPosition(Vector3(x, y, z))` but `node.SetPosition({x, y, z})`

eugeneko: I'd like to avoid imperative rules because our old codebase may break them.
It is still good point, shall probably re-write.

OvermindDL1:
To follow modern C++ conventions then this document should be followed:  https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md

eugeneko:
@OvermindDL1 It's nice idea if someone starts the project from scratch. We don't have such freedom of choise. Also, it's a bit too large for check-list of things.

OvermindDL1:
> @OvermindDL1 It's nice idea if someone starts the project from scratch. We don't have such freedom of choise. Also, it's a bit too large for check-list of things.

Not necessarily.  It is designed to be introduced incrementally into existing large projects.  There is even beginning work on an auto-linter that uses (and refers back to) these rules with some safe auto-translations to the proper forms as well that can be enabled on a set-by-set basis.  It is recommended that all 'new' code, even code introduced to an old codebase, should follow these rules as have been defined by Bjarne Stroustrup and Herb Sutter (the 2 primary C++ language designers).  There is also a helper library that follows this standard as to the letter as possible for a variety of 0-cost when possible helpers at (which is also being worked on by (Bjarne Stroustrup and Herb Sutter):  https://github.com/Microsoft/GSL

eugeneko:
@OvermindDL1 

> It is designed to be introduced incrementally into existing large projects. 

It's a kind of self-delusion. It's just impossible to keep code consistent and incrementally integrate GSL guidelines. How can we e.g. move to `enum class`es "incrementally"?

It's possible to follow core guidelines in the new community modules and keep Urho core consistent.

OvermindDL1:
> How can we e.g. move to enum classes "incrementally"?

It's been specifically stated, just keep existing ones as they are (unless there is a reason to refactor it), but new ones should use the style specified in the guidelines.  :-)

eugeneko:
@OvermindDL1 So we will end up with mixed codestyle unless we make widespread API-breaking refactoring to rename all enums and items. Do you really think that it worths benefits of `enum class`?

```
enum FaceCameraMode
{
    FC_NONE = 0,
    FC_ROTATE_XYZ,
...
};
enum ShadowQuality
{
    Simple16Bit = 0,
    Simple24Bit,
    PCF16Bit,
    PCF24Bit,
    VSM,
    BlurVSM
};
...
foo(FC_ROTATE_XYZ);
bar(ShadowQuality::Simple24Bit);
```

OvermindDL1:
> @OvermindDL1 So we will end up with mixed codestyle unless we make widespread API-breaking refactoring to rename all enums and items. Do you really think that it worths benefits of enum class?

I personally think so, the benefits in type checking is a great boon.  Though if someone wants to go through and refactor some old code I'd not say no, but I'd not press it either unless I actually came to refactoring it.  :-)

eugeneko:
@OvermindDL1 Satisfactory solution is to prerare script that migrates enum names. Something like massive Find&Replaces batch. In this case we just migrate to `enum class` at some point and deploy migration script. At the same moment the guidline item about enums will be changed.

eugeneko:
Then, I updated some items.

I'm also going to remove item about old/new style constructor, because
1) I don't think it's really important;
2) the best rule is absence of rule.

I am really interested in you opinion about all this stuff, @cadaver

cadaver:
I don't have a specific, strong opinion, but looks OK.

PR #2089, opened 2017-08-21
Remove URHO3D_CXX11 guards.

eugeneko:


CMake/Modules/UrhoCommon.cmake
```diff
@@ -423,17 +421,6 @@ if (NOT URHO3D_LIB_TYPE STREQUAL SHARED AND NOT URHO3D_LIB_TYPE STREQUAL MODULE)
...
 endif ()
 
-# Force C++11 standard (required by the generic bindings generation) if using AngelScript on Web and 64-bit ARM platforms
-if (URHO3D_ANGELSCRIPT AND (EMSCRIPTEN OR (ARM AND URHO3D_64BIT)))
-    set (URHO3D_C++11 1)
-endif ()
-
-# Force C++11 standard (required by nanodbc library) if using ODBC
-if (URHO3D_DATABASE_ODBC)
-    find_package (ODBC REQUIRED)
```
weitjong: The `find_package()` for ODBC is still required.

CMake/Modules/UrhoCommon.cmake
```diff
@@ -475,30 +462,28 @@ if (WIN32 AND NOT CMAKE_PROJECT_NAME MATCHES ^Urho3D-ExternalProject-)
...
-                execute_process (COMMAND ${CMAKE_COMMAND} -E echo COMMAND ${CMAKE_CXX_COMPILER} -std=${STANDARD} -E - RESULT_VARIABLE GCC_EXIT_CODE OUTPUT_QUIET ERROR_QUIET)
-                if (GCC_EXIT_CODE EQUAL 0)
-                    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=${STANDARD}")
-                    set (VERIFIED_SUPPORTED_STANDARD ${STANDARD} CACHE INTERNAL "GNU extension of C++11 standard that is verified to be supported by the chosen compiler")
-                    break ()
-                endif ()
-            endforeach ()
-            if (NOT GCC_EXIT_CODE EQUAL 0)
-                message (FATAL_ERROR "Your GCC version ${CMAKE_CXX_COMPILER_VERSION} is too old to enable C++11 standard")
+add_definitions (-DURHO3D_CXX11)   # Note the define is NOT 'URHO3D_C++11'!
```
weitjong: Don't need this line anymore.

Source/Urho3D/Engine/Console.cpp
```diff
@@ -376,14 +376,10 @@ void Console::HandleTextFinished(StringHash eventType, VariantMap& eventData)
         // Send the command as an event for script subsystem
         using namespace ConsoleCommand;
 
-#if URHO3D_CXX11
-        SendEvent(E_CONSOLECOMMAND, P_COMMAND, line, P_ID, static_cast<Text*>(interpreters_->GetSelectedItem())->GetText());
-#else
         VariantMap& newEventData = GetEventDataMap();
```
weitjong: Should delete all the code block from the `else` side. For example, these lines will cause the same event being sent twice, one in modern way, one in old way.

weitjong:
Btw, since you are one the core devs, you can actually push your commit in a separate new branch instead of pushing it as a PR. The difference is, branches are subjected to full battery of CI matrix while PR only subjected to a small fraction of it.

CMake/Modules/UrhoCommon.cmake
eugeneko: Wow! I am so inattentive. Will revert.

CMake/Modules/UrhoCommon.cmake
eugeneko: I want to leave the define for compatibility with user code that may use it.

Source/Urho3D/Engine/Console.cpp
eugeneko: I... I don't understand the issue, sorry. I removed block under #if guard because we don't use variadic SendEvent in any other place than this.

CMake/Modules/UrhoCommon.cmake
weitjong: Although I see the point but I don't agree with it. We should keep our code clean. If our code internally don't use it anymore then it should be removed sooner rather than later. The next release will be API breaking anyway, so let's the downstream deal with it. They could temporarily add the compiler define in their code if they don't have time to do the necessary, but it's their problem.

Source/Urho3D/Engine/Console.cpp
weitjong: I was the one adding the sample usage there in the first place to demonstrate how it would be done in the future when we have the new standard enabled. Check the commit history and also the rather heated discussion over variadic template implementation in the issue tracker. I didn't change all the rest because I don't want to peppered the code with ifdef/endif everywhere. But to me, now it is the time to rewrite everything and to remove all the old code that look like boilerplate.

eugeneko: I see.
Could you please share a link to discussion?
I haven't managed to find it.

weitjong: I was using my iPhone to review the PR. The small viewport and the mobile version has limit my ability to do the searches. Earlier I thought you just remove the guard and causing the event being sent twice. Only now I see you just remove the `then` block. My apology. Perhaps you could try search for "variadic" or "template initialization" or something like that in the issue tracker and may be also in the forum. Not really sure where the discussion took place anymore or with whom.

eugeneko: Here it is
https://discourse.urho3d.io/t/aggregate-initialization-with-urho3d-containers/1901
I re-thinked the subject... I am still against merging all event parameters into the single line. This is like with variable declarations, you know.
However, I think now that variadic `SendEvent` is better than non-variadic because it doesn't expose internal map to sender.

eugeneko:
Updates are in the branch
https://github.com/urho3d/Urho3D/tree/remove-cxx11-guards

PR #2091, opened 2017-08-22
Attribute refactoring

### LINKED ISSUE #1509: Undefined Behavior in attributes
URHO3D_ATTRIBUTE and URHO3D_ENUM_ATTRIBUTE use offsetof, but offsetof is allowed only for PODs (or standard layout types in C++11)
Actually, any class derived from Object is neither POD nor stdlayout. So, all serialization in Urho3D is based on UB.
http://en.cppreference.com/w/cpp/types/offsetof

Let's think about getting rid of it.

### LINKED ISSUE #1766: Serializable attributes meta information.
Inspired by Unity and #1741
I plan to implement a kind of attribute meta information.

Introduce new (probably virtual) class (e.g. `AttributeMetaInfo`).
Add optional pointer to `AttributeInfo` that specifies attribute meta information.
By design, meta information mustn't be used in Urho core and affect important processes like serialization.
However, it can (and will) be freely used by customers who need some facultative information about attributes.

Examples of meta info:
- Vector information from #1742
- Bitmask view flag
- Realtime apply control (now relatime apply is disabled for some hard-coded attributes)
- (New) Value ranges for numeric attributes and many other things.

I am going to work on it when I publish usable version of my Editor because I am going to use this feature for it. I am __not__ going to add any features to current script Editor.

Opened issue for discussion and as reminder.

eugeneko:
Fixes #1509
Resolves #1766

### DIFF
--- a/Source/Urho3D/Physics/Constraint.cpp
+++ b/Source/Urho3D/Physics/Constraint.cpp
@@ -84,42 +84,17 @@ void Constraint::RegisterObject(Context* context)
     context->RegisterFactory<Constraint>(PHYSICS_CATEGORY);
 
     URHO3D_ACCESSOR_ATTRIBUTE("Is Enabled", IsEnabled, SetEnabled, bool, true, AM_DEFAULT);
-    URHO3D_ENUM_ATTRIBUTE("Constraint Type", constraintType_, typeNames, CONSTRAINT_POINT, AM_DEFAULT);
-    URHO3D_ATTRIBUTE("Position", Vector3, position_, Vector3::ZERO, AM_DEFAULT);
-    URHO3D_ATTRIBUTE("Rotation", Quaternion, rotation_, Quaternion::IDENTITY, AM_DEFAULT);
-    URHO3D_ATTRIBUTE("Other Body Position", Vector3, otherPosition_, Vector3::ZERO, AM_DEFAULT);
-    URHO3D_ATTRIBUTE("Other Body Rotation", Quaternion, otherRotation_, Quaternion::IDENTITY, AM_DEFAULT);
-    URHO3D_ATTRIBUTE("Other Body NodeID", unsigned, otherBodyNodeID_, 0, AM_DEFAULT | AM_NODEID);
+    URHO3D_ENUM_ATTRIBUTE_EX("Constraint Type", constraintType_, MarkConstraintDirty, typeNames, CONSTRAINT_POINT, AM_DEFAULT);
+    URHO3D_ATTRIBUTE_EX("Position", Vector3, position_, AdjustOtherBodyPosition, Vector3::ZERO, AM_DEFAULT);
+    URHO3D_ATTRIBUTE_EX("Rotation", Quaternion, rotation_, MarkFramesDirty, Quaternion::IDENTITY, AM_DEFAULT);
+    URHO3D_ATTRIBUTE_EX("Other Body Position", Vector3, otherPosition_, MarkFramesDirty, Vector3::ZERO, AM_DEFAULT);
+    URHO3D_ATTRIBUTE_EX("Other Body Rotation", Quaternion, otherRotation_, MarkFramesDirty, Quaternion::IDENTITY, AM_DEFAULT);
+    URHO3D_ATTRIBUTE_EX("Other Body NodeID", unsigned, otherBodyNodeID_, MarkConstraintDirty, 0, AM_DEFAULT | AM_NODEID);
     URHO3D_ACCESSOR_ATTRIBUTE("High Limit", GetHighLimit, SetHighLimit, Vector2, Vector2::ZERO, AM_DEFAULT);
     URHO3D_ACCESSOR_ATTRIBUTE("Low Limit", GetLowLimit, SetLowLimit, Vector2, Vector2::ZERO, AM_DEFAULT);
     URHO3D_ACCESSOR_ATTRIBUTE("ERP Parameter", GetERP, SetERP, float, 0.0f, AM_DEFAULT);
     URHO3D_ACCESSOR_ATTRIBUTE("CFM Parameter", GetCFM, SetCFM, float, 0.0f, AM_DEFAULT);
-    URHO3D_ATTRIBUTE("Disable Collision", bool, disableCollision_, false, AM_DEFAULT);
-}
-
-void Constraint::OnSetAttribute(const AttributeInfo& attr, const Variant& src)
-{
-    Serializable::OnSetAttribute(attr, src);
-
-    if (!attr.accessor_)
-    {
-        // Convenience for editing static constraints: if not connected to another body, adjust world position to match local
-        // (when deserializing, the proper other body position will be read after own position, so this calculation is safely
-        // overridden and does not accumulate constraint error
-        if (attr.offset_ == offsetof(Constraint, position_) && constraint_ && !otherBody_)
-        {
-            btTransform ownBody = constraint_->getRigidBodyA().getWorldTransform();
-            btVector3 worldPos = ownBody * ToBtVector3(position_ * cachedWorldScale_ - ownBody_->GetCenterOfMass());
-            otherPosition_ = ToVector3(worldPos);
-        }
-
-        // Certain attribute changes require recreation of the constraint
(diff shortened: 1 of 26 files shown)


eugeneko:
@weitjong Any ideas why I am getting C++11-related error here?
https://travis-ci.org/urho3d/Urho3D/jobs/267357727

weitjong:
Sorry for the late response. Didn't have a chance to sit in front of my workstation these few days.

Source/Urho3D/Physics/Constraint.cpp
```diff
@@ -84,42 +92,42 @@ void Constraint::RegisterObject(Context* context)
...
     URHO3D_ACCESSOR_ATTRIBUTE("High Limit", GetHighLimit, SetHighLimit, Vector2, Vector2::ZERO, AM_DEFAULT);
     URHO3D_ACCESSOR_ATTRIBUTE("Low Limit", GetLowLimit, SetLowLimit, Vector2, Vector2::ZERO, AM_DEFAULT);
     URHO3D_ACCESSOR_ATTRIBUTE("ERP Parameter", GetERP, SetERP, float, 0.0f, AM_DEFAULT);
     URHO3D_ACCESSOR_ATTRIBUTE("CFM Parameter", GetCFM, SetCFM, float, 0.0f, AM_DEFAULT);
-    URHO3D_ATTRIBUTE("Disable Collision", bool, disableCollision_, false, AM_DEFAULT);
+    URHO3D_ATTRIBUTE("Disable Collision", bool, disableCollision_, false, AM_DEFAULT)
+        .SetMetadata(AttributeMetadata::P_CONSTRAINT_RECREATE, true);
 }
 
 void Constraint::OnSetAttribute(const AttributeInfo& attr, const Variant& src)
```
eugeneko: TODO: Revise this place. Seems suspicious.

eugeneko:
I'm back. Here is new version of my attribute refactoring.
I temporarily added some code in HelloWorld to test things, you could check it for example.

**Important changed:**

- offset_ is removed, all attributes are working via accessors now.

- So, identifying attribute inside OnSetAttribute by offset doesn't work anymore.

- I added `URHO3D_ATTRIBUTE_EX` and `URHO3D_ENUM_ATTRIBUTE_EX` that call specified member function when attribute is set (similar to accessor attributes, but this function doesn't have input arguments). Metadata or attribute name could also be used to identify the attribute, but I decided to avoid this for now. Could be changed.

- I replaced all `URHO3D_*_FREE` functions with `URHO3D_CUSTOM_ATTRIBUTE` and `URHO3D_CUSTOM_ENUM_ATTRIBUTE` that work with any getter/setter functional objects. Note that these new functions works with Variant type due to performance reasons. Could be changed later.

**Benefits:**

- No UB in attributes anymore!

- No extra calls (or even less calls) comparing to old version. Every attribute is set or get via the following chain of calls: `AttributeAccessor::Set(virtual)->[internal lambda](probably inline)->member access or setter/getter call`

- No extra memory allocation even for lambda getters/setters

- No extra copying. Attribute actions are performed directly with Variant came from Serializable code.

- Exact type matching is not required anymore. It's enough for member attribure to be _convertible_ to corresponding Variant type. Same for getters and setters, no need in mixed accessors anymore.

- No more tons of accessors and traits. Compiler resolves all type stuff on its own.

- Any lambda could be passed as attribute accessors. Note: this lambda should work with Variant directly.

- Attribute with explicit post-set action is more straightforward than custom logic in OnSetAttribute.

**Disadvantages:**

- Old offset-based attribute identifying doesn't work anymore.

- Old type hacks like for Light cascade splits don't work enymore.

- Someone may dislike macro magic used to make lambdas. Probably will change it in the future.

eugeneko:
BTW, @weitjong, what do you think about this version?
You didn't like my previous attempt to get rid of offset. I didn't like it too, to be fair...
My current approach seems far more local.

weitjong:
I don't see any major issue so far. In fact if I understood it correctly, your proposed changes is pretty cool. I wonder have you compared the compiled binary size between your version and the old using offset?

eugeneko:
Wow, didn't thought about it at all, thanks.
I've just checked.
Urho3D.lib is 5% fatter in both Debug and Release.
Urho3DPlayer.exe is 3% fatter in both Debug and Release.
I checked release disassembly in both cases. In new version all indirect functional calls are inlined, as expected. So, I think it worth.

eugeneko:
I reviewed it once more time. So I'm going to write docs, stop hestitate and merge it.
Let's fix problems as they arise.

weitjong:
The 3% to 5% for desktop platform may be still OK, but may be not so on an mobile or web platforms. So, we should think it carefully, especially this increase in size does not bring in any new feature and the existing approach actually works on all the supported platform so far, UB aside.

eugeneko:
This 5% size increase is caused by changes in _offset_ attributes because only 26% of Urho attributes are offset-based. The main idea of my approach is to make _every_ attribute setters and getters more inline to get all benefits listed above. This inlining results in increased binary size and _slightly_ better performance.

(I tested also samples 18 and 42, size growth is the same)

It would be very easy to "get back" most of this extra size: just re-add all these removed traits and accessors in Serializable.h and re-map macros.

I don't think that we should be afraid of small growth of Urho binary after some changes, unless it cause some real problems. 5% is just 250kb.
Have we ever traced how Urho binary sizes change when some 3rd-party is updated or some new feature (like GImpactMesh) is used? I suppose no.

> especially this increase in size does not bring in any new feature
Actualy it does. Minor features, but features. It relaxes type requirements, so attributes are easier to use. And it also brings type safety to member attributes, so it wouldn't accidently explode in run-time if user touch malformed attribute.

eugeneko:
I suggest a compromise ;)
If people notice this size growth and prove that they really suffer from it, I'll re-add non-offset accessors.
We would have _only_ type-safe offset attributes, so size growth would be about 1.5%-2% instead of 3%-5%. I have just tested it.
However, the code would be much less clean IMO. I could even store "tiny" version guarded off inside Serializable. But I don't like it, yeah.

weitjong:
> I don't think that we should be afraid of small growth of Urho binary after some changes, unless it cause some real problems. 5% is just 250kb for Urho player. This 5% is probably smaller than difference caused by changing toolset.
Have we ever traced how Urho binary size changes when some 3rd-party is updated or some new feature (like GImpactMesh) is used? I suppose no.

I think we would also reject a feature if it just bloats the size without the possibility to disable it. This is why most of our 3rd-party libs are optional, they can be turned off if the user does not want it. Also it does not really matter how many symbols and big the size the 3rd party added into the library but if the symbols is not used then they are not linked in the final app (when static library is used). So, we only "pay" for what we use.

> Actualy it does. Minor features, but features. It relaxes type requirements, so accessor attributes are more user-friendly. And it also brings type safety to member attributes, so it wouldn't accidently explode in run-time if user touch malformed attribute.

That is a valid point.

> However, the code would be much less clean IMO.

Thanks for taking the time to do that. Not sure how ugly it will become without looking at the code, but probably we have the same code taste, so may be not a good idea to compromise.

Well, I just voiced out my opinion since you asked for it. Aside from this little concern of increasing in size, I have no issue with what you have proposed, like I already said. However, it would be interested to find out how big the sample app increases in size for other platforms. For instance on Web platform where Emscripten may not be able to inline those assessors.

eugeneko:
Thanks for review, @weitjong !
So I'll merge it as is and I also push "semi-fat" attributes into some separate branch just to keep code somewhere.

weitjong:
You are welcome. If you can wait, I will make the size comparison for web and android platforms tomorrow. I like to have all the fact before we made the final decision.

eugeneko:
Wow, sorry. I've merged it about 30 seconds before your message.
But you still could check things and I'll change other things if needed.

weitjong:
No worry :)

eugeneko:
Safe member attributes are IMO quite important to tolerate with 1-2% bigger binaries.
On the other hand, accessor attributes could always be easily tuned.

I think even if we decide to keep upstream as is, one who needs 3% smaller binaries could always cherry-pick commits from legacy-attributes branch.

weitjong:
I have just done some comparison for Android and WASM platforms. The final binary size increment there is surprisingly consistent with your observation (I suppose on VS Windows platform), which means both Android-NDK and Emscripten, to their credit, are also probably inlining those accessor calls. i.e. the more they are used the bigger the size would be. Alas, we don't have a full blown apps to do a more meaningful comparison.

Issue #2064, opened 2017-08-10
Engine hangs when removing compound shape


pqftgs:
To replicate:
1. Create a compound shape by adding two or more CollisionShape components to a node.
2. Wait for the next frame
3. Begin deleting shapes with node.RemoveComponent()
4. Engine will freeze when only 1 shape remains.

```
// First shape
shape_ = box_->CreateComponent<CollisionShape>();
shape_->SetBox(Vector3::ONE);

// Second shape
shape_ = box_->CreateComponent<CollisionShape>();
shape_->SetBox(Vector3::ONE);

// Trigger the freeze sometime later
box_->RemoveComponent(shape_);
```

Modified 11_Physics to show the problem
http://pasteall.org/504185

And I get this in debug mode
```
Assert colObjWrap->getCollisionShape()->isCompound() in line 244, file /home/user/repos/Urho3D/Source/ThirdParty/Bullet/src/BulletCollision/CollisionDispatch/btCompoundCollisionAlgorithm.cpp
Trace/breakpoint trap
```

EDIT
Workaround = remove the last 2 collision shapes at the same time and re-create the original.

eugeneko:
Reproduced. It seems that `rigitBody.setCollisionShape makes some internal bullet cache dirty and rigid body shall probably be re-added to the physical world.
http://bulletphysics.org/Bullet/phpBB3/viewtopic.php?t=5194

cadaver:
Possibly broken logic, or logic misusing Bullet, when returning to *not* using a compound for optimization. Can look into this too.

eugeneko:
It's possible to either remove this optimization or do this in `RigidBody::UpdateMass`:
```
    // Rigid body shall be re-added to physics world if shape is changed
    btCollisionShape* newCollistionShape = useCompound ? shiftedCompoundShape_.Get() : shiftedCompoundShape_->getChildShape(0);
    if (newCollistionShape != body_->getCollisionShape())
    {
        btDiscreteDynamicsWorld* world = physicsWorld_->GetWorld();
        world->removeRigidBody(body_.Get());
        body_->setCollisionShape(newCollistionShape);
        world->addRigidBody(body_.Get(), (short)collisionLayer_, (short)collisionMask_);
    }
```
What's better?

cadaver:
Should be fixed. Downside of fix can be that the readd happens to every body when they initially add their components, so there can be some lost performance.

Lumak:
[...]

PR #2096, opened 2017-08-25
Inherit transform mode

1vanK:
In some situations child node should inherit parent position but ingnore scaling of parent node.

p.s. I could not test this for LUA. Bite operations added in LUA 5.3 but urho used lua 5.1,
so
```
childNode.inheritMode = INHERIT_ROTATION | INHERIT_POSITION
```
just not works

### DIFF
--- a/Source/Urho3D/Physics/RaycastVehicle.cpp
+++ b/Source/Urho3D/Physics/RaycastVehicle.cpp
@@ -330,7 +330,7 @@ void RaycastVehicle::FixedPostUpdate(float timeStep)
             }
             if (skidInfoCumulative_[i] > 0.05f)
             {
-                skidInfoCumulative_[i] -= 0.002;
+                skidInfoCumulative_[i] -= 0.002f;
             }
         }
         else
--- a/Source/Urho3D/AngelScript/SceneAPI.cpp
+++ b/Source/Urho3D/AngelScript/SceneAPI.cpp
@@ -161,6 +161,15 @@ static void RegisterNode(asIScriptEngine* engine)
     RegisterComponent<DebugRenderer>(engine, "DebugRenderer", true, false);
     engine->RegisterObjectMethod("Component", "void DrawDebugGeometry(DebugRenderer@+, bool)", asMETHOD(Component, DrawDebugGeometry), asCALL_THISCALL);
     engine->RegisterObjectMethod("DebugRenderer", "void DrawDebugGeometry(DebugRenderer@+, bool)", asMETHOD(DebugRenderer, DrawDebugGeometry), asCALL_THISCALL);
+
+    engine->RegisterGlobalProperty("const uint INHERIT_NONE", (void*)&INHERIT_NONE);
+    engine->RegisterGlobalProperty("const uint INHERIT_POSITION", (void*)&INHERIT_POSITION);
+    engine->RegisterGlobalProperty("const uint INHERIT_ROTATION", (void*)&INHERIT_ROTATION);
+    engine->RegisterGlobalProperty("const uint INHERIT_SCALE", (void*)&INHERIT_SCALE);
+    engine->RegisterGlobalProperty("const uint INHERIT_ALL", (void*)&INHERIT_ALL);
+
+    engine->RegisterObjectMethod("Node", "void set_inheritMode(uint)", asMETHOD(Node, SetInheritMode), asCALL_THISCALL);
+    engine->RegisterObjectMethod("Node", "uint get_inheritMode() const", asMETHOD(Node, GetInheritMode), asCALL_THISCALL);
 }
 
 static bool SceneLoadXML(File* file, Scene* ptr)
--- a/Source/Urho3D/LuaScript/pkgs/Scene/Node.pkg
+++ b/Source/Urho3D/LuaScript/pkgs/Scene/Node.pkg
@@ -17,6 +17,12 @@ enum TransformSpace
     TS_WORLD
 };
 
+static const unsigned INHERIT_NONE;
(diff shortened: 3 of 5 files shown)


eugeneko:
That's interesting.. I like this feature, but don't like names.

1vanK:
Rename as you like, I just use names from blender

Source/Urho3D/Physics/RaycastVehicle.cpp
```diff
@@ -330,7 +330,7 @@ void RaycastVehicle::FixedPostUpdate(float timeStep)
             }
             if (skidInfoCumulative_[i] > 0.05f)
             {
-                skidInfoCumulative_[i] -= 0.002;
+                skidInfoCumulative_[i] -= 0.002f;
```
eugeneko: BTW, what's this?

1vanK: fixed warning double to float runcation

eugeneko: Damn, sorry. I thought you've changed the value.

eugeneko:
> names from blender

Then, names are fine for me xD
I doubt whether the new flag shall be named like other flags, e.g. IP[=Inherit Parent]_NONE, IP_POSITION and so on...

1vanK:
I do not mind, but we many exceptions already
```
enum WindowDragMode
{
    DRAG_NONE,
    DRAG_MOVE,
    DRAG_RESIZE_TOPLEFT,
    DRAG_RESIZE_TOP,
    DRAG_RESIZE_TOPRIGHT,
    DRAG_RESIZE_RIGHT,
    DRAG_RESIZE_BOTTOMRIGHT,
    DRAG_RESIZE_BOTTOM,
    DRAG_RESIZE_BOTTOMLEFT,
    DRAG_RESIZE_LEFT
};
```

eugeneko:
I'll merge it today unless someone write something important.

Lumak:
Can we not include unrelated changes to the check-ins please. This is really a bad habit. Please check in the changes to RaycastVehicle.cpp separately.

1vanK:
I'll take into account the wish for following PRs

Lumak:
> I'll take into account the wish for following PRs

It should be removed from this PR. How would someone know that the warning for raycastvehicle was corrected? Would you reply "oh, that was slipped in with another unrelated PR"?

Also, this check-in. I don't use Blender, so I never used this type of different inheritance pos/rot/scale.  If it's already something that you do from Blender, does everyone need this code change or just you?

1vanK:
> Also, this check-in. I don't use Blender, so I never used this type of different inheritance pos/rot/scale. If it's already something that you do from Blender, does everyone need this code change or just you?

For example vehicle example add wheels to scene root instead as child to body. When all the nodes in the root, this slows down the search.

Also in my app I have canvas in scene (scaled plane) and сursor on it (which should not be scaled).

> It should be removed from this PR.

If this is so important, then you can take the code and create a new PR. Even Cadaver included fixes of typos into their commits.

1vanK:
for example https://github.com/urho3d/Urho3D/commit/fe2d6cf21a38f6964a9ea9a3b58883cb9c0784bf

Lumak:
This seems to break the upward node inheritance chain, for example, GetWorldXXX() wouldn't know anything about inheritance_flags.  Still, this seems very specific case that only you would use.

Fixing typos and slipping in unrelated code change are same to you?

1vanK:
In fact, I'm too lazy to prove the need for PR longer than I did it.

Lumak:
You should leave it open. I voiced my opinion, but others may have a different opinion.

eugeneko:
> This seems to break the upward node inheritance chain, for example, GetWorldXXX() wouldn't know anything about inheritance_flags. Still, this seems very specific case that only you would use.

Could you elaborate? I probably missed the case when smth is broken

Lumak:
one example would be, say you set a hand bone to a specific scale using an inheritance_flag. Calling GetWorldScale() on that bone would climb the node chain and return the scale value as it typically would without even knowing about the inheritance_flag.

eugeneko:
1. `hand.SetInheritMode(POS | ROT)`
2. hand is marked dirty
3. Call `hand.GetWorldScale()`
4. `UpdateWorldTransform()` here, worldTransform_ is updated and has non-inherited scale
5. `GetWorldScale()` yields true world scale.

Lumak:
Ah, yeah I see that. Just the SetWorldXXX as you mentioned and Translate, Rotate, LookAt, etc.

pqftgs:
+1: Would be very convenient. Seems to pop up at least once every project.

-1: It hurts to see small bits of overhead for uncommon(ish) scenarios that we've grown used to solving with code.

I like this a lot... but also fear it. Or rather - the precedent that it sets.
Just my 2 coins.

Lumak:
If you like this then I say pursue it. Just because a PR gets included in the master branch doesn't mean that everyone has to merge with it or use it.  For example, I rip out the changes for https://github.com/urho3d/Urho3D/issues/2064 when I downloaded the 1.7 tag as I didn't want code that I'll never use to eat performance.

yushli:
@1vanK  I think this feature is convenient for some use cases, e.g., mimic camera following the character, so the camera only inherits the character's position. Thank you for your contribution. My suggestion is to move the newly added data member to NodeImpl instead of directly add to Node class. After all only a very small amount of nodes would use this flag.
Would you like to  reopen this PR?

1vanK:
> Would you like to reopen this PR?

No any plans, but anyone can fix PR and send by himself, if he wants

PR #2097, opened 2017-08-26
Refactor Variant (part #1)

eugeneko:
I removed those nasty reinterpret casts from Variant. Also added int64 to float/double casts in getters and fixed cast in `op == (ULL)`

### DIFF
--- a/Source/Urho3D/Core/Variant.cpp
+++ b/Source/Urho3D/Core/Variant.cpp
@@ -70,6 +70,8 @@ static const char* typeNames[] =
     0
 };
 
+static_assert(sizeof(typeNames) / sizeof(const char*) == MAX_VAR_TYPES + 1, "Variant type name array is out-of-date");
+
 Variant& Variant::operator =(const Variant& rhs)
 {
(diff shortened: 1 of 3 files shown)


yushli:
Looks like much cleaner.

iSLC:
Since c++11 is allowed now. I remember something about unrestricted unions. Which should allow the usual approach:
```cpp
struct VariantValue
{
    union
    {
        unsigned char storage_[VARIANT_VALUE_SIZE];
        int int_;
        bool bool_;
        float float_;
        //...
    };
    /// Construct uninitialized.
    VariantValue() { }
};
```

To be simply:
```cpp
union VariantValue
{
    unsigned char storage_[VARIANT_VALUE_SIZE];
    int int_;
    bool bool_;
    float float_;
    //...

    /// Construct uninitialized.
    VariantValue() { }
};
```

eugeneko:
@iSLC I thought about it, but somewhy I was afraid to change `VariantValue` from `struct` to `union`. Will try now.

PR #2098, opened 2017-08-26
Added operator overloads for multiplying IntVectors

Modanung:
This PR allows for `IntVector2` and `IntVector3` to also be multiplied by vectors of the same type instead of only scalars. With `Vector2` and `Vector3` this was already possible.


eugeneko:
Could you write bindings?
I know, everybody hates writing binding, but...

Modanung:
Can now! :wink: 
Are there more bindings required apart from the AS ones?

eugeneko:
Probably you could also add non-assignment ops to Vector2.pkg and Vector3.pkg

Modanung:
Done!
Also added IntVector2 constants and removed line 19 from Vector2.pkg, which occured twice:
```
Vector2 operator / (const Vector2& rhs) const;
```

Issue #2100, opened 2017-08-26
[Editor] "Parent to last" keybinding closes quick action menu


zombiegonsdude:
Was wondering if it would be possible to remap the binding. Currently closes the quick action menu. Prevents searching for commands like "Run script..."

https://github.com/urho3d/Urho3D/blob/a523e604dc7197efce7c7444d8bf6aa938502d65/bin/Data/Scripts/Editor/EditorUI.as#L420

PR #2101, opened 2017-08-26
Disregard Menu accelerator keys when a LineEdit is focused

cadaver:
Typing keys that are an accelerator key would fire up Menu actions. In the case of issue #2100, U or Ctrl-U would be one way to cause the Editor quick action window to be closed, due to logic code in the Editor that causes each menu selection to close it.

Not sure if this is wanted in each case, but would seem to make sense, particularly if there are accelerators bound without qualifier.


eugeneko:
In fact, I haven't managed to reproduce the problem, but...
I just trust you here.

PR #2103, opened 2017-08-27
Make Variant able to store any type.

eugeneko:
1. Varaint is able to store any custom type
2. Types are stored without memory allocation if `sizeof(Type) <= sizeof(void*)*2`
3. It is guaranteed that there is no internal allocations if assign the same type again (either directly or inside another Variant)
4. It is guaranteed that assignment operator is called in this case
5. ~~Custom types cannot be compared to either custom type or zero~~
6. Custom types cannot serialized
7. Custom types can be converted to string, compared to zero or each other if traits are specialized

The main purpose was to type-safely store pointers in the Variant.

### DIFF
--- a/Source/Urho3D/Core/Context.cpp
+++ b/Source/Urho3D/Core/Context.cpp
@@ -196,7 +196,8 @@ void Context::RemoveSubsystem(StringHash objectType)
 void Context::RegisterAttribute(StringHash objectType, const AttributeInfo& attr)
 {
     // None or pointer types can not be supported
-    if (attr.type_ == VAR_NONE || attr.type_ == VAR_VOIDPTR || attr.type_ == VAR_PTR)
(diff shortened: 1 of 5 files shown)


1vanK:
It is very good feature, recently I just needed this

iSLC:
I think that functionality that depends on `std::type_info` should be wrapped under some macros guards in case the engine is compiled with something like `-fno-rtti`.

eugeneko:
@iSLC Have you ever tried to compile Urho with `-fno-rtti`?
I doubt whether it compile, and it definetely won't work as expected.

iSLC:
@eugeneko to be honest, I never did. Just thought in case someone might. But I get your point.

eugeneko:
I've added ability to comapre custom types, moved the specialization to traits, removed some hacks, updated things...

PR #2104, opened 2017-08-27
Fix too strict assert in kNet.

henu:
I don't know if I should make pull request to Urho or kNet, so I'll first try this place :)

If you call `kNet::MessageConnection::sendMessage()` with zero length data, one of the assertions will fail. The reason for this is that `kNet::NetworkMessage::Resize()` does not allocate memory for `data` if length is zero, which is perfectly logical. However the assertion in `kNet::MessageConnection::sendMessage()` does not take this into account. This PR fixes the assertion, so it does not care about NULL if length is also zero.

### DIFF
--- a/Source/ThirdParty/kNet/src/MessageConnection.cpp
+++ b/Source/ThirdParty/kNet/src/MessageConnection.cpp
@@ -12,6 +12,8 @@
    See the License for the specific language governing permissions and
    limitations under the License. */
 
+// Modified by Henrik Heino for Urho3D
+
 /** @file MessageConnection.cpp
 	@brief */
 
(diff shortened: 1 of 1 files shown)


eugeneko:
BTW, do you really find zero-length messages useful? If so, I'll merge it.

henu:
I think with some simple message IDs like "exit world", "stop following" you don't need any parameters. And especially when the network code is not complete, there might be many situations where no data is needed, but you are still developing the project with debug flags and the assert kicks in.

weitjong:
The only thing that missing is the modified by who and why marker. This is so that we wouldn't accidentally override the internal fixes when we upgrade the library from upstream later. See examples from other modified 3rd-libraries.

henu:
@weitjong Something like that? I couldn't find any "by who" markers :/

weitjong:
You got it. Thanks.

PR #2106, opened 2017-08-27
Modernized sample 01_HelloWorld

Enhex:


1vanK:
I'm against
```
auto helloText = MakeShared<Text>(context_);
```

1vanK:
I will explain my point of view. Examples are primarily intended for beginners. Beginners should not guess the type returned MakeShared(). For example, I have never use this function and I would not have guessed, without text on the left "-    SharedPtr<Text> helloText(new Text(context_));"

Enhex:
That's not a real reason. It's obvious Make**Shared** makes a shared pointer.
Also if you explicitly write a type that doesn't mean it's the function's return type, implicit conversion is possible.
If someone wants to be sure about the return type he must inspect the declaration anyway.

This is the standard C++ approach and part of the C++ Core Guidelines:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rh-make_shared

1vanK:
I tell my impressions. I see ```auto helloText = MakeShared<Text>(context_);``` and think "Wtf?" Then I look at the left side and I understand what it is.

Enhex:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-auto

1vanK:
I expressed my point of view and do not want to argue. I don't know who will take the final decision.

1vanK:
In fact, I would recommend minimizing the use of auto to only interators

eugeneko:
Yeah, delicious auto holywar, as expected...
Well, I don't like these changes, but not because of the autos, like @1vanK 

Urho has almost 230'000 lines of code, and you modernized only 150 of them. It's about 0.07% of total codebase. I don't like the idea of such modernization itself because it requires _tremendous_ amount of work and gives approximately zero benefit. So I don't even want it to be started.

I am going to use _automatic_ tools to fix things like `override`, `nullptr` and `using`.
Then, I am going to embed inplace member initialization file by file. Here the community may help me.
I don't think we ever need to change anything in the old code except these items.

Of course, it makes sence to modernize some function if you change it (this is your words, as I recall).

Enhex:
You think it's a good idea to modernize everything in a single commit instead of step by step?

There are benefits. It shrinks and simplifies the code, making working with it easier.

1vanK:
When copyright date changes in all files, this is not a problem. Once we also changed all LOG to URHO3D_LOG. So we can make a series of commits each of which perform one task. For example adding override to all virtual functions.

eugeneko:
@1vanK 
> When copyright date changes in all files, this is not a problem

There is big difference between automatic and manual changes...

@Enhex 
> You think it's a good idea to modernize everything in a single commit instead of step by step?

I think that most of things don't have to be aggresively modernized.
And other things (nullptr, override and using) could be modernized automatically.

And yes, I thing it's good idea. I push my syntax-related changes:
1. Either together with functional code changes (e.g. my last Variant refactoring)
2. Or in the single commit for the entire refactoring (e.g. incoming automatic refactoring for nullptr, override and using)

This keeps git history clean and prevents people from wasting their time (see below)

> There are benefits. It shrinks and simplifies the code, making working with it easier.

How much time did you spent on this PR? I suppose, 5-10 minutes to review and fix the code.
So, it will take about 120-300 hours to update the whole Urho codebase. This shrinked and simplified code doesn't worth 300h of my time (or your, or anybody's)

Modanung:
The good thing about starting with a small piece of non-engine code, like @Enhex did, is that it helps the discussion by providing an example. I think it is better to make a small modification first and fine-tune it until everyone agrees on a new standard then to slam a personal opinion on the entire codebase. Once everyone agrees it will take much less effort. The discussion that may seem pointless, but it is the solution to itself.

@eugeneko Do not mistake discussion for war. I would say this use-case of `auto` is in a grey area and I hear what @1vanK is saying. But I would **not** use a deviant standard in the samples. _If_ this is a no-auto case, it should be the standard.

eugeneko:
@Modanung Good point. I didn't even thought about this code as proposal.

However, my opinion about the PR is the same: we shouldn't change code unless functional changes are made.

nick70:
I think we could write very functional but very ugly code. If it were the case then changes would obviously be necessary.

That being said unless the engine is abandoning C++03 these changes aren't appropriate.

weitjong:
We have been busy modernizing our code base to C++11 standard in the master branch and we have also set C++11 as the minimum standard. That is, URHO3D_C++11 is not a build option anymore.

bvanevery:
Ok, so, what happened with this?  Did it get merged?  Doesn't look like it.  I've been reviewing all Issues and Pull Requests to understand the full dynamics of the Urho3D project.  If a simple HelloWorld program can't have changes committed in a timely manner, it's a sign that something is wrong with the process.  Too much politics, policy, and procedure, and not enough action on straightforward matters.  Unless I'm missing something about why a HelloWorld would be remotely difficult.

If it did get merged, please close this, so that future generations of possible maintainers and contributors do not have to wade through things that are resolved.  Same sensibility would apply to many Issues I've seen from 2016 and earlier.  I have no idea if they're still relevant, and I wouldn't want to have to know.

weitjong:
You are not actually helping by making that comment.

Modanung:
Often the first step of solving a problem is acknowledging its existence.
Since Urho devs don't see one, I decided to create another:

[](https://gitlab.com/luckeyproductions/dry)

PR #2107, opened 2017-08-28
Skyplane technique fixed. 

gleblebedev:
Now it is rendering a fullscreeen backgroud image as (i guess) it was intended.

It looks like "skyplane" previously wasn't implemented the same way as other "sky*" techniques which seemed as a mistake. This one could be useful for a background images like camera feed, etc.

### DIFF
--- a/bin/CoreData/Shaders/GLSL/Skyplane.glsl
+++ b/bin/CoreData/Shaders/GLSL/Skyplane.glsl
@@ -1,0 +1,16 @@
+#include "Uniforms.glsl"
+#include "Samplers.glsl"
+#include "Transform.glsl"
+
+varying vec2 vTexCoord;
+
+void VS()
+{
+    gl_Position = vec4(iPos.x*2.0, iPos.z*2.0, 1.0 ,1.0);
+    vTexCoord = iTexCoord.xy;
+}
+
+void PS()
+{
+    gl_FragColor = texture2D(sDiffMap, vTexCoord);
+}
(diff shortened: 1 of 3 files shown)


eugeneko:
> it was intended.

It wasn't. Skyplane is old sky technique with the single horizontal textured plane rendered.
I don't know how exactly it should work and, I suppose, it's probably better to be deleted.

gleblebedev:
Please don't delete it :) It's very useful for AR applications with a camera feed as background or for menu with a single background image.

cadaver:
Skyplane was intended to be taken literally, ie. a plane suspended in the air, like for clouds. The technique is used in NinjaSnowWar, though it's nothing more than an unlit mesh rendering in the appropriate render queue.

This change would break the NinjaSnowWar example, so I'd recommend calling it something else, like SkyQuad for example.

gleblebedev:
SkyImage or, maybe, BackgroundImage?

weitjong:
You can use a custom render path to render a quad for your camera's video.

gleblebedev:
All Sky* techniques use postopaque. Why would I need a custom one?

weitjong:
When you said "BackgroundImage", I figure you might be using the SkyPlane for rendering background video. If so, then I suggest you use custom render path approach. If not, then just ignore me.

1vanK:
Actually for custom render path also need create a new shader https://urho3d.prophpbb.com/topic1746.html#p14946

weitjong:
The point is, there is nothing wrong with this shader code then.

gleblebedev:
I'll rename and resubmit. I think SkyImage is acceptable since all background techniques are called SkySometing.

eugeneko:
If you want to submit new shader/technique, do it in separate PR please.

PR #2108, opened 2017-08-28
C++11 upgrade: typedef to using, 0 to nullptr, added override

eugeneko:
Here is C++11 upgrade of Urho core.
Regardless of code standard discussion, I updated three items that are unquestionably better than C++98 counterparts.
~~Samples and Tools will be upgraded tomorrow.~~

1st and 3rd commit are made by Visual Assist/Code Refactoring/Quick Fix.
2nd is manual.
4th is regex replace + manual check.


weitjong:
Just in case you don't know. Instead of switching between topic branch and PR to control the CI, you can just make commits to the branch and control the CI with additional keyword commit messages. Something like `[ci only: cpp11-upgrade$]` to limit CI to the topic branch and skip the rest of mirror branches.

1vanK:
we actually need both virtual and override keywords for functions?

eugeneko:
@1vanK This PR has both override and virtual because of:
0) ~~My personal tastes~~
1) Using `override`s is required, but omitting `virtual`s is questionable. So leaving both keywords is win-win way, at least now.
2) This is not my code, actually. Automatic tool decided to leave both keywords, so I followed this decision.

About my personal opinion... I read code from left to right. So, `virtual` is much more noticeable mark of virtual function than `override`. In other words, `virtual` is for reader and `override` is for compiler.
I prefer both keywords regardless of Google/Core Guidelines best practices. My teammates too. However, I admit that this is questionable point and there are people that disagree with me.

PS. I consider this as the comittee fault. If they had introduced `override` and `final` keywords as replacement of `virtual`, C++ would be much better. They just didn't have enough balls to declare `override` and `final` as _real_ keywords.

OvermindDL1:
_/me is personally a fan of using `virtual` with `override` as well and does so in own code_

eugeneko:
All builds are green now, so I'm going to push it.

PR #2112, opened 2017-08-31
fix fixtureless 2D shape scaling; fixes #2111

### LINKED ISSUE #2111: Scale not applied to CollisionShape2D unless it has a fixture
CollisionShape2D does not reflect the right scale when the following sequence of operations is applied:

1. create collision shape (e.g. box) in node N
2. scale node N
3. create collision body in node N

ricab:
`ApplyNodeWorldScale` works well in the fixtureless case: it calls `RecreateFixture` which scales the shape, as expected, but ignores the fixture and the body when they are not there to start with: https://github.com/urho3d/Urho3D/blob/14bdd868e8db33c2e9193ed26ac57dbca510142d/Source/Urho3D/Urho2D/CollisionShape2D.cpp#L237 and https://github.com/urho3d/Urho3D/blob/14bdd868e8db33c2e9193ed26ac57dbca510142d/Source/Urho3D/Urho2D/CollisionShape2D.cpp#L215.

### DIFF
--- a/Source/Urho3D/Urho2D/CollisionShape2D.cpp
+++ b/Source/Urho3D/Urho2D/CollisionShape2D.cpp
@@ -320,8 +320,7 @@ void CollisionShape2D::OnMarkedDirty(Node* node)
 
     cachedWorldScale_ = newWorldScale;
 
-    if (fixture_)
-        ApplyNodeWorldScale();
+    ApplyNodeWorldScale();
 }
 
 }


ricab:
Hello, could I please get some feedback on this? Thanks

eugeneko:
I've just reproduced the problem and tested the fix.
Seems fine, so I merge it.

PR #2120, opened 2017-09-08
Template primitives

rokups:
This PR turns following classes into templates:

* `Vector2` -> `BaseVector2<T>`
* `Vector3` -> `BaseVector3<T>`
* `Vector4` -> `BaseVector4<T>`
* `Rect` -> `BaseRect<T>`

Original `IntVector2`, `IntVector3`, `IntVector4` and `IntRect` were deleted.

`Vector2`, `IntVector2`, `Vector3`, `IntVector3`, `Vector4`, `IntVector4`, `Rect` and `IntRect` are defined as specializations of their respective base template classes.

A new type `IntVector4` is introduced. `Rect` can be defined from `Vector4` therefore `IntRect` needed `IntVector4` counterpart. It is not exactly necessary, but i did it for the sake of completeness.

`IntRect` now holds data in two `BaseVector2<T>` vectors `min_` and `max_`, just like `Rect`. Structs and unions were added to give access to `left_`, `top_`, `right_` and `bottom_` to both `Rect` and `IntRect`. `Rect` also gained `Width()` and `Height()` methods that were originally present only in `IntRect`.

All of these classes were provided with explicit template constructor that defines object from an array of arithmetic types.

Now these types share common code base and are of similar layout. It was very awkward to work in cases where `Rect` and `IntRect` were mixed, because of their differences in storing values. It was no less awkward to find yourself in need of a certain operator in `IntVector3` only to realize it is missing, while it is defined in `Vector3`. Now defining operator for base type will provide it to all specializations of that type. And as a bonus user can easily define his own versions of vectors that fit perfectly with the engine. `Int64Vector3` now can be a reality.

I also tested every example, including native, angelscript and lua and editor as well. Editor could use some extra testing though. I did look for features that `Int*` classes had and their `float`-based counterparts did not, but it is possible i missed something.

And that is about it. Lets hope CI will be happy 🤞


eugeneko:
I'll try to review it...

Source/Urho3D/Math/Rect.h
```diff
@@ -181,123 +192,53 @@ class URHO3D_API Rect
...
 
-    /// Minimum vector.
-    Vector2 min_;
-    /// Maximum vector.
-    Vector2 max_;
+    union
+    {
+        /// Minimum vector.
+        BaseVector2<T> min_;
+        struct
```
eugeneko: There is no anonymous structs in C++

Source/Urho3D/Math/Rect.h
```diff
@@ -181,123 +192,53 @@ class URHO3D_API Rect
...
     /// Rect in the range (0, 0) - (1, 1)
-    static const Rect POSITIVE;
+    static const BaseRect POSITIVE;
     /// Zero-sized rect.
-    static const Rect ZERO;
+    static const BaseRect ZERO;
 };
 
-/// Two-dimensional bounding rectangle with integer values.
-class URHO3D_API IntRect
```
eugeneko: IntRect logic has changed in many places.
1. Default ctor values
2. Clip and Merge conditions and so on

Source/Urho3D/Math/Vector3.h
```diff
@@ -435,61 +277,63 @@ class URHO3D_API Vector3
     unsigned ToHash() const
     {
         unsigned hash = 37;
-        hash = 37 * hash + FloatToRawIntBits(x_);
-        hash = 37 * hash + FloatToRawIntBits(y_);
-        hash = 37 * hash + FloatToRawIntBits(z_);
-
+        hash = 37 + (x_ * hash);
```
eugeneko: Why changed? Such changed could potentially break someone's caches or so on.

Source/Urho3D/Math/Rect.h
rokups: Compilers tend to support it. This was also used out of necessity, in order to avoid change rippling through big part of the codebase. Do you have a better idea how this could be solved?

Source/Urho3D/Math/Rect.h
rokups: Please elaborate.

Source/Urho3D/Math/Vector3.h
rokups: Because it can be `BaseVector3<char>` or `BaseVector3<double>` or `BaseVector3<int64_t>`. Use of `FloatToRawIntBits()` becomes invalid in many contexts.

Source/Urho3D/Math/Rect.h
eugeneko: It somehow worked without anonymous unnamed structs before, so I just thought it could continue working...

Source/Urho3D/Math/Rect.h
eugeneko: IntRect() was (0, 0, 0, 0), become (2kkk, 2kkk, -2kkk, -2kkk).
So, IntRect is no longer empty, Width and Height are broken. Clip is broken too.
Then, the IntRect has exclusive upper border and Rect has inclusive one. So literally _all_ places with > and < signs are broken for IntRect now. Clip, Merge, IsInside...

Source/Urho3D/Math/Vector3.h
eugeneko: However, the fix is unacceptable.
1. It's bad to change hashing algo, people may have assets that rely on it.
2. The new algo is bad itself. (0, 0, 0) and (0.01, 0, 0) will have the same hashes. (1kkkk, 0, 0) will just explode hashing algo and end up in UB.

Source/Urho3D/AngelScript/MathAPI.cpp
```diff
@@ -218,17 +218,17 @@ static void RegisterIntVector2(asIScriptEngine* engine)
     engine->RegisterObjectMethod("IntVector2", "IntVector2& opAddAssign(const IntVector2&in)", asMETHOD(IntVector2, operator +=), asCALL_THISCALL);
     engine->RegisterObjectMethod("IntVector2", "IntVector2& opSubAssign(const IntVector2&in)", asMETHOD(IntVector2, operator -=), asCALL_THISCALL);
     engine->RegisterObjectMethod("IntVector2", "IntVector2& opMulAssign(const IntVector2&in)", asMETHODPR(IntVector2, operator *=, (const IntVector2&), IntVector2&), asCALL_THISCALL);
-    engine->RegisterObjectMethod("IntVector2", "IntVector2& opMulAssign(int)", asMETHODPR(IntVector2, operator *=, (int), IntVector2&), asCALL_THISCALL);
```
eugeneko: 1. float -> int conversion is losing. Try IntVector2(1, 1) * 2kkk before and after the commit.
2. Legacy code will spam warnings after this change

Source/Urho3D/Math/Rect.h
rokups: > IntRect() was (0, 0, 0, 0), become (2kkk, 2kkk, -2kkk, -2kkk).

Is there a good reason for `Rect` to define infinitely large rect? Maybe it should become `(0, 0, 0, 0)` instead? Does infinitly large rect have a purpose?

> Then, the IntRect has exclusive upper border and Rect has inclusive one.
> Clip, Merge, IsInside...

Good spot. Ill add template specialization restoring old behavior to `IntRect`. For the record: one of these rects has a bug. Not sure if we want to fix it though.

Source/Urho3D/AngelScript/MathAPI.cpp
rokups: Yup already saw it on CI and pushed a commit addressing it. Left scalar multiplication and division operators to always take `float` as parameter.

Source/Urho3D/Math/Rect.h
eugeneko: > Maybe it should become (0, 0, 0, 0) instead?

Rect (0, 0, 0, 0) is considered as valid rect with zero area. Point (0, 0) is inside the rect.
Rect(inf, inf, -inf, -inf) is considered as invalid rect. Point (0, 0) is outside the rect because there is no rect.
So we should keep old behavior here.

> For the record: one of these rects has a bug. Not sure if we want to fix it though.

Huh, what bug do you mean?

Source/Urho3D/AngelScript/MathAPI.cpp
eugeneko: > Left scalar multiplication and division operators to always take float as parameter

I meant, use float for scalar ops with IntVector2/3 is bad too.

rokups:
Any idea why CI fails to build erroring about undefined references? Everything builds fine both static and shared on my machine (linux).

Source/Urho3D/Math/Rect.h
rokups: Yeah see issue is that `Rect` was defined as `Vector2 min_; Vector2 max_`, but `IntRect` was defined as `int left_; int top_; int right_; int bottom_;`. I used anonymous structs here in order to maintain backwards compatibility with `IntRect`. If we were to throw anonymous structs out then i would have to go through code of all engine and change `.left_` to `.min.x_` and so on. Do we really want that?

Source/Urho3D/Math/Vector3.h
rokups: Hmm true. I will add one more template specialization restoring old algorithm for floats here. For other types it should be sufficient to pick large enough prime right? How about `16777619` and  `1099511628211` used by fnv?

Source/Urho3D/Math/Rect.h
rokups: > Huh, what bug do you mean?

I meant inconsistent `IsInside()` checks, but i guess all of the inconsistent behavior should be called a bug.

eugeneko:
> Any idea why CI fails to build erroring about undefined references?

I am not sure, but explicit specialization may requre `extern` keyword.
http://en.cppreference.com/w/cpp/language/class_template
However, I don't think you really need it at all.

Source/Urho3D/Math/Rect.h
eugeneko: Rect over discrete space (int) is exclusive, rect over continious space (float) is inclusive. This is the most convinient way of defining rects, so it's not a bug and will never be changed.

In the perfect world, template BaseRect should behave as Rect for real types and as IntRect for integer ones.

rokups: Got it. I think i can make that happen.

eugeneko:
Summary.
I like the idea of code reuse, but this PR seems unstable and overcomplicated now.
I'll think later what to do here...

rokups:
Please allow me to finish it.

eugeneko:
I suggest you to think if there is the simpler way to reach the same goal.
Then, I'll try to do the same this weekend.

eugeneko:
In the current imlementation this PR now have more cons than pros now. Let me try another way...

What do you think about e.g. this approach?
https://github.com/eugeneko/Urho3D/commit/a33dabdb32c80aeb34b3638ba69cdd084e2b363d
Bindings are broken, but it's not a problem now.

1vanK:
may be just move to https://github.com/g-truc/glm ?

eugeneko:
How to move to glm and don't break Urho API simultaneously?

1vanK:
Personally, I do not care about backward compatibility, but this is just my opinion

eugeneko:
> Personally, I do not care about backward compatibility, but this is just my opinion

But we **must** care.
Breaking _all_ math code for every Urho user is like saying 'I don't care how and when you are going to update your stupid code. Fuck you'

1vanK:
Users can still use old version of engine, but new project start with new version

rokups:
> What do you think about e.g. this approach?

Oh i was not aware we could do that. TIL. This is definitely better. Do you plan to push it forward or want me to finish it?

Also what about `Rect`/`IntRect`? I was thinking maybe it would be viable approach to have these classes:

* `BaseRect<T>` - would define common methods
* `IntegralRect<T> : public BaseRect<T>` - would define methods from `IntRect`
* `ScalarRect<T> : public BaseRect<T>` - would define methods from `Rect`
* `Rect<float> : public ScalarRect<float>`
* `IntRect<int> : public IntegralRect<int>`

And customize their behavior through overloads. Does that sound reasonable?

> may be just move to https://github.com/g-truc/glm ?

I think outsourcing work is always a great idea, but it is nice to have data types fit well into the engine. Moving to external lib for these things would make urho more frankenstein.

> backward compatibility

Existing classes could be turned into a thin wrappers around glm in order to retain old interfaces. That would not even cost any performance.

eugeneko:
> Oh i was not aware we could do that. TIL. This is definitely better. Do you plan to push it forward or want me to finish it?

To be honest... I am not sure whether we should change anything at all and what to change if we should. Subsystem refactoring is major change, so I don't want to hurry.

> Also what about Rect/IntRect? 

I doubt that we ever need those classes template. Actually, I could say the same about vectors.
It's better to decide what we want to reach before implementing anything. The final implemetnation will depend on goals. Maybe it's really better to move to GLM library, who knows?

rokups:
My motivation for this PR was to:

1. Solve the problem of missing methods between float-based and int-based variants.
2. Solve interface inconsistencies and ease conversion between int/float variants.
3. Allow defining vectors with different data types, basically with doubles or int64.

GLM appears to satisfy these all. Or can we solve these ourselves somehow?

eugeneko:
I see.

1 and 2 could be solved inside Urho, yep. Some template refactoring and reusing will make code simpler.

The 3rd item is subtle. Imagine that we have an ability to define double vectors. How to use them? They are not exported into script API, Urho subsystems couldn't handle it, even Variant is unable to consume them without recently added VAR_CUSTOM and memory allocations per object. Even matrices and quaternions are unable to work with new vectors.
If you want to reach the 3rd goal, I suppose that outsource math is the only viable way. This doesn't solve all problems, but at least Urho.Math would be consistent.

Rects... Huh, rects are not math, actually. So I'd like to leave them untouched now, and return if (when?) we solve _other_ problems.

If you have enough time and passion, you could try to _pilot_ GLM math in Urho. Or maybe I'll try it at some point.
My view is that we could try alias out mat types to GLM types (without any inheritance) and make all member functions like Angle/Normalize/etc free. Then, check how many places in the engine are affected... And ananlyze whether such backward compatibility violation is acceptable or not.

rokups:
> The 3rd item is subtle. Imagine that we have an ability to define double vectors. How to use them? They are not exported into script API, Urho subsystems couldn't handle it, even Variant is unable to consume them without recently added VAR_CUSTOM and memory allocations per object. Even matrices and quaternions are unable to work with new vectors.

Engine has no business using them though. It is purely for user convenience. Vector could hold that too after increasing `VARIANT_VALUE_SIZE`.

> Rects... Huh, rects are not math, actually. So I'd like to leave them untouched now, and return if (when?) we solve other problems.

Can we at least normalize their data storage interfaces? Have them both do this:

Rect:

```cpp
    union
    {
        /// Minimum vector.
        Vector2 min_;
        struct
        {
            /// Left coordinate.
            float left_;
            /// Top coordinate.
            float top_;
        };
    };

    union
    {
        /// Maximum vector.
        Vector2 max_;
        struct
        {
            /// Right coordinate.
            float right_;
            /// Bottom coordinate.
            float bottom_;
        };
};
```

IntRect:

```cpp
    union
    {
        /// Minimum vector.
        IntVector2 min_;
        struct
        {
            /// Left coordinate.
            int left_;
            /// Top coordinate.
            int top_;
        };
    };

    union
    {
        /// Maximum vector.
        IntVector2 max_;
        struct
        {
            /// Right coordinate.
            int right_;
            /// Bottom coordinate.
            int bottom_;
        };
};
```

This would be a real quality of life improvement.

rokups:
@eugeneko ping. Please read my last post.

eugeneko:
Despite this trick is pretty common in C++ community, it relies on compiler extension and leads to UB. I could accept usage of such "almost standard" hacks if they give some noticeable benefit. Here they don't.

If we decide to break backward compatibility with e.g. migration to GLM, it's better to make IntRect hold two IntVector2 instead of making union-struct chimera. That's why I want to solve problems with math core before fixing rects.

rokups:
I see. My issue with this is possible migration to GLM is both uncertain and definitely quite a while away if ever. I want to ease things for myself and others today. Would it be acceptable if i replaced ints with 2 x IntVector2 and make a PR now?

eugeneko:
> My issue with this is possible migration to GLM is both uncertain and definitely quite a while away if ever

C'mon @rokups, don't be so pessimistic!

>  I want to ease things for myself and others today. Would it be acceptable if i replaced ints with 2 x IntVector2 and make a PR now?

To be honest, I don't understand why are you pushing this change. From my point of view it is minor visual change (like renaming). Do you have some reason for this change that I don't understand now?

Ok, then constructively...

1. Safe way: make some getters and setters in Rects. E.g. Left/SetLeft for both Rect and IntRect and so on.
2. Rect way: use IntVector2 in IntRect, add left/right/top/bottom getters;
3. IntRect way: use 4*float in Rect, add min/max getters.

What way do you like more? What way will end up with minimal legacy changes (in Urho core)? What way will produce more clean Rect/IntRect code?

PS. It would be good if someone except me say something about the subject.

rokups:
> Do you have some reason for this change that I don't understand now?

I just want a consistent interface because these differences is a code smell. It is more convenient to use `Rect` instead of `IntRect` in some cases due to `IntRect` missing some stuff, but then again mixing them is pain due to missmatch of interfaces.

> What way do you like more?

Anything that is consistent between `Rect` and `IntRect`.

> What way will end up with minimal legacy changes (in Urho core)?

Union-struct chimera gives that. That is why i proposed it.

> What way will produce more clean Rect/IntRect code?

Giving 2 x `IntVector2` to IntRect and adding getters/setters for left|right|top|bottom to both rects is probably most kosher way on paper, but it would cause cascade of changes through large part of the engine. It would also make code less readable because frankly `rect.left_ = val;` is more obvious than `rect.min_.x_ = val;`. `rect.SetLeft(val)`; is still a downgrade from plain assignment, but at least it is easier to instantly see what code does.

eugeneko:
> I just want a consistent interface because these differences is a code smell.

Yeah, it smells... But is e.g. your chimera better? There is one detail that you are probably missing.
`IntRect` was designed to be screen rect. X axis is left-right and Y axis is up-down. So, (left, top) <= (rigth, bottom).
Then, `Rect` was designed as generic rect for sprites and so on. X axis is left-right and Y axis is *down-up*. So, `rect.top` is actually *below* `rect.bottom`. It smells, doesn't it?
If change names, we will have `intRect.top <= intRect.bottom` but `rect.top >= rect.bottom`. It smells too.

rokups:
Ayyy right. Then mixing them is bad idea to begin with. If mixing them is a bad idea then there is little point in consistent interface i guess? This really needs to be documented as its easy to trip on this without extensive engine knowledge.

In this case i suppose there is little to do with rects other than adding missing operators and what not.

Thank you for a great code review btw ;) 👍

eugeneko:
NP.

Sometimes I behave quite conservatively, but it is needed to avoid pitfalls and detect possibly bad design decisions. E.g. I recalled about this up-down stuff only 30 mins ago during this discussion.

eugeneko:
@rokups @1vanK Minor update: I tried to replace Urho Math with GLM.

* I removed underscores from xyzw memebers via regex replace `(\W)([xyzw])_(\W)` to `\1\2\3`. Pretty simple breaking change that could be easily fixed in downstream projects, so it is acceptable for me.

* It's impossible to wrap GLM into new classes. It just doesn't work stable enough.

* It's very hard to rewrite engine code onto GLM, @1vanK. About 200-500 places that have to be manually pathced only for vectors. Moreover, Urho has very convinient interfaces that will get uglier with pure GLM.

* Script bindings will get destroyed anyway. They have to be written from scratch.

* The most viable way is to patch GLM itself to make it fit legacy code.

* I used `#include`-s to mix-in code into GLM vectors. As you can see in branch, it results in very localized changes in the Engine.

* The same (I hope) could be done for Quaternion and Matrix-es. Plus something like replace `m23_` to `m(2, 3)` for matrix element accessors.

* The most demotivating is to rewrite script bindings. I didn't even touch it.

So, migration to GLM is doable in this way, but requires some effort.
Benefit is to have powerful math library features out of box.

My experiments are available here:
https://github.com/eugeneko/Urho3D/tree/glm-math-3
(compile with AS&Lua disabled)

rokups:
@eugeneko this looks great, but does it build? I tried and it does not build for me. Also few observations:

> VectorLerp
> VectorMin
> VectorMax

Why excess verbosity when type system ensures these are called for vectors?

> static vec FromData(const T* data)

Why this instead of injecting a constructor?

> DotProduct()
> Length()
> Normalize()
> others..

glm defines these. Should we not just wrap glm functions instead?

eugeneko:
>  I tried and it does not build for me.

Have you disabled script bindings?
I tried only MSVC15, so I am not sure that it is buildable everywhere. Plus you probably have to symlink glm into `include/Urho3D/ThirdParty`.

> Why excess verbosity when type system ensures these are called for vectors?

I don't really like mixing templates. It ends up with occasional hard-to-debug and hard-to-fix name resolving errors. However, it may work aster some design or SFINAE-ing.

> Should we not just wrap glm functions instead?

It is doable if functions bodies are extraced from class body mixin header. BTW, GLM has some guards that prevent e.g. `glm::dot` from calling with integer vectors. This logic shall be off.

rokups:
I tried to use your method while maintaining max backwards compatibility and i got pretty good results: https://github.com/rokups/Urho3D/commit/c382f3c54932f445deb4ce86faeef0bca7b754da

So far its just `IntVector3` as it was easiest, will go for others now. We actually can even temporary keep backwards-compatibility by doing [this](https://github.com/rokups/Urho3D/commit/0a467352168f7bb89ac385d2ee5b448b2875a87e#diff-dc9fdf2267511607653db89e2cbc2adaR45). I am feeling optimistic about this. Anyhow do you want to continue implementing this or should i?

eugeneko:
> Anyhow do you want to continue implementing this or should i?

I just piloted GLM integration. I consider this task as low-priority and will not push it in the nearest future. So, feel free to work on it if you are interested in. However, I could help you somewhere.

I hope if we achieve good enough level of compatibility, it could be merged.
BTW, @weitjong, what do you think about the subject?

-----

I have some thoughts about implementation, let'em be here:

* Make such mix-ins for vectors, matrixes and quaternions, don't touch other types for now;
* Avoid any implementation stuff in original GLM headers, move into glm/urho as most as possible;
* Script bindings shall have 100% backward compatibility. I'm almost sure that it could be done. I also hope it's possible to generalize bindings, at least for AS;
* Finally, fork GLM into Urho account and maintain it like e.g. SDL.

> This is probably ok, no?

No, it's bullshit xD
The only motivation of GLM author is to "keep GLM functionality close to GLSL maths blah blah blah".
The author has even added a define guard to disable these annoying asserts, but not everywhere.

eugeneko:
Ahh, one more thing. GLM matrices are defaulted to all-zeros, Urho matrices are defaulted to identity. This shall be changed too.

weitjong:
Thanks for asking. I have a mixed feeling about this. I don't like:
- the "x_" becomes "x". It is not about how trivial the downstream projects can adapt the changes, but it is more about consistency. All our class member variables have trailing underscore. I saw Rokas tried to address that already.
- the GLM header is being exposed in Urho3D headers. Usually we tried to hide the implementation detail by not exposing the 3rd-party headers in our own headers. There are exception to this rule, however, like SDL and Bullet. If we choose to expose it then it is harder to change the implementation later.
- We also try not to modify too much of the 3rd-party codes unless it is absolutely necessary, like fixing a bug. So far, I already see changes being made just for sake of integrating it. The more we change the more difficult to sync with the upstream changes later, even when git subtree is being used.

Although using a well tested math library is a good thing, but I still don't see why we should do it. Our math classes has done their job well and nothing is broken at this moment. What is the tangible benefit of replacing what we currently have with GLM? Aside from replacing arguably a little code smell of IntRect/Rect with even bigger one :). Why should I have to read and learn another manual just to implement a new math function/classes? Not that I need to by the way because like I said, our existing implementation has served us quite well already.

eugeneko:
> I have a mixed feeling about this

The same. So I consider it as low-priority questionable enhancement, but I don't block it unconditionally.

I suggested to add EASTL into Urho 3rdParty some time ago. Now I have similar rationalies for GLM. Both GLM and EASTL don't give us immediate benefit, but they will make things simpler in the future. If one day someone needs circular buffer or gauss random in Urho, he will only need to use already existing 3rdparty instead of re-inventing or copy-pasting functionality.

1vanK:
After some thought I agree with weitjong. Changing the working code is not the best waste of time.

PR #2122, opened 2017-09-09
Hot-fix to AngelScript: fixed compile with address sanitizer.

rainerdeyke:
A quick fix to AngelScript that was preventing me from using address sanitizer with Urho3D, fixed upstream in revision 2410.  See https://www.gamedev.net/forums/topic/691763-error-compiling-on-gccx64-with-address-sanitizer .


weitjong:
Thanks for that. As the same change is also available in the upstream, we can merge your PR as is without any marker.

PR #2125, opened 2017-09-14
Add ability to specify an image offset in Button, to enable graying-out buttons that are disabled.

JTippetts:
Possibly, it might be more appropriate to add this offset to the BorderImage class instead, although I can't think of a personal use-case for doing so. If others disagree, I can withdraw this PR and add the functionality to BorderImage instead.

### DIFF
--- a/Source/Urho3D/AngelScript/APITemplates.h
+++ b/Source/Urho3D/AngelScript/APITemplates.h
@@ -1338,10 +1338,13 @@ template <class T> void RegisterButton(asIScriptEngine* engine, const char* clas
 {
     RegisterBorderImage<T>(engine, className);
     engine->RegisterObjectMethod(className, "void SetPressedOffset(int, int)", asMETHODPR(T, SetPressedOffset, (int, int), void), asCALL_THISCALL);
+    engine->RegisterObjectMethod(className, "void SetDisabledOffset(int, int)", asMETHODPR(T, SetDisabledOffset, (int, int), void), asCALL_THISCALL);
     engine->RegisterObjectMethod(className, "void SetPressedChildOffset(int, int)", asMETHODPR(T, SetPressedChildOffset, (int, int), void), asCALL_THISCALL);
     engine->RegisterObjectMethod(className, "void SetRepeat(float, float)", asMETHOD(T, SetRepeat), asCALL_THISCALL);
     engine->RegisterObjectMethod(className, "void set_pressedOffset(const IntVector2&in)", asMETHODPR(T, SetPressedOffset, (const IntVector2&), void), asCALL_THISCALL);
     engine->RegisterObjectMethod(className, "const IntVector2& get_pressedOffset() const", asMETHOD(T, GetPressedOffset), asCALL_THISCALL);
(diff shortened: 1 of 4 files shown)


JTippetts:
Is there in issue with this? I mean, being able to gray an inactive button isn't exactly unprecedented... Or is there something I'm missing?

eugeneko:
Looks fine for me.
TODO: Fix tab in Button.pkg

JTippetts:
Darn those tabs.

SuperWangKai:
I though we have these states offsets already...
Normal/Hover/Pressed/Disabled should be in the `BorderImage `I think. Thus we get every widget aware of the four states.

eugeneko:
@SuperWangKai You are probably right... I didn't have so much experience in Urho UI tho. So I could miss such things.

JTippetts:
Hover offset is in border image, pressed is in Button. There currently is no offset for disabled. My reasoning for putting it in Button was that's the typical use case for it. However, I'm not opposed to moving it into border image instead. When I get some time, I might make a pull request. Or I might not, as it kinda seems like nobody is really steering this ship, and I might just continue letting my private fork diverge.

eugeneko:
Am I right that now Urho doesn't support separate display for disabled widgets?

weitjong:
Why is it closed? It's a pity. I agreed with the Eugene's initial assessment that the PR looks fine and didn't say anything to object it.

eugeneko:
I think it could be merged as is and generalized later.

weitjong:
Yes, agree.

eugeneko:
[offtop] What's better: rebase&merge or merge commit?

weitjong:
> offtop] What's better: rebase&merge or merge commit?

IMHO, they both do slightly different things although the end result looks the same. So, it depends on the situation. In general when a PR is automatically merge-able then a press of a button to merge it is sufficient.

OvermindDL1:
In my opinion, *Never* rebase&merge, it destroys some of the history that makes git so valuable.

OvermindDL1:
My major experiences all you can get some commits occurring after others in the timeline yet before them temporally, this means that if a bug was introduced after the time that a commit was done and before when it was rebased on the master branch then git bisect breaks in interesting ways.

A quick google search pops this up at the top:  http://geekblog.oneandoneis2.org/index.php/2013/04/30/please-stay-away-from-rebase
But there are plenty *plenty* more about why rebase'ing should not be done.

weitjong:
I think this is getting out of topic. But to me, merge and rebase are two different operations that have their own use case. So, I won't rule out any of them as option to get the thing done. Rebase changes the commit history, but that may be exactly what you want it at time. Usually it is never a good idea to do that in a public facing/default branch, but it is perfectly ok for PR or topic/feature branch.

OvermindDL1:
> Usually it is never a good idea to do that in a public facing/default branch, but it is perfectly ok for PR or topic/feature branch.

Precisely this, I was referencing it in using it on the public github (any branch), you should never ever ever rebase on a public branch ever (excepting very few breaking cases).

PR #2128, opened 2017-09-19
Fix crash in AS named object constructor & minor refactoring

eugeneko:
* Fix crash in AS named object constructor
* Make ValueAnimation::GetAnimationValue const

### DIFF
--- a/Source/Urho3D/AngelScript/APITemplates.h
+++ b/Source/Urho3D/AngelScript/APITemplates.h
@@ -380,12 +380,15 @@ template <class T> void RegisterObject(asIScriptEngine* engine, const char* clas
 
 template <class T> T* ConstructObject()
 {
-    return new T(GetScriptContext());
+    T* object = new T(GetScriptContext());
+    object->AddRef();
+    return object;
 }
 
 template <class T> T* ConstructNamedObject(const String& name)
 {
     T* object = new T(GetScriptContext());
+    object->AddRef();
     object->SetName(name);
     return object;
 }
@@ -393,14 +396,14 @@ template <class T> T* ConstructNamedObject(const String& name)
 /// Template function for registering a default constructor for a class derived from Object.
 template <class T> void RegisterObjectConstructor(asIScriptEngine* engine, const char* className)
 {
-    String declFactory(String(className) + "@+ f()");
+    String declFactory(String(className) + "@ f()");
     engine->RegisterObjectBehaviour(className, asBEHAVE_FACTORY, declFactory.CString(), asFUNCTION(ConstructObject<T>), asCALL_CDECL);
 }
 
 /// Template function for registering a named constructor for a class derived from Object.
 template <class T> void RegisterNamedObjectConstructor(asIScriptEngine* engine, const char* className)
 {
-    String declFactoryWithName(String(className) + "@+ f(const String&in)");
+    String declFactoryWithName(String(className) + "@ f(const String&in)");
     engine->RegisterObjectBehaviour(className, asBEHAVE_FACTORY, declFactoryWithName.CString(), asFUNCTION(ConstructNamedObject<T>), asCALL_CDECL);
 }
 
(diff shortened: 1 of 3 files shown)


Source/Urho3D/AngelScript/APITemplates.h
```diff
@@ -380,27 +380,30 @@ template <class T> void RegisterObject(asIScriptEngine* engine, const char* clas
 
 template <class T> T* ConstructObject()
 {
-    return new T(GetScriptContext());
+    T* object = new T(GetScriptContext());
```
weitjong: I have my reservation on this change and the ones followed it. These codes that use AS "auto handles" have been there more than 6 or 7 years ago, so I don't believe they are the culprit of the crash. Also I don't see how replacing them with the non auto handles version will of any help. According to my understanding, both your proposed new code and the old one do exactly the same thing. It just that one being implicitly (automatically) done and one being explicitly in handling the reference count.

eugeneko: Auto handles are incremented after function return.
Classic handles are incremented manually within the function.

I think we shouldn't use auto handles for any newly created object returning from the function.
It's a kind of malformed code because we have to return an semi-dead object with zero references.
The bug was in ConstructNamedObject, but I changed both for consistency.

weitjong: I beg going to differ. If you what you said is true then indeed we could not use `@+` for binding at all.

weitjong: * I beg to differ. Correct iPad auto correct. :)

eugeneko: > I beg going to differ. If you what you said is true then indeed we could not use @+ for binding at all.

Nope, `@+` references are fine in _most_ cases.

The only problematic case is when we return some newly created object that is not referenced anywhere yet.
Such object will have single reference after `@+` evaluation, so it should have no references when returned from function and _this_ is malformed code.

Search for `@ ` usage, it is already used in the same cases overall the bindings.

weitjong: I am not saying they are wrong. I am just saying your proposed change and the old code do the same thing. I just don't understand your rationale that latter is malformed. The function simply return a pointer to an object with intrusive reference counting, whether the reference count is increased by the script engine or by the function before returning make no difference to me. I would really be surprise when this will be the crash and equally surprised when my understanding is wrong and that this is genuinely a bug and we have to wait 7 years for the bug to manifest itself.

eugeneko: @weitjong Sorry, I lost the thread. Are you talking about `ConstructNamedObject`, `ConstructObject` or both?

weitjong: Both

eugeneko: Well, the problem with semi-dead (0 refs) objects is that such objects could be accidentally destroyed by code that works with the object. `SetName` call in `ConstructNamedObject` has non-trivial routine: it send the event. So, event handler could accidentally destroy the object before function return.

weitjong: Don't you think if the event handler could do potentially stupid thing then the problem is with the "user" event handler in question? But I get your point. Now the event handler needs to do even more stupid thing to crash it.

eugeneko: > Don't you think if the event handler could do potentially stupid thing then the problem is with the "user" event handler in question?

No, I don't.
Create&destroy temporary SharedPtr or AS handle from passed object is enough to crash.
For example, call `GetEventSender` from `EditorMainHandleNodeNameChanged` that does this implicitly.

weitjong: Again, if it is so prone to crash then why do we only hear it now? Anyway, I have done reviewing with this code and voiced out my concerns. I have no strong objection if you want to merge this as it is.

eugeneko: > Again, if it is so prone to crash then why do we only hear it now?

I think it's pretty rare crash. Does anybody except the Editor ever interested in `NameChanged` event?

PR #2130, opened 2017-09-22
improve Android SDLActivity to check for no shared libraries found wh…

c4augustus:
…en loading and alert instead of null crashing


weitjong:
Thanks for your PR.

PR #2140, opened 2017-10-03
Add internal layout functions for ListView.

eugeneko:
Some helpers that make massive ListView item instertion 10 times faster
```
    Text* rootItem = new Text(context_);
    rootItem->SetText("Root");
    rootItem->SetStyleAuto();
    listView->DisableInternalLayoutUpdate();
    listView->InsertItem(M_MAX_UNSIGNED, rootItem, nullptr);
    for (int j = 0; j < 15; ++j)
    {
        auto tp1 = std::chrono::high_resolution_clock::now();
        for (int i = 0; i < 1000; ++i)
        {
            Text* childItem = new Text(context_);
            childItem->SetStyleAuto();
            childItem->SetText("Level" + String(j) + "Item" + String(i));
            listView->InsertItem(1000 * j / 2, childItem, rootItem);
        }
        auto tp2 = std::chrono::high_resolution_clock::now();
        unsigned dur = std::chrono::duration_cast<std::chrono::milliseconds>(tp2 - tp1).count();
        URHO3D_LOGINFOF("j=%d t=%u", j, dur);
    }
    listView->EnableInternalLayoutUpdate();
    listView->UpdateInternalLayout();
```

### DIFF
--- a/Source/Urho3D/AngelScript/UIAPI.cpp
+++ b/Source/Urho3D/AngelScript/UIAPI.cpp
@@ -309,6 +309,9 @@ static void RegisterListView(asIScriptEngine* engine)
     engine->RegisterEnumValue("HighlightMode", "HM_ALWAYS", HM_ALWAYS);
 
     RegisterUIElement<ListView>(engine, "ListView");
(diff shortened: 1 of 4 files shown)


weitjong:
Nice, but still missing the script API binding so we can replace the same logic in the Editor (Hierarchy window) with the new API call.

eugeneko:
I am going to test performance of Editor and use this feature there.

PR #2143, opened 2017-10-10
fix typo in Quaternion opSub

lieff:


eugeneko:
Good catch, but...
How did you notice it?

lieff:
Just trying to use Quaternions from script :)

lieff:
For example, for rotation interpolation: Qres = a*Q1 + (1-a)*Q2.

eugeneko:
I wouldn't expect any predictable result from linear interpolation of quaternions. `Slerp/Nlerp` shall be used instead.
Indeed, these functions are implemented via some kind of linear interpolation internally, but I see no reason to use linear interpolation in user code.

lieff:
Yes, It's not linear interpolation of angles (scale interpolated normally though). But sometimes this is what we need.

PR #2151, opened 2017-10-14
Fixing shared pointer release bug

TheComet:
Granted, this is an extremely rare bug, but even so: The following code will not behave as expected:
```cpp
struct Foo : public RefCounted {
    SharedPtr<Foo> child_;
};

// --- elsewhere ---

SharedPtr<Foo> foo = new Foo;  // #1
foo->child_ = new Foo;  // #2
foo->child_->child_ = new Foo;  // #3

foo = foo->child_->child_;  // Deletes all 3 Foo instances
```

The expected behaviour is for Foo #3 to now be assigned to "foo" with a refcount of 1. In reality, all three Foo objects are released and "foo" is NULL. The reason why this happens can be seen in the assignment operator of SharedPtr:

```cpp
    SharedPtr<T>& operator =(const SharedPtr<T>& rhs)
    {
        if (ptr_ == rhs.ptr_)
            return *this;

        ReleaseRef();  // BUG!
        ptr_ = rhs.ptr_;
        AddRef();

        return *this;
    }
```

Releasing the refcount of this->ptr_ potentially executes destruct code, which can in very rare cases cause rhs.ptr_ to be deleted (this is such a case) before acquiring rhs.ptr_ and incrementing its refcount.

The correct implementation is:

```cpp
    SharedPtr<T>& operator =(const SharedPtr<T>& rhs)
    {
        if (ptr_ == rhs.ptr_)
            return *this;

        T* tmp = ptr_;
        ptr_ = rhs.ptr_;
        AddRef();
        if (tmp) tmp->ReleaseRef();

        return *this;
    }
```

### DIFF
--- a/Source/Urho3D/Container/Ptr.h
+++ b/Source/Urho3D/Container/Ptr.h
@@ -80,9 +80,10 @@ public:
         if (ptr_ == rhs.ptr_)
             return *this;
 
-        ReleaseRef();
+        T* tmp = ptr_;
         ptr_ = rhs.ptr_;
         AddRef();
+        if (tmp) tmp->ReleaseRef();
 
         return *this;
     }
@@ -93,9 +94,10 @@ public:
         if (ptr_ == rhs.ptr_)
             return *this;
 
-        ReleaseRef();
+        T* tmp = ptr_;
         ptr_ = rhs.ptr_;
         AddRef();
+        if (tmp) tmp->ReleaseRef();
 
         return *this;
     }
@@ -106,9 +108,10 @@ public:
         if (ptr_ == ptr)
             return *this;
 
-        ReleaseRef();
+        T* tmp = ptr_;
         ptr_ = ptr;
         AddRef();
+        if (tmp) tmp->ReleaseRef();
 
         return *this;
     }
@@ -166,17 +169,19 @@ public:
     /// Perform a static cast from a shared pointer of another type.
     template <class U> void StaticCast(const SharedPtr<U>& rhs)
     {
-        ReleaseRef();
+        T* tmp = ptr_;
         ptr_ = static_cast<T*>(rhs.Get());
         AddRef();
+        if (tmp) tmp->ReleaseRef();
     }
 
     /// Perform a dynamic cast from a shared pointer of another type.
     template <class U> void DynamicCast(const SharedPtr<U>& rhs)
     {
-        ReleaseRef();
-        ptr_ = dynamic_cast<T*>(rhs.Get());
+        T* tmp = ptr_;
+        ptr_ = static_cast<T*>(rhs.Get());
         AddRef();
+        if (tmp) tmp->ReleaseRef();
     }
 
     /// Check if the pointer is null.


weitjong:
There is a mistake in the last change where the dynamic cast has been converted to static cast. Besides that, IMHO, the proposed changes do not that clean to me. Instead of making the guards everywhere, can't we just fix the issue as its root in the assignment operator? I think something like this should work (haven't tested it yet).

```
    SharedPtr<T>& operator =(const SharedPtr<T>& rhs)
    {
        if (ptr_ == rhs.ptr_)
            return *this;

        SharedPtr<T> copy(rhs);
        ReleaseRef();
        ptr_ = copy.ptr_;
        AddRef();

        return *this;
    }
```

But then again, I also think that this `copy`  should really be done outside of the assignment operator all together. In other words, I believe this is programming error.

```
SharedPtr<Foo> copy(foo->child_->child_);
foo = copy;
```

eugeneko:
Good catch! I fixed similar problem with vector self-assignment some time ago...

@weitjong
Self-assignment check almost never fixes any problem. It is common example of bad assignment operator implementation, see Sutter or other C++ guys.
It also doesn't fix the problem from PR description.

weitjong:
I didn't add the self-assignment check. It is the existing self-check in the code. What I proposed is just to fix the assignment operator by having a "copy" of the rhs using another shared pointer. That should fix the PR description without touching other parts of the class. Since a copy must be made, the assignment operator rhs parameter signature might as well change from a reference to a value object,  so no need to make a copy explicitly as I first proposed.

I don't want to get into the discussion on whether the self-check is bad or not. What I do know is we don't want to do something in the operator that is equivalent to "do nothing" for performance reason. Those C++ guru can talk about exception safety, but in Urho I believe we only have one try-catch block at the beginning of the engine initialization In other words, we do not use C++ exception feature at all.

eugeneko:
@weitjong Sorry, I misread your post... Shame on me /_-

Well, I suggest almost the same as you have suggested.
I mentioned Sutter because they propose nice pattern of such ops:

        SharedPtr<T> copy(rhs);
        Swap(copy);

I think we should use this pattern without those nasty add&release calls everywhere.

weitjong:
I am not too sure about your last statement. Although the swap() looks very concise on the outside, internally it need to do some move and pointer assignments. Not entirely which one wins out without doing any thorough testing.

eugeneko:
To swap SharedPtr it's enough to swap pointers inside, because all refcounding is already done during construction/destruction. So SharedPtr could have zero-cost Swap member function like UniquePtr.

All the difference is that manual `AddRef` call is replaced with temporary `SharedPtr` ctor and manual `ReleaseRef` call is replaced with one's destructor.

weitjong:
I was referring to your last statement "I think we should use this pattern without those nasty add&release calls **everywhere**",  with the added stress on the offending keyword. For the SharedPtr class, it is easier to convince because it's constructor/destructor is quite simple. But I am not convince for using the copy-and-swap idiom everywhere, until there is benchmark result proofing that there is no impact to the speed. The performance is paramount. If I what I have read is correct, even the std::vector implementation does not use that idiom internally in its assignment operator for performance reason.

Back to SharedPtr class. I am still ambivalent to replace the code with swap. To me, It hides too many things behind our back. At least with the user-defined version, we take full control of what should be performed efficiently. As I mentioned before, the aspect of exception safety is less of a concern for Urho. Having said that, if there is no objection from other core devs or contributors then I am OK if swap is used for this case.

eugeneko:
Well, I meant "everywhere in SharedPtr": assigments, resets, casts...

yushli:
I think the code in this PR is clean and fix the problem. And has the best performance. Why not just merge it now and then wait for a better solution to come up?

eugeneko:
Because it gives nothing but unconvinience.
This defect is not so urgent to push the first working revision of fix without review.

PS: That's how I see the fix
https://github.com/eugeneko/Urho3D/commit/04c615822c3e0d13f0b47e961d0197392084614d

weitjong:
@eugeneko In the earlier discussion, I thought we are talking about std::swap().

eugeneko:
Well, I've just pushed my fix. I think it's much simpler and faster than do formal review with follow-ups.
Thanks for report, @TheComet! I am sorry for stealing your PR.

OvermindDL1:
I'm curious, doesn't Urho3D require C++11+ now?  If so why not use `std::shared_ptr` as it should be able to do everything Urho3D's can and is well tested (and in some cases optimized by specific compilers)?

eugeneko:
> it should be able to do everything Urho3D's can

It's definetely not because SharedPtr is _not_ shared ptr in C++ meaning.

OvermindDL1:
> It's definetely not because SharedPtr is not shared ptr in C++ meaning.

Hmm, what is it in the vein of?

eugeneko:
`SharedPtr` is the hybrid of `std::shared_ptr` and `boost::intrusive_ptr`, extended version of `std::shared_ptr & std::shared_from_this`. There is no such class in any standard library.

TheComet:
@eugeneko Ok!

@OvermindDL1 
> Hmm, what is it in the vein of?

Urho3D::SharedPtr is very different from std::shared_ptr. Urho3D uses intrusive reference counting (this means: The object holding the reference count information is stored in the object itself, by inheriting from Urho3D::RefCounted). std::shared_ptr stores the reference count in the std::shared_ptr class instead of in the object its wrapping.

I talked on the forums about this before that the name "SharedPtr" is really bad, it should be "Reference". The *object being wrapped* is the shared pointer, not the class that manages its reference counting.

Intrusive reference counting has a lot of benefits over std::shared_ptr as far as performance goes. For example:

```cpp
struct Thing : public RefCounted {};

void foo(Thing* thing) {
    bar(thing);
}
void bar(Thing* thing) {
    SharedPtr<Thing> ref = thing;  // The original "thing" object now has 2 refs
    // ...
}

int main() {
    SharedPtr<Thing> thing = new Thing;
    foo(thing);
}
```

Notice how we are able to pass just raw pointers to ```foo()``` and ```bar()``` and are still able to gain a reference to the object. If you were using std::shared_ptr you have the problem of having to *pass the ref count information down the entire callstack*:

```cpp
struct Thing : public RefCounted {};

void foo(std::shared_ptr<Thing> thing) {
    bar(thing);
}
void bar(std::shared_ptr<Thing> thing) {
    SharedPtr<Thing> ref = thing;  // The original "thing" object now has 5 refs
    // ...
}

int main() {
    std::shared_ptr<Thing> thing(new Thing);
    foo(thing);
}
```

You could optimise it to use ```const std::shared_ptr<Thing>&``` but that has other problems.

OvermindDL1:
> Urho3D::SharedPtr is very different from std::shared_ptr. Urho3D uses intrusive reference counting (this means: The object holding the reference count information is stored in the object itself, by inheriting from Urho3D::RefCounted). std::shared_ptr stores the reference count in the std::shared_ptr class instead of in the object its wrapping.

Isn't that precisely the usage of `std::enabled_shared_from_this` combined with `std::make_shared`?  `std::enabled_shared_from_this` makes it so that the object itself holds a weak shared pointer to itself that it can acquire trivially and `std::make_shared` makes the shared pointer intrusive with the object itself, in addition if you `make_shared` on an object that implements `enabled_shared_from_this` then it hoists them more efficiently together as well.

> Notice how we are able to pass just raw pointers to foo() and bar() and are still able to gain a reference to the object. If you were using std::shared_ptr you have the problem of having to pass the ref count information down the entire callstack:

Correct, this is the point of using the above things I mentioned.

In addition I would not have something take an std::shared_ptr as an argument unless it truly needs to increment the reference count 'now' (a raw pointer is fine, it means 'I do not own this and will not persist this', an `std::shared_ptr` passed in means `I need this reference count incremented *now* before I start operating on this (important in some template code), and `const std::shared_ptr&` means that I may or may not need to increment the reference count either directly or via some other call further within).

But still what you state that you need seems already possible with the standard as it is via using `std::enabled_shared_from_this` and `std::make_shared`?  What about those do not allow for what you need?  (The standard has some other things that could help other use-cases if you have others as well).

eugeneko:
> What about those do not allow for what you need? 

For example, they don't allow to change counter manually. They don't allow to release object without destruction.

OvermindDL1:
> they don't allow to change counter manually.

Well other than sounding horrifically unsafe, can't you do that by just cloning one of the shared pointers and storing it somewhere (even on itself, then reset it when you need to clear it)?

> They don't allow to release object without destruction.

If you mean to release a shared_ptr object, just `reset` it.  If you mean to remove the object from all shared_ptr handling, just set an empty `deleter`.

eugeneko:
> Well other than sounding horrifically unsafe

It's prefectly safe if you are careful enough. It isn't much worse than AddChild/RemoveChild pair or like this.

> can't you do that by just cloning one of the shared pointers and storing it somewhere (even on itself, then reset it when you need to clear it)?

So, you suggest to store vector of shared_ptrs in every single Object just to emulate refcounting?
It doesn't sound very good.

> If you mean to remove the object from all shared_ptr handling, just set an empty deleter

I have no idea how to do it for existing shared_ptr

OvermindDL1:
> if you are careful enough

`if` you are careful yes.  ^.^;

> So, you suggest to store vector of shared_ptrs in every single Object just to emulate refcounting?

Actually I'd propose storing the cloned shared_ptr in wherever it needs to be used, I.E. whatever incremented the ref-count.  Even if that is, say, lua, you could store that object in the metatable then `reset` it when GC'ing or so.

> I have no idea how to do it for existing shared_ptr

Something I find very worrying about this phrase is it implies needing to remove a pointer from a shared pointer while keeping it alive in possible different areas (I.E. you want it to destruct in some callpaths and not destruct in others), that I would say is a bug in wait.

However, if you want to do that there are two ways:

* The correct way:  Use more weak_pointers or owned_pointers, depending on if single or multi-owner.
* A hacky, but still safe way:  Instance a new shared_ptr from the pointer to the original active shared_ptr, along with a deleter that does nothing but `reset` the base shared_ptr (or deallocates it if it is the sole owner of it at that point, but better to keep its memory somewhere on the stack or in another object if possible of course).

And there are a couple of unsafe ways that I'd really not like to mention as I'd run in horror if I saw them used in production code (from accessing its memory indirectly to runnin the constructor or destructure functions of the shared_ptr manually on a non-scope object, which are, honestly, fine in 'some' cases...).

This really *really* sounds like an instance of [The XY Problem](http://xyproblem.info/) though.  What is being accomplished and where by manual reference handling and why cannot the object be kept itself (it is a tiny object after all), or why cannot a raw pointer be used if the object is 'owned' elsewhere for certain (or a weak pointer if it is owned elsewhere but could die at some random point)?

yushli:
The current Urho3D SharedPtr and WeakPtr look good to me. The code is simple and the usage is clean.

eugeneko:
> Actually I'd propose storing the cloned shared_ptr in wherever it needs to be used

If we were able to store SharedPtr, we wouldn't have used AddRef/ReleaseRef.
E.g. AngelScript work with raw pointers everywhere.
Switching to shared_ptr means wide-scale refactoring.

> This really really sounds like an instance of The XY Problem though

Yes. But Y is user code that is linked with Urho. And X is functionality provided by Urho. I don't let myself guess about Y. I am just maintaining X.

PS. shared_from_this doesn't work in ctors and I can't bet that it's not used that way.
PPS. Of course, we couldn't deal with global backward compatibility breakage, so migration to STL most probably will stay in our thoughts forever.

OvermindDL1:
> E.g. AngelScript work with raw pointers everywhere.

Wow really?  How does AS handle ownership, even LUA's API allows full information of ownership...  o.O

_/me stays on the C++ side..._

eugeneko:
> How does AS handle ownership

AS has its own refcounting mechanism that is mapped onto intrusive counters of Object so it could work together with SharedPtr. There is no chance to make somethink like this with shared_ptr.

OvermindDL1:
> AS has its own refcounting mechanism that is mapped onto intrusive counters of Object so it could work together with SharedPtr. There is no chance to make somethink like this with shared_ptr.

How does it refcount `Object`, if it is via a callback then those callbacks could just call the constructor destructor of `std::shared_ptr` on it's own instance as a quick change.

eugeneko:
> those callbacks could just call the constructor destructor of std::shared_ptr

AS doesn't keep any counter. It just say AddRef and ReleaseRef when needed.

OvermindDL1:
> It just say AddRef and ReleaseRef when needed.

How does it say that, are they function calls?  Macro calls?  In what scope are they called?  Is ADT taking place?

eugeneko:
AS has AddRef and ReleaseRef operators that are connected to member functions during registration.

OvermindDL1:
Ah, those could be wired up then, maybe an idea if a rewrite occurs.

eugeneko:
You have something like this and you cannot touch the signature:
```
void AddRef(Object* obj);
void ReleaseRef(Object* obj);
```
I have no idea how to make it work together with shared_ptrs in C++ code.

OvermindDL1:
Disclaimer:  Feel free to ignore the long post, it's not like changing the shared_ptr is backwards compatible anyway, but it is still an interesting conversation point.  Given that:

> I have no idea how to make it work together with shared_ptrs in C++ code.

Ah in that case it would not be hard at all, just implement those ref functions as something kind of like (pseudo-C++):

```cpp
// If `Object` is setup as defined before with `enable_shared_from_this` and so forth
// Skipping error checking for succinctness

void AddRef(Object *obj) {
  auto p = obj->shared_from_this();
  new (&p) shared_ptr<Object>(); // no memory allocation, just re-calling the constructor
}

void ReleaseRef(Object *obj) {
  auto p = obj->shared_from_this();
  p.~shared_ptr<Object>();
  // You might need to guard `p` to prevent it from being optimized away in high optimization modes
  // Thus assumes that the destructor will not zero out the shared pointers memory, which no current
  // implementation does that I've seen as it would be just a useless performance lose, otherwise copy
  // copy the bytes into a stack buffer and cast and destruct it to be safe about it, which only copies 16
  // bytes anyway on 64-bit systems so is plenty cheap enough especially to the stack.
}
```

This is of course entirely a hack, though should work everywhere (but still why on earth does something manually increment/decrement instead of handling the objects itself or at least some type of ownership...), a more 'proper' thing would be to put a `shared_ptr<Object>` on `Object` itself along with an integer (probably an `std::atomic_integer` if possibly multi-threaded calls) and just increment that integer (if 0 first then instance the shared pointer) and decrement it on release (when hit 0 then `reset()` the shared pointer).  There are a few others ways as well of course.  All of these seems rife with a mis-design though.  Even in a game engine there should be a single primary owner of a given allocated pointer in 99% of the cases, so the fact that shared_ptr's of any form exist on such a root class seems inefficient and dangerous to me...  Though personally I'd probably use a Handle of some form (probably 64-bit with 32-bit of it being a handle and the other 32-bit being a generational id) as that is far far safer and although it has a single assembly instruction more to access it (which is efficient on modern CPU's anyway) it would likely still be far faster anyway considering a singular owner would control.

I.E. instead of an Object class hierarchy why not have a singular 'entity' be a 64-bit integer, which is just a handle to various components.  As an example take a, oh, let's do an OctreeComponents and a MeshComponents.  The OctreeComponents would be a class (a singleton to a given engine instance, though could be other instances of the engine was instanced multiple times somehow) that holds a table of data mapping the entity ID's to the data (being the parent entity in the octree most likely, position could be yet another component, or pack them together if they are often used that way).  The table form could be whatever is efficient for a given component type, for something common like both OctreeComponents and MeshComponents it could just be a packed array with the entity id being an index into it.  This is what I'd been doing in my side engine for a long while now and it works fantastically (with so little memory thrashing as well).  I 'kind of' pulled out the style from the rest of the engine into it's own repo as I was showing how to do it a while back to some people, I put it at https://github.com/overminddl1/overecs but it is not complete (lacks a lot of helpers and the rest of the event system), however it does show the component/system design a bit (though if you don't mind another language, the Rust library `specs` is substantially more 'pure' than mine while even outperforming it in many cases, and mine is still one of the fastest C++ ECS examples out).

I've always been curious why Urho3D went with an object hierarchy where components run their own code, it seems entirely counter-intuitive that things get scattered all over memory like that with a *ton* of indirect function calls that cannot be optimized out (take the 'update' calls on instanced components, each individual one is a virtual call that references `this` that are jumping all over memory with unknown code location, crazy hard for the compiler to optimize, this issue bit me pretty early on in Urho3D actually, thousand papercuts kind of thing, though I tend to run little test games that handle a few tens of thousands to hundreds of thousands of objects routinely, which my style handles just fine).  But the gist of it it, not only do I have no un-owned pointers but a handle is trivially marked as invalid just by incrementing its generation value in the primary store (which may ripple out to caches in the various systems, Rusts spec library does it in an entirely different and interesting way).

PR #2154, opened 2017-10-19
In scripting optimize searching node's script object by their type name

orefkov:
Current implementation search only in object's base classes, compare by name, and not search in implemented interfaces. I was single search typeinfo and call AngelScript's API methods "Derived" and "Implement".


weitjong:
It looks OK to me to merge without testing. But could you change the parameter of `GetTypeInfoByName()` to `className.CString()` as it is more idiomatic that way in our code base. While you are it, please also fix the spacing to follow our coding convention. Finally, squash all the commits into one. Thanks!

eugeneko:
It looks OK (2)
Except broken compilation.

PS. This PR definetely should be squashed.

PR #2156, opened 2017-10-23
Improve serialization of ScriptInstance

orefkov:
Added serialization of Array<String> and Array<Variant> attributes of ScriptInstance.
Added method Swap to script type Array<T> for swap content of two arrays to avoid copying.


eugeneko:
This makes sence, I'll try it later. Variant[] and String[] will be automatically serializable, am I right?

Source/Urho3D/AngelScript/Addons.cpp
```diff
@@ -1112,6 +1112,18 @@ int CScriptArray::FindByRef(asUINT startAt, void *ref) const
     return -1;
 }
 
+bool CScriptArray::Swap(CScriptArray& other)
+{
+    if (&other != this && other.GetArrayObjectType() == GetArrayObjectType())
+    {
+        SArrayBuffer* t = buffer;
```
eugeneko: `Urho3D::Swap(t, buffer);` is better, I think.
Also I think that `&other != this` is redundant.

Source/Urho3D/AngelScript/Addons.cpp
```diff
@@ -1647,7 +1659,8 @@ void RegisterArray(asIScriptEngine* engine)
     engine->RegisterObjectMethod("Array<T>", "int Find(uint, const T&in) const", asMETHODPR(CScriptArray, Find, (asUINT, void*) const, int), asCALL_THISCALL);
     engine->RegisterObjectMethod("Array<T>", "int FindByRef(const T&in) const", asMETHODPR(CScriptArray, FindByRef, (void*) const, int), asCALL_THISCALL);
     engine->RegisterObjectMethod("Array<T>", "int FindByRef(uint, const T&in) const", asMETHODPR(CScriptArray, FindByRef, (asUINT, void*) const, int), asCALL_THISCALL);
-    engine->RegisterObjectMethod("Array<T>", "bool opEquals(const Array<T>&in) const", asMETHOD(CScriptArray, operator==), asCALL_THISCALL);
+    engine->RegisterObjectMethod("Array<T>", "bool Swap(Array<T>&in)", asMETHOD(CScriptArray, Swap), asCALL_THISCALL);
+	engine->RegisterObjectMethod("Array<T>", "bool opEquals(const Array<T>&in) const", asMETHOD(CScriptArray, operator==), asCALL_THISCALL);
```
eugeneko: Tab to space

Source/Urho3D/AngelScript/ScriptInstance.cpp
```diff
@@ -642,6 +675,10 @@ void ScriptInstance::GetScriptAttributes()
                 break;
 
             default:
+                if (0 == typeName.Compare("Variant[]"))
```
eugeneko: Why not `typeName == "Variant[]"`?

eugeneko:
Thanks, I'll test your code later.

orefkov:
About serialization of Variant[] and String[].
In documentation writed:
"After instantiation, the script object's public member variables that can be converted into Variant, and that don't begin with an underscore are automatically available as attributes of the ScriptInstance, and will be serialized."
In scripts we can convert to Variant  VariantMap, Variant[], String[].
But VariantMap is auto serialized in ScriptInstance, and Variant[] and String[] both not.
Imho it wrong.

orefkov:
I was found and fix error when loaded scene in background.
In VectorToArray not possible get active script context, and it return null.

eugeneko:
Merged. Thanks for patience.
BTW, I'm going to make some minor post-review changes.
The most important that Array::Swap IMO shall take inout& reference so you can't call it with const argument.

PR #2160, opened 2017-11-01
Fix Android joysticks

orefkov:
Fix bug in SDL cause poor joystick detection in Android. On my device Urho3D detect joystick only 1-2 times from 10-20 runs.
Now joysticks worked perfectly - in tests, I connected two joysticks at the same time - one through bluetooth, another through USB - both were visible in the game.


Lumak:
could it be because of your android API level?

>  Support for device connect/disconnect is API >= 16 only,

Lumak:
I'm curious, how is JoystickByDevIndex() being called, or which fn calls it?

orefkov:
Lumak, problem not in dynamic connection, SDL not finded joysticks, already connected to phone before running programm. In my tests I was subscribed to all joustick events, and receiving nothing.
There is 2 bugs, first - if system running from 24 to 48 days - SDL_GetTicks return number grater 2147483648, and in SDL_SYS_JoystickDetect - SDL_TICKS_PASSED return FALSE on timeout = 0, and Android_JNI_PollInputDevices never called.
Second - in function JoystickByDeviceId after first search in list, if SDL_SYS_JoystickDetect found new joystick and add it to list - before second search varible "item" not reinitialized and always stayed NULL, and JoystickByDeviceId return NULL instead new founded joystick.

Lumak:
In this fn

```
JoystickByDeviceId(int device_id)
{
    SDL_joylist_item *item = SDL_joylist;

    while (item != NULL) {
        if (item->device_id == device_id) {
            return item;
        }
        item = item->next;
    }
    
    /* Joystick not found, try adding it */
    SDL_SYS_JoystickDetect();
    
    while (item != NULL) {
        if (item->device_id == device_id) {
            return item;
        }
        item = item->next;
    }

    return NULL;
}

```
I agree that the 2nd while loop is meaningless and should be removed since the local variable, **item** never gets reset to the head of link list. I'm thinking it was just an oversight on whoever wrote it. The function should just exit after the call to SDL_SYS_JoystickDetect().

If you look at the catlog and the timestamp in Android Studio, you'll notice that JNI calls are not immediate.  There are few milliseconds, sometimes tens of milliseconds of delay before the java thread processes queued messages. This means that anything after the call to SDL_SYS_JoystickDetect() or even a call to Android_JNI_PollInputDevices() there is a delay, and the remaining code below those calls are moot and stepped through on the SDL thread.  Response to the queries are made on the callback fns.

I understand your frustration with the joystick detection, but if I were in your shoes, I'd manage the calls to Android_JNI_PollInputDevices() in my app, or write a wrapper to make that fn call instead of making SDL changes in the manner that you're proposing.  You might make an appeal to SDL (not sure if they're on git, github, or where ever) but I'd be curious is the problem that you're seeing with an external joystick is common.

Lumak:
If you're happy with that then cool you got your fix.  But in light of this, I decided to make my own changes due to your finding.

The way I look at it is that if JoystickByDeviceId() is called due to a **device input event and the device that caused it is yet in the SDL_joylist**, I force calling the Android_JNI_PollInputDevices() fn because the detection should be imminent.

weitjong:
@orefkov Thanks for the PR. We will get the bug fix directly from the SDL upstream repo.

PR #2161, opened 2017-11-02
Combine GetOSVersion() into GetPlatform()

Florastamine:
I noticed there's functionality duplication of `GetPlatform()` and `GetOSVersion()`. While the former simply returns the platform name, `GetOSVersion()` returns platform name plus version information. Did a PR to combine both into one, from now on `GetPlatform()` returns platform + also version (if version checking is implemented for that platform). No need for `GetOSVersion()`.


eugeneko:
Please ensure that code that uses `GetPlatform` is not broken.

Florastamine:
Yep. I've also figured that out. Existing code use something like `GetPlatform() == "Windows"` would break as the string now contains version information. And the editor contains such line. 

Nevermind, I'm going to close it then.

PR #2163, opened 2017-11-06
remove verbose logging of window position and size.

tommy3:
every time the window is moved or resized, the new position or size is logged. in my opinion this logging is too verbose, even for debug logging.

### DIFF
--- a/Source/Urho3D/Graphics/Direct3D11/D3D11Graphics.cpp
+++ b/Source/Urho3D/Graphics/Direct3D11/D3D11Graphics.cpp
@@ -1862,8 +1862,6 @@ void Graphics::OnWindowResized()
     // Reset rendertargets and viewport for the new screen size
     ResetRenderTargets();
 
-    URHO3D_LOGDEBUGF("Window was resized to %dx%d", width_, height_);
-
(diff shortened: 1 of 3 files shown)


weitjong:
It is easy to remove it when one doesn't need it, however it won't be that obvious where to put it back again when someone new needs it in his/her debugging session.

tommy3:
i know what you mean. may be i was to ruthless just removing this. i thought it was a remainder from development and was forgotten to be removed. :)

what do you think is the best way? leave it as it is or just comment out the log-statements so that one can find it?

may be this is a job for a new finer-grained logging level?

weitjong:
> may be this is a job for a new finer-grained logging level?

I don't have any objection to have another level after DEBUG. In Java world, after DEBUG there is TRACE level.

eugeneko:
`OnWindowMoved` is fine to be `TRACE`d.
However, `OnWindowResized` is important and heavy event due to viewports re-creation and other GAPI things. Unsure about it.

weitjong:
I agree with @eugeneko . So, this PR won't be merged as is. However, a new verbose log level for tracing the mouse move will still be added.

PR #2168, opened 2017-11-10
Misc fixes and features

rokups:
This is a mega-PR of various changes i did in the past and when working on the [editor](https://discourse.urho3d.io/t/yet-another-editor/3718). I know PRs are supposed to be about one single change, but that would simply be too much of a trouble. Most of changes should be benign and totally acceptable. Some will certainly be discussed. I will rewrite commit history cleaning up commits and adding/removing required bits if that will be necessary.

I will not write list of changes as commit messages detail them. Most of commits are simple enough that no explanation is needed. Some have broader explanation. So please see commit messages.


eugeneko:
Wow... It's bigger than I expected. I'm starting reviewing...

Source/Urho3D/Core/Context.cpp
```diff
@@ -53,6 +67,7 @@ static void HandleIKLog(const char* msg)
 }
 #endif
 
+static Context* contextInstance_ = nullptr;
```
eugeneko: I don't think that global context should go into Urho core, at least in this form. It's too restrictive now.
I think you should either move these static things into user code/Toolbox or make explicit Context::Set/GetGlobalContext to give user choice.

rokups: Maybe while at it, [this](https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/IO/Log.cpp#L57) should be changed to make use of that global context?

Source/Urho3D/Core/Object.cpp
```diff
@@ -72,7 +87,8 @@ bool TypeInfo::IsTypeOf(const TypeInfo* typeInfo) const
 }
 
 Object::Object(Context* context) :
-    context_(context)
+    context_(context),
+    blockEvents_(false)
```
eugeneko: I'm unsure that you really need this hack. Event suppression is usually better to make locally, shuting down undesired events one-by-one. Please, re-think this design decision. However, I'm not strictly against it.

rokups: I implemented it for a game state switching. At times this is very convenient, at other times unsubscribing/resubscribing events is more suitable. Just one more tool giving more freedom to developer.

Source/Urho3D/Core/WorkQueue.h
```diff
@@ -53,7 +53,7 @@ struct WorkItem : public RefCounted
     }
 
     /// Work function. Called with the work item and thread index (0 = main thread) as parameters.
-    void (* workFunction_)(const WorkItem*, unsigned);
+    std::function<void(const WorkItem*, unsigned)> workFunction_;
```
eugeneko: [Thoughtful review needed]
How do you use it.

Source/Urho3D/Graphics/ConstantBuffer.h
```diff
@@ -61,6 +61,7 @@ class URHO3D_API ConstantBuffer : public Object, public GPUObject
     /// Return whether has unapplied data.
     bool IsDirty() const { return dirty_; }
 
+    using GPUObject::GetGraphics;
```
eugeneko: How do you use it?

Source/Urho3D/Graphics/DebugRenderer.h
```diff
@@ -140,9 +140,11 @@ class URHO3D_API DebugRenderer : public Component
     /// Add a skeleton.
     void AddSkeleton(const Skeleton& skeleton, const Color& color, bool depthTest = true);
     /// Add a triangle mesh.
-    void AddTriangleMesh
```
eugeneko: API-breaking change. Please keep old version shortcut.

Source/Urho3D/Graphics/DebugRenderer.cpp
```diff
@@ -432,6 +435,24 @@ void DebugRenderer::AddTriangleMesh(const void* vertexData, unsigned vertexSize,
...
+{
+    if (auto staticModel = node->GetComponent<StaticModel>())
+    {
+        for (auto index = 0; index < staticModel->GetBatches().Size(); index++)
+        {
+            const auto& geometry = staticModel->GetLodGeometry(index, -1);
+            const auto& ib = geometry->GetIndexBuffer();
+            for (const auto& vb : geometry->GetVertexBuffers())
+            {
+                AddTriangleMesh(vb->GetShadowData(), vb->GetVertexSize(), geometry->GetVertexStart(),
```
eugeneko: What if data is not shadowed? Should be checked probably.

Source/Urho3D/Core/WorkQueue.h
rokups: https://github.com/rokups/Urho3D/blob/master/Source/Tools/Editor/SceneSettings.cpp#L189-L205

In order to put scene settings into inspector i created new class that registers settings as Urho3D attributes. Registering them as "free" attributes and using lambdas there is very convenient as it allows me to easily copy some relevant objects into scope of lambda instead of complicating code with storing of those objects and later accessing them through Serializable that is passed to getter/setter.

Source/Urho3D/Graphics/ConstantBuffer.h
rokups: Introducing getters for core subsystems caused name collision because both `Object` and `GPUObject` now have `GetGraphics()` which returns graphics subsystem. This statement makes compiler prefer `GPUObject::GetGraphics()`. Without this statement compiler produces ambiguity error.

Source/Urho3D/Graphics/Terrain.cpp
```diff
@@ -524,6 +525,10 @@ void Terrain::SetOccludee(bool enable)
 
     MarkNetworkUpdate();
 }
+void Terrain::SetEnableDebug(bool enable)
+{
+    debugGeometry_ = enable;
```
eugeneko: Is it needed to dirty anything?

rokups: This should be called before creating any terrain. All this flag does is make `TerrainPatch` vertex buffers shadowed upon (re)creation. So i guess no?

Source/Urho3D/Scene/Serializable.cpp
```diff
@@ -69,99 +69,102 @@ void Serializable::OnSetAttribute(const AttributeInfo& attr, const Variant& src)
 {
     // Check for accessor function mode
     if (attr.accessor_)
-    {
         attr.accessor_->Set(this, src);
-        return;
-    }
+    else
```
eugeneko: If you don't touch ifs, changes would be smaller.

Source/Urho3D/Scene/Serializable.h
```diff
@@ -186,8 +186,8 @@ template <typename T, typename U> class EnumAttributeAccessorImpl : public Attri
 template <typename T, typename U> class EnumAttributeAccessorFreeImpl : public AttributeAccessor
 {
 public:
-    using GetFunctionPtr = U(*)(const T*);
```
eugeneko: [Thoughtful review needed]
How do you use it?

rokups: Same as with previous attribute change: https://github.com/rokups/Urho3D/blob/master/Source/Tools/Editor/SceneSettings.cpp#L102-L125

Attribute inspector creates dropdowns for enum attributes and custom based on Variant type for anything else. So this is used for utilizing lambdas as getters/setters when creating enum attribute.

Source/Urho3D/Core/Context.cpp
eugeneko: Let's fix Log it later. At least, Log is localized and optional.

I mean that if some user application use subsystems that need global context, it should set the context explicitly. Context itself shouldn't touch it.

Source/Urho3D/Core/WorkQueue.h
eugeneko: I understand Attribute changes. Where do you use WorkQueue?

Source/Urho3D/Graphics/Terrain.cpp
eugeneko: I think it may be re-written like debug Attribute that could be toggled on and off dynamically. There already are places in Urho that behave like that.

Source/Urho3D/Core/WorkQueue.h
rokups: Oh got confused about this one.

Used in some experiments. Again it is very convenient to use lambdas to copy some pointers into the scope of lambda and sidestep using `start_`, `end_` and `aux_` pointers. To be honest i think at some point internal engine work functions should be upgraded to use lambdas or `std::bind()` instead of those raw pointers.

eugeneko:
I suppose I have to return to my attribute refactoring and see if I can do anything here.
I also have some style issues, but they are not so important now.

rokups:
Is there anything wrong with using std::function<> for storing attribute getters/setters?

yushli:
std::function<> uses much more memory than a simple function pointer.

Source/Urho3D/Core/WorkQueue.h
eugeneko: Yeah, I think so. But I don't think that this is required in _this_ PR.
Probably should be separate refactoring.

rokups:
@yushli it should not if you use to store plain lambda or function pointer. It sure does if you start binding arguments (using std::bind or lambda capture), however that is only acceptable.

eugeneko:
My only concern is about performance of std::function call itself, especially in debug mode.
I'll take a look on attributes this evening, then decide.

rokups:
Addressed requested changes:

* Added `static void SetContext(Context*)` and removed context instance setting from constructor/destructor. Now these statics are opt-in by user.
* Removed `std::function<>` for `WorkItem`
* Cleaned up `Serializable` fix to be a smaller diff (at the expense of small code duplication)
* Replaced `Terrain::SetDebugEnabled()` with attribute that enables debugging. Though i am unsure how it will work now as debugging is likely to be enabled after terrain is created. I probably need to test it a little more.

And still forgot DebugRenderer checks apparently.

eugeneko:
Huh, about network update.
Your change seems logical, but why this wasn't done in Urho long time ago?

rokups:
Maybe noone really used networked scene replication eh?

Also cant make terrain debug rendering work properly so i will nuke it. It does not work in all situations as expected and is just not worth spending time on it..

eugeneko:
> Maybe noone really used networked scene replication eh?

Every setter explicitly marks object as dirty for Network. It looks like intended desing decision. @cadaver, do you rememeber why it is done this way?

cadaver:
@eugeneko Because network replication also needed to happen when just calling the setter functions directly instead of going through attributes, and to mark explicitly the set of objects that the server would need to check for replication message generation, instead of having to checking all replicated nodes of the scene. If all setters are correctly written to call MarkNetworkDirty(), then the addition is unnecessary, it just makes that sure in the case they aren't.

eugeneko:
Oh thanks, I completely missed that point /_-
Thanks!

rokups:
Awesome, will remove that change

Lumak:
Good to see all these enhancements, but I'll have to omit the changes in the Object class when this is added. I just don't care to have all those system call methods included in every class derived from Object: all UI classes, Component, Serializable, and all custom classes that I create. If I want to make a specific system call, I can add it.

eugeneko:
@rokups I made the following changes:

- Renamed Graphics::RaiseWindow to Graphics::Raise due to consistency with Minimize/Maximize;
- Removed Graphics::GetSDLWindow, Graphics::GetNumMonitors and Graphics::GetMonitorResolution due to duplicate functionality;
- Reverted changes in Serializable and merged fresh master (to keep this branch feature-complete)
- Minor non-functional naming and style changes.

Then, I have few questions:
- DebugRenderer::AddTriangleMesh (StaticModel version) doesn't seem valid in general case. I'm going to try smth more reliable than this.
- Image::SetSubimage. Is it really sane that host image and subimage may have different number of components? How does it work in this case?
- Why unsigned long long in Int64 serialization?

And... I'm partially agree with @Lumak that these changes in Object are a bit too bloating, especially because of very common names of getters. What do you think about keeping these accessors in Context only? Should be short enough and avoid adding dozen of new getters to every class in hierarchy.

rokups:
> Image::SetSubimage. Is it really sane that host image and subimage may have different number of components? How does it work in this case?

This method was imported from Atomic. Cant comment on implementation. If its not good enough can just drop it.

> Why unsigned long long in Int64 serialization?

No good reason. Must be a habit i guess. It would work the same way with signed or unsigned type as far as i can tell.

> What do you think about keeping these accessors in Context only?

Implementing these getters i had two objectives:

* Side-step unnecessary hashmap lookups
* Provide convenient access to subsystems so that we could use getters everywhere instead of caching subsystem pointer in a local scope.

First point was already destroyed by having me to drop pointer caching. By leaving getters only in the `Context` second point would be destroyed as well. `GetSubsystem<UI>()` isnt that much longer than `context_->GetUI()`. Besides renaming `GetCache()` to `GetResourceCache()` also is very inconvenient. I am totally ok with having long `GetFileSystem()` getter since it isnt used that often. Resource cache on the other hand is used quite a lot and i think shorter name should be preferred to the longer one.

Anyhow if you think these getters are too bloating (i completely disagree) then there is little point in keeping them in `Context` while dropping them from `Object` as they no longer complete goals they were implemented with.

eugeneko:
I'm about to merge this PR. If anybody wants to comment, it's time.

weitjong:
I have a few comments, if you can wait.

eugeneko:
That's why I'm asking. Take your time.

weitjong:
- There are still missing binding for some of newly added C++ URHO3D_API.
- Is it really necessary to save the style name in the base class of UI-element? I have my reservation for this one.

Minor things are:
- New import for the DebugRenderer class not properly sorted alphabetically.
- FileSystem::GetTemporaryPath() looks like a duplicate of FileSystem::GetAppPreferencesDir(), and also forget to free the resource from SDL. Perhaps should drop this one.
- Color::ToUIntArgb() is copied from Atomic which is only added in Atomic to suit their profiler format which has nothing to do with us. BTW, if the code is not the original work from the contributor then I suppose it needs to be clearly stated so, perhaps in a comment line above the code. We have done that for a few of utility method in math class. But in this case, I think we could just drop this new method.
- Wrong comment for the `IntVector2(const float* data)`.

Lumak:
In regards to block event, isn't that block **ALL** events?  Does it have common use?

rokups:
> Is it really necessary to save the style name in the base class of UI-element? I have my reservation for this one.

It is not necessary but is very convenient. Originally i implemented it for UIEditor, so that opened UI layout looks correct instantly. Without it there would have to be a two-step layout opening procedure. 1. Load style and 2. load layout. I plan to merge UIEditor into Editor so we may rethink this.

> FileSystem::GetTemporaryPath() looks like a duplicate of FileSystem::GetAppPreferencesDir(), and also forget to free the resource from SDL. Perhaps should drop this one.

This was originally intended to return platform temporary directory, but i stopped bit short of that as i realized i do not know enough about mobile platforms to do that. Maybe should still proceed to implement it as per original goal?

> Color::ToUIntArgb() is copied from Atomic which is only added in Atomic to suit their profiler format which has nothing to do with us. BTW, if the code is not the original work from the contributor then I suppose it needs to be clearly stated so, perhaps in a comment line above the code. We have done that for a few of utility method in math class. But in this case, I think we could just drop this new method.

This method is written by me, but `Image::SetSubImage()` isnt, so i will add comment there. `Color::ToUIntArgb()` can indeed be dropped as in latest profiler incarnation it is no longer needed. Originally `easy_profiler` API was exposed and their macros were used, therefore compatible color format was needed. Now that API is hidden from public headers this method indeed has no use. I will remove it.

> There are still missing binding for some of newly added C++ URHO3D_API.

Old bindings are not broken and i think it would only be fair for contributors to make sure of that. However i will not be writing any bindings as i am not interested in spending my free time on that.

> In regards to block event, isn't that block ALL events? Does it have common use?

I use it for suspending/resuming states, where state is scene and all of it's UI.

weitjong:
Until we have changed the contribution check list or that we have the automated binding mechanism in place, I would not agree to accept PR without complete binding. By your logic, since I only need to make the engine works for Linux then I could just don't give a damn to other platforms.

weitjong:
That did not come out quite right. What I want to say is, I will only agree to merge the PR after one of the core dev complete the bindings. If you want the PR to be merged faster than you can help to do the binding work. Our time is as valuable as yours.

eugeneko:
@weitjong I've written some bindings, then I can write more of them. But,,,
Do you have some secret techinque of writing AS bindings?
I find it extremely demotivating because I cannot verify such bindings without manual testing every single f~n method...

rokups:
> Our time is as valuable as yours.

This is exactly why it is a good idea to reconsider binding policy. We all could do more if we did not have to do that manual labour noone wants to do.

@eugeneko are you saying that bindings, even if they compile, do not necessarily work?

eugeneko:
@rokups That's the exact reason why I want to push binding policy.

If you make a mistake in Lua binding, it won't compile.
If you make a mistake in AS binding, you will either get runtime binding error or crash on binding call. Or hidden bug.

I already pushed such broken bindings some time ago, Then fixed it in few days or weeks...

weitjong:
> Do you have some secret techinque of writing AS bindings?

Unfortunately I don't have. Usually I just search for similar looking method as used that as template. Launching Urho3DPlayer will also usually help to spot binding mistake. In the past when in debug build config, it would check all the registered bound classes. But I think that logic has been removed.

eugeneko:
> Usually I just search for similar looking method as used that as template. 

The same. However, such templates looks too complicated sometimes (e.g. involve static wrappers).
And I'm starting to hestitate whether to copy-paste these ugly wrappers or use simpler binding with higher risks (maybe I'm missing something). That's why I added only few bindings in Graphics for now.

> Launching Urho3DPlayer will also usually help to spot binding mistake

It would help in case of syntax error in binding string, but didn't help with malformed binding or wrong signature.

rokups:
I should probably have rebased later for easier change review.. I realized it being mid-way, sorry :|

Changes addressed in latest push:

* Removed `Color::ToUIntArgb()`
* Updated `IntVector2` cosntructor comment which was incorrect
* Removed extra includes from `DebugRenderer.cpp`, now no new includes are added by this PR
* Updated `FileSystem::GetTemporaryPath()` to return platform temporary paths when possible. It still uses `SDL_GetPrefPath()` on mobile and TVOS. I could not find a clear answer on what temporary path should be on those OS or getting that path required objective-c code.
* Added comment noting authorship of `Image::SetSubImage()`

rokups:
Bindings added. Renamed `GetTemporaryPath()` to `GetTemporaryDrir()` for consistency with other methods returning dirs. Added `monitor` parameter to `Graphics::GetDisplayDPI()` too.

eugeneko:
I'm fine with this PR, again.

Source/Urho3D/IO/FileSystem.cpp
```diff
@@ -1067,4 +1067,34 @@ bool IsAbsolutePath(const String& pathName)
     return false;
 }
 
+String FileSystem::GetTemporaryDir() const
```
weitjong: Ok. Now I understood more what you wants for this new method. To me, this method implementation is too convoluted just to return a platform-specific temp dir. The fallback "./temp" directory name is of no use if it is not exists also. For all *NIX, the tmp dir should be:
- The environment variable TMPDIR, if it is defined.
- The value of the P_tmpdir macro, if it is defined. Note that importing the `stdio.h` still does not guarantee this.
- The last fallback is `/tmp`.

Source/Urho3D/Graphics/Graphics.h
```diff
@@ -379,7 +379,7 @@ class URHO3D_API Graphics : public Object
     /// Returns true if window is maximized or runs in full screen mode.
     bool GetMaximized() const;
     /// Return display dpi information: (hdpi, vdpi, ddpi). On failure returns zero vector.
-    Vector3 GetDisplayDPI() const;
+    Vector3 GetDisplayDPI(int monitor=0) const;
```
weitjong: spacing

Source/Urho3D/LuaScript/pkgs/Graphics/RenderPath.pkg
```diff
@@ -97,6 +97,8 @@ class RenderPath
     bool Append(XMLFile* file);
 
     void SetEnabled(const String tag, bool active);
+    bool IsEnabled(const String& tag) const;
```
weitjong: Lua binding has a quirk. It cannot use String reference.

Source/Urho3D/UI/UIElement.cpp
```diff
@@ -220,6 +220,27 @@ bool UIElement::LoadXML(const XMLElement& source, bool setInstanceDefault)
 
 bool UIElement::LoadXML(const XMLElement& source, XMLFile* styleFile, bool setInstanceDefault)
 {
+    // If this is a root element get style file from which style is loaded.
```
weitjong: As commented before, I don't think it is wise to mix loading of UI-layout with UI-style file. The LoadXML() has assumed too much on how the style file could be loaded.

Source/Urho3D/IO/FileSystem.cpp
rokups: > For all *NIX, the tmp dir should be

Are you sure that will also work on Android and iOS?

Source/Urho3D/UI/UIElement.cpp
rokups: Any idea what problems could this cause?

Source/Urho3D/IO/FileSystem.cpp
weitjong: I didn't say that. The embedded platform is very close to Linux but it is not POSIX compliance, so they probably have its own system call to get the temp dir, similar to how Windows platform has one.

Source/Urho3D/UI/UIElement.cpp
weitjong: What if I made a mistake to forget to load the style before loading the UI-layout and I want it to fail instead of conveniently auto-loading one from resource because I could have my style constructed from memory or by using any other means.

Source/Urho3D/IO/FileSystem.cpp
rokups: Mobile platforms are a reason why my implementation was as it was - it was unclear to me how exactly they behave regarding temp dir. My android has `TMPDIR` set and it points to `/data/local/tmp` so that will probably work. Not sure what happens on iOS. No device to test.

Source/Urho3D/UI/UIElement.cpp
rokups: Then your layout xml would not have a `styleFile` parameter and it would behave just like it does without this patch. It is totally optional.

Source/Urho3D/LuaScript/pkgs/Graphics/RenderPath.pkg
rokups: Really? I used existing bindings as example.. And to verify i ran a search for `const String&` on `LuaScript/pkgs` dir and there are many results. Are all those other bindings broken too?

weitjong: Most probably. I just did a search and there are 57 in total. Compared to 500+ that don't use reference. I suppose the issue that it may cause is quite subtle.

Source/Urho3D/UI/UIElement.cpp
weitjong: I am still not convince. We already have "defaultStyle_" member variable and now we also have "defaultStyleFileName_". So, which one is the ultimate source of truth. Have you looked at the implementation of the `UI::LoadLayout()`? If you modify both of these method overloads to optionally read the stylefile name attribute when it is available and use it to override the default style of the root element before calling the `UIElement::LoadXML()` then I am fine with that.

rokups: I will remove this change. While something along these lines would be nice, there is no dire need of it. For now ill copy behaviour of current editor - pre-selecting default style before loading UI layout. We can revisit this matter in the future should there be a need.

eugeneko:
@weitjong Please let me know when you have no more questions about this PR.
Or just merge it on your own.

weitjong:
Strange, for some reason GitHub did not send me notification that new pushes have been made. Just have a quick glance on the revised `GetTemporaryDir()` method. I believe another `#ifdef` may be needed for Emscripten. My workspace is switch to SDL upgrade branch right now with some pending changes, so I could not investigate how it should be handled for that code branch. Perhaps, need to verify for iOS/tvOS platforms as well.

Other than that, I am fine with this PR. You can merge it as it is now if you want it and then fix `GetTemporaryDir()` later.

rokups:
Emscripten has `S_tmpdir` defined. Maybe it already handles this dir in it's virtual file system? https://github.com/kripken/emscripten/blob/master/system/include/libc/stdio.h#L157

I do not have any of apple devices so i can not verify iOS/tvOS so in case anyone has such device would be great if you could check if `TMPDIR` env variable is defined there.

weitjong:
For Emscripten we need a runtime temp dir when the app runs in the browser, I suppose. The link you provided seems to be a temp dir during compile time.

eugeneko:
Ok, I'll merge it now. I think temp dir issues on some platforms could be fiexd later.

PR #2169, opened 2017-11-11
Remove headers few headers from Urho3DAll.h

rokups:
* `Urho3D/IK/IKConverters.h` - includes headers from third party ik library which is not bundled in the SDK.
* `Graphics/GraphicsImpl.h` - includes headers from third party GLEW library which is not bundled in the SDK.

`IKConverters.h` removal should be safe, as far as i can tell converters are not needed for using IK subsystem.

I am unsure about `GraphicsImpl.h` though. Are there cases where user would need to use whatever this header provides? If so - maybe third party libs should include GLEW instead of removing `GraphicsImpl.h` include from mega-header?


weitjong:
Thanks for your PR. I will probably not merge this in as it is, but will instead modify the regex to exclude those two out.

rokups:
I can do that ;)

rokups:
Also removed `|Urho3D|` from regex, it had no effect on generated header

weitjong:
Thanks. I don't think you need to put the directory as the file name should be unique enough, also to make it consistent with others. And also I won't be so sure about removing the `Urho3D`. Although I could not remember exactly now the use case where it is needed, but it IS needed.

weitjong:
One more thing. Could you also list them in the right order.

rokups:
Whats the right order?

weitjong:
The list is in alphabetically order.

PR #2170, opened 2017-11-12
add_subdirectory() support

rokups:
This change revives support for add_subdirectory().

#### Things worth mentioning

`define_dependency_libs()` macro was removed, instead each target defines it's own dependency libs. They are still appended to `LIBS` list used by other build system macros.

Added `pkg_check_modules_internal()` macro to `FindUrho3D.cmake`. Macro leverages the fact that all SDK builds include `Urho3D.pc` with engine build flags and link libraries.  This file gets parsed for engine link libraries and compile definitions instead of having `FindUrho3D.cmake` guess them. This lifts a limitation where project build had to have exact same cmake parameters defined as sdk build. For example not defining `URHO3D_IK=OFF` in project build, when SDK was built with `URHO3D_IK=ON`, resulted in build errors. `FindUrho3D.cmake` now returns `URHO3D_DEFINES` variable too.

Removed use of `Urho3D::WString` from `stb_image_write.h` as it was causing circular dependency. It resulted in build errors on windows platforms in shared library builds. It was replaced with manual conversion of multibyte string to wide string using `MultiByteToWideChar()` windows API.

This PR does not solve all the problems, but is a step to the right direction. For example we still must use `find_package(Urho3D)` in order to link to library. Linking directly to Urho3D target is not possible because target does not export required include directories and compile definitions through `target_include_directories()` and `target_compile_definitions()`.

Example on using engine as a in-tree dependency:

```cmake
# We can use our own project name as one would expect
project(Urho3DAsLibraryExample)

# Add game engine to the project.
add_subdirectory(Urho3D)
# Still need to append module path so UrhoCommon can be found
list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/Urho3D/CMake/Modules)
# This is required, but once engine build system starts properly using target_include_directories() / target_compile_options() / target_link_libraries() this should be no longer needed.
find_package(Urho3D REQUIRED)
```
Manually:
```cmake
set(SOURCE_FILES main.cpp)
add_executable(Urho3DAsLibraryExample WIN32 ${SOURCE_FILES})
target_include_directories(Urho3DAsLibraryExample PRIVATE ${URHO3D_INCLUDE_DIRS})
target_link_libraries (Urho3DAsLibraryExample ${URHO3D_LIBRARIES} -lpthread)
```
Using Urho3D macros:
```cmake
set (TARGET_NAME Urho3DAsLibraryExample)
define_source_files ()
# This also automatically links to Urho3D libraries found by find_package()
setup_main_executable ()
```


weitjong:
Thanks for the PR, I will review it after I complete with SDL-2.0.7 upgrade.

rokups:
Odd that one build failed on appveyor. Could you trigger a rebuild? Very same commit built in my branch fine.. https://ci.appveyor.com/project/rokups/urho3d/build/233

weitjong:
For Appveyor it is not possible to just rerun one build job. The retrigger will rerun the whole four build jobs again unfortunately. Anyway, it is OK. We are aware that our CI providers (both Travis and AppVeyor) are a bit wonky at time. We know that a lone failed/errored CI job could just mean nothing. That is why our jobs are usually configured in pairs.

OvermindDL1:
Although I was a fan for `add_subdirectory` support before, I'm thinking that something like https://hunter.sh/ might be better, ergo adding urho3d would just be like adding any other library, cached builds and no submodules and all.  I think it would take just needing one extra metadata file in the repo if I recall right.

`add_subdirectory` is definitely still nice though and would fix a lot of the hacks I currently use. ^.^;

weitjong:
Two or three years ago there was an attempt by a few of their devs to bring in Urho to hunter, but I believe they have aborted the attempt. I am not sure what is their difficulties are nor that I want to know them. For sure there is no one canonical way to build Urho.

OvermindDL1:
> Two or three years ago there was an attempt by a few of their devs to bring in Urho to hunter, but I believe they have aborted the attempt. I am not sure what is their difficulties are nor that I want to know them. 

I wonder what that could be, assuming it just needed a `cmake <dir> && make` then it should work, and options could be defined in the get package interface as well, hmm...  If I get time I may be tempted to try it myself.

weitjong:
Why there is an unrelated commit in this PR? Kindly rebase to remove that commit and submit it separately, if necessary.

I think I have mentioned this before somewhere (forget where) that I do not see the dependency to the Urho3D.pc is the right to do. There must be another way to solve the problem.

I did not complete the review. But I am willing to test run the PR against the whole CI build battery.

rokups:
All of commits are related. If you mean `WString` commit - circular dependency breaks when linking on some systems with these build system changes.

In defence of `Urho3D.pc` - current system is way worse than defending on parsing this file. .pc file is available in all SDK installations and current system does sort of double-guessing of linking flags.

I guess part of the problem can be solved by adding all engine defines to Urho3D.h, much like it is done now, but only for some flags. This does not solve double-guessing linked libraries and other compiler flags though. Do you have any ideas?

Also why parsing `Urho3D.pc` is bad? It serves well for pkg-config for exactly same purpose. I could have used `FindPkgConfig` cmake module, but it depends on `pkg-config` executable which is not present on some systems, so parsing this file manually is cross-platform way of getting info out of it. If you fear things would break - i will be here to fix it as well.

weitjong:
It is bad because it is one more dependency we don’t actually need. I feel it will make the build system even more brittle than current. Second reason, the script that generates the Urho3D.pc was done by me and I cannot guarantee I have it right for all the cases and also it is less tested. After all we do not expect our user to use pkg-config but only support that as last resort. Lastly, our existing build system is already unorthodox. The next iteration I plan to make it as standard as possible. Putting a dependency on pkg-config is the last thing i would want.

rokups:
How do you propose to solve this while not breaking old build system and making changes as little invasive as possible?

weitjong:
I have done some reading on how modern CMake scripts should look like, but it is quite some time ago, and I think the new way can solve all the problems we have without resorting to unorthodox way. Having said that, thinking about solving it is one thing and actually finding time to solve it is another.

I agree that some of changes you made in the PR is a step in the right direction. You can continue to work on the problem if you still interested in solving it, i.e. making it works for all the cross-compiling build scenarios that we support. Until then, we can simply say we do not support embedding Urho lib use case just yet.

rokups:
I definitely want to see this PR through.

At the moment main problem i see is `FindUrho3D.cmake` file. It has to provide compile definitions, compiler flags and linking libraries. This file contains lots of duplicate effort declaring same things in this file and in properties of targets.

What if `FindUrho3D.cmake` was configured at a build time and build system would just insert all the libraries/defines/flags obtained directly from Urho3D target? That would solve double work and relieve us from doing same change in multiple places. Do you find it acceptable?

Also to make things more kosher i would go as far as removing `setup_exectable()`/`setup_main_executable()`/`setup_library()` and work directly with CMake API to create targets. Is that acceptable?

weitjong:
> What if FindUrho3D.cmake was configured at a build time

No. The module should not be programmatically generated, IMHO. Currently we can point to one CMake module directory and just changing the URHO3D_HOME for different target platform and I want to keep that.

> Also to make things more kosher i would go as far as removing setup_exectable()/setup_main_executable()/setup_library() and work directly with CMake API to create targets. Is that acceptable?

If you can achieve what the macros currently do behind the scene then by all mean. And by that, the macros work their magic regardless of what supported platforms you are targeting. The truth is, originally we have it setup exactly like that and found out that we have to put if/endif here and there to cater for all the possibilities and the logic is identical and repeated all over the places and one fine day we said hey, let make it a macro to safe us the trouble. So if we have a bug or enhancement, we just need to fix it in one single place rather than going through all the places. If you still see there is wisdom of doing it without those macros after learning this then I won't stop you.

rokups:
> No. The module should not be programmatically generated, IMHO. Currently we can point to one CMake module directory and just changing the URHO3D_HOME for different target platform and I want to keep that.

Then i do not know how to avoid repetition in this case.

ruslo:
Hello guys,

> Two or three years ago there was an attempt by a few of their devs to bring in
> Urho to hunter, but I believe they have aborted the attempt.

Just for your information I've finished adding Urho3D few days ago:
* https://github.com/ruslo/hunter/issues/389#issuecomment-350399303
* https://docs.hunter.sh/en/latest/packages/pkg/Urho3D.html

For now it has a limited configuration `URHO3D_{ODBC|SQLITE|LUA}=OFF` and only
host Windows/Linux/OSX is working but I'm planning to add sqlite/lua and
Android/iOS.

> I am not sure what
> is their difficulties are nor that I want to know them. For sure there is no
> one canonical way to build Urho.

Biggest difficulties was not about Urho3D itself but about dependencies.
Urho3D trying to do the things that designed to be done with the help of package manager.

All the https://github.com/urho3d/Urho3D/tree/master/Source/ThirdParty stuff
should be managed by Hunter. Assuming that I had to rewrite CMake code of 3rd
parties to allow loading them as a stand-alone projects, for example:
* https://github.com/hunter-packages/Box2D/commit/0e1e8def120a628c3f01506897cf6c483ad330a8

Plus I have to apply patches that you've made to dependencies, e.g.:
* https://github.com/hunter-packages/SDL2/commit/dc9c443ad88f037c43bfb19ba32c015a2f061378

> What if FindUrho3D.cmake was configured at a build time and build system would
> just insert all the libraries/defines/flags obtained directly from Urho3D
> target?

`Find*.cmake` files are not designed to be installed by the package.
Project should generate and install `*Config.cmake` file
and CMake will add all needed defines and dependent libraries.

For Urho3D it will look like this:
* https://github.com/hunter-packages/Urho3D/blob/3fb779949f83792462692bba2be4c2d7e634306c/Source/cmake/Config.cmake.in#L1-L43

PS I'm not proposing or suggesting anything, just dropping few words that you might find helpful :)

weitjong:
I think I understand what you are saying about the 3rd party libs. Ideally if we don’t have any local patches then we could just use Bullet or Box2D or SDL dev packages provided by the package manager, and by package manager I really meant “apt” or “yum” or “dnf”. There is really no need to build them ourselves and not even via Hunter. Unfortunately though, in practice there are always things that we need to patch up quickly like to fix a bug or to make it fit/integrate with Urho, within our own release cycle. In doing so, we lose the ability to let our user to upgrade 3rd-party libs from external available source. It is a trade off.

ruslo:
> In doing so, we lose the ability to let our user to upgrade
> 3rd-party libs from external available source.

Yes, you're locking your users with internal version of dependency.

For example let's take a look at Bullet.

You do install headers to `<root>/include/Urho3D/ThirdParty/Bullet/LinearMath/btVector3.h` but keep using `#include <Bullet/LinearMath/btVector3.h>` in `<root>/include/Urho3D/Physics/PhysicsUtils.h` meaning that logically it's the same as installing to `<root>/include/Bullet/LinearMath/btVector3.h>`. And if I want to use custom version of Bullet we got conflict. Note that "I use custom Bullet" includes case when I use product X and product X choose same bundle dependency strategy about Bullet.

Similar problem on the symbols level:
```
> nm --defined-only --demangle lib/Urho3D/libUrho3D.a | grep btConvexTriangleMeshShape::getEdge
0000000000000000 T btConvexTriangleMeshShape::getEdge(int, btVector3&, btVector3&) const
```

You're using exactly the same symbols that stand-alone Bullet provides.

If you will protect bundled 3rd parties on header level (like #include <Urho3D/external/Bullet.h>) and symbols level (like `namespace Urho3D { namespace external { namespace Bullet ...`) then you will increase final size of executable by double-sizing usage of Bullet dependency: my custom Bullet and quite the same code from Urho3D but by different name.

Sending your patches upstream is the only optimal strategy. There is no tool to resolve this conflicts, this is how language is working. I don't think you have a lot of other options here.

weitjong:
I agree on your last statement that we should become a good citizen by submitting our relevant local patches upstream. However, even if we do so, I feel there is no easy way to avoid the thing we do right now to encapsulate some of our dependency libs. The upstream may or may not accept our PR in time. And timing is actually key. From the past experience, all the local bug fixes that we made for SDL would eventually fixed by SDL developers over time. Anyway, I think we have gone off topic for this issue.

rokups:
@weitjong so what do you think i should do with `FindUrho3D.cmake` in this PR?

weitjong:
I think I have already hinted you that the next iteration of our build system should we have time to do it, will be a total revamp, replacing all the old CMake 2.8.x scripts with the more modern way. If you haven't done catching up on that then I suggest you do that reading first. Personally I have not had a chance to practice what I have learnt yet. Perhaps you would like to start with this one.

https://cmake.org/Wiki/CMake/Tutorials/Exporting_and_Importing_Targets

I also think this is the approach used by @ruslo team in Hunter.

rokups:
I wasnt sure you would find a complete build system replacement PR as acceptable. I already do have a build system replacement in my fork repository. It still does not have some stuff current build system has, like running samples as tests or iOS builds.

Also was not aware of target exporting in cmake, looks like good stuff. I will experiment with it in my branch and if i manage to make it a feasible replacement for `FindUrho3D.cmake` then i will update this branch. If not.. then i will close this PR.

weitjong:
I would not mind scrap the old one if the newer and more modern one can do better jobs than the old one.

rokups:
I took a look at possible options and arrived to a conclusion that the only sensible way to export Urho3D target requires to convert all third party dependencies to object libraries. Reason for this is that exporting target requires exporting all of it's dependencies, meaning almost entire ThirdParty folder. If ThirdParty targets were converted to object libraries then they would not have to be exported. In addition to that conversion would immediately create monolithic engine library and currently employed hacks to achieve that would no longer be needed. This is rather big undertaking so for the time being i am closing this PR. If i succeed this conversion in my fork i will revisit this PR.

weitjong:
Thanks for update.

rokups:
I implemented proof of concept in my fork. I would like to share what i found out.

As i said all third party libraries (including header-only libs) were converted to CMake `OBJECT` libraries. However due to incompleteness of CMake such libraries can not have dependencies. In order to mitigate these limitations and allow dependencies between third party libraries i had to create [a hackish macro](https://github.com/rokups/Urho3D/blob/master/CMake/Modules/UrhoCommon.cmake#L219-L280). `target_link_objects()` macro is like `target_link_libraries()` except it should be used for linking to `OBJECT` targets or linking dependencies to `OBJECT` targets. Macro pulls in dependencies/includes/defines.

With this done we can produce a true monolithic static library and no hacks are needed. However situation with monolithic shared library is trickier. I updated bunch of third party libs to properly export their interface from shared library. In case of Box2D i modified headers to add api defines that export Box2D classes. In case of Bullet it is not possible and i had to build bullet as `STATIC` instead of `OBJECT`, and export static library of Bullet even in shared engine builds. This makes it possible for user to use Bullet APIs with shared library builds of the engine at a cost of bullet code duplication.

Exporting Urho3D target works. Since bullet is always a static library it has to be exported as well. Other third party dependencies are not required to be exported because they are `OBJECT` libs.

This isnt a completely clean solution, but it is working. If similar solution was to be worked out for a PR, how do you think Bullet situation should be handled?

weitjong:
Here are my two cents. Since I only grasped what you are doing by your explanation above without looking at the actual work itself (I am commuting now), it sounded like you are replacing a CMake hack with yet another hack. However, if your proposed hack involves altering the 3rd-party libs in any way then I would say it is less attractive than the current existing one.

rokups:
Hey @weitjong im revisiting topic a little and having issues with CI builds. Is current CI setup supposed to work with repos other than main urho3d one? Im experiencing odd ruby script failure there. If you could take a peek at error and point me to the right direction it would be very helpful: https://travis-ci.org/rokups/Urho3D-Upstream/jobs/400897145

P.S. Got Urho3D cmake target exported and working (more or less).

weitjong:
Those scripts are designed specifically for Urho3D github project. So they won’t work out of the box outside that context. Just delete the scripted instruction  that causing any problem for you for a start.

rokups:
Unfortunately that keeps failing with more obscure errors. It would be extremely useful if we could use CI setup with our own accounts while urho3d-specific stuff would be disabled if build does not belong to main project account. Would that be possible?

weitjong:
Anything is possible. In fact before I set them up for Urho3D main github project, I have those setup tested on my own fork first. It was quite some time back. And I myself don’t have the full recollection of what is needed unless I am there looking at those old setup and past emails that I had with Lasse containing the secret credentials information required for the setup. I don’t think I have time to help you to do that now.

weitjong:
Also you may want to set thing up differently. Nowadays people use docker for anything cloud, including CI build test.

PR #2176, opened 2017-11-19
Updated 2D samples from pull request #641

thesquib:
I've updated the samples @Mike3D issued in pull request #641 so they compile and work with Urho3D master. There are two samples included, a 2d platformer, and a 2d isometric dungeon. 

Only minor changes were required to get the samples working. 

Additionally, I added sound and collisions similar to the isometric demo, very similar to the mechanics from the 2d platformer. I also fixed support for compiling on Windows. 

Thanks to @Mike3D for creating these.


eugeneko:
@weitjong Any idea why it fails? According to @thesquib, samples terminate fine with testing turned on.

thesquib:
I've built this in test mode and run these on Ubuntu 16.04, passing the flag 'timeout' to the samples, with a value of 5. The samples terminate as expected. Unsure what the problem is here, unless there is a crash which I'm not seeing. I'll try a few repeat build and runs.

thesquib:
@weitjong also there was not much feedback on the previous PR so I don't know if there was other work to do?
I haven't found any issue yet which would cause the failure.

Source/Samples/49_Urho2DIsometricDemo/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+set (TARGET_NAME 49_Urho2DIsometricDemo)
+
+# Define source files
+define_source_files (EXTRA_H_FILES ${COMMON_SAMPLE_H_FILES} ../Utilities2D/Sample2D.h ../Utilities2D/Sample2D.cpp ../Utilities2D/Mover.h ../Utilities2D/Mover.cpp)
+
+# Setup target with resource copying
+setup_main_executable ()
+
+# Setup test cases
+add_test (NAME ${TARGET_NAME} COMMAND ${TARGET_NAME} -timeout ${URHO3D_TEST_TIME_OUT})
```
weitjong: This line is wrong. It should just call "setup_test ()" macro like the rest.

weitjong:
My apology. I just added one liner comment which should fix your CI build test issue. Haven't got time to review the PR as whole though.

Source/Samples/49_Urho2DIsometricDemo/CMakeLists.txt
thesquib: Thanks

thesquib:
Thanks @weitjong, the problem is fixed.

eugeneko:
I don't have enough passion to thoughtfully review this PR, but I tested these samples a bit and have no objections about merging it.
(Will be merged in three days if no activity here)

eugeneko:
@weitjong Do you want to review this PR?

1vanK:
It seems physics interaction processed in Update instead FixedUpdate. At the moment Update and FixedUpdate is equivalent for Box2D, but it is bug and can be redone someday

weitjong:
Thanks for asking. Do what you think it’s best. I will check out the code later.

thesquib:
@1vanK thanks I'll check that out soon and update it

eugeneko:
I'm merging it. Don't want to hold these samples one more time. Please make separate PR if you want to change anything.

thesquib:
Good idea, thanks @eugeneko

PR #2185, opened 2017-11-22
3D UI cleanup and update

rokups:
Turns out 3D UI implementation was not flexible enough. This change allows ultimate freedom by splitting up responsibilities correctly.

UIComponent is no longer responsible for storing batches and vertex data. This information was moved to new `RenderToTextureData` struct inside `UI` subsystem. `UI` and `UIComponent` are no longer friends of each other.

New class `UIElement3D` was created. It is used as root UI element of UI rendered in a 3D scene. It implements `ScreenToElement()` method which is used for translating screen coordinates to element coordinates. Previously `UIComponent` did translation of these coordinates. `UIElement3D::ElementToScreen()` is not implemented at this time and returns `{-1, -1}` (invalid coordinates). It is also possible to set a custom viewport to root element of `UIComponent` and this viewport will be used for processing user input.

In order to render UIElement to a texture it is enough to call SetRenderTexture(texture), where texture is set up as `TEXTURE_RENDERTARGET`. In order to properly support input of `UIElement` rendered to a texture user should subclass `UIElement` and implement `ScreenToElement()` method which properly translates global screen coordinates to local element coordinates.

New API:
* UIElement::SetRenderTexture(Texture2D*)
* UI::SetRenderTexture(UIElement*, Texture2D*)

Removed API:
* UI::SetRenderToTexture(UIComponent*, bool)

### DIFF
--- a/Source/Urho3D/UI/UI.cpp
+++ b/Source/Urho3D/UI/UI.cpp
@@ -445,8 +445,12 @@ void UI::RenderUpdate()
     {
         RenderToTextureData& data = it->second_;
         if (data.rootElement_.Expired())
+        {
             it = renderToTexture_.Erase(it);
-        else if (data.rootElement_->IsEnabled())
+            continue;
+        }
+
+        if (data.rootElement_->IsEnabled())
         {
             data.batches_.Clear();
             data.vertexData_.Clear();
@@ -465,10 +469,8 @@ void UI::RenderUpdate()
                 batch.AddQuad(scissor.left_, scissor.top_, scissor.right_, scissor.bottom_, 0, 0);
                 data.batches_.Push(batch);
             }
-            ++it;
         }
-        else
-            ++it;
+        ++it;
     }
 }
 
@@ -503,9 +505,9 @@ void UI::Render(bool renderUICommand)
     // Render to UIComponent textures. This is skipped when called from the RENDERUI command
     if (!renderUICommand)
     {
-        for (auto it = renderToTexture_.Begin(); it != renderToTexture_.End(); it++)
+        for (auto& item : renderToTexture_)
         {
-            RenderToTextureData& data = it->second_;
+            RenderToTextureData& data = item.second_;
             if (data.rootElement_->IsEnabled())
             {
                 SetVertexData(data.vertexBuffer_, data.vertexData_);
@@ -548,9 +550,9 @@ void UI::DebugDraw(UIElement* element)
             element->GetDebugDrawBatches(debugDrawBatches_, debugVertexData_, scissor);
         else
         {
-            for (auto it = renderToTexture_.Begin(); it != renderToTexture_.End(); it++)
+            for (auto& item : renderToTexture_)
             {
-                RenderToTextureData& data = it->second_;
+                RenderToTextureData& data = item.second_;
                 if (!data.rootElement_.Expired() && data.rootElement_ == root && data.rootElement_->IsEnabled())
                 {
                     element->GetDebugDrawBatches(data.debugDrawBatches_, data.debugVertexData_, scissor);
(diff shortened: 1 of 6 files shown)


Source/Urho3D/UI/UI.cpp
```diff
@@ -2076,16 +2078,31 @@ IntVector2 UI::GetEffectiveRootElementSize(bool applyScale) const
     return size;
 }
 
-void UI::SetRenderToTexture(UIComponent* component, bool enable)
+void UI::SetRenderTexture(UIElement* element, Texture2D* texture)
```
eugeneko: Don't really like this name. Maybe, `SetElementRenderTexture` is more meaninful... What do you think?

rokups: Sounds good to me

Source/Urho3D/UI/UIComponent.cpp
```diff
@@ -182,25 +193,51 @@ bool UIComponent::ScreenToUIPosition(IntVector2 screenPos, IntVector2& result)
...
                 continue;
-            return false;
+            return result;
         }
 
         Vector2& uv = queryResult.textureUV_;
-        result = IntVector2(static_cast<int>(uv.x_ * rootElement_->GetWidth()),
-                            static_cast<int>(uv.y_ * rootElement_->GetHeight()));
-        return true;
+        result = IntVector2(static_cast<int>(uv.x_ * component_->GetRoot()->GetWidth()),
```
eugeneko: But `component_->GetRoot()` is always `this`, am I right?

Source/Urho3D/UI/UIComponent.h
```diff
@@ -61,29 +65,41 @@ class URHO3D_API UIComponent : public Component
...
+    /// Construct.
+    UIElement3D(Context* context);
+    /// Destruct.
+    virtual ~UIElement3D() override = default;
+    /// Register object factory.
+    static void RegisterObject(Context* context);
 
-    friend class UI;
+    /// Set UIComponent which is using this element as root element.
+    void SetUIComponent(UIComponent* component);
```
eugeneko: Looks like hierarchy flaw because `UIElement3D` has bidirectional connection with `UIComponent`. What's the point of splitting one class into two if they both still _require_ each other to work?

Source/Urho3D/UI/UIComponent.cpp
rokups: `component_->GetRoot()` returns `UIElement` which acts as a root element. This is same `UIElement` that is passed to `UI::SetRenderTexture()`

Source/Urho3D/UI/UIComponent.h
rokups: `UIElement3D` acts as a root UI element for layout rendered to a texture. It handles screen -> element mouse translation. `UIComponent` handles UI texture rendering on to an object. They do different things. Problem is that in order to do mouse coordinates translation we need to know where object is rendered. So.. Whats the alternative? UI subsystem needs to query translated coordinates and this is by far cleanest way i could think of.

rokups:
So should i replace `UIElement3D::SetUIComponent()` with `UIElement3D::SetDrawable()`?
And can i move `UIElement3D` to `UIComponent.cpp`?

eugeneko:
> So should i replace UIElement3D::SetUIComponent() with UIElement3D::SetDrawable()?

Nah, let's keep this dependency as it. It's simpler than anything else.

> And can i move UIElement3D to UIComponent.cpp?

Do you mean class declaration? Yes, if possible.

rokups:
But if class is moved to cpp file then interface does not really matter. Considering `UIElement3D` requires only drawable and scene i think change would make code more kosher..

eugeneko:
@rokups To pass drawable into `UIElement3D` you have to ensure that this drawable is always up-to-date. If it's easy, let's do it. If not, let's just drop it.

To be honest, I consider the `UIComponent` a bit hacky.
It should be a kind of dynamic material, not a component. But Urho doesn't have asset pipeline for dynamic materials. So, I prefer to keep things as is unless we have better solution.

eugeneko:
Will be merged (with minor changes and a bit more review) in three days if no activity here

Issue #2186, opened 2017-11-23
Auto Scale UI when in high DPI mode


thesquib:
At the moment the UI in the samples is half size on high dpi displays (e.g MacBook Pro retina screen). The editor code does scale the UI however. 

The scaling should be done by default, as this behaviour can be changed if it is not desired.

weitjong:
I believe the auto-scaling is already implemented for Retina Display (High DPI mode).

PR #2187, opened 2017-11-23
Auto scale UI when in high dpi mode

thesquib:
Scale UI when in high DPI mode (issue #2186).


weitjong:
Thanks for your PR. IMHO, we cannot safely set the UI scale just based on the engine parameter. We can only safely do so in the Graphics subsystem after we have confirmed that the display device indeed support Retina resolution.

thesquib:
Thanks, I'll think on this some more.

PR #2188, opened 2017-11-23
PugiXml crash fix

rokups:
Fixes a crash in PugiXml when removing a node which does not belong to a parent.

### DIFF
--- a/Source/ThirdParty/PugiXml/src/pugixml.cpp
+++ b/Source/ThirdParty/PugiXml/src/pugixml.cpp
@@ -1288,6 +1288,10 @@ PUGI__NS_BEGIN
 	{
 		xml_node_struct* parent = node->parent;
 
+		// Urho3D fix: fixes crash when node does not belong to a parent.
+		if (parent == 0)
+			return;
+
 		if (node->next_sibling)
 			node->next_sibling->prev_sibling_c = node->prev_sibling_c;
 		else


eugeneko:
Could you explain Urho use case that lead to the crash?

rokups:
Calling `XMLelement::Remove()` on element that is already removed causes the crash. It happens frequently during undo/redo when editing UI style in new editor.

eugeneko:
There should be some check in either Urho or Pugi that disallow user to remove child from not-a-parent node. Will debug it a bit...

rokups:
I believe this is where pugi should disallow removal. No method should ever crash due to incorrect parameters passed to it.

eugeneko:
What happens if we remove element from not its parent? How shall it behave? Your case looks like a subcase of this generic case.

rokups:
Pugixml does verify parent owning the child before doing removal: https://github.com/rokups/Urho3D/blob/master/Source/ThirdParty/PugiXml/src/pugixml.cpp#L5915

eugeneko:
@rokups Could you please provide crashing example? Something like that works fine for me

    XMLFile file(context_);
    auto root = file.CreateRoot("root");
    auto child1 = root.CreateChild("child1");
    auto child2 = root.CreateChild("child2");
    root.RemoveChild("child1");
    child2.Remove();
    child1.Remove();
    child2.Remove();
    child1.Remove();

Just want to reproduce crash before fix.

rokups:
Hmm this is odd. Anyhow do this:

1. Build editor `develop` branch with https://github.com/urho3d/Urho3D/pull/2185 as SDK
2. In resource browser open (double-click) `UI/MessageBox.xml`
3. Select main window element
4. In attribute inspector right-click `Image Rect` (below `Texture`)
5. Select `Remove from style`
6. Select `Save to style`
8. Do `ctrl+z` 2 times

eugeneko:
PugiXML is fine. The root cause is the bug in user code.

rokups:
If PugiXML is fine, then does it mean that there can not be a node without parent? If node without parent can exist then this is still a bug, even if it is not worth keeping fix downstream.

Also there is a glaring issue with `XMLElement` - copy-constructor does not make a actual copy. This works counter-intuitively and causes subtle bugs as in this case. Copy-constructor comment even says `Copy-construct from another element.` but if element copy was constructed from gets removed - now both instances of `XMLElement` become invalid. This is very bad.

PR #2192, opened 2017-11-27
Drag components in editor

orefkov:
This changes in editor allow drag components from hierarchy window to inspector's attributes editor.
It allow set ScriptInstance's attributes to component ID, for reference to component.
Early it worked only with nodes.


Lumak:
I think you've included some SDL changes by accident, and I'm still skeptical as to why the SDL_TICKS_PASSED() condition doesn't evaluate to true on init.

orefkov:
That bug already fixed in SDL - https://bugzilla.libsdl.org/show_bug.cgi?id=3935
In SDL_TICKS_PASSED used signed comparation, and after approx 24 days of uptime ticks in signed form become negative, and 0 always grater than negative value, and comparation is always false.
SDL_TICKS_PASSED must be used with previous saved ticks, not with 0.

Lumak:
My question is, how did you manage to set a condition to test this 24-days-of-uptime-ticks?

Issue #2196, opened 2017-12-04
Add DoubleClickMaxDistance to UI Subsytem


TrevorCash:
The ability to specify a maximum unit distance between the first click and second click would be a nice setting in the UI Subsystem.  

Currently double clicking registers even if you move the mouse large distances between clicks.  It would be nice if there was a facility to turn this down.

PR #2202, opened 2017-12-10
added UI::SetMaxDoubleClickDist and UI::GetMaxDoubleClickDist

TrevorCash:
Adds a maximum distance parameter to the UI subsystem to enable filtering of double clicks where the first click and second click are far away from each other on the same UIElement.  The distance is defaulted to very large so existing functionality should not be broken.


TrevorCash:
Fixes #2196

Source/Urho3D/UI/UI.h
```diff
@@ -402,6 +411,10 @@ class URHO3D_API UI : public Object
     Timer clickTimer_;
     /// UI element last clicked for tracking double clicks.
     WeakPtr<UIElement> doubleClickElement_;
+	/// Screen position of first mouse click for double click distance checking.
```
eugeneko: Tabs to spaces

Source/Urho3D/UI/UIEvents.h
```diff
@@ -56,6 +56,8 @@ URHO3D_EVENT(E_UIMOUSEDOUBLECLICK, UIMouseDoubleClick)
     URHO3D_PARAM(P_ELEMENT, Element);              // UIElement pointer
     URHO3D_PARAM(P_X, X);                          // int
     URHO3D_PARAM(P_Y, Y);                          // int
+	URHO3D_PARAM(P_XBegin, XBegin);                // int
```
eugeneko: Please use CAPS for parameter names, as everywhere

Source/Urho3D/UI/UIEvents.h
```diff
@@ -56,8 +56,8 @@ URHO3D_EVENT(E_UIMOUSEDOUBLECLICK, UIMouseDoubleClick)
     URHO3D_PARAM(P_ELEMENT, Element);              // UIElement pointer
     URHO3D_PARAM(P_X, X);                          // int
     URHO3D_PARAM(P_Y, Y);                          // int
-	URHO3D_PARAM(P_XBegin, XBegin);                // int
-	URHO3D_PARAM(P_YBegin, YBegin);                // int
+    URHO3D_PARAM(P_XBEGIN, XBEGIN);                // int
```
eugeneko: Sorry for confusing. I mean `(P_XBEGIN, XBegin)`

Source/Urho3D/UI/UI.cpp
```diff
@@ -95,7 +95,7 @@ UI::UI(Context* context) :
     dragBeginDistance_(DEFAULT_DRAGBEGIN_DISTANCE),
     mouseButtons_(0),
     lastMouseButtons_(0),
-	maxDoubleClickDist_(FLT_MAX),	
+    maxDoubleClickDist_(99999.0f),
```
eugeneko: There is `M_LARGE_VALUE` constant. Is it acceptable?

Source/Urho3D/UI/UIEvents.h
TrevorCash: Ah yes - I see that now.

Lumak:
Is this necessary? Clicks are registered if within the bounds of UIElement, and it's working as designed.  If you need a smaller bound then you'd typically add a smaller bound UIElement, or is this a work around so you don't have to add a smaller bound UIElement?

TrevorCash:
Its helpful. not necessary.  If I want to detect double clicks say, in the root ui element (which is very big) I want to know that I'm clicking in the same general area when a double click is registered.  I find this most helpful when not considering UIElements at all and just wanting metrics on mouse actions in the canvas.

Lumak:
To get mouse position, all you need is call:

`Input::IntVector2 GetMousePosition() const;
`
without having the need to click on anything.

Source/Urho3D/UI/UI.cpp
TrevorCash: Awesome! - did not realize that was there - just updated

Source/Urho3D/UI/UI.cpp
```diff
@@ -1635,14 +1642,37 @@ void UI::SendClickEvent(StringHash eventType, UIElement* beginElement, UIElement
...
             endElement->SendEvent(E_CLICKEND, eventData);
-        else if (eventType == E_UIMOUSEDOUBLECLICK)
-            endElement->SendEvent(E_DOUBLECLICK, eventData);
     }
 
     // Send the global event from the UI subsystem last
     SendEvent(eventType, eventData);
 }
 
+void UI::SendDoubleClickEvent(StringHash eventType, UIElement* beginElement, UIElement* endElement, const IntVector2& firstPos, const IntVector2& secondPos, int button,
```
eugeneko: `eventType` parameters seems useless. Could you remove it?

eugeneko:
Well, I'm fine with this PR.
I'll merge it if there is no comments for three days.

TrevorCash:
Cool! I dont plan on making any alterations to this branch.  Thanks Eugene!

Get Outlook for Android<https://aka.ms/ghei36>

Source/Urho3D/UI/UI.h
```diff
@@ -163,7 +165,10 @@ class URHO3D_API UI : public Object
 
     /// Return UI element double click interval in seconds.
     float GetDoubleClickInterval() const { return doubleClickInterval_; }
-
+    
+    /// Get max screen distance for double clicks to register.  default is FLT_MAX.
```
weitjong: Comment not properly case. The constant value seems to be different than the implementation.

PR #2203, opened 2017-12-11
Simple pollable timeout functionality to Timer and HResTimer

TrevorCash:


Lumak:
This could be useful, but I'd make some changes.

weitjong:
I agree with @Lumak.

Don’t forget our contribution checklis: spacing convention, script binding, etc. GetHasTimedOut() sounds weird too, perhaps IsTimedOut() is better?

Lumak:
The `timeoutTime_` is a constantly moving target, something that you can do without. Instead of storing `timeoutTime_`, store `duration_`.  Eval of timeout would then be: `curTime - startTime_ > duration_` then you don't have to readjust the `duration_` in Reset() every time it's called, like what happens with `timeoutTime_`.  And yes, I was going to say the same about `IsTimedOut()`.

TrevorCash:
@Lumak sounds like a great simplification!

Lumak:
I find the interfaces some what quirky for me, maybe it's a new way of writing things.  But here are the interfaces that I expect:
* interface to set timeout duration directly
* being able to do the above, the new constructor, Timer(duration) is not needed
* not having timeout duration nullified if not passed in GetMSec(), GetUSec(), and Reset(). keeping them as they were would be more ideal and change timeout manually if I need in the 1st bullet item.

eugeneko:
Agree. Functionality is OK, code style shall be fixed.

TrevorCash:
Is there anything still keeping this from merging in?  Just let me know and I'll fix it.

weitjong:
Thanks for your PR. We are a little short handed, hopefully we have more free time in the coming weeks to review and merge the backlog.

urnenfeld:
Myself have a subclass of Timer to implement some of the changes proposed in this PR.

Should this PR be simplified or recreated?

We could introduce just a `bool Timer::IsTimedOut(unsiged ms)` which would return **true** if **ms** miliseconds have elapsed since last Reset.

PR #2206, opened 2017-12-16
SPARK particle engine

fredakilla:
Spark particle engine integration as third party lib + add particles sample.

Alternative particles engine to Urho3D built-in particles.


fredakilla:
Any idea why compilation fail for WIN32 SHARED lib (link errors) ? miss something in CMakeLists.txt ?

1vanK:
class **URHO3D_API** ...

weitjong:
Nice! Will have a look at the soonest.

fredakilla:
I think i miss something in the build process, because it works when I explictly link the sample exe with Spark.lib (when I add Spark.lib to the linker in visual studio sample project). 
How others libs are linked to Urho ?

weitjong:
I am sorry I haven't got time to review your PR. All the 3rd party libs are built as static and therefore always statically linked to Urho3D lib regardless of whether it is built as shared or static lib type.

weitjong:
We will try to get it accepted and merged before the year end. I hope you don’t mind if we make changes to your PR including the bits in the 3rd party lib when necessary.

fredakilla:
ok, no pb.

weitjong:
I just pulled your PR into a dev branch to warm up the CI build cache and also to see what else break.

weitjong:
Slightly off-topic, but does anyone here have MinGW-W64 7.2.0 and seeing similar problem? This is the latest version provided by Fedora 27 package manager. It is my first MinGW cross-compiling build after upgrading my host system recently and it failed miserably (even when building the master branch).  These auto-detection results are reported as "found" when using older MinGW version, but no so in 7.2.0.

```
-- Looking for sys/types.h
-- Looking for sys/types.h - not found
-- Looking for stdio.h
-- Looking for stdio.h - not found
-- Looking for stdlib.h
-- Looking for stdlib.h - not found
-- Looking for stddef.h
-- Looking for stddef.h - not found
-- Looking for stdarg.h
-- Looking for stdarg.h - not found
-- Looking for memory.h
-- Looking for memory.h - not found
-- Looking for string.h
-- Looking for string.h - not found
-- Looking for limits.h
-- Looking for limits.h - not found
-- Looking for strings.h
-- Looking for strings.h - not found
-- Looking for wchar.h
-- Looking for wchar.h - not found
-- Looking for ctype.h
-- Looking for ctype.h - not found
-- Looking for math.h
-- Looking for math.h - not found
```

Florastamine:
I've been using 7.2.0 (provided by MSYS2) for a few months now without any issues. I'm on Windows though, and not doing any cross-compiling.

weitjong:
Thanks for your feedback. I will try to reproduce the issue in my Ubuntu 17.10 VM later. Hope to figure this one out and get back to fixing the MinGW build.

weitjong:
In a fresh Ubuntu host system, installing `mingw-w64` software package is enough. In a fresh Fedora host system, installing `mingw64-gcc-c++` software package would only get the shared "pthread" library, which CMake/Urho/SDL does not use; so installing `mingw64-winpthreads-static.noarch` is required. Probably we should add this to our documentation.

weitjong:
Unfortunately, IMHO, I have to say this PR is not yet merge ready. Aside from usual lacking of binding and documentation, I see that you have not finished of adapting the the classes under "Urho3D/Spark" to follow our coding convention: method and class names are using camel case with initial capital letter. The biggest problem is, you does not encapsulate the "ThirdParty/Spark" library. The third-party headers are exposed to the Urho library users. Arguably, that could be fine also, if that is what you really want and we could (force) install the Spark 3rd-party headers also alongside with other exposed 3rd-party headers (i.e. it is not first to do so). Another big problem is, your implementation classes do not reference most of the symbols in the Spark 3rd-party lib, so when building a shared lib, most of the symbols got optimized away, I believe. I have to do a quick and dirty linker flags hack to keep all the symbols around in the Urho3D.dll. However, those flags are MinGW-specific, which means the problem is only solved when cross-compiling. Building shared lib using VS will still fail after this.

weitjong:
The macOS build still failed. Will have a look when I happen to boot into my Mac. The latest Debug build on Linux failed for STATIC build configuration is caused by the disk space issue on the Travis-CI server. So, those are non-issue.

@fredakilla Let us know if you will keep pushing to address the concerns mentioned in my previous comment.

fredakilla:
ok, maybe its' better to cancel the PR if there is many things to fix. For now I need to finish another project and I will come back to spark soon. I will try to fix mentioned problems and I'll push it again. Thanks for infos and your effort.

PR #2207, opened 2017-12-16
Add some Skeleton and Math helper functions.

eugeneko:
(waiting for green)

### DIFF
--- a/Source/Urho3D/Graphics/Skeleton.h
+++ b/Source/Urho3D/Graphics/Skeleton.h
@@ -110,9 +110,11 @@ public:
 
     /// Return root bone.
     Bone* GetRootBone();
-    /// Return index of the bone by name hash. Return number of bones if not found.
+    /// Return index of the bone by name. Return M_MAX_UNSIGNED if not found.
+    unsigned GetBoneIndex(const String& boneName) const;
+    /// Return index of the bone by name hash. Return M_MAX_UNSIGNED if not found.
     unsigned GetBoneIndex(const StringHash& boneNameHash) const;
-    /// Return index of the bone by the bone pointer.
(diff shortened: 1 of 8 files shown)


Source/Urho3D/Graphics/Skeleton.h
```diff
@@ -110,14 +110,20 @@ class URHO3D_API Skeleton
 
     /// Return root bone.
     Bone* GetRootBone();
+    /// Return index of the bone by name hash. Return number of bones if not found.
```
weitjong: It is customary for Urho to return M_MAX_UNSIGNED when something is not found.

Source/Urho3D/Graphics/Skeleton.h
```diff
@@ -110,14 +110,20 @@ class URHO3D_API Skeleton
 
     /// Return root bone.
     Bone* GetRootBone();
+    /// Return index of the bone by name hash. Return number of bones if not found.
+    unsigned GetBoneIndex(const StringHash& boneNameHash) const;
```
weitjong: Missing LuaScript and AngelScript bindings.

Source/Urho3D/Graphics/Skeleton.h
eugeneko: Ok I'll do it.
BTW, `Vector::IndexOf` behaves like iterator, so it returns numElements. Is't probably an exception

weitjong: I don’t think so. Also it’s easier for caller to check against the constant rather than making another call to get the actual size in order to know if the return value is out of bound.

Source/Urho3D/Graphics/Skeleton.h
eugeneko: Yeah, probably _some_ of these functions could be binded...

PR #2210, opened 2017-12-18
File Abstraction

SirNate0:
Inspired by #2198, I went ahead and abstracted the packed and unpacked behavior of the File class into two separate classes, and created an abstract base for PackageFile to allow the creation of different kinds of these FileSources (say for a different method of compression or encryption). Presently, I think it is in working order (it hasn't been thoroughly tested, but I checked that some of the samples ran as well as the Editor and NinjaSnowWar), and it only minimally breaks the API. I'm planning on maybe making an example other file source for gzipped files or something, but that hasn't yet been done.

Notable changes required include:

- Cannot construct a `File` as it is now an abstract class -- must use a `SystemFile` instead (assuming this is for reading/writing to the file system, which is the usual use for constructing a File directly).
- Certain resource cache operations are now `FileSource` based, not `PackageFile` based (e.g. `GetPackageFiles()` presently just returns the vector of FileSources.
- `IsPackaged()` was removed from the File class.

### DIFF
--- a/Source/Urho3D/IO/FileSource.h
+++ b/Source/Urho3D/IO/FileSource.h
@@ -51,8 +51,8 @@ public:
     /// Opens a file with the given name and returns a file shared pointer from the source.
     SharedPtr<File> GetFile(const String& fileName, FileMode mode = FILE_READ);
 
-    /// Open the package file, if applicable.
-    virtual bool Open(const String& filename) { return true; }
+    /// Open the package file, if applicable. Returns true on success.
+    virtual bool Open(const String& filename) = 0;
 
     // TODO: NEL: Possibly replace the -1 with a 0 and add a separate [Is]Loaded() or LoadSuccessful() function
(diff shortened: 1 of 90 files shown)


SirNate0:
I created an example project [here](https://github.com/SirNate0/Urho-GZIP-File-Source) (same basic format as the examples, based on the code from one of them) that creates new file source and file classes to read and write gzipped files.

Source/Urho3D/IO/FileSource.h
```diff
@@ -0,0 +1,91 @@
...
+{
+    URHO3D_OBJECT(FileSource, Object);
+
+public:
+    /// Construct.
+    FileSource(Context* context);
+    /// Destruct.
+    virtual ~FileSource() override { }
+
+    /// Check if a file exists within the package file. This should be case-insensitive on Windows and case-sensitive on other platforms.
```
thesquib: Comments still refer to Package File  instead of File Source

PR #2217, opened 2017-12-24
Expose body_->GetContactList

cirosantilli-lovechina:
This provides a convenient method to get the list of bodies currently
colliding with the current body.

This is already possible through events, but this method is sometimes
more convenient.

Sample usage:

```
Vector<std::tuple<Node*, Vector2, Vector<Vector2>>> contactList;
this->playerBody->GetContactList(contactList);
for (auto& contact : contactList) {
    Node *node;
    Vector2 normal;
    Vector<Vector2> positions;
    std::tie(node, normal, positions) = contact;
    if (positions.Size() > 0) {
        std::cout << this->steps << std::endl;
        std::cout << "name: " << node->GetName().CString() << std::endl;
        std::cout << "normal: " << normal.ToString().CString() << std::endl;
        for (auto& position : positions) {
            std::cout << "position: " << position.ToString().CString() << std::endl;
        }
        std::cout << std::endl;
    }
}
```

If you guys want this, I could clean up the patch further to return a results class instead of tuple, which is more forward portable.

### DIFF
--- a/Source/Urho3D/Urho2D/RigidBody2D.cpp
+++ b/Source/Urho3D/Urho2D/RigidBody2D.cpp
@@ -610,4 +610,25 @@ void RigidBody2D::OnMarkedDirty(Node* node)
         body_->SetTransform(newPosition, newAngle);
 }
 
+void RigidBody2D::GetContactList(Vector<std::tuple<Node*, Vector2, Vector<Vector2>>> &contactList) const
+{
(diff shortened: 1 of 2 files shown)


cirosantilli-lovechina:
@weitjong hi, can you explain why this received wontfix, so I can make better contributions in the future? Thanks.

PR #2218, opened 2017-12-25
Fixed can't save ValueAnimation by SaveJSON() and can't load ValueAnimation by LoadJSON()

spwork:
Fixed can't save ValueAnimation by SaveJSON() and can't load ValueAnimation by LoadJSON()

there is 3 small bug in that 2 functions.


weitjong:
For some reasons I/we have overlooked this PR. The double increment is surely a bug at a glance.

weitjong:
The double increment bug is already fixed in another PR in Jan. Nevertheless, the rest of the fixes that your proposed seem still valid. So, merged now. Thanks.

PR #2219, opened 2017-12-25
Update the STB Image Write library.

iSLC:
This allows the JO_JPEG library to be removed from the dependencies list since it was "_recently_" integrated into the STB Image Write library (github.com/nothings/stb/pull/432).


iSLC:
Sorry, the libcpuid was not meant to be part of this pull request (yet) for consistency. Any idea how to separate it? I just pushed my changes to github. Didn't explicitly open a pull request for that.

weitjong:
In your master branch

`git reset HEAD~ && git push-f`

Followed by below to commit your last change to a new separate branch

`git stash && git checkout -b ryzen-fix origin/master && git stash pop && git commit -am “Update the libcpuid...”`

Typing this while commuting, so please double-check the command first. For PRs it is easier to manage them if you have separate branches for each.

weitjong:
I have question for you. Did you just drop-in the newer version of the code in without first checking whether the older version have local patch made by us? Or you intentionally replace them because they are not relevant anymore.

iSLC:
If you're talking about this modification:

```cpp
   // Urho3D: proper UTF8 handling for Windows, requires Urho3D WString class
#ifndef _WIN32
   FILE *f = fopen(filename, "wb");
#else
    Urho3D::WString wstr(filename);
    FILE *f = _wfopen(wstr.CString(), L"wb");
#endif
```

Yes, that's still included. Whenever I update an third party library in Urho I usually do a search for `Urho` to see what comes out.

weitjong:
In this case I believe we need to merge the bits from upstream and from our local patch together. The merge is done. Thanks.

PR #2220, opened 2017-12-26
Update the libcpuid library.

iSLC:
On Ryzen CPUs, the engine (according to its own rules) mistakes virtual cores for physical.


weitjong:
I took the chance to migrate the subdir into git subtree. The new subtree is based on the upstream repository directly. So, I didn't merge the code from this PR. Thanks for your PR anyway.

PR #2226, opened 2017-12-31
kNet: Prevent SIGPIPE signal when other end closes connection.

henu:
I'm using kNet in my multiplayer game, but it often crashes the whole server with SIGPIPE signal. I don't see any point in receiving the signal, so this PR just disables it. The error handling that already exists in the code nicely catches and handles the error.

If you have some information what would be the better way to handle this, please let me know. To my knowledge, this happens in a thread created by the kNet lib, so I don't know if it is possible to somehow catch the signal in the code of my game and ignore it.

I have only tested this on Linux, so I have no idea if it works on Windows or Mac. [Some guy claimed  Windows does not have SIGPIPE](https://stackoverflow.com/questions/23911929/writing-to-a-close-socket-didnt-raise-a-sigpipe-as-expected#comment36851122_23911929), and [according to documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/ms740149\(v=vs.85\).aspx) it has no MSG_NOSIGNAL either, so I just ignored this on Windows. 

I also added Lasse Öörni's name to comment, as he has also modified Socket.cpp for Urho3D, but there was no mention about him. Last time when I created a pull request modifying third party lib somebody told me those mentions should always be there :)

### DIFF
--- a/Source/ThirdParty/kNet/src/Socket.cpp
+++ b/Source/ThirdParty/kNet/src/Socket.cpp
@@ -15,6 +15,8 @@
 /** @file Socket.cpp
 	@brief */
 
+// Modified by Lasse Oorni and Henrik Heino for Urho3D
+
 #include <string>
 #include <cassert>
 #include <utility>
@@ -135,6 +137,14 @@ writeOpen(true), readOpen(true)
 	SetSendBufferSize(512 * 1024);
 	SetReceiveBufferSize(512 * 1024);
 	udpPeerAddress = remoteEndPoint.ToSockAddrIn();
+
+	// Urho3D: Prevent SIGPIPE signal when other end closes connection
+#ifdef __APPLE__
+	if (connection != INVALID_SOCKET) {
+		int set = 1;
+		setsockopt(connection, SOL_SOCKET, SO_NOSIGPIPE, (void *)&set, sizeof(int));
+	}
+#endif
 }
 
 Socket::Socket(const Socket &rhs)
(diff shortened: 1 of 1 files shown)


weitjong:
Just done some googling now. It seems that `MSG_NOSIGNAL` flag may be specific to Linux and some UNIX but not BSD/Darwin (read MacOS), if I understood it correctly. If so then we need another code branch for macOS using `SO_NOSIGPIPE` option as follows.

```
int set = 1;
setsockopt(sd, SOL_SOCKET, SO_NOSIGPIPE, (void *)&set, sizeof(int));
```

Which is not so nice. Have you tried `signal(SIGPIPE, SIG_IGN)`? Say in the `Network::StartServer()`. That should work for all the *Nix including MacOS, so it just need to be guarded against `_WIN32`.

henu:
Finally a reply! :D

In my opinion your solution looks nice. I guess it only needs to be called when socket is created?

But I can try the ignoring if you like that more. I already tried it in my own code, but not in the library code. I can try that at some point during this weekend.

weitjong:
Thanks for your patience :smile:

I reckon the Network::StartServer() is only being called once per server app, so it would be good place to put the code for ignoring the SIGPIPE.

weitjong:
The `SO_NOSIGPIPE` option only works for some BSD flavors of *Nix and not works for Linux. The option need to be set when start listening to the socket, as you already guess. So using this, still require the bit you have already proposed for Linux. Which is why I don’t like it. And prefer the latter approach I mentioned. I hope that clarifies.

weitjong:
@henu Any update?

henu:
@weitjong Ah sorry, I forgot this whole thing and I have been doing other projects :(

I tested it now, but I was unable to prevent the crash with signal ignoring. I tried to put it in the `Network::StartServer()` and since it didn't work, I also tried `NetworkWorkerThread::MainLoop()` which is the actual thread where the crash happens. But no luck.

I suggest I make those BSD/Darwin modifications you mentioned, but unfortunately I cannot test them.

weitjong:
Thanks for checking it and making a new commit.

PR #2228, opened 2018-01-06
Revise smart pointers

eugeneko:
I always lacked an ability to use `SharedPtr` with non-`RefCounted` objects.
Especially when I needed to store some objects by interface and so on.

I added `ExternalPtr` that is similar to `std::shared_ptr`.
`ExternalPtr` could hold object of any type together with `SharedPtr`s with external `RefCount` struct.
`ExternalPtr` is constructible from `SharedPtr` and convertible to `WeakPtr`.
`ExternalPtr` is also constructble from raw pointer and reference counter, so it's possible to make one from raw `RefCounted` pointer.
The brand new `ExternalPtr` could be created via `MakeExternal`. ~~Don't do it for `RefCounded` things.~~ Intusive instance of `RefCount` will be used if possible.

`ExternalPtr` is more like holder, so it may lack some sugar casts or methods. I prefer to go this way to avoid dozens of overloaded methods that accept different pointer types.

_I know_ that in the perfect world `ExternalPtr` should be `SharedPtr` and `SharedPtr` should be `IntrusivePtr`, but in this world this is too hard to keep things compatible with this approach.

PS. I also removed operator [] from SharedPtr and WeakPtr that is completely useless.

Example:
```
struct InterfaceA
{
    virtual ~InterfaceA() {}
    String a_ = "a";
};

struct InterfaceB
{
    virtual ~InterfaceB() {}
    String b_ = "b";
};

struct RefCountedObject : public RefCounted, public InterfaceA, public InterfaceB
{
    String object_ = "object";
};

    {
        auto obj = MakeShared<RefCountedObject>(); // 1 strong, 1 weak
        ExternalPtr<RefCountedObject> objPtr = obj; // 2 strong, 2 weak
        ExternalPtr<InterfaceA> aPtr = obj; // 3 strong, 3 weak
        ExternalPtr<InterfaceB> bPtr = obj; // 4 strong, 4 weak
        WeakPtr<RefCountedObject> weak(obj); // 4 strong, 5 weak
        WeakPtr<RefCountedObject> ptrWeak(objPtr); // 4 strong, 6 weak
        WeakPtr<InterfaceA> aPtrWeak(aPtr); // 4 strong, 7 weak
        WeakPtr<InterfaceB> bPtrWeak(bPtr); // 4 strong, 8 weak
        obj.Reset(); // 3 strong, 8 weak
        objPtr.Reset(); // 2 strong, 7 weak
        aPtr.Reset(); // 1 strong, 6 weak
        bPtr.Reset(); // expired, 4 weak
        weak.Reset(); // expired, 3 weak
        ptrWeak.Reset(); // expired, 2 weak
        aPtrWeak.Reset(); // expired, 1 weak
        bPtrWeak.Reset(); // expired, cleaned
    }
    {
        auto obj = MakeExternal<int>(42); // 1 strong, 1 weak
        WeakPtr<int> objWeak(obj); // 1 strong, 2 weak
        obj.Reset(); // expired, 1 weak
        objWeak.Reset(); // expired, cleaned
    }
    {
        auto obj = MakeExternal<RefCountedObject>(); // 1 strong, 2 weak
        WeakPtr<RefCountedObject> objWeak(obj); // 1 strong, 3 weak
        SharedPtr<RefCountedObject> objShared(obj.Get()); // 2 strong, 3 weak
        obj.Reset(); // 1 strong, 2 weak
        objShared.Reset(); // expired, 1 weak
        objWeak.Reset(); // expired, cleaned
    }

```

### DIFF
--- a/Source/Urho3D/Container/Ptr.h
+++ b/Source/Urho3D/Container/Ptr.h
@@ -125,13 +125,6 @@ public:
         return *ptr_;
     }
 
-    /// Subscript the object if applicable.
-    T& operator [](const int index)
-    {
-        assert(ptr_);
-        return ptr_[index];
-    }
-
     /// Test for less than with another shared pointer.
     template <class U> bool operator <(const SharedPtr<U>& rhs) const { return ptr_ < rhs.ptr_; }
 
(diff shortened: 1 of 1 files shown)


iSLC:
Might as well call it `SmartPtr` since that's what it is.

eugeneko:
@iSLC Yeah, but such name is too generic.

Since I'm unsure in these changes and names, I'm not pushing this PR.
Just want to keep the code somewhere.

dev-plvlml:
The most common name for shared pointer used only with RefCounted objects is IntrusivePtr.
Maybe the name for the one in this PR (the opposite) should be NonIntrusivePtr?

PR #2229, opened 2018-01-07
Remove misplace explicits from smart pointers.

eugeneko:
@weitjong Want to revert some autorefactored things...

### DIFF
--- a/Source/Urho3D/Container/Ptr.h
+++ b/Source/Urho3D/Container/Ptr.h
@@ -56,7 +56,7 @@ public:
     }
 
     /// Copy-construct from another shared pointer allowing implicit upcasting.
-    template <class U> explicit SharedPtr(const SharedPtr<U>& rhs) noexcept :
+    template <class U> SharedPtr(const SharedPtr<U>& rhs) noexcept :    // NOLINT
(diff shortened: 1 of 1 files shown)


weitjong:
I only put "NOLINT" for types that I consider can be used interchangeably, with the exception case like construct from null pointer and helper operator. So, I am agree that some of the auto-fixed `explicit` specifier should be reverted back (those were probably done during the rush `git rebase` before the year ended. However, I am not sure about the potentially upcasting copy-ctor to be excluded from the linter as well. Do you think it is a good idea to treat those types as interchangeable?

eugeneko:
Implicit upcasting from derived to base is natural for raw pointers. STL pointers do the same.
So I think that Urho's smart pointers could expose the same functionality.

weitjong:
Just to be sure we are talking about the same thing. The specifier doesn't prevent any of this. It just forces developer to be explicit about it when it is intended.

If you are sure it is better this way then I have no objection.

eugeneko:
I see. Yep, I think that implicit conversions are better here.
More handy to e.g. return pointers from functions or push 'em to containers and so on.. And no pitfalls except minor performance cost of copying.

PR #2235, opened 2018-01-13
Added SetSpriteTextureEdgeOffset method for TmxFile2D.

kostik1337:
A convenience method to remove texture bleeding from tilemap


eugeneko:
Bindings.
And, please, use range-based `for` loop. It would be two times shorter.
Thanks for PR!

weitjong:
Thanks for your PR.

Issue #2246, opened 2018-01-30
Triage and fix coverity issues


ricab:
To track _Coverity_ issue triaging and fixing.

ricab:
This issue follows some discussion in an [unrelated PR](https://github.com/urho3d/Urho3D/pull/2214) which went too off-topic. I am planning to triage/fix a couple of Coverity issues once in a while, as I find time (no strings attached). Others may want to do the same or just discuss related matters here.

weitjong:
Something is still fishy on the Coverity-Scan service. One of our CI has submitted the Urho3D for analysis a few days ago. At first it was put into a queue and finally when it is our turn to run, it just stuck there in the running days. At least for 3 days already, if I counted correctly. 

https://scan.coverity.com/projects/urho3d-urho3d

ricab:
OK, thanks for the info. Where can you see that? I can only see that the last scan was on Feb 19. I suppose their system is overloaded after being down for so long. Perhaps it will get sorted out in a few more days.

ricab:
I can't see it, but I believe you.

ricab:
Perhaps only admins can see running instances? I can see the "passed" badge only: <a href="https://scan.coverity.com/projects/urho3d-urho3d">
  <img alt="Coverity Scan Build Status"
       src="https://scan.coverity.com/projects/4954/badge.svg"/>
</a>

Maybe this merits a question to scan-admin@coverity.com. Perhaps it could be manually restarted or something.

weitjong:
I just did and now we wait for their reply.

weitjong:
The last run result is in. The defect density dropped a tiny little.

ricab:
It is still a positive balance :slightly_smiling_face: I suspect most variation comes from Assimp

ricab:
For curiosity, how often does Coverity run on Urho?

weitjong:
On our side we have set it up for CI to scan on every single commit to master branch (only). However, on Coverity-Scan side they have a quota system that limits the scan (determined based on the project size = lines of code), which effectively reduces our scan to maximum once a day, no matter how many commits we make within a day.

PR #2247, opened 2018-01-30
fix RibbonTrail coverity issues; CID: 127543, 216057, 216060, 216066

ricab:
Addresses #2246

### DIFF
--- a/Source/Urho3D/Graphics/RibbonTrail.cpp
+++ b/Source/Urho3D/Graphics/RibbonTrail.cpp
@@ -53,6 +53,22 @@ inline bool CompareTails(TrailPoint* lhs, TrailPoint* rhs)
     return lhs->sortDistance_ > rhs->sortDistance_;
 }
 
+TrailPoint::TrailPoint() :
+    TrailPoint{Vector3::ZERO, Vector3::ZERO}
+{
+}
+
+TrailPoint::TrailPoint(const Vector3& position, const Vector3& forward) :
+    position_{position},
+    forward_{forward},
+    parentPos_{Vector3::ZERO},
+    elapsedLength_{0.0f},
+    next_{nullptr},
+    lifetime_{0.0f},
+    sortDistance_{0.0f}
+{
+}
+
 RibbonTrail::RibbonTrail(Context* context) :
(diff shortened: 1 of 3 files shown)


Source/Urho3D/Graphics/RibbonTrail.cpp
```diff
@@ -53,13 +53,31 @@ inline bool CompareTails(TrailPoint* lhs, TrailPoint* rhs)
...
+TrailPoint::TrailPoint() :
+    TrailPoint{Vector3::ZERO, Vector3::ZERO}
+{
+}
+
+TrailPoint::TrailPoint(const Vector3& position, const Vector3& forward) :
+    position_{position},
+    forward_{forward},
+    parentPos_{Vector3::ZERO},
+    elapsedLength_{0.0f},
```
eugeneko: I suggest to move scalar initializers into header, mark `TrailPoint()` as `=default`, and keep only `TrailPoint(const Vector3& position, const Vector3& forward) : position_(...), forward_(...) {}`

ricab: Ok sure, will do that.

ricab:
@eugeneko is this what you meant?

eugeneko:
Exactly. Thanks.

weitjong:
This PR can be merged as is then. However, it would be great if you can add a period to end the doxygen comment and also at the unrelated LUA script comment that you have touched. Only inline comments do not end with period. That’s current convention. And while you at it, may be also could change the member init in the header using `{}` instead of `= 0` or `= nullptr`.

ricab:
Done. I will try to remember the periods in non-inline comments. BTW, the reason why I changed the lua file was that, while searching for TrailPoint occurrences that could be affected in script, I noticed the wrong comment, and though another PR would be a bit of an overkill for that. But now I noticed none of the other comments in it have periods at the end, so another PR may actually be needed :laughing:

weitjong:
Thanks for that. Yeah, I guess only the doxygen comments are more glaring because they will end up in our online documentation. Also you are right that sometime we also make such mistakes ourselves or may not catch them while reviewing PR every time. But that’s not an excuse.

PR #2249, opened 2018-02-01
Allow arbitrary coordinate systems for RaycastVehicle.

nick70:
This is useful for vehicles that need to be reoriented in order to drive correctly.

### DIFF
--- a/Source/Urho3D/Physics/RaycastVehicle.cpp
+++ b/Source/Urho3D/Physics/RaycastVehicle.cpp
@@ -34,6 +34,8 @@
 namespace Urho3D
 {
 
+const IntVector3 RaycastVehicle::DEFAULT_COORDINATE_SYSTEM(0, 1, 2);
+
 struct RaycastVehicleData
 {
     RaycastVehicleData()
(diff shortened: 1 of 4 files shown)


Source/Urho3D/Physics/RaycastVehicle.h
```diff
@@ -168,6 +168,8 @@ class URHO3D_API RaycastVehicle : public LogicComponent
     WeakPtr<RigidBody> hullBody_;
     /// Opaque Bullet data hidden from public
     RaycastVehicleData* vehicleData_;
+    /// Coordinate system
```
weitjong: Terminate the doxygen comment with a period. Also, may be declare an enumeration or constants for the coordinate system. They should be more readable than 0, 1, and 2.

weitjong:
Thanks for your PR.

nick70:
I think that should work. Changes in the attribute won't take effect until SetCoordinateSystem gets called again. It gets called eventually in ApplyAttributes.

Source/Urho3D/Physics/RaycastVehicle.cpp
```diff
@@ -670,6 +677,17 @@ float RaycastVehicle::GetInAirRPM() const
     return inAirRPM_;
 }
 
+void RaycastVehicle::SetCoordinateSystem(const IntVector3& coordinateSystem)
+{
+    coordinateSystem_ = coordinateSystem;
+    vehicleData_->SetCoordinateSystem(coordinateSystem_);
+}
+
+IntVector3 RaycastVehicle::GetCoordinateSystem()
```
weitjong: The getter should be declared as `const`.

weitjong: And move to header.

Source/Urho3D/Physics/RaycastVehicle.cpp
```diff
@@ -34,6 +34,8 @@
 namespace Urho3D
 {
 
+const IntVector3 RaycastVehicle::DEFAULT_COORDINATE_SYSTEM(0, 1, 2);
```
weitjong: Since 3P3 = 6, do you think it makes sense to declare all the possible permutations here?

weitjong:
This is merged now. Thanks again.

PR #2255, opened 2018-02-06
Coverity shader parameters CID127650

ricab:
Addresses  #2246.

I was unfortunately unable to test the DirectX cases, as I am on Linux here.  It would be great if someone could run the PBR sample with D3D9 and D3D11, with these changes, just to confirm everything is ok.


ricab:
Note: there is no CID for the D3D cases as coverity is not setup for that, IIUC, but the 2 latter commits address analogous issues to the one detected for the OpenGL case.

weitjong:
Thanks for the PR.

Issue #2150, opened 2017-10-13
App not being updated on window resize


uniquejack:
Whenever I try to resize the window, the application stops to update until I release the mouse button. 
Here's an image of what happens:
![immagine](https://user-images.githubusercontent.com/6839120/31567316-379566dc-b06f-11e7-9cb6-68af4f7ac9b1.jpg)

It looks ugly because it doesn't even clear the screen. I know in some cases the engine should stop rendering but I don't think this should be the case since the device don't get lost and the window still has the mouse focus. What I would like to do is to render while I'm resizing the window so I can see how my UI responses to a resize in realtime without the need to wait the mouse button release. Is it possible?

weitjong:
This looks like a platform-specific issue. Just tested on Linux and could not reproduce the issue there (using one of the sample app and editor). If you build your project using DEBUG build configuration, you should be able to see SDL fires the windows resized event while it is still in progress and not after the mouse has been released. The Urho's input subsystem handles these events by delegating to the Graphics subsystem which performs the actual work and then adds a debug log entry among others. So, you could at least quickly verify whether the window resized events come while you resizing or not on Windows platform. It it doesn't then it is beyond us (it is SDL/Windows problem). However, if it does come multiple times then you can try to switch between D3D9 and D3D11 to see whether both Graphics API produce the same issue or not.

weitjong:
Close. Feel free to reopen if you have more to add.

PR #2256, opened 2018-02-09
Windows repaint while resizing window.

JSandusky:
Pretty bare-bones, does what it says.

Tested OpenGL, D3D9, and D3D11 via the editor. OpenGL is definitely the worst (occasional white void spaces), both DX's repaint about the same and the issues are basically application specific (ie. the editor's lower bar for the window-position setting like to go nuts during layout) - possibly an issue with UI but I didn't look to see how it was positioned as everything else was fine.

The `SDL_SetWindowData(window, "URHO3D_CONTEXT", GetContext());` should probably really go somewhere else as that should probably happen regardless.

### DIFF
--- a/Source/Urho3D/Input/Input.cpp
+++ b/Source/Urho3D/Input/Input.cpp
@@ -38,6 +38,12 @@
 #include "../UI/Text.h"
 #include "../UI/UI.h"
 
+#ifdef _WIN32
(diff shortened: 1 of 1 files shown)


weitjong:
Thanks for your PR. And, welcome back to participate actively.

weitjong:
Probably fix #2150.

Issue #2165, opened 2017-11-07
Direct3D11 backend: Texture2D.SetSize fails when using TEXTURE_DYNAMIC


andrekoehler:
This bug was already reported in issue #1153 and was fixed, but it seems like commit 46b9a90fb1df053860adc9ce219ac05b2b295310 removed the fix later on.

The following code fails (added to the end of the HelloWorld::Start() method in the 01_HelloWorld sample):
```
Graphics &graphics = *GetSubsystem<Graphics>();
Urho3D::Texture2D tex(context_);
bool success = tex.SetSize(
	256, 256,
	graphics.GetFormat(Urho3D::CompressedFormat::CF_RGBA),
	Urho3D::TextureUsage::TEXTURE_DYNAMIC);
```

I debugged it and the error occurs in D3D11Texture2D.cpp in the bool Texture2D::Create() method at the following line:
https://github.com/urho3d/Urho3D/blob/37cba9522d6ce0d47445951469291ea078ac9440/Source/Urho3D/Graphics/Direct3D11/D3D11Texture2D.cpp#L416

textureDesc has the following content:
-	textureDesc	{Width=256 Height=256 MipLevels=9 ...}	D3D11_TEXTURE2D_DESC
		Width	256	unsigned int
		Height	256	unsigned int
		MipLevels	9	unsigned int
		ArraySize	1	unsigned int
		Format	DXGI_FORMAT_R8G8B8A8_UNORM (28)	DXGI_FORMAT
	-	SampleDesc	{Count=1 Quality=0 }	DXGI_SAMPLE_DESC
			Count	1	unsigned int
			Quality	0	unsigned int
		Usage	D3D11_USAGE_DYNAMIC (2)	D3D11_USAGE
		BindFlags	8	unsigned int
		CPUAccessFlags	65536	unsigned int
		MiscFlags	0	unsigned int

hr has the following value after CreateTexture2D returns:
hr	E_INVALIDARG One or more arguments are invalid.	HRESULT

The DirectX debug runtime prints the following error explanation: D3D11 ERROR: ID3D11Device::CreateTexture2D: A D3D11_USAGE_DYNAMIC Resource must have MipLevels equal to 1. [ STATE_CREATION ERROR 102: CREATETEXTURE2D_INVALIDMIPLEVELS]

eugeneko:
Yeah.
The second part of the condition seems to be removed
[...]

PR #2257, opened 2018-02-11
Use mipmap levels of 1 when texture usage is TEXTURE_DYNAMIC

### LINKED ISSUE #2253: Texture2D::SetData on dynamic texture causes black texture
I modified the RenderToTexture example to test how I can write CPU-generated pixels to a texture.
Before:
https://github.com/urho3d/Urho3D/blob/1.7/Source/Samples/10_RenderToTexture/RenderToTexture.cpp#L183-L203

After:
```
SharedPtr<Texture2D> renderTexture(new Texture2D(context_));
//renderTexture->SetSize(1024, 768, Graphics::GetRGBFormat(), TEXTURE_RENDERTARGET);
renderTexture->SetSize(128, 128, Graphics::GetRGBFormat(), TEXTURE_DYNAMIC);
unsigned char buffer[128 * 128 * 3];
// generate white pixels 0xFFFFFF
for (int i = 0; i < (128 * 128 * 3); ++i) {
	buffer[i] = 0xFF;
}
renderTexture->SetData(0, 0, 0, 128, 128, buffer);
renderTexture->SetFilterMode(FILTER_BILINEAR);

...

//RenderSurface* surface = renderTexture->GetRenderSurface();
//SharedPtr<Viewport> rttViewport(new Viewport(context_, rttScene_, rttCameraNode_->GetComponent<Camera>()));
//surface->SetViewport(0, rttViewport);
```

The expected outcome is a plane with a 128x128 texture that contains only white pixels, but this result can only be observed with the Direct3D9 backend. When compiling the same code for the OpenGL or Direct3D11 backends, the plane is black instead of white.

Direct3D9:
correct:
d3d9

Direct3D11:
wrong:
urho3d_d3d11

OpenGL:
wrong:
urho3d_opengl

I verified that this issue is not caused by device loss because Texture2D::IsDataLost() returns false for each frame in the sample's update loop.

JSandusky:
Closes #2253 and #2165

As #2165 mentions, these checks used to be there but were removed when textures were last reorganized some.

### DIFF
--- a/Source/Urho3D/Graphics/Direct3D11/D3D11Texture2D.cpp
+++ b/Source/Urho3D/Graphics/Direct3D11/D3D11Texture2D.cpp
@@ -396,7 +396,7 @@ bool Texture2D::Create()
     textureDesc.Width = (UINT)width_;
     textureDesc.Height = (UINT)height_;
     // Disable mip levels from the multisample texture. Rather create them to the resolve texture
-    textureDesc.MipLevels = multiSample_ == 1 ? levels_ : 1;
+    textureDesc.MipLevels = (multiSample_ == 1 && usage_ != TEXTURE_DYNAMIC) ? levels_ : 1;
     textureDesc.ArraySize = 1;
     textureDesc.SampleDesc.Count = (UINT)multiSample_;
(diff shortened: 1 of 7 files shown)


weitjong:
Thanks for the fix. Will merge at the soonest.

weitjong:
Close #2165.

PR #2260, opened 2018-02-13
String usage optimisation

QBkGames:
Related to discussion:
https://discourse.urho3d.io/t/memory-fragmentation-concerns/3994

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,6 +64,18 @@ generated/
 /*.project
 /*.workspace
 
+# Visual Studio files
+.vs
+*.vcxproj*
+*.sln
+*.sdf
+*.suo
+/ipch/
+*.tlog
+Debug/
+Release/
+Win32/
+
 # Android project settings
 /Android/local.properties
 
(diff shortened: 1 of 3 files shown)


eugeneko:
This PR is good example of counter-OOP design.

Derived classes should extend base classes instead of putting extra limitations.
In perfect world there should be base class with read-only methods (e.g. `StringBase`, like `VectorBase`) and `String` as derived class.

`ConstString` has too many hidden restrictions, so it'd be a pitfall for not-very-experienced users.
We already have bs like PODVector that may shoot somebody's leg. I don't want to have one more such class _without good reason_ to.

QBkGames:
I agree it's not a particularly good solution, but more like a hack, it's just the simplest solution with minimal impact on the existing code.

I still vote for using "const char*" instead of "const String&" in function definitions.
I'm curious of what other solution get proposed.

eugeneko:
> I still vote for using "const char*" instead of "const String&" in function definitions.

I vote not to change code unless it gives some practical benefit.
I'm sure you heard these words before, but want to write'em there too.

PR #2270, opened 2018-03-02
Make LuaFunction::PushLuaTable able to lookup nested tables.

eugeneko:
Small fix requested via gitter


weitjong:
@eugeneko Does the latest commit fix the shadowing issue? If yes then I won't spend time to look into it.

eugeneko:
Yeah, true.
I didn’t want to mess around with branches, so these two fixes are in the
single PR.

вс, 4 марта 2018 г. в 4:53, Yao Wei Tjong <notifications@github.com>:

PR #2272, opened 2018-03-04
HighDpi Unification

TrevorCash:
Unifies UI and Input Subsystem with common scale factor obtained from Graphics

### DIFF
--- a/Source/Urho3D/Graphics/Direct3D11/D3D11Graphics.cpp
+++ b/Source/Urho3D/Graphics/Direct3D11/D3D11Graphics.cpp
@@ -200,6 +200,7 @@ Graphics::Graphics(Context* context) :
     Object(context),
     impl_(new GraphicsImpl()),
     position_(SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED),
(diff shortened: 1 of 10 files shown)


weitjong:
Thanks for your PR. I don't have any PC/Mac with Retina display to test or review this, however, personally I don't think the coupling of those scales is actually necessary. Currently although it is more cumbersome to set the scale manually with the API call from the respective subsystems but the devs have absolute freedom to set the scale one way or another that suit their need.

In SDL they differentiate between size in "screen coordinate" and size in pixel. In Urho currently we have not adopted the same terminology and always returns things in pixel. We only have minimal support for High DPI at the moment due to reason I have mentioned above. But it is safe to say we would likely follow the terminology and approach from SDL closely.

TrevorCash:
I dont see an Input()->SetScale method of any sorts

weitjong:
Could you rebase your branch on the latest Urho3D master branch and then force-push it to your remote? That will automatically trigger another round of CI build jobs with the lint phase out of your way so you could, hoperfully, see the build issue clearer. And thanks for your effort. I am not the best person to review your PR though as I don't have the hardware.

PR #2273, opened 2018-03-04
Some move constructors and assignment operators

dev-plvlml:


dev-plvlml:
Hmm, some old commits wormed into this. Maybe should close and I will somehow reopen clean.

iSLC:
Not an expert on this but, how do move semantics work over DLL boundaries? I was under the impression that sharing heap memory and/or non-pod heap allocated objects over DLL boundaries was not "_a great_" idea. Perhaps some macro guards to include this feature only on static builds if that's the case?

dev-plvlml:
Sounds like a good point.

Issue #2245, opened 2018-01-27
For CI - move the lint stage as the first stage in the primary CI


weitjong:
Also enable Clang-Format in the lint stage. As the lint stage is in the primary CI, all future PRs will be subjected to the Clang-Tidy and Clang-Format checks. The CI mirror branches will only be created when the lint stage passed.

weitjong:
The lint stage is already moved to the top now. And we just have our new shiny Clang-Format version 6 available for our CI use. Will experiment with it in the next few days.

weitjong:
It appears there is a drawback of having the lint stage as the first build stage that I had not anticipated before. When the submitted PR has a general build issue and not the code convention issue then the linter will not be able to show the error due to its standard error stream has been redirected for further post-processing in the current setup. Since a build stage failure stops any further stages, it is impossible for the submitter to find out the actual build error message from the build log. Probably will revert to move the lint stage to where it was.

PR #2274, opened 2018-03-04
Added move constructors and move assignment operators to some classes

dev-plvlml:
Particularly to String and SharedPtr.

### DIFF
--- a/Source/Urho3D/Container/Ptr.h
+++ b/Source/Urho3D/Container/Ptr.h
@@ -62,6 +62,20 @@ public:
         AddRef();
     }
 
+    /// Move-construct from another shared pointer.
+    SharedPtr(SharedPtr<T>&& rhs) noexcept :
+        ptr_(rhs.ptr_)
+    {
+        rhs.ptr_ = nullptr;
+    }
+
+    /// Move-construct from another shared pointer allowing implicit upcasting.
+    template <class U> SharedPtr(SharedPtr<U>&& rhs) noexcept :    // NOLINT(google-explicit-constructor)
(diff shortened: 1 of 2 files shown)


eugeneko:
IDK why LINT has failed, I haven't managed to find constructive errors there.

However, as long as you touched this code, I suggest you to remove non-template move ctor, move assigment, copy ctor and classic assigment.

weitjong:
It was a build failure. The linter has not kicked in yet. The failure could be due to stale ccache and/or PCH. It is a known issue that some time GCC has failed to detect that it needs to rebuild PCH or it did detect it but the cached object get into the way. In the past this occasionally happens when core headers in the PCH are touched.

I am in the middle of git rebasing for Assimp and not able to switch branch. Will move the Lint phase back to where it was instead of the first build phase as soon as I am able to do so. See https://github.com/urho3d/Urho3D/issues/2245#issuecomment-370193037.

Also it is possible that we change the compiler toolchain to Clang instead of GCC to avoid the above problem because I have observed Clang does not or less susceptible to this.

In the meantime you can try to push a commit with a commit message containing this keyword `[cache clear]` to workaround the stale cache, if indeed it is the problem.

weitjong:
@dev-plvlml Could you rebase your branch on the latest Urho3D master branch and then force-push it to your remote? That will automatically trigger another round of CI build jobs with the lint phase out of your way so you could, hoperfully, see the build issue clearer. Or alternatively, do it with a long shot at the same time, amend the commit message of your last commit to include ```[cache clear]``` keyword before force-pushing.

PR #2281, opened 2018-03-26
Scope guard

ricab:
Add scope guard and use it to fix CID 126870

### DIFF
--- a/Source/ThirdParty/scope_guard/scope_guard.hpp
+++ b/Source/ThirdParty/scope_guard/scope_guard.hpp
@@ -1,0 +1,177 @@
+/*
+ *  Created on: 13/02/2018
+ *      Author: ricab
+ *
+ * See README.md for documentation of this header's public interface.
+ */
+
+#ifndef SCOPE_GUARD_HPP_
+#define SCOPE_GUARD_HPP_
+
+#include <type_traits>
+#include <utility>
+
+#if __cplusplus >= 201703L && defined(SG_REQUIRE_NOEXCEPT_IN_CPP17)
+#define SG_REQUIRE_NOEXCEPT
+#endif
+
+namespace sg
+{
+  namespace detail
+  {
+    /* --- Some custom type traits --- */
+
+    // Type trait determining whether a type is callable with no arguments
+    template<typename T, typename = void>
+    struct is_noarg_callable_t
+      : public std::false_type
+    {}; // in general, false
+
+    template<typename T>
+    struct is_noarg_callable_t<T, decltype(std::declval<T&&>()())>
+      : public std::true_type
+    {}; // only true when call expression valid
+
+    // Type trait determining whether a no-argument callable returns void
+    template<typename T>
+    struct returns_void_t
+      : public std::is_same<void, decltype(std::declval<T&&>()())>
+    {};
+
+    /* Type trait determining whether a no-arg callable is nothrow invocable if
+    required. This is where SG_REQUIRE_NOEXCEPT logic is encapsulated. */
+    template<typename T>
+    struct is_nothrow_invocable_if_required_t
+      : public
+#ifdef SG_REQUIRE_NOEXCEPT
+          std::is_nothrow_invocable<T> /* Note: _r variants not enough to
+                                          confirm void return: any return can be
+                                          discarded so all returns are
+                                          compatible with void */
+#else
+          std::true_type
+#endif
+    {};
+
+    // logic AND of two or more type traits
+    template<typename A, typename B, typename... C>
+    struct and_t : public and_t<A, and_t<B, C...>>
+    {}; // for more than two arguments
+
+    template<typename A, typename B>
+    struct and_t<A, B> : public std::conditional<A::value, B, A>::type
+    {}; // for two arguments
+
+    // Type trait determining whether a type is a proper scope_guard callback.
+    template<typename T>
+    struct is_proper_sg_callback_t
+      : public and_t<is_noarg_callable_t<T>,
+                     returns_void_t<T>,
+                     is_nothrow_invocable_if_required_t<T>,
+                     std::is_nothrow_destructible<T>>
+    {};
+
+
+    /* --- The actual scope_guard template --- */
+
+    template<typename Callback,
+             typename = typename std::enable_if<
+               is_proper_sg_callback_t<Callback>::value>::type>
+    class scope_guard;
+
+  } // namespace detail
+
+  /* --- Now the single public maker function --- */
+
+  template<typename Callback>
+  detail::scope_guard<Callback> make_scope_guard(Callback&& callback)
+  noexcept(std::is_nothrow_constructible<Callback, Callback&&>::value); /* we
+                            need this declared here to make it a friend below */
+
+  namespace detail
+  {
+
+    /* --- The template specialization that defines the interface --- */
+
+    template<typename Callback>
+    class scope_guard<Callback> final
+    {
+    public:
+      typedef Callback callback_type;
+
+      scope_guard(scope_guard&& other)
+      noexcept(std::is_nothrow_constructible<Callback, Callback&&>::value);
+
+      ~scope_guard() noexcept; // highlight noexcept dtor
+
+      void dismiss() noexcept;
+
+    public:
+      scope_guard() = delete;
+      scope_guard(const scope_guard&) = delete;
+      scope_guard& operator=(const scope_guard&) = delete;
+      scope_guard& operator=(scope_guard&&) = delete;
+
+    private:
+      explicit scope_guard(Callback&& callback)
+      noexcept(std::is_nothrow_constructible<Callback, Callback&&>::value); /*
+                                                      meant for friends only */
+
+      friend scope_guard<Callback> make_scope_guard<Callback>(Callback&&)
+      noexcept(std::is_nothrow_constructible<Callback, Callback&&>::value); /*
+        only make_scope_guard can create guards from scratch (i.e. non-move) */
+
+    private:
+      Callback m_callback;
+      bool m_active;
+
+    };
+
+  } // namespace detail
+
+
+} // namespace sg
+
+////////////////////////////////////////////////////////////////////////////////
+template<typename Callback>
+sg::detail::scope_guard<Callback>::scope_guard(Callback&& callback)
+noexcept(std::is_nothrow_constructible<Callback, Callback&&>::value)
+  : m_callback{std::forward<Callback>(callback)}
+  , m_active{true}
+{}
+
+////////////////////////////////////////////////////////////////////////////////
+template<typename Callback>
+sg::detail::scope_guard<Callback>::~scope_guard() noexcept
+{
+  if(m_active)
+    m_callback();
+}
+
+////////////////////////////////////////////////////////////////////////////////
+template<typename Callback>
+sg::detail::scope_guard<Callback>::scope_guard(scope_guard&& other)
+noexcept(std::is_nothrow_constructible<Callback, Callback&&>::value)
+  : m_callback{std::forward<Callback>(other.m_callback)}
+  , m_active{std::move(other.m_active)}
+{
+  other.m_active = false;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+template<typename Callback>
+inline void sg::detail::scope_guard<Callback>::dismiss() noexcept
+{
+  m_active = false;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+template<typename Callback>
+inline auto sg::make_scope_guard(Callback&& callback)
+noexcept(std::is_nothrow_constructible<Callback, Callback&&>::value)
+-> detail::scope_guard<Callback>
+{
+  return detail::scope_guard<Callback>{std::forward<Callback>(callback)};
+}
+
+#endif /* SCOPE_GUARD_HPP_ */
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -64,7 +64,7 @@ if (URHO3D_LIB_TYPE STREQUAL SHARED AND NOT WIN32 AND NOT ANDROID AND NOT IOS AN
 endif ()
 
 # Add targets
-foreach (TARGET FreeType LZ4 PugiXml rapidjson SDL StanHull STB)
+foreach (TARGET FreeType LZ4 PugiXml rapidjson SDL StanHull STB scope_guard)
     add_subdirectory (ThirdParty/${TARGET})
 endforeach ()
 
--- a/Source/ThirdParty/scope_guard/CMakeLists.txt
+++ b/Source/ThirdParty/scope_guard/CMakeLists.txt
@@ -1,0 +1,30 @@
+#
+# Copyright (c) 2008-2018 the Urho3D project.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+# Define target name
+set (TARGET_NAME scope_guard)
+
+# Define dependency libs
+set (INCLUDE_DIRS .)
+
+# Install headers for building the Urho3D library
+install_header_files (DIRECTORY ./ DESTINATION ${DEST_INCLUDE_DIR}/ThirdParty/scope_guard FILES_MATCHING PATTERN *.hpp BUILD_TREE_ONLY)  # Note: the trailing slash is significant
--- a/Source/ThirdParty/scope_guard/LICENSE
+++ b/Source/ThirdParty/scope_guard/LICENSE
@@ -1,0 +1,24 @@
+This is free and unencumbered software released into the public domain.
+
+Anyone is free to copy, modify, publish, use, compile, sell, or
+distribute this software, either in source code form or as a compiled
+binary, for any purpose, commercial or non-commercial, and by any
+means.
+
+In jurisdictions that recognize copyright laws, the author or authors
+of this software dedicate any and all copyright interest in the
+software to the public domain. We make this dedication for the benefit
+of the public at large and to the detriment of our heirs and
+successors. We intend this dedication to be an overt act of
+relinquishment in perpetuity of all present and future rights to this
+software under copyright law.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+OTHER DEALINGS IN THE SOFTWARE.
+
+For more information, please refer to <http://unlicense.org>
--- a/Source/ThirdParty/scope_guard/README.md
+++ b/Source/ThirdParty/scope_guard/README.md
@@ -1,0 +1,674 @@
+# scope_guard
+
+A public, general, simple, and fast C++11 scope guard that
+defends against implicitly ignored returns and optionally enforces `noexcept`
+at compile time (in C++17), all in a SFINAE-friendly way.
+
+##### TLDR
+
+Get it [here](https://github.com/ricab/scope_guard/blob/master/scope_guard.hpp).
+Usage is simple:
+
+```c++
+#include "scope_guard.hpp"
+...
+{
+  ...
+  auto guard = make_scope_guard(my_callback);
+    ...
+} // my_callback is invoked
+```
+
+## Outline
+
+- [scope_guard](#scope_guard)
+  * [Outline](#outline)
+  * [Introduction](#introduction)
+  * [Acknowledgments](#acknowledgments)
+  * [Features](#features)
+  * [Setup](#setup)
+  * [Client interface](#client-interface)
+  * [Preconditions in detail](#preconditions-in-detail)
+  * [Design choices and concepts](#design-choices-and-concepts)
+  * [Tests](#tests)
+
+## Introduction
+
+A scope guard is an object that employs RAII to execute a
+provided callback when leaving scope, be it through a _fall-through_, a return,
+or an exception. That callback can be a a function, a function pointer, a
+functor, a lambda, a bind result, a std::function, a reference to any of
+these, or any other callable, as long as it respects a few
+[preconditions](#preconditions-in-detail) &ndash; most of which are enforced
+during compilation and the rest being hopefully intuitive.
+
+All necessary code is provided in a [single header](scope_guard.hpp)
+(the remaining code is for tests).
+
+##### Notes
+
+<sub>This project follows [Semantic Versioning](https://semver.org/).</sub>
+
+<sup>_The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL
+NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this
+document are to be interpreted as described in RFC 2119._</sup>
+
+## Acknowledgments
+
+The concept of "scope guard" was [proposed](http://drdobbs.com/184403758)
+by Andrei Alexandrescu and Petru Marginean and it is well known in the
+C++ community. It was later proposed for standardization (see
+[N4189](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4189.pdf))
+but is still not part of the standard library, as of March 2018. While there
+are several implementations available, I did not find any with the
+characteristics I aimed for here.
+
+## Features
+
+### Main features
+- [x] &ge; C++11
+- [x] Reduced interface
+- [x] Thin callback wrapping: no added `std::function` or virtual table
+penalties
+- [x] General: accepts any callable that respects a few
+[preconditions](#preconditions-in-detail)
+- [x] No implicitly ignored return (see [below](#void-return))
+- [x] Option to enforce `noexcept` in C++17
+(see [below](#compilation-option-sg_require_noexcept_in_cpp17))
+- [x] _SFINAE friendliness_ (see [below](#sfinae-friendliness))
+- [x] Exposes careful exception specifications (`noexcept` with conditions when
+necessary)
+
+### Other characteristics
+- [x] No dependencies to use (besides &ge;C++11 compiler and standard library)
+- [x] No macros to make guard &ndash; just write explicit lambda or bind or
+what have you
+- [x] Extensively tested, with both
+[compile time tests](compile_time_tests.cpp) and
+[runtime-tests](catch_tests.cpp)
+- [x] Carefully documented (and adheres to
+[RFC2119](https://tools.ietf.org/html/rfc2119))
+- [x] Unlicense'd
+- [x] `snake_case` style
+
+### Issues
+
+Bug reports and suggestions are welcome. If you find that something is incorrect
+or could be improved, feel free to open an issue.
+
+## Setup
+
+Setup consists merely of making the [header file](scope_guard.hpp) available to
+the compiler. That can be achieved by any of the following options:
+
+- placing it directly in the client project's include path
+- placing it in a central include path that is known to the compiler
+- placing it in an arbitrary path and configuring the compiler to include that
+path
+
+The preprocessor definition `SG_REQUIRE_NOEXCEPT_IN_CPP17` MAY be provided
+to the compiler. The effect of this option is explained
+[below](#compilation-option-sg_require_noexcept_in_cpp17).
+
+## Client interface
+
+The public interface consists of a template function to create scope guard
+objects, a few members of those objects, and one boolean compilation option that
+can be activated with a preprocessor macro definition.
+
+Here is an outline of the client interface:
+
+- [Maker function template](#maker-function-template)
+- [Scope guard objects](#scope-guard-objects)
+  * [Invariants](#invariants)
+  * [List of available public members](#list-of-available-public-members)
+  * [List of deleted public members](#list-of-deleted-public-members)
+  * [Member type `calback_type`](#member-type-calback_type)
+  * [Member function `dismiss`](#member-function-dismiss)
+  * [Member move constructor](#member-move-constructor)
+  * [Member destructor](#member-destructor)
+- [Compilation option `SG_REQUIRE_NOEXCEPT_IN_CPP17`](#compilation-option-sg_require_noexcept_in_cpp17)
+
+### Maker function template
+
+The free function template `make_scope_guard` is the primary element of the
+public interface &ndash; most uses do not require anything else. It provides a
+way for clients to create a scope guard object with the specified callback as
+_associated callback_.
+
+Scope guards execute their _associated callback_ when they are destroyed, unless
+they were meanwhile [dismissed](#member-function-dismiss) or
+[moved](#member-move-constructor). Like other variables, scope guards with
+automatic storage are destroyed when they go out of scope, which explains their
+name.
+
+This function template is [SFINAE-friendly](#sfinae-friendliness).
+
+###### Function signature:
+
+```c++
+  template<typename Callback>
+  /* unspecified */ make_scope_guard(Callback&& callback)
+  noexcept(std::is_nothrow_constructible<Callback, Callback&&>::value);
+```
+
+###### Preconditions:
+
+The template and function arguments need to respect certain preconditions. They
+are listed here and discussed in more detail ahead.
+
+- [invocable with no arguments](#invocable-with-no-arguments)
+- [void return](#void-return)
+- [_nothrow_-invocable](#nothrow-invocable)
+- [_nothrow_-destructible if non-reference](#nothrow-destructible-if-non-reference-template-argument)
+template argument
+- [const-invocable if const reference](#const-invocable-if-const-reference)
+- [appropriate lifetime if lvalue reference](#appropriate-lifetime-if-lvalue-reference)
+- [movable or copyable if non-reference](#movable-or-copyable-if-non-reference)
+
+###### Postconditions:
+
+A scope guard object is returned with
+
+- the provided callback as _associated callback_
+- _active_ state
+
+###### Exception specification:
+
+As the signature shows, instances of this function template are `noexcept` _iff_
+`Callback` can be _nothrow_ constructed from `Callback&&` (after reference
+collapsing). Notice this is always the case if `Callback` is a reference type.
+
+###### Example:
+
+```c++
+const auto guard = make_scope_guard([]() noexcept { /* do something */ });
+```
+
+### Scope guard objects
+
+Scope guard objects have some unspecified type with:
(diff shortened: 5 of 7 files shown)


eugeneko:
You know... `std::shared_ptr` with custom deleter fixes this issue too. 1kLOC to fix something that could be fixed in 4 lines looks a bit overkill.

ricab:
@eugeneko the main goal of the PR is to add scope guard (notice it is really <200 LOC, the rest is basically the readme). The coverity issue is a use-case.

A scope guard is a complementary utility to pointers. It is not limited to memory ownership and may call whatever you want upon leaving. But I agree a `unique_ptr` would work in the image code as well (although it is a bit harder to get right). BTW, is it OK to use those with urho's code?

ricab:
I just looked for returns with manual clean-up code other than memory freeing. This is _not_ to suggest changing existing code, but suppose you were writing the File class. It often seeks to the beginning when something fails. With a scope guard, you could use a transactional approach: create a scope guard that cleans up and dismiss it only in the single case where everything succeeds. Now if someone adds a failure case and forgets to seek, no problem, it's done for them. It leads to less repetition and less error-prone code.

You could imagine all sorts of other things that go beyond memory, e.g. closing a file on exit, undoing a DB change, dumping debug info to a file, notifying an observer of failure, you name it. It is not a panacea, but still an utility worth having IMO.

weitjong:
A Java "finally".

ricab:
Sort of, yes, but one that you can dismiss if you want (finally could also have a condition).

ricab:
Concerning the failures in Visual Studio... I admit I hadn't tried to compile in windows and did not expect a compiler error that clang and gcc did not see... I am now logged into Windows for the first time in a month and I am trying to setup a C++ development environment to investigate.

eugeneko:
> the main goal of the PR is to add scope guard

Scope Guard itself is good too, but... It's still looks overkill.
https://ideone.com/d89nbU
This Scope Guard is 30 LOC and has the same functionality as the one in the PR.
There's a lot of code that does nothing but redoing the job already done by compiler.

> BTW, is it OK to use those with urho's code?

IMO it's OK to use _any_ library in cpp files, including STL.

weitjong:
I just have to time to look at the implementation detail for the proposed scope guard. I have to agree with Eugene because I am actually having difficulty to understand it at a few places. Do they serve special purpose to be written in such a complex way?

ricab:
> This Scope Guard is 30 LOC and has the same functionality

Not really. Differences:

- defense against implicitly ignored return
- [correct exception specifications](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Re-noexcept)
- option to enforce noexcept callbacks in C++17
- documented
- extensively tested, which brings me to the last point...
- ... that one has a bug... try to find it

What you deem so complex is probably caused in part by:
- not being used to the snake_case style with 2 spaces
- not being used to separation of interface and implementation
- not feeling like looking at the template stuff

But I can explain a few details in the code directly.

Source/ThirdParty/scope_guard/scope_guard.hpp
```diff
@@ -0,0 +1,177 @@
...
+ * See README.md for documentation of this header's public interface.
+ */
+
+#ifndef SCOPE_GUARD_HPP_
+#define SCOPE_GUARD_HPP_
+
+#include <type_traits>
+#include <utility>
+
+#if __cplusplus >= 201703L && defined(SG_REQUIRE_NOEXCEPT_IN_CPP17)
```
ricab: In C++17, exception specifications are part of the type. That allows rejecting, at compile time, callbacks that throw - to avoid throwing from a destructor.

Source/ThirdParty/scope_guard/scope_guard.hpp
```diff
@@ -0,0 +1,177 @@
...
+#define SG_REQUIRE_NOEXCEPT
+#endif
+
+namespace sg
```
ricab: I try to avoid implementations in this namespace and leave that to the end of the file. This way you can look at the code and see the signatures without being distracted by implementation details

Source/ThirdParty/scope_guard/scope_guard.hpp
```diff
@@ -0,0 +1,177 @@
...
+{
+  namespace detail
+  {
+    /* --- Some custom type traits --- */
+
+    // Type trait determining whether a type is callable with no arguments
+    template<typename T, typename = void>
+    struct is_noarg_callable_t
```
ricab: The compiler would enforce this, yes, but with an immediate error (no SFINAE)

Source/ThirdParty/scope_guard/scope_guard.hpp
```diff
@@ -0,0 +1,177 @@
...
+      : public std::false_type
+    {}; // in general, false
+
+    template<typename T>
+    struct is_noarg_callable_t<T, decltype(std::declval<T&&>()())>
```
ricab: Just a specialization for types which are indeed callable with no args. BTW, this is [available in C++17](http://en.cppreference.com/w/cpp/types/is_invocable), but unfortunately not in 11.

Source/ThirdParty/scope_guard/scope_guard.hpp
```diff
@@ -0,0 +1,177 @@
...
+      : public std::true_type
+    {}; // only true when call expression valid
+
+    // Type trait determining whether a no-argument callable returns void
+    template<typename T>
+    struct returns_void_t
```
ricab: To reject returns in a SFINAE friendly manner. https://github.com/ricab/scope_guard#void-return

Source/ThirdParty/scope_guard/scope_guard.hpp
```diff
@@ -0,0 +1,177 @@
...
+      : public std::is_same<void, decltype(std::declval<T&&>()())>
+    {};
+
+    /* Type trait determining whether a no-arg callable is nothrow invocable if
+    required. This is where SG_REQUIRE_NOEXCEPT logic is encapsulated. */
+    template<typename T>
+    struct is_nothrow_invocable_if_required_t
```
ricab: This encapsulates requiring noexcept callback when requested in C++17

Source/ThirdParty/scope_guard/scope_guard.hpp
```diff
@@ -0,0 +1,177 @@
...
+                                          discarded so all returns are
+                                          compatible with void */
+#else
+          std::true_type
+#endif
+    {};
+
+    // logic AND of two or more type traits
+    template<typename A, typename B, typename... C>
+    struct and_t : public and_t<A, and_t<B, C...>>
```
ricab: A conjunction of boolean types - [provided in C++17](http://en.cppreference.com/w/cpp/types/conjunction), but not before

Source/ThirdParty/scope_guard/scope_guard.hpp
```diff
@@ -0,0 +1,177 @@
...
+    {}; // for more than two arguments
+
+    template<typename A, typename B>
+    struct and_t<A, B> : public std::conditional<A::value, B, A>::type
+    {}; // for two arguments
+
+    // Type trait determining whether a type is a proper scope_guard callback.
+    template<typename T>
+    struct is_proper_sg_callback_t
+      : public and_t<is_noarg_callable_t<T>,
```
ricab: The type trait that encapsulates all compile-time preconditions for the callback

Source/ThirdParty/scope_guard/scope_guard.hpp
```diff
@@ -0,0 +1,177 @@
...
+                     std::is_nothrow_destructible<T>>
+    {};
+
+
+    /* --- The actual scope_guard template --- */
+
+    template<typename Callback,
+             typename = typename std::enable_if<
+               is_proper_sg_callback_t<Callback>::value>::type>
+    class scope_guard;
```
ricab: The scope_guard type is only enabled if the callback respects the preconditions (see http://en.cppreference.com/w/cpp/types/enable_if). We don't implement yet, because we need the declaration of make_scope_guard to make it a friend, so we only forward declare here

Source/ThirdParty/scope_guard/scope_guard.hpp
```diff
@@ -0,0 +1,177 @@
...
+
+  } // namespace detail
+
+  /* --- Now the single public maker function --- */
+
+  template<typename Callback>
+  detail::scope_guard<Callback> make_scope_guard(Callback&& callback)
```
ricab: this also needs the declaration of scope_guard to exist already, hence the forward declaration

Source/ThirdParty/scope_guard/scope_guard.hpp
```diff
@@ -0,0 +1,177 @@
...
+  noexcept(std::is_nothrow_constructible<Callback, Callback&&>::value); /* we
+                            need this declared here to make it a friend below */
+
+  namespace detail
+  {
+
+    /* --- The template specialization that defines the interface --- */
+
+    template<typename Callback>
+    class scope_guard<Callback> final
```
ricab: the actual scope_guard class, with no implementations here so that all the interface is immediately visible

Source/ThirdParty/scope_guard/scope_guard.hpp
```diff
@@ -0,0 +1,177 @@
...
+    {
+    public:
+      typedef Callback callback_type;
+
+      scope_guard(scope_guard&& other)
+      noexcept(std::is_nothrow_constructible<Callback, Callback&&>::value);
```
ricab: we make operations noexcept if at all possible

Source/ThirdParty/scope_guard/scope_guard.hpp
```diff
@@ -0,0 +1,177 @@
...
+
+      ~scope_guard() noexcept; // highlight noexcept dtor
+
+      void dismiss() noexcept;
+
+    public:
```
ricab: explicit deleted special members

Source/ThirdParty/scope_guard/scope_guard.hpp
```diff
@@ -0,0 +1,177 @@
...
+      scope_guard() = delete;
+      scope_guard(const scope_guard&) = delete;
+      scope_guard& operator=(const scope_guard&) = delete;
+      scope_guard& operator=(scope_guard&&) = delete;
+
+    private:
+      explicit scope_guard(Callback&& callback)
```
ricab: the constructor that is used by `make_scope_guard` is not meant for public usage, as we want type deduction and [universal references](https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers)

Source/ThirdParty/scope_guard/scope_guard.hpp
```diff
@@ -0,0 +1,177 @@
...
+    private:
+      Callback m_callback;
+      bool m_active;
+
+    };
+
+  } // namespace detail
+
+
+} // namespace sg
```
ricab: Done with the declarations at this point. Below only the implementations of what was implemented above. Let me know if any further clarification is needed

Source/ThirdParty/scope_guard/scope_guard.hpp
ricab: https://github.com/ricab/scope_guard#compilation-option-sg_require_noexcept_in_cpp17

Source/ThirdParty/scope_guard/scope_guard.hpp
ricab: https://github.com/ricab/scope_guard#nothrow-invocable

Source/ThirdParty/scope_guard/scope_guard.hpp
ricab: See https://github.com/ricab/scope_guard#sfinae-friendliness

ricab:
> IMO it's OK to use any library in cpp files, including STL.

OK, nice to know. I thought they should not be used where urho alternatives exist

eugeneko:
> defense against implicitly ignored return

Nah. There's 1000 and 1 easy way to shot one's leg in C++, but implicily ignored return is not the case. I hardly could imagine a bug injected this way.

> SFINAE friendly

To be honest, SFINAE is one of the ugliest things in C++. Concepts will fix it sometimes... Sometimes.

> correct exception specifications

Nowhere in Urho code it's treated carefully, since Urho itself is _almost_ `noexcept` (ofc except `bad_alloc`).

> option to enforce noexcept callbacks in C++17

Same. Good theoretically, useless practically.

> ... that one has a bug... try to find it

Are you talking about miscopypasted move-assignment?

weitjong:
Thanks for your explanation. Now I am home and in front of my desktop with wide screen monitor to review your code again.... unfortunately you are right about how see those template thingy as my head still hurt after re-reading them. I am afraid it is beyond me to comment further on it. Hopefully we get feedback from other C++ experts here. But since the header file is sourced from a 3rd-party lib (maintained by you externally), I believe it is not a show-stopper even if it has more features than what we need, provided if other feedback (if any) are all good.

I do have an issue with how the scope guard being integrated though. Normally, we would like to hide the 3rd-party libs from our library users, if possible at all. That is, we want to keep their references on our implementation only and not in Urho3D headers. Your proposed PR has exposed the underlying 3rd-party header via `ScopeGuard.h` and I think that is bad.

weitjong:
> > IMO it's OK to use any library in cpp files, including STL.

> OK, nice to know. I thought they should not be used where urho alternatives exist

See the last point. https://urho3d.github.io/documentation/HEAD/_contribution_checklist.html#ContributionThirdParty

ricab:
@weitjong Ahh, thanks for the link, I thought I had read something about std...

I am having a bit of trouble understanding what I did wrong in isolating Urho3D from third party. I tried to do that with a Urho3D header `Source/Urho3D/Core/ScopeGuard.h` so that other urho code does not have to include `scope_guard/scope_guard.hpp`. What should I change?

weitjong:
The problem is the `Source/Urho3D/Core/ScopeGuard.h` itself will be "installed" later as `$URHO3D_HOME/include/Urho3D/Core/ScopeGuard.h`. So, the library users could potentially include that header in their own project, but of course this won't work because the underlying 3rd-party lib (your lib) is not configured to be "installed". If you look at other header-only library, like rapidjson, you can see we never reference its header and the things it declared in any of our *.h header files, but only in our *.cpp class implementation file.

```
$ git grep rapidjson -- Source/Urho3D
```

Something like, include your 3rd-party header-only in the `Image.cpp` and call the `sg::make_scope_guard()` directly there.

Hope it clarifies.

ricab:
@eugeneko I think there is a misconception. You seem to regard this as something that I wrote _for Urho_ that has a bunch of features that Urho does not need. It was not so. I wrote it for the general case, aiming at something that was general, safe, tested, documented, etc. I think Urho benefits from some features, others not so much. That is usually the case when you include a library.

> [...]implicily ignored return[...] I hardly could imagine a bug injected this way.

I have trouble understanding how you could claim that. Whenever a function signals status/errors through the return...? The C API is full of those. And, incidentally, Urho is full of those (hence the pervasive pattern `if(!something){ log_error; return false;}`)

> SFINAE is one of the ugliest things in C++

Uhh, ok... it is an opinion. We are using C++ here though. Of course, Urho uses it (just think `MakeUnique` for instance). _Wishing for_ concepts does not change anything. They are still a while a way, in a world that is still transitioning to C++11.

> Urho itself is almost noexcept

Urho not needing noexcept does not mean rejecting libraries because they provide noexcept, does it?

> Same. Good theoretically, useless practically.

What exactly is the difference between the theory and the practice to justify that?

> miscopypasted move-assignment

Yes, which shows the benefit of having something that is already extensively tested rather than just copying code.

ricab:
@weitjong got it, will fix that

eugeneko:
> You seem to regard this as something that I wrote for Urho that has a bunch of features that Urho does not need. It was not so.

I understood this. The question is: what are benefits of using generic ScopeGuard library comparing to 30 LOC ScopeGuard in Urho Core?

This is like Urho containers versus [EA]STL containers.
You must know about Pareto principle. 80% of result is done by 20% of code.
Urho containters stay at these 80%. They may lack some move semantic sugar and things, but they are simple, readable, perfectly usable in 95% of cases and sometimes even better than STL counterparts.

Same for ScopeGuard. Your library is good. This code maybe worth placing somewhere in EASTL with some feature `#define`-guards. The question is: does Urho _need_ such extra dependency when 30 LOC ScopeGuard is fine too? And more readable. And inlined into Urho core.

[This](https://www.npmjs.com/package/is-positive) is definetely an antipattern.

ricab:
> This is definetely an antipattern.

I don't see how that relates to this discussion. I also don't quite follow your EASTL arguments. Anyway, if you want to compare LOC we need similar styles, so [here](https://ideone.com/WKuekr) is a modified version of my code:

- erased comments
- less newlines (same approach as your code)
- definitions inside the class
- erased namespaces

Is your pick with any of the above?

Otherwise, we have 89 LOC vs. 39 LOC. With additional 50 LOC for added features, some of which are useful for urho (e.g. inaccessible direct ctor, enforced void return). It is template code, so it is complex, but that is just the nature of the beast. Templates are complex and those are actually much simpler than corresponding std implementations, which account for many different cases.

Now, does it make client code complex? On the contrary: the client needs not worry about details because I made sure the compiler does it for him. Does this provide something useful that Urho does not yet have? Yes, and apparently you agree it does. 

And yet, when I propose it, you prefer writing a different one (with "miscopypasted" bugs) to reject mine? You know, I did not detect that bug merely by chance, I actually put a lot work into this so I knew where to look. So why do you insist on disregarding it? What is you thing really? Sorry to say but the only thing I can see is a "not invented by me" syndrome.

eugeneko:
> Sorry to say but the only thing I can see is a "not invented by me" syndrome.

Wrong. It's "small 3rdparty allergy" plus "non-functional garbage allergy".

[First] The beauty of 3rdparty grows as O(x^2) where x is functionality. I'd prefer complex 3rdparty like Detour/Bullet/etc (even IK) over manual solution almost without thinking.
The smaller 3rdparty is, the less benefit comes from its integration. At some point benefit becomes less than cost of extra dependency. is-positive is an example of such 3rd party. I don't say it's your case, but you are really close to it.

[Second] Even with style changes, your library is 50% code and 50% airbag for noobs. And here comes some constructive criticism of the library.

> inaccessible direct ctor

In fact, it is almsot inaccessible even in my code, unless one have named functor. It doesn't give any benefit anyway. **However, it will disturb people who use scope guard as class member**. IDK why one want to do it, but still.

> enforced void return

I accept that it could safe somebody from some nasty bug once a life. **However, it prevents people from intentional using named functors with return values**.
Try to pass into your guard something like
```
auto finalizeObject = [](Object& object) { ...; return true; };
auto guard = MakeScopeGuard(std::bind(finalizeObject, std::ref(object)));
```

> is_noarg_callable_t

Compiler already checks it w/o any code. And it would say "I cannot call this thing!!" instead of "meh, I cannot find scope_guard that meet some blah blah blah". That's why SFINAE sucks in this case. Compiler just say bullshit instead of clean error message. Actually, single-line `static_assert` with `is_invocable` inside `make_scope_guard` is much better than these SFINAE things. C++17, sadly. But you could emulate this thing manually with few helpers.

> is_nothrow_invocable

`nothrow` in C++ sucks. Let's just accept it. It's doesn't have any compile-time checks, it just crashes program. It's needed for some freak classes like 100%-thread-safe-containers, but... This check actually say "hey developer! you must now write noexcept after every single lambda in every single guard! It will give you literally nothing, but I don't let you compile your code without it!".

>With additional 50 LOC for added features

With additional 50 LOC of questionable improvements and synthetic restrictions (except callable check, callable check is fine, but w/o SFINAE)

ricab:
> "small 3rdparty allergy" plus "non-functional garbage allergy"

That is indeed very similar to "not invented here" syndrome... Perhaps "not invented here when we could have" is more your thing?

> The smaller 3rdparty is, the less benefit comes from its integration. At some point benefit becomes less than cost of extra dependency.

What cost?

> is-positive

... does not add anything. You need a good imagination jump to say it is almost the same.

> airbag for noobs

It is not for noobs. Anyone can forget to check a return. Compiler errors help noobs and professionals alike. Humans make mistakes. Preventing them at compilation is a good thing.

> However, it will disturb people who use scope guard as class member

Are you referring to the default ctor, which yours lacks too? I was referring to the ctor taking a callback. There is no point in having 2 ways of doing the same thing, and the ctor can easily be misused (you would be basically be giving up the universal reference). 

> it prevents people from intentional using named functors with return values

No it doesn't! It just forces you to be explicit about it, thereby asserting you made a conscious choice to both the compiler and readers:

```
auto g = make_scope_guard([&returns]() { /* return ignored on purpose */ returns(); } );
```

> you must now write noexcept after every single lambda in every single guard!

Which is a good thing, just like const! - you are helping the compiler help you. You wouldn't though, unless you chose a certain pp define in C++17, but that is not the default.

> SFINAE sucks
> nothrow in C++ sucks

You seem to have a very particular view of C++ and programming. It is your right, of course, but you should know it is contrary to what is generally advised. It is hardly fair ground to shoot down a contribution though.

The basic principle in question (which people usually agree with) is **make interfaces easy to use but hard to misuse**. That protects everyone, not just noobs. Defending the opposite does not make you look more pro.

iSLC:
Not sure how flexible this implementation actually is but since you can inherit from lambdas. You could reduce this whole code to just:

```c++

template < class L > struct ScopeGuard : public L
{
    using L::operator();
    // Constructor.
    ScopeGuard(L l) noexcept
        : L(std::move(l))
    {
    }
    // Destructor.
    ~ScopeGuard()
    {
        (*this)();
    }
};
/// Create a scope guard from the given functor.
template < class L > auto MakeScopeGuard(L && l) noexcept -> ScopeGuard< typename std::decay< L >::type >
{
    return ScopeGuard< typename std::decay< L >::type >(std::forward< L >(l));
}

int main()
{
    puts("A");
    {
        puts("B");
        auto sg = MakeScopeGuard([](){ puts("guard"); });
        puts("C");
    }
    puts("D");

    return 0;
}

```

ricab:
@iSLC inheriting from lambdas is an interesting idea that I had not thought of. Still, that approach lacks most of the features I argue make a better scope guard (in particular, the _easy to use but hard to misuse_ part).

I do not mean to claim I have the less code at all (I suppose that trophy belongs to manual function calls). There are certainly alternatives with less code. But I did write a document arguing why I think this is a better approach than the alternatives I could find. In short, the idea is having something that is carefully coded to provide generality, safety, correctness, performance, etc. You may be interested to read a bit [here](https://github.com/ricab/scope_guard#features) and [here](https://github.com/ricab/scope_guard#design-choices-and-concepts).

ricab:
Wait, so you don't want small libraries, but you want less lines of code... but you don't want to shoot down the PR...  just want me to change the library... to make it smaller... Just less third party code... Not more third party code... you know, because of the cost... lets just copy-paste buggy code from the internet, because that is costless... forget testing and design and all that useless zero-functionality bullshit crap...

ricab:
Oops, early return on that keyboard

ricab:
The license could be a problem, except that in this case it happens to be the Unlicense. Meaning it is not even needed...

> It is obvious for every C++ developer than he could change function return type from void to something else and it won't break anything.

That's just plain absurd! If you suddenly start returning status codes from code that clients already use that is obviously wrong.

> Forced restrictions suck. They suck twice in C++ because this language is designed to let people do whatever they want regardless of the danger.

Wrong again. One of the main reasons for C++ is dealing with the lack of safety of C. With how easy it is to make mistakes there. That is why there are constructors, destructors, const, noexcept, and all those complex little things...

The return value has **always** been important, and probably one of the top 5 bug causes of all time is forgetting to check it. That is probably the main reason why C++ introduced exceptions. What use is 10 years experience if you insist on ignoring [universally accepted](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i10-use-exceptions-to-signal-a-failure-to-perform-a-required-task) [expert advise](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i10-use-exceptions-to-signal-a-failure-to-perform-a-required-task) (don't read just the titles, there is more in there).

It is a shame to have to come down to this but, if that's what matters to you, I also have more than 10 years of experience in C++, you know... and if you are entitled to your opinions, I am entitled to mine and I actually think you are fundamentally wrong about what makes good programming...

But no, your 10 years C++ experience taught you sfinae sucks, noexcept sucks. Where it does not suck is allowing you to ignore returns. That is what is great. Give me an inexperienced programmer who is willing to learn any day...

ricab:
Anyway, I think this is going too far. I fundamentally disagree with you but I am tired of arguing. You have the authority so, if you don't like the contribution, go ahead and reject the PR.

weitjong:
Good morning! I am sure you can imagine how excited I was, reading those comments before my morning coffee. I think both of you made some valid arguments. It is ok to disagree but no need to get too personal. Despite that, it shows that both of you are passionate about this project. And that is what we need at the moment, passionate developers in the project. 

Since I am not able to contribute any opinion on the subject matter, I would like to share my view on the other aspect. I don’t think there is anything wrong with the `is-positive` package. On the contrary, I believe it is because all those small packages in the NPM working together that makes JS development community out grows all the rest. The concept is not new, however. On *Nix we also have many small utility commands that at first glance look stupid. See the man pages for `true` command for example. It’s a command for “doing nothing, successfully” and yet it is indispensable.

eugeneko:
> you don't want small libraries, but you want less lines of code... 

Nope. I don't like small 3rdparties because they tend to make projects messy. Regardless of LOCs.

> just want me to change the library... to make it smaller... Just less third party code

Nope. I look at your 3rdparty and notice things that I usually hate in 3rdparties when I use them.
So I want you to fix it, as "customer" of 3rdparty. Regardless of LOCs.

> If you suddenly start returning status codes from code that clients already use that is obviously wrong.

Nope. It _may_ be wrong, may be not.
If it's important for user (for me it's not), he must turn on appropriate warning.
And then _compiler_ will say if somebody pass lambda with return value into guard.
Why are you redoing the job of compiler and reinventing exiting mechanism?

> What use is 10 years experience if you insist on ignoring universally accepted expert advise 

Destructor _is_ exception to this advice.
You may want to return some status from some deinitializer functions,
But you cannot process it anyway. So you just do it and ignore all errors. That's how all destructors are written just cause there's no mechanism of error propagation.

> your 10 years C++ experience taught you sfinae sucks, noexcept sucks

You have intentinanally ignored my questions.
If you don't think they suck, tell me why not. I don't understand, really.

> How much times in your life have you gained benefit from noexcept?
> If you don't think that SFINAE sucks in your case, please give me at least one reason to use it instead of static_assert.

eugeneko:
Since this discussion is getting messy, I'd like to structurize my issues:

1) SFINAE is harmful.

- It gives zero benefit;
- It makes error output _less_ readable;
- It's verbose and makes code more ugly;

It's actually misuse of SFINAE. SFINAE is _not_ for error checking. It's for switching behaviour.
scope_guard has zero switching behaviour and a lot of error checking.

2) Hiding scope_guard is meaningless.

- It gives zero benefit, just enforces users to use `make_*` form even if they don't need it.
- It is not what is expected. While choosing between two designs, more familiar design is usually better. So, it's better for make_scope_guard to be like make_shared and make_unique since there's no obstacles.

3) Treating return value is counter-intuitive.

- It's almost meaningless. You have forbidden implicit return values in 1 place, but they are allowed in other 1000 places. So your code actually makes things 0.1% safer.
- It is not what is expected. Again, more familiar design is usually better. If scope_guard accepts some function to be called, it's better to make it alike std::function, std::thread and other STL classes. Do they warn about missing return values?..

Last two points are mostly about _predictability_ of interface and quite questionable.

eugeneko:
Well, now I worry a bit that I've scared away some contribution..

@weitjong Just curious. What's the pipeline for adding installable 3rdparty dependencies?
I know it's now very good when 3rdparty requires installation, but...
Sometimes it's just needed. Like SDL.

> So, the library users could potentially include that header in their own project, but of course this won't work because the underlying 3rd-party lib (your lib) is not configured to be "installed".

ricab:
@weitjong thank you for the diplomatic words. I hope you can agree the words below are chosen to be respectful and honest, but frank. I must say though I am at the limit of what I will put up with. I am tired and wasting wayyyy too much time here for no meaningful outcome.

@eugeneko

> Since this discussion is getting messy

Your 1st comment:

> 1kLOC to fix something that could be fixed in 4 lines

Next:

> This [buggy] Scope Guard is 30 LOC and has the same functionality as the one in the PR

Sorry, did you even bother with a quick look at the doc, where I went to the trouble of explaining how it is different from existing implementations? Or do you assume you don't need to, because only you are smart and everyone else is always dumb? Put yourself in my shoes for a minute. Why should I even continue this discussion? How do you expect any valid discussion to follow?

Suppose instead of "zero-benefit" you had said "what does this do that existing implementations don't?" Then I could point you to [this](https://github.com/ricab/scope_guard#features) or [this](https://github.com/ricab/scope_guard#design-choices-and-concepts) and, assuming you had an interest in what I had to say, you could read it and then criticize objectively: "I think feature X or design choice Y is not worth the cost W because of Z". Starting with "adds nothing", whatever the actual words, is bound to lead nowhere fast (even if you had valid points).

> I'd like to structurize my issues

Thank you! Please do try to do that systematically. I agree that would be a much wiser way to discuss, but for that you need to justify your arguments, rather than just saying "sucks", "garbage", "zero benefit" without sustaining it. If you do have some valid criticism, you have to expose it. Saying "zero benefit" does not expose it. Unsustained criticism that just dismisses everything does not do anybody any good. Not me, not you, not other people, not Urho... What makes you think anyone should even bother?

> error output less readable and verbose and code more ugly

I agree those are downsides and I would be happy to hear proposals of techniques to mitigate that. I don't agree they are reason enough to set SFINAE aside when, on the other side of the scale, you have:

> "[SFINAE] gives zero benefit".

[These tests](https://github.com/ricab/scope_guard/blob/master/catch_tests.cpp#L1603) show examples of use-cases that are enabled by SFINAE-friendliness in scope_guard. It is _just not true_ that it is used for error checking as you claim. You are showing again a misconception that arises from a predisposition to reject.

>    Hiding scope_guard is meaningless.

Again that kind of formulation. Meaningless... How expressive and constructive. What if I kick that ball back: exposing scope_guard is meaningless...

>    It gives zero benefit

False. I already mentioned a design principle saying "make interfaces easy to use and hard to misuse". That is why door knobs do not turn the wrong way. It is why electrical plugs require a certain shape. `make_scope_guard` provides easy usage with deduction and universal references. The alternative does not and is easy to misuse: 

```
scope_guard<A&&>{std::move(a)}; /* oops, nothing is moved, may end up with
dangling ref (should be scope_guard<A>, which the compiler deduces when 
using make) */
```

> It is not what is expected. 

Do you know the type of lambda? Can you stick it into a class? It is not uncommon to have unspecified types when there is no need to specify them. Same approach here.

> While choosing between two designs, more familiar design is usually better. 

"Usually" is the operative word here. All other things being equal, sure, go for consistency. But don't make bad choices in the name of consistency. Here is a [nice article](http://slashslash.info/2018/02/a-foolish-consistency/) discussing precisely that.

>    Treating return value is counter-intuitive.

Don't agree.

>    It's almost meaningless. You have forbidden implicit return values in 1 place, but they are allowed in other 1000 places. So your code actually makes things 0.1% safer.

What an argument... why do anything correctly then? You are only doing it correctly at that single correct place... A piece of code should be safe, it should not make 1000 other places safe.

> It is not what is expected

Consistency again. See reply above.

@eugeneko please understand that, after pouring many hours into this discussion, I have no interest in continuing discussing with you. You don't have to like my proposal nor agree with my opinions, but you cannot force me into unending loops of "sucks", "garbage", "meaningless". I am pretty much done with this discussion at this point. If you want to accept the PR, I will address weitjong's request (or whatever you decide concerning that). If not, as I said before, feel free to close it.

eugeneko:
Well, my code is'n really buggy. I'm sure it doesn't crashes the program or compiles when it shouldn't. Just small non-functional typo. I've spent 10 minutes on this code, so I have an excuse.

> These tests show examples of use-cases that are enabled by SFINAE-friendliness in scope_guard.

I see the point now. It is... surprising. I suppose you know about KISS. Do you have use case in which you need this SFINAE-friendliness?
I'd prefer to use library with some `static_assert`s that will just say me "hey, you function has wrong arguments, fix it!" instead of "meh shit happened here" (hi SFINAE!).
You sacrifice _real_ usability (quality of error reporting) for some weird feature (ability to suppress compilation error in the outer code).
So I wonder. Do you have any real use case that requires such feature? If so, what's it?

> why do anything correctly then? You are only doing it correctly at that single correct place...

It's okay to block obviously incorrect things.
Implicitly ignoring return values is something that happens and will happen everywhere in C++ code.
And your score_guard breaks compilation just because _you personally don't like it_.
This thing is similar to forced noexcept (artificial restriction), so it should be 1) guarded with define and 2) IMO defaulted to OFF. Less unexpected errors, behaves like everything in STL. If one want strict checks, then just turn it on.

Well, it's my bad that I started treating your code in the same way as code written in Urho and for Urho. But your code have absolutely different underlying philosophy that I didn't get till now. Your code is trying to be academically perfect (I don't say it's bad). I tend to write the code that is as simple as possible and just works (like all code in Urho, actually [except some code that doesn't work]).

I have actually one good reason for it. Compilers (especially MS) tend to suck when parsing metaprogramming things. You see, you code doesn't even compile for VS2015 because it just cannot deal with some friendship plus noexcept tags bullshit... Once you start to do metaprogramming, issues multiplies with number of compilers you have to support. So I decided to avoid any metaprogramming whenever possible.

BTW, I also heard that templates sometimes _slow down_ the program because generated code is getting too complex and optimizer can't deal with it. It's not your case at all, just wanted to mention it.

> I will address weitjong's request

How are you going to do it?

> you need to justify your arguments, rather than just saying "sucks"

It's offtopic, but sometimes things just suck, like `noexcept`. I suppose you haven't answered my question because you never gained any benefit from `noexcept`. Well, like me. Like everybody. People just write noexcept because it's modern and cool and sutters and straws troops say that it's modern and cool to write such things... And once a millenium, when somebody is writing super-fool-exception-proof-yet-another-standard-library... Then `noexcept` becomes _meaningfull_.

ricab:
@eugeneko you have invested yourself in attacking someone else's code with repeated irrational arguments that require you to systematically dismiss universally acclaimed C++ features and best practices in an interminable, stubborn, and insufferable way. There is a name for that: trolling. It is beneath me and I will not have it. This PR is closed.

weitjong:
Well, that is too bad but I respect your decision.

eugeneko:
@ricab I'm sure I had 0% of trolling in my last two posts here. Unsure about previous ones tho.

> You however have invested yourself in attacking my code with repeated irrational arguments that require you to systematically dismiss universally acclaimed C++ features and best practices in an interminable, stubborn, and insufferable way.

Well, your code just doesn't compile because of these so called "best practices".
What could be more rational than this?
I managed to fix your code only by cutting out some non-functional safety features.

> There is a name for that: trolling. It is beneath me and I will not have it.

Ignoring unconvinient questions when you don't have any argument to say why I am wrong is also feels like trolling, you know...

ricab:
I have replied to you repeatedly, wasting many hours in here, even though I wrote a document that you showed multiple times you did not bother reading. I told you, I am tired of being forced to repeat arguments that you ignore, deconstructing irrationalities... This is at 55 comments now...

- why do I need to reply to "just because you personally don't like it"? 
- why should I need to explain 10 minutes is not an excuse to say buggy code is the same?
- why do I have to bother pointing out MSVC compile errors are to be fixed, not a sufficient reason to say the code should not exist?
- why do I need to repeat that your claims of "non-functional" are false, just as with the mechanism preventing a door knob from turning the wrong way?
- why do I have to reply to arguments of KISS when you confuse keep it simple, stupid with _keep it stupid, simple_?

I don't. There are certainly valid criticisms and rational discussions to be had about my code. You just cannot manage to articulate any. If anything, you proved unable to argue logically. That should be patently clear to anyone reading by now. If it isn't, they are not going to be convinced anyway.

At some point it is just pointless. If you start arguing that "vaccines do not work" or "the earth is flat", I am not going to bother with that either... And by now I would not be surprised did.

Trying to force me to discuss again and again when I have told you I am not interested [is trolling](https://en.wikipedia.org/wiki/Internet_troll). I have safely concluded your opinion is unworthy of any more of my time so please stop bothering me. You are not exactly paying me for my time here are you? I told you [2 days ago](https://github.com/urho3d/Urho3D/pull/2281#issuecomment-376712856) to decide on the PR because I did not want to continue the discussion. You couldn't even own it! I had to close it myself. 

**Do not confuse the right to an opinion, which you have, with the right to have others pay any attention to it, which you don't have**.

eugeneko:
> I have replied to you repeatedly, wasting many hours in here, even though I wrote a document that you showed multiple times you did not bother reading.

I already told you that I misunderstood the philosophy behind the library, so a lot of my points are invaild. It seems that you just ignored 50% of my text. I can't blame you for it tho. This thread has _too_ much text and I'm not the best speaker.

> There are certainly valid criticisms and rational discussions to be had about my code. You just cannot manage to articulate any.

There are two criticisms that stay valid so far.

1) You've sacrifised error reporting in order to make your code SFIANE-friendly. It's at least questionable design decision.

2) Your code is already broken, in some meaning. I haven't managed to fix it without featurecut. That's an issue.

ricab:
> You've sacrifised error reporting in order to make your code SFIANE-friendly. It's at least questionable design decision.

I have conceded error output is worse. But it is something worth paying for all the features. That is my opinion, which you are under no obligation to share. Notice it is not because of SFINAE alone though. Remove line 68, no more SFINAE friendliness, all other features remain, error output still the same.

> Your code is already broken, in some meaning.

I did not have the time yet, but MSVC errors would obviously have to be fixed before this proposal was valid for Urho. I though that was obvious from [comment 7](https://github.com/urho3d/Urho3D/pull/2281#issuecomment-376361157). I was of course not expecting it to be integrated until then.

eugeneko:
> But it is something worth paying for all the features.

I really wonder what kind of use case could require such feature as SFINAE-friendly scope guard.
I couldn't even imagine such a case. Could you?

> MSVC errors would obviously have to be fixed

I wonder if you'll manage to find some nice fix. Shall probably check the library later.

PR #2285, opened 2018-03-29
Added URHO3D_URHO2D guard that seems omitted

BlueMagnificent:
Samples 49_Urho2DIsometricDemo and 50_Urho2DPlatformer fail in the build process when URHO3D_URHO2D is not selected in CMake configuration. This pull request resolves it.


weitjong:
Thanks for the PR.

Issue #2286, opened 2018-03-29
[OpenGL backend] Multisampled RenderSurface is black after Graphics::SetMode


andrekoehler:
To demonstrate this bug, I modified the RenderToTexture.cpp example:
In the RenderToTexture::CreateScene function modify the line https://github.com/urho3d/Urho3D/blob/master/Source/Samples/10_RenderToTexture/RenderToTexture.cpp#L184 to enable multisampling:
```
SharedPtr<Texture2D> renderTexture(new Texture2D(context_));
renderTexture->SetSize(1024, 768, Graphics::GetRGBFormat(), TEXTURE_RENDERTARGET,
	// added: multisampling
	4);
```
And at the end of the RenderToTexture::MoveCamera function (https://github.com/urho3d/Urho3D/blob/master/Source/Samples/10_RenderToTexture/RenderToTexture.cpp#L272) add the following lines:
```
// added
if (input->GetKeyDown(KEY_R))
{
	GetSubsystem<Graphics>()->SetMode(1024, 1024);
}
```
Full Source Code: 
[RenderToTexture.zip](https://github.com/urho3d/Urho3D/files/1860924/RenderToTexture.zip)

When I run the modified example, everything looks fine (anti-aliasing works):
![good](https://user-images.githubusercontent.com/22099714/38099796-ca28b988-337b-11e8-839f-83c37d7cbf22.png)

However, when I press the "R" key to change the window size, the render texture breaks and will be rendered as a full black surface from now on:
![bad](https://user-images.githubusercontent.com/22099714/38099844-e8eedb04-337b-11e8-83db-5529023de5e2.png)

This only happens with the OpenGL backend. The Direct3D11 backend works fine.
Tested with UrhoSharp 1.8.93 (OpenGL and Direct3D11 backend) and Urho native 7508b9003bd6c9c6fa2e99ec527c8df30b5f1049 (OpenGL and Direct3D11 backend).

weitjong:
I actually don't know whether this can be considered as a bug or not, however, it does again just show there are differences between DirectX and OpenGL API. If you modify your testing code to call `SetMode(024, 1024)` just before the line where the render target texture is being created then that new RTT works fine in OpenGL too. So, the question is, whether engine should automatically take care to recreate the RTT after the mode change; or whether it is responsibility of the library user to make the URHO3D_API calls in the right order.

andrekoehler:
[...]

PR #2289, opened 2018-03-31
Fix OpenGL multisample resolve not working after context loss

cadaver:
This fixes issue #2286. The FBO object names were not being set to zero, so they were never recreated.


weitjong:
Thanks Lasse. We miss you!

weitjong:
Merged.

PR #2290, opened 2018-04-01
Created new GUI skin

### LINKED ISSUE #2288: Your repo includes copyrighted Microsoft content?
bin/Data/Textures/UI.png  -- this contains various Microsoft material such as the standard "hourglass" cursor from earlier versions of Windows.

Modanung:
Resolves #2288


weitjong:
Thanks for the PR. If I may make a suggestion, could you please replace the wait cursor with more modern spinner? Something that could apply a fixed rotation to give the impression it is spinning. Anything but not beach ball :smiley:

Modanung:
@weitjong You mean replacing the hourglass? Since rotating tiny rasterized images tends to look bad this would be done best using separate frames or by adding SVG support to the engine.
In either case I would suggest saving this for another PR. Another thing I plan to do is crisping up the editor's _icons_.

weitjong:
Probably you are right, if the texture has too much detail. What I have in mind is just a simple spinners, like two or three small dots (with fixed/ increasing size) encircle a center point and rotate around that center point. Having said that, I like your SVG idea more.

eugeneko:
@weitjong But we don't have animated cursors... Hourglass looks nice when static, but spinners are probably not.

@Modanung IMO arrow tail is _too_ small and _too_ round. Also make checkbox V _a bit_ smaller so it doesn't push the border.

weitjong:
We don’t have now doesn’t mean we cannot have animated cursor, and the spinner won’t even need Urho2D subsystem. Using hourglass as wait cursor is fine too but it’s so retro. Since we are at it, I just throw the idea.

Modanung:
@eugeneko Unfortunately I disagree. I did my best to make them look as good as possible and this is the result. Have you _tried_ the UI or just looked at the new texture?

eugeneko:
@Modanung Nope, I didn't. Just first impression from texture. If you think it looks better "ingame"... Ok, I'll try.

1vanK:
Do not forget about updating SourceAssets\UITemplate.xcf

Modanung:
@eugeneko I gave in to more conventionally confined check marks after all.

@1vanK Done!

Modanung:
I unsubscribed from this discussion.

eugeneko:
In general, I have no objections against this PR.
If you are going to make any further changes or tuning, let us know.

Modanung:
Looks even better now! :sunglasses:

1vanK:
Is cursor clipped Win 10 cursor? (its coincide per pixel)

Modanung:
@1vanK Nope

PR #2294, opened 2018-04-03
Use non-zero default edge offset for Sprite2D.

eugeneko:
There was some bug noticed by @kostik1337 
There were same artifacts in newly added demos by @thesquib
I suppose that small non-zero default edge (0.00005f) should be used by default, since it most likely doesn't disturb any rendering, but prevent atlases from leaking edges (as far as I could see).

### DIFF
--- a/Source/Urho3D/Urho2D/Sprite2D.cpp
+++ b/Source/Urho3D/Urho2D/Sprite2D.cpp
@@ -39,7 +39,7 @@ Sprite2D::Sprite2D(Context* context) :
     Resource(context),
     hotSpot_(0.5f, 0.5f),
     offset_(0, 0),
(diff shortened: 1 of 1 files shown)


weitjong:
Just want to comment that I did not observe any leaking edge problem in OpenGL, at least not on my NVIDIA OpenGL driver implementation. I did notice other kind of rendering artifact on the last demo when the app is toggled to full screen.

eugeneko:
> I did notice other kind of rendering artifact on the last demo when the app is toggled to full screen.

@weitjong Could you elaborate, please?

weitjong:
It not related to your PR. The artifact is visible in the original version of the demo after pressing Alt+Enter (full screen).

PR #2297, opened 2018-04-06
Fix: reverted bad changes to RaycastSingleSegmented()

Enhex:
Wrong rounding that causes it to have 0 segments if the raycast length is smaller than segment length.
And in general the changes are only likely to cause performance regression and have no benefit, so these changes should be reverted.

### DIFF
--- a/Source/Urho3D/AngelScript/PhysicsAPI.cpp
+++ b/Source/Urho3D/AngelScript/PhysicsAPI.cpp
@@ -243,10 +243,10 @@ static PhysicsRaycastResult PhysicsWorldRaycastSingle(const Ray& ray, float maxD
     return result;
 }
 
-static PhysicsRaycastResult PhysicsWorldRaycastSingleSegmented(const Ray& ray, float maxDistance, float segmentDistance, unsigned collisionMask, PhysicsWorld* ptr)
+static PhysicsRaycastResult PhysicsWorldRaycastSingleSegmented(const Ray& ray, float maxDistance, float segmentDistance, unsigned collisionMask, float overlap_distance, PhysicsWorld* ptr)
 {
     PhysicsRaycastResult result;
-    ptr->RaycastSingleSegmented(result, ray, maxDistance, segmentDistance, collisionMask);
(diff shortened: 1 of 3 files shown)


eugeneko:
I suppose in original commit it was intended to be `CeilToInt`.
Revert is fine too tho.

Enhex:
There actually might be a bug with my old implementation related to this linter error.
I remember that sometimes segmented raycasts would miss walls in the way, and that could be related to precision error since the likelihood increased with smaller segment length.
I don't think the for loop itself is related to the problem, since it won't cause gaps between raycast segments. I'm more suspicious about the start/end points. I'll look into it.

Enhex:
this is a false-positive with the linter since `rayDistance` is calculated directly from the integer, not by adding floats.

For the record the need for segment overlapping seems to be related to Bullet's raycasting internals somehow creating gaps, since even with relatively high overlap distance like `0.01f` and with floating point values close to each other (everything between 0.f to 100.f), so there's no precision problems, there were still intersection misses in the segmented raycasting.

weitjong:
I don’t agree it is a false positive. That check exactly discourages the use of floating point number as loop control variable. 3 times 0.333... not necessary equals to 1. So, whether you increment by 0.333... in the loop or by using multiplication with loop index make no difference, IMHO, if the limit is 1 you will not sure it will loop 3 or 4 times.

But since this method is contributed by you, I have no issue if you deem your code is correct. If you must add NOLINT to silent the linter, please add the name of check in the bracket, so we know which check is being bypassed.

Enhex:
In this case it doesn't really matter how many times you loop. All that matters is covering all the distance.
With the old code I think it was possible for `segmentDistance` to be too small for decrementing `remainingDistance` so you'd have the infinite loop case.

eugeneko:
TBH I prefer initial version of code (with Ceil instead of Round) and overlap like it is implemented now.
Any reason why it's impossible?
Then, Urho doesn't use snake_case variables.
___
This is my personal opinion, but `auto` in AFTER doesn't make things better. It actually makes things worse. I spent about ~5 seconds thinking whether the `overlap` is vector or scalar. It's like tripping up for code reading.

eugeneko:
Merged manually with some amending. Hope this works.

PR #2302, opened 2018-04-12
kNet replaced with SLikeNet (RakNet fork)

ArnisLielturks:
- Existing functionality that has been provided by kNet is still available.
- Added NAT punchtrough functionality.
- Added LAN discovery
- Added client-server authentication (clients can connect only with password if server has enabled it)
- Angel script bindings added to support new functionality

Things still in my to-do list:
- Add samples for LAN discovery
- Add samples for NAT punchtrough
- Do code cleanup & style fixes
- Add github wiki documentation later to describe new functionality


ArnisLielturks:
I have tested this on Ubuntu 16.04 and Windows 10 machines, would be nice to test this on Android iOS and OSX platforms.

eugeneko:
Mother of God.. o_o
Well, I probably have to check it out and review using Desktop tools.

2018-04-12 13:56 GMT+03:00 Arnis Lielturks <notifications@github.com>:

eugeneko:
@ArnisLielturks Could you please split this commit into two (or even three commits)?
1) Copied sources of "RakNet", just as they are downloaded from repo;
2) All your manual changes;
3) Removing kNet;
Just for sake of beauty of the world.
It takes only about 5mins if you are familiar with git.

ArnisLielturks:
@eugeneko Yes, of course. I was originally doing this implemetation for other project but for this PR commited everything in one go. 

Also I noticed that MinGW builds are failing so I will have to dig into that.

eugeneko:
@ArnisLielturks Thanks, I've started reviewing changes.

Commit history is still messy. In case you don't know how to clean it:
1) Rename _local_ `master` branch
2) Create new _local_ `master` branch
3) Cherry-pick all interesting commits from old master (ab97c5f, 2c07275, a4b2bd8, 1fefda6 in your case)
4) Force push this local master to remote master

IDK how do you use git so I can't give more details. I prefer GitExtensions.

Source/Urho3D/Network/Connection.cpp
```diff
@@ -60,30 +62,25 @@ PackageUpload::PackageUpload() :
...
-    address_ = Urho3D::ToString("%d.%d.%d.%d", endPoint.ip[0], endPoint.ip[1], endPoint.ip[2], endPoint.ip[3]);
-    port_ = endPoint.port;
+    port_ = address.systemAddress.GetPort();
 }
 
 Connection::~Connection()
 {
     // Reset scene (remove possible owner references), as this connection is about to be destroyed
-    SetScene(nullptr);
+    SetScene(0);
```
eugeneko: Revert, please.

Source/Urho3D/Network/Connection.cpp
```diff
@@ -146,7 +134,7 @@ void Connection::SendRemoteEvent(Node* node, StringHash eventType, bool inOrder,
         URHO3D_LOGERROR("Sender node is not in the connection's scene, can not send remote node event");
         return;
     }
-    if (!node->IsReplicated())
+    if (node->IsReplicated())
```
eugeneko: This is wrong.

Source/Urho3D/Network/Connection.cpp
```diff
@@ -1180,7 +1185,7 @@ void Connection::ProcessNewNode(Node* node)
     {
         Component* component = components[i];
         // Check if component is not to be replicated
-        if (!component->IsReplicated())
+        if (component->IsReplicated())
```
eugeneko: This is wrong too

Source/Urho3D/Network/Connection.cpp
```diff
@@ -1348,7 +1353,7 @@ void Connection::ProcessExistingNode(Node* node, NodeReplicationState& nodeState
         {
             Component* component = components[i];
             // Check if component is not to be replicated
-            if (!component->IsReplicated())
+            if (component->IsReplicated())
```
eugeneko: And this is wrong too. Yes, I'm going to write comments for _all_ places with functional defects, regardless of similarity. Just to be sure they won't be lost.

Source/Urho3D/Network/Network.h
```diff
@@ -51,16 +52,20 @@ class URHO3D_API Network : public Object, public kNet::IMessageHandler, public k
...
-        (kNet::MessageConnection* source, kNet::packet_id_t packetId, kNet::message_id_t msgId, const char* data, size_t numBytes) override;
-    /// Compute the content ID for a message.
-    u32 ComputeContentID(kNet::message_id_t msgId, const char* data, size_t numBytes) override;
-    /// Handle a new client connection.
-    void NewConnectionEstablished(kNet::MessageConnection* connection) override;
-    /// Handle a client disconnection.
-    void ClientDisconnected(kNet::MessageConnection* connection) override;
 
+    /// Handle an inbound message.
+    virtual void HandleMessage(const SLNet::AddressOrGUID& source, int packetID, int msgID, const char* data, size_t numBytes);
```
eugeneko: Why virtual?

Source/Urho3D/Network/Network.h
```diff
@@ -69,19 +74,22 @@ class URHO3D_API Network : public Object, public kNet::IMessageHandler, public k
     bool StartServer(unsigned short port);
     /// Stop the server.
     void StopServer();
+    /// Start NAT punchtrough client to allow remote connections
+    void StartNATClient();
+    /// Get local server GUID
+    const String& GetGUID();
+    // Attempt to connect to NAT server
```
eugeneko: One more /

Source/Urho3D/Network/Network.h
```diff
@@ -137,18 +143,22 @@ class URHO3D_API Network : public Object, public kNet::IMessageHandler, public k
     /// Handle render update frame event.
     void HandleRenderUpdate(StringHash eventType, VariantMap& eventData);
     /// Handle server connection.
-    void OnServerConnected();
+    void OnServerConnected(const SLNet::AddressOrGUID& address);
     /// Handle server disconnection.
     void OnServerDisconnected();
     /// Reconfigure network simulator parameters on all existing connections.
     void ConfigureNetworkSimulator();
+    //All incoming packages are handled here
```
eugeneko: Fix comment style

Source/Urho3D/Network/Network.h
ArnisLielturks: I believe this is an artifact left from kNet implementation. Network previously implemented kNet listener classes. Same for NewConnectionEstablished and ClientDisconnected methods

Source/Urho3D/Network/Network.cpp
```diff
@@ -200,40 +326,69 @@ void Network::ClientDisconnected(kNet::MessageConnection* connection)
...
+    rakPeer_->SetOfflinePingResponse((const char*)buffer.GetData(), buffer.GetSize());
+}
+
+void Network::DiscoverHosts(unsigned port)
+{
+    // JSandusky: Contrary to the manual, we actually do have to perform Startup first before we can Ping
+    if (rakPeer_ && !rakPeer_->IsActive())
+    {
+        SLNet::SocketDescriptor socket;
+        rakPeer_->Startup(1, &socket, 1);
```
eugeneko: It'd be nice to have comments for magic numbers here and below, like `/* very important stuff */`

Source/Urho3D/Network/Network.cpp
```diff
@@ -200,40 +326,69 @@ void Network::ClientDisconnected(kNet::MessageConnection* connection)
...
+{
+    rakPeer_->SetIncomingPassword(password.CString(), password.Length());
+    password_ = password;
+}
+
 bool Network::Connect(const String& address, unsigned short port, Scene* scene, const VariantMap& identity)
 {
     URHO3D_PROFILE(Connect);
 
-    // If a previous connection already exists, disconnect it and wait for some time for the connection to terminate
```
eugeneko: Who sends `OnServerDisconnected` in AFTER version if `Connect` is called while connection is active?

Source/Urho3D/Network/Network.cpp
```diff
@@ -200,40 +326,69 @@ void Network::ClientDisconnected(kNet::MessageConnection* connection)
...
-        URHO3D_LOGERROR("Failed to connect to server " + address + ":" + String(port));
-        SendEvent(E_CONNECTFAILED);
-        return false;
-    }
 }
 
 void Network::Disconnect(int waitMSec)
 {
-    if (!serverConnection_)
+    if (serverConnection_.Refs() == 0)
```
eugeneko: Should be the same, isn't it?

Source/Urho3D/Network/Network.cpp
```diff
@@ -279,30 +483,34 @@ void Network::BroadcastMessage(int msgID, bool reliable, bool inOrder, const Vec
 void Network::BroadcastMessage(int msgID, bool reliable, bool inOrder, const unsigned char* data, unsigned numBytes,
     unsigned contentID)
 {
+    if (!rakPeer_) 
+        return;
     // Make sure not to use kNet internal message ID's
-    if (msgID <= 0x4 || msgID >= 0x3ffffffe)
+    if (msgID < ID_USER_PACKET_ENUM || msgID >= 255)
```
eugeneko: What's behind these limitations?

Source/Urho3D/Network/Network.cpp
```diff
@@ -317,14 +525,14 @@ void Network::BroadcastRemoteEvent(Node* node, StringHash eventType, bool inOrde
         URHO3D_LOGERROR("Null sender node for remote node event");
         return;
     }
-    if (!node->IsReplicated())
+    if (node->IsReplicated())
```
eugeneko: Wrong

Source/Urho3D/UI/UISelectable.h
```diff
@@ -22,6 +22,11 @@
 
 #pragma once
 
+/// RakNet libraries include windows headers which may conflict with Color::TRANSPARENT
```
eugeneko: This is nothing wrong with UISelectable. Such undefs shall be placed after RakNet includes in `Network/`

eugeneko:
In general, I do like this migration. Maybe a bit risky, but should also be benefical.
Final style cleanup and merge artifact reverting is required, tho.
Changes in `linux_adapter.h` looks suspicious. What are they needed for?

ArnisLielturks:
@eugeneko This file originally comes from the SLikeNet library ([https://github.com/SLikeSoft/SLikeNet/blob/master/Source/include/slikenet/linux_adapter.h](https://github.com/SLikeSoft/SLikeNet/blob/master/Source/include/slikenet/linux_adapter.h))
File is used to allow "safe" (Microsoft) C library functions to be executed on non-windows platforms. 
I did add additional functionality that should also support MinGW compiler (SLikeNet doesn't support it by default) with additional functions which are not part of MinGW yet, as far as I understood. These functions are `inet_ntop`, `inet_pton4`, `inet_pton6` and `inet_pton`.

Also I would not like to rush these changes in master. I want to add samples and do additional testing, probably will have to ask for help in the forums to cover platforms that I don't have access to.

eugeneko:
> I did add additional functionality that should also support MinGW compiler (SLikeNet doesn't support it by default) with additional functions which are not part of MinGW yet, as far as I understood. These functions are inet_ntop, inet_pton4, inet_pton6 and inet_pton.

@ArnisLielturks In this case, it worth to try PR these changes into SLikeNet repo, doesn't it? When it starts working.

ArnisLielturks:
@eugeneko Was thinking about that too

ArnisLielturks:
@eugeneko I run some additional tests and it turns out that MinGW compiler works for 16.04 and not 14.04 at least without additional modifications. 

Ubuntu 14.04 by default uses MinGW with this GCC version
```
$: x86_64-w64-mingw32-g++ --version
x86_64-w64-mingw32-g++ (GCC) 4.8.2
```

Ubuntu 16.04
```
$: x86_64-w64-mingw32-g++ --version
x86_64-w64-mingw32-g++ (GCC) 5.3.1 20160211
```

What should be the proposed way to handle this? I can dig a bit deeper and try to fix the builds for 14.04 which could possibly include the option to build MinGW manually if that kind of package does not already exist.

weitjong:
> What should be the proposed way to handle this? I can dig a bit deeper and try to fix the builds for 14.04 which could possibly include the option to build MinGW manually if that kind of package does not already exist.

I hope that won't be necessary. Urho3D should support any compilers that are C++11 compliant; and GCC/MinGW version 4.8.1 and above is in that category.

weitjong:
Thanks for the hard work. There are a few new linter errors introduced by your proposed changed. See https://travis-ci.org/urho3d/Urho3D/jobs/366820582#L3705-L3722. Once those are cleared then we can merge the PR into a test branch to get it build tested by all the CI builds.

ArnisLielturks:
@weitjong Before the merge I would also like to add LAN discovery sample.

weitjong:
Sure. After you have finished your work then. Also I meant merging to new test branch, not master branch.

weitjong:
Done. It is merged into `ArnisLielturks-SLikeNet` branch. You can continue to submit new commits into your PR and the commits can be easily pulled into this test branch.

ArnisLielturks:
@weitjong Should I create new PR for changes to be merged in `ArnisLielturks-SLikeNet` branch?

weitjong:
No. Use the same PR to push new commits your side. However, you have to wait one of the core devs to pull those new commits into the test branch. I should be able to do that once I am home. Alternatively, if you need a faster turn around then I have no issue to grant you a direct push privilege to the test branch only, if it is possible at all.

weitjong:
The build logs on Travis for MacOS and iOS are harder to discern because they are being filtered. Usually I reproduce the build error locally using unfiltered build output and error streams. They usually are quite verbose and that’s why being filtered on Travis.

If you don’t have a Mac development environment then you could probably look at the last two build logs in Travis which uses Makefile and so their output/error are shown in the usual way (search for “error:” keyword). Let us know if you need more assistance on this one.

ArnisLielturks:
@weitjong Thanks for the additional info! Unfortunately I don't have a MacOS development environment available but will try to figure out the solution by reading Travis logs.

weitjong:
I have rebased just the failed test branch and pushed it for a quick test.

ArnisLielturks:
@weitjong I guess I was wrong about fixing the OSX build problem. I believe I do need assistance to fix this.

weitjong:
I should have time to look at it tomorrow. Thanks again for the effort.

weitjong:
I spot a mistake in the implementation which causes the SLikeNet to be exposed to Urho3D library user. I believe it is also the root cause of the Lua API binding error for the new Network subsystem. Thus, fixing the former should fix the latter as well.

ArnisLielturks:
@weitjong  I think I know what you mean. Basically I have to move the SLikeNet includes to .cpp files and keep header files as clean as possible. Did I understood you correctly?

weitjong:
Yes, you understood it correctly.

weitjong:
Thanks for that. I will pull those new commits now. I notice you have different indentation convention. Before merging to the master branch, those out of alignment code need to be fixed first. Whenever makes sense, after a "delete" we also set the raw pointer to null to avoid double deletion potential problem.

weitjong:
Great. This PR is merge ready. As this PR is rather big, for others who are reading this, please help to test the dev branch if you have time. Although we can fix any bug later after it is merged to master, we would like to keep master branch as stable as possible.

eugeneko:
Just tested samples locally on my laptops (Chat, Scene Replication and LAN Discovery), works fine.
Cannot really test NAT punchthrough on my own...

Issue: if I connect to nonexisting server, e.g. in Chat, it stays for a long time/forever trying to connect. Old version prompts error message.

ArnisLielturks:
@eugeneko Thanks for the feedback! I was able to trace down the problem and it should be fixed in the latest commit.

ArnisLielturks:
@elix22 Thanks for the fix!

elix22:
I tried 16_Chat on my iPad and got an exception .
This is a general issue for both iOS and macOS , getaddrinfo() returns NULL if passing the hostname.
But on my Mac I fixed it by adding my hostname to /etc/hosts
Anyway I fixed it by modifying the function below , the code is compiled to support IPv4 only.
If you will compile the code to support IPv6 (RAKNET_SUPPORT_IPV6==1) you will also have to modify the function GetMyIP_Windows_Linux_IPV4And6().
With the below fix you won't need to modify /etc/hosts on macOS.

Basically so far I verified it on 3 devices connected to my WIFI LAN .
Windows laptop
Mac laptop
iPad

The chat app runs on all 3 , with the 2 fixes , all 3 can send/receive messages to/from each other.


```
void GetMyIP_Windows_Linux_IPV4( SystemAddress addresses[MAXIMUM_NUMBER_OF_INTERNAL_IDS] )
{



	int idx=0;
	char ac[ 80 ];
	int err = gethostname( ac, sizeof( ac ) );
    (void) err;
	RakAssert(err != -1);
    const char *localhost_str="localhost";
	
	struct addrinfo *curAddress = NULL;
	err = getaddrinfo(ac, NULL, NULL, &curAddress);

    if(curAddress == NULL)
    {
        err = getaddrinfo(localhost_str, NULL, NULL, &curAddress);
    }
    
	if ( err != 0 || curAddress == 0 )
	{
		RakAssert(false);
		return ;
	}
	while (curAddress != NULL && idx < MAXIMUM_NUMBER_OF_INTERNAL_IDS)
	{
		if (curAddress->ai_family == AF_INET) {
			addresses[idx].address.addr4 = *((struct sockaddr_in *)curAddress->ai_addr);
			++idx;
		}
		curAddress = curAddress->ai_next;
	}
	
	while (idx < MAXIMUM_NUMBER_OF_INTERNAL_IDS)
	{
		addresses[idx]=UNASSIGNED_SYSTEM_ADDRESS;
		idx++;
	}

}



```

ArnisLielturks:
@elix22 Unfortunately I don't have access to OSX and iOS platforms so I don't know how to proceed with this.  @weitjong Any suggestions? I can create a commit with these changes so others can test it.

weitjong:
If those functions behaves similarly on Linux and on MacOS/iOS then I don't think the proposed fix will get the actual IP address. I quickly test the function on my Linux host with Ruby binding. This is what I got.

```
$ irb
irb(main):001:0> require 'socket'
=> true
irb(main):002:0> Addrinfo.foreach('localhost', nil) { |x| p x }
#<Addrinfo: ::1 TCP (localhost)>
#<Addrinfo: ::1 UDP (localhost)>
#<Addrinfo: ::1 SOCK_RAW (localhost)>
#<Addrinfo: 127.0.0.1 TCP (localhost)>
#<Addrinfo: 127.0.0.1 UDP (localhost)>
#<Addrinfo: 127.0.0.1 SOCK_RAW (localhost)>
=> [#<Addrinfo: ::1 TCP (localhost)>, #<Addrinfo: ::1 UDP (localhost)>, #<Addrinfo: ::1 SOCK_RAW (localhost)>, #<Addrinfo: 127.0.0.1 TCP (localhost)>, #<Addrinfo: 127.0.0.1 UDP (localhost)>, #<Addrinfo: 127.0.0.1 SOCK_RAW (localhost)>]
```

The "::1" is inet6 while "127.0.0.1" is inet. Both are not my actual IP on my LAN. Only when I substitute the "localhost" with my actual host name then the true IP shown up.

As for your question, I think @elix22 should be able to test his proposed code to test without your help. If the result is not "127.0.0.1" or "::1" on MacOS/iOS (which I doubt) then you can take his code. However, please mark any local modification of the 3rd-party source codes, if you do so.

weitjong:
Sorry, pressing the wrong button.

elix22:
The proposed fix won't give the actual IP .
It's a fallback  avoiding an crash .
The following line of code fails on my Mac and iPad when passing the actual hostname.
**err = getaddrinfo(ac, NULL, NULL, &curAddress);
curAddress is NULL both on my Mac and iPad.**
I verified that it works fine on Windows.
It might not the optimal fix but can be considered as a workaround.
Anyway it actually works both on my MacBook Pro and iPad .
using 3 devices connected to my Wifi LAN both 16_chat ,and 17_SceneReplication are working fine.

You can wait until you will receive more feedback from other OSX and iOS users , 
Keep this fix as your backup plan .

weitjong:
I have just pulled your last commit into the test branch.

If we cannot get the samples to work without crashing on macOS/iOS then we could just add a //TODO marker for now and exclude them from being build for these platforms.

elix22:
I would suggest also to apply my suggested fix/workaround to avoid crashing on macOS/iOS

weitjong:
It was late when I made the last comment. What I really wanted to say is, leave the original 3rd-party code alone for now, based on its name it does not sound like it’s intended for Apple platform anyway. Wait for someone else to do the more proper fix.

ArnisLielturks:
@weitjong Looking at the SLikeNet library code it doesn't look like the explicitly say which parts of the code should work on macOS/iOS. I assume that this function is intended for all desktop platforms but the naming for them is bad in this case. The proposed fix is actually a good solution.

Could you post what's the default hosts file content on the OSX platforms?
On ubuntu it already contains 
```
127.0.0.1 myhostname
```

SLikeNet functionality uses `gethostname` function to resolve the IP address, but since the record doesn't exist in hosts file, this function fails.

But searching on the web (since I don't have a OSX device) it seems that OSX platforms doesn't have this line in the hosts file by default. If so, I would suggest to implement the solution by @elix22

Also you can test this by using the ruby script that you provided, but change the "localhost" value to your actual hostname.

weitjong:
I could be wrong but there may not be equivalent of /etc/hosts in iOS. So at best the approach will probably only work for macOS (desktop), like you said, if a hostname is set at all. 

On Apple platform there should be API to get the IP directly instead of hopping from getting the hostname then getting the address info. Albeit it might require to write the code in Objective C and wrap with extern C.

ArnisLielturks:
Proposed fix tries to resolve the device hostname and if it doesn't succeed, it tries the same thing using 'localhost' hostname which should always point to 127.0.0.1 or ::1 on any device. Based on @elix22 tests this solution do work with iOS devices too.

elix22:
> On Apple platform there should be API to get the IP directly instead of hopping from getting the hostname then getting the address info. Albeit it might require to write the code in Objective C and wrap with extern C.

@weitjong Do you know which API ?  
Not promising anything , I might try making a more genuine fix at my spare time during one of  the weekends.

@ArnisLielturks . you can try approaching the SLikeNet owners  letting them know that there is an issue with **GetMyIP_Windows_Linux_IPV4()** on macOS/iOS , maybe they will be able to provide a better solution , you can also provide them my solution so they will merge it into the main branch.

ArnisLielturks:
@elix22 Submitted a issue on the SLikeNet library. Could you please review it just to make sure I got everything right?

elix22:
Yes , looks right to me .
I found some link with different options ,I need some spare time to play with it (won't  in the coming weeks) .

[(https://developer.apple.com/library/content/technotes/tn1145/_index.html#//apple_ref/doc/uid/DTS10002984-CH1-SECGETTINGIPLIST)]

ArnisLielturks:
Update:

Guys from SLikeNet is planning to look at it in the next few days with the appropriate fix, @weitjong they might be planning to use the OSX/iOS API that you mentioned.
I think we can wait few more days for this to be resolved.
Also @elix22 can rest without digging into this problem, but if you do find a solution, let the SLikeNet guys know about the proposed fix.

ArnisLielturks:
Guys from SLikeNet have implemented fixes to solve this but unfortunately the changes breaks the integration a bit with MinGW builds. I'm unable to fix this. 
The problem is that library depends on `iphlpapi` static lib which isn't linked automatically and so the following error appears:

```
../../../lib/libUrho3D.a(RakNetSocket2.cpp.obj):RakNetSocket2.cpp:(.text+0xb5): undefined reference to `GetAdaptersAddresses'
```

Library is visible here `/usr/i686-w64-mingw32/lib/libiphlpapi.a`
What is the correct way to link static libraries to the subprojects?

I tried adding this line in the SLikeNet CMakeLists.txt file:
```
target_link_libraries(${TARGET_NAME} iphlpapi)
```

This didn't help but it worked when I added the same line in Urho3DPlayer subproject. At least then the Urho3DPlayer was able to compile but other subprojects still threw the same error.

Hopefully the problem is really easy to fix but I just don't have that much knowledge about the building pipeline of the MinGW projects.

weitjong:
Thanks for the update. I will have a look at the soonest.

weitjong:
Can you let me know what is release number or tag number of the SLikeNet that you use? Or the commit SHA in case you use their master branch?

ArnisLielturks:
@weitjong https://github.com/SLikeSoft/SLikeNet/commit/bdbae10f457d5c74d6ad10c347d77428cd02c279

weitjong:
Thanks for your patience. I hope you don't mind we make some changes to your test branch.

ArnisLielturks:
@weitjong Totally fine by me!

weitjong:
Don’t worry, will try to workaround this problem differently without touching the upstream code unnecessarily, if possible at all.

weitjong:
It looks like we are the pilot tester, especially for Android platform. Usually that is not a problem when I can try and fix build related issue locally, however, at the moment I am in middle of revamping my local Android build environment. So right now I can only rely on Travis CI's build log to fix the issue one by one. Which come back to my original assessment, it does not seem the code from master branch is being tested for Android at all.

ArnisLielturks:
@weitjong I did run the samples on Android, before the fix provided by SLikeNet. I was able to run one of my more complicated projects which utilizes networking a lot and it ran without any issues. One thing to mention tho is that the full testing of the new SLikeNet functionality is planned before the release of version 0.2.0. I would suggest to leave this PR open till the 0.2.0 is released. I will keep this PR up-to-date till then. As you said, we're the pilot testers for the provided fixes and if it's problem on the 3rd party library side, then there's nothing we can do about it till it's fixed.

ArnisLielturks:
And also thanks a lot for helping out!I appreciate it!

weitjong:
I will get back to work on this test branch to fix the Android build once I have finished revamping the Android build environment with Gradle build system. We could try to send our local changes to upstream SLikeNet as PR too, if by that time the upstream hasn’t released a new version with the fixes for Android platform.

ArnisLielturks:
I actually already did the fixes for Android platform, but when merged in the latest changes accidentally replaced them since I copied over the source code from the SLikeNet repo. I would say that you can leave that to me if you want. 
Also I'm glad to hear that graddle build system might be available in the nearest future.

ArnisLielturks:
@weitjong I see that you were able to fix all the isues related to latest changes in the test branch. Is it ok if I include them in this PR?

weitjong:
Should be OK. I expect git to be smart enough to handle it.

ArnisLielturks:
Is there anything needed from me then?

weitjong:
Android build still fails on the test branch. I thought your last comment said you will be fixing it. Is the fix available in your PR already?

weitjong:
I am working on *other* test branch for Gradle build system. That test branch does not have any commits related to your PR.

weitjong:
Thanks for your patience. Will test your Android fix in the branch. If all goes well then I guess we should merge it to the master branch.

ArnisLielturks:
@weitjong Thanks! Should I also create PR in the https://github.com/urho3d/urho3d.github.io repo regarding the documentation updates?

weitjong:
No. The content of our online documentation come from “Docs” directory in the main Urho repo. So, you can just add more commits to this PR that updates the doxygen files there.

weitjong:
It seems there are no more build issues on all the platforms! If there are no objections then I will merge the test branch to the master master branch after a few cleanups.

@ArnisLielturks do you want to be the maintainer of this network subsystem? We could use your help as a team member.

ArnisLielturks:
Yeah, I saw the errors. Will look into it. I do have gained a lot of knowledge about it and it would be an honor to join!

weitjong:
Great. I will add you in at the soonest. 

All your commits in the PR have been merged in the test branch together with my minor fix/cleanup and the refactor to source the upstream SlikeNet code using a git subtree. So all that need to be done is to merge the test branch. All the CI build from test branch were green now.

weitjong:
Since this is quite a big merge and complicated by an initial git subtree split operation, the merge cannot be done with a push of a button. But will try to allocate some time to get it done by end of the week.

@ArnisLielturks will also pass you some notes via PM on how to maintain our SlikeNet subtree to track the upstream.

ArnisLielturks:
Sounds good, thanks!

weitjong:
I just checked the latest coverity scan report this morning while commuting. It flagged up 64 new (potential) defects and one of them is in Urho3D own network subsystem. 

```
** CID 309664:  Control flow issues  (DEADCODE)
/Source/Urho3D/Network/Network.cpp: 783 in Urho3D::Network::HandleIncomingPacket(SLNet::Packet *, bool)()
```

It is caused by duplicated if condition.

The rest of the defects should not be fixed by us.

weitjong:
Just to clarify, the rest of new defects are in SLikeNet.

weitjong:
I have removed the deadcode in the `Network` class. BTW, I hope you have time to do those TODOs before the next Urho3D release though :smiley:

ArnisLielturks:
When the next release is planned? At the moment I'm in the process of moving so I would say that I can do those in the next 2 weeks.

weitjong:
Take your time. There is no plan release date yet.

1vanK:
https://github.com/urho3d/Urho3D/blob/master/README.md still to contain kNet link

green-zone:
Compilation - MinGW, Windows XP 32bit
Master branch from 7 Dec 2018

With URHO3D_NETWORK
Compilation - OK
RunTime:
An entry point to the _vsnprintf_s procedure was not found in the msvcrt.dll library

Without URHO3D_NETWORK - there are no runtime errors

CMake scripts or code use msvcrt.dll - version for Vista+ OS
But SLikeNet library say:
OS support: Windows XP (SP3), Windows XP x64 (SP2)
https://github.com/SLikeSoft/SLikeNet

Also strange combination MinGW and Visual Studio runtime library (msvcrt.dll)

iSLC:
@green-zone mingw uses the runtimes too. Especially for functions like those.

You should specify which MinGW distribution and/or version you've used. Also which MSVC++ runtimes were installed on that XP machine.

green-zone:
XP SP3 32bit
Visual C++ Redistributable 2005, 2008, 2010
MinGW
version : MinGW-W64-builds-4.3.0
user    : nixman
date    : 07.23.2017- 6:26:18 AM
args    : --mode=gcc-7.1.0

https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win32/Personal%20Builds/mingw-builds/7.1.0/threads-posix/dwarf/

He has a libstdc++-6.dll binary library in his folder:
(MinGWInstallDir)/bin
and
(MinGWInstallDir)/i686-w64-mingw32/lib

===============================
slikenet 
linux_adapter.h
osx_adapter.h
FormatString.cpp and others .cpp files
use vsnprintf_s (not _vsnprintf_s) function.
Maybe the problem is in the build scripts?

weitjong:
@green-zone Thanks for your report. As much as we would like to continue the support for Windows XP, we just don't have the extra resource and machine with XP running anymore to test it. I am not the maintainer for the network subsystem and especially not very familiar with SLikeNet yet, but I believe the issue could be due to Windows XP even with SP3 still only shipped runtime DLL without the secure "_s" version. So in order to make the code portable to WinXP SP3, we probably need to perform a check during CMake initial configuration and dynamically set compiler define somewhere so that the code would branch accordingly. If you can submit a patch for doing that, we will be glad to review it.

weiyuemin:
I encountered crash on Android in SLikeNet module and it can be reproduced. Whenever I call Connect it crashes. Tested on Huawei and Virtual Device.

Stacks:
interpretAddr
interpret
...
SLNet::RakPeer::FillIPList
SLNet::RakPeer::Startup
Urho3D::Network::Connect

stack is said corrupted when in interpretAddr

Will have more test results when i have time.

ArnisLielturks:
@weiyuemin Could you please create issue for this? Will take a look.

PR #2308, opened 2018-04-27
Fix for include order issues on Windows

a-sennov:
The following sequence in includes leads to compilation errors on Windows:

#include <Urho3D/Engine/Application.h>
#include <Urho3D/UI/Text.h>

This is because of wingdi.h included from inside windows.h, that has some nasty macros defined (like TRANSPARENT). To prevent it from loading I've put NOGDI macro into Core/Main.h before include of windows.h

WIN32_LEAN_AND_MEAN are here just to not define what may happily be skipped :)


a-sennov:
Maybe that's because Core/Main.h and UI/UIElement.h  never meet in one .cpp in this particular order ? :)

I think most of users just use the Sample framework where they're protected from this when they simply extend Sample class and do it in another .h/.cpp pair like it's done in the samples. But if you'll try to make app from scratch and also use some UI from the beginning, then you'll have to include Engine/Application.h in your header to derive your app class from it and then one of UI classes into .cpp to make use of UI, so this error will trigger up.

UrhoPlayer also doesn't include any of UI in it's .cpp

eugeneko:
> Also why we don’t receive compiler error report from others all this while?

TBH it was simpler to locally patch compilation error (caused by TRANSPARENT) unstead of properly fixing it.

weitjong:
I see. In that case then I would prefer to add the definition in the CMakeLists.txt.

a-sennov:
Then you should be careful to **not** add WIN32_LEAN_AND_MEAN to ProcessUtils.cpp

weitjong:
Could you kindly try this patch to see if it works for you. Thanks.

[export-header.patch.zip](https://github.com/urho3d/Urho3D/files/1957433/export-header.patch.zip)

eugeneko:
@weitjong I have offtopic question. If I want to add configurable CMake option that will be exposed as `#define`, what should I do?

I found these places:
https://github.com/urho3d/Urho3D/blob/70049ba58210ad26d8402363cf4d864f6c294155/CMake/Modules/exportheader.cmake.in#L26
https://github.com/urho3d/Urho3D/blob/70049ba58210ad26d8402363cf4d864f6c294155/CMake/Modules/FindUrho3D.cmake#L56
https://github.com/urho3d/Urho3D/blob/20c047b743002716f5decd84e5b2fac09b819240/CMake/Modules/UrhoCommon.cmake#L199
https://github.com/urho3d/Urho3D/blob/830436a694af0c5ffb26cdd3e467f030b6509479/Docs/GettingStarted.dox#L97

eugeneko:
> Or both when building and using?

This one. Define that affects engine functionality, but doesn't affect list of source files.

I added `StringHash` debugging things that allows to reverse `StringHash` (in VS2015 debugger watch too) and also warns about StringHash collisions.

This _may_ slow down things since there are mutexes and so on... Of course, it should be optional.

weitjong:
It sounds like a clear cut build option for building Urho3D library to me. So the define should be added into the local scope in the "Source/Urho3D/CMakeLists.txt". But of course, I am just guessing here. If you still have questions, better ask them in the forum in its own topic.

a-sennov:
@weitjong No, patch doesn't work - as I said, ProcessUtils needs full version of windows.h

weitjong:
Can you post a zip file that contain a sample app that failed to build. Locally I have tested using one of modified sample and it did not tripped on the problem you just mentioned. I was just MinGW cross-compiling build on my Linux host to test though. So, it would be great if I can have a failed sample (presumably on VS) to test later.

1vanK:
> Can you post a zip file that contain a sample app that failed to build

Problem when using <Urho3D/Urho3DAll.h>

```
#include <Urho3D/Urho3DAll.h>

class Game : public Application
{
    URHO3D_OBJECT(Game, Application);

public:
    Game(Context* context) : Application(context)
    {
    }
};

URHO3D_DEFINE_APPLICATION_MAIN(Game)
```
failed to building

weitjong:
Aha, that could be the missing ingredient. I have never used the generated Urho3DAll.h. I will try that later.

weitjong:
@1vanK I have tried it on MinGW cross-compiler on my Linux host. Against the current master branch, the build failed. After applying the the patch that I have attached in one of my comment above and it built successfully. Can you help to test the patch on your build environment?

weitjong:
Forget it. I just went to the dark side and confirmed that my patch didn't work for VS.

eugeneko:
@weitjong Please remind me why we cannot kill TRANSPARENT every time when `Windows.h` is included?

weitjong:
> Please remind me why we cannot kill TRANSPARENT every time when Windows.h is included?

I didn't say it cannot be done. But I just ask myself is that the cleanest way to fix this problem?

1vanK:
Acrtually TRANSPARENT s not a very clear identifier^ (1, 0, 0, 0) - trasparent, (1, 1, 1, 0) - transparent, (0, 0, 0, 0) - transparent

weitjong:
> or just Color::ZERO

Although I like that, it does not sound like a color name to me. I am about to change it to ```Color::TRANSPARENT_BLACK```.

PR #2309, opened 2018-04-29
Add StringHash reverse support.

eugeneko:
Added extra StringHash functionality under `URHO3D_HASH_DEBUG` guard (off by default).
- New method StringHash::Reverse that gives `String` from hash
- Warning in log if two strings have same hash
- Some minor helpers that I found useful for debuging things

Why? Because of this:

@weitjong IDK if it's good because there's no CMake `option` in this file...

> It sounds like a clear cut build option for building Urho3D library to me. So the define should be added into the local scope in the "Source/Urho3D/CMakeLists.txt"


weitjong:
There is always the first time. When I started there was even not a single add definition call in this file, if my memory serves me right. Anyway, we can discuss about the right place later. I actually don’t like the fact that the changes only benefited VS user (if so, the new option should have been marked as VS-only and probably should have been DEBUG-build-only). 

The reverse StringHash mapping is a common need. It is needed in the event profiler and also in the Editor, to name a few. Both have their own makeshift way to meet that need. Do you think it’s possible to unify them all into a single more proper solution? Will optionally (based on compiler define) storing the original string as member variable of StringHash class work?

eugeneko:
> I actually don’t like the fact that the changes only benefited VS user 

There's no deep "VS-only" limitiation, and people can use this functionality for e.g. logging and string hash collision detection. The last one is unlikely, but it doesn't mean we should just ignore this problem...

The main benefit is VS-only just because nobody have bothered about making debug visualizer for GDB.

> Do you think it’s possible to unify them all into a single more proper solution?

I just looked at Editor's `globalVarNames` and `EventProfiler`.
If we are talking about code reuse, YES. There isn't any huge benefit, but code will become a bit cleaner.
If we are talking about reusing _this specific global map_ that I added in this PR, then unlikely. This map may grow really big for some user code, and locking mutex may be slow. So I want to make "the main `StringHash` reverser" optional and let specific places use specific solutions.

> Will optionally (based on compiler define) storing the original string as member variable of StringHash class work?

I thought about it. It would have much biger performance and memory impact and make `#ifdef` guarded code more nasty. If object that is known to be small and fast, become fat and slow... Well, at least the `Variant` wont be able to store some types w/o heap allocation.

eugeneko:
@weitjong Please help me shut linter.
I tried to place NOLINT in three ways, no one worked.
RAGE.

eugeneko:
@weitjong Could you help me with linter here? I tried NOLINT in multiple forms and places, but the error is still reporting. I could put some code explicitly into dtor, but it's more like hack.

weitjong:
In the CI build log it shows the name of check in square bracket. You need to pass the name of the check as the argument to the `NOLINT()` marker. In your case it seems both "modernize-use-equals-default" and "hicpp-use-equals-default" apply. So, try putting both as comma separated list argument.

eugeneko:
Oh thanks. I didn't notice that there were two different checks for the same line. Things are green, finally.
I reused hash reverse helper for events. For Editor it doesn't really make sense because I'd have to write more lines of code than I safe.

eugeneko:
@weitjong Good catch! Huh... I think, that is another usecase. StringHash reverser uses HashMap internally and that's the best way to go. However, Resource does _not_ need to reverse hashes. Resource needs to somehow order metadata for nice serialization, so it has to store Vector instead of other containers.

weitjong:
With the current implementation, I still don't believe the URHO3D_HASH_DEBUG option need to be "baked" into the URHO3D export header. The compiler define is only referenced inside the `Source/Urho3D/Math/StringHash.cpp` and nowhere else, which means it is only needed when building the Urho3D library. There is no point to pollute the downstream projects with this compiler define.

Unless, you also modify the `Source/Urho3D/Math/StringHash.h` to also reference this compiler define, say to conditionally make `StringHash::Reverse()` and `StringHash::GetGlobalStringHashRegister()` available or not available. And as such, the downstream projects could also use the baked compiler define in their own code to safely make call to those two methods.

Current implementation is like neither here nor there.

eugeneko:
> With the current implementation, I still don't believe the URHO3D_HASH_DEBUG option need to be "baked" into the URHO3D export header

@weitjong Actually, I started to think the same way when I prepared my last commits.
So, I have to put `add_def` into Urho3D CMake instead, am I right?

weitjong:
Yes, that is the right place to minimize the scope.

eugeneko:
@weitjong is it correct to do it this way?.. https://github.com/urho3d/Urho3D/pull/2309/commits/025714ec4d5385b8066f0bed71784096411c9a71

weitjong:
Yes, the last commit looks good to me. If you read again my initial comment in this PR, I also want the new build option in the UrhoCommon to be made conditional dependent on Debug build config (and for VS) only. This is to prevent user from accidentally enable it in other build config.

weitjong:
But I still don’t see anyone want to do what you said in a Release build.

eugeneko:
> But I still don’t see anyone want to do what you said in a Release build.

My first thought was about `RelWithDebInfo` that I always use and refer to it as "release build".
The difference between `Debug` and `RelWithDebInfo` is only the degree of optimization.
Both configurations are debuggable. So, they both may need hash debugging too.

But then I recalled about true `Release` configuration.
For this configuration it makes sense.

How to make such a guard that checks for `Release`, but not for `RelWithDebInfo`?

weitjong:
You can get the inspiration from this line

https://github.com/urho3d/Urho3D/blob/master/CMake/Modules/UrhoCommon.cmake#L280

Either use an OR with two expressions; or use a NOT in front of a single expression for Release.

One more thing, also add that new option description into the corresponding dox file.

eugeneko:
> Either use an OR with two expressions; or use a NOT in front of a single expression for Release.

I just figured out that this doesn't work.
`add_definitions` always applies to all `CMAKE_BUILD_TYPE` for multi-configuration platforms because it's executed once at configuration stage.
The only possible solution is override `CMAKE_CXX_FLAGS_*` for Debug and RelWithDebInfo, but it will be compiler-specific.

I don't think it makes sence to fight with CMake and postpone this PR anymore.
I'm going to fix the last item and merge this PR.
Please let me know if you know good way to do it.

> One more thing, also add that new option description into the corresponding dox file.

weitjong:
For multi-config generator, our build system unsets the CMAKE_BUILD_TYPE variable at the very beginning. Thus, a "NOT CMAKE_BUILD_TYPE STREQUAL Release" is a good expression to use for both single-config and multi-config generators. No?

weitjong:
Anyway, you can just merge the PR in master branch or any other branch first then I can help to fix it.

weitjong:
I think you misunderstood me. If you use:

```
cmake_dependent_option (URHO3D_HASH_DEBUG "comment here" FALSE "NOT expression here" FALSE)
```

Then for the single-config generator user the option is not available when it is a Release build config. Otherwise the option is available and user can set it to TRUE or FALSE.

For the multi-config generator user, however, the option is always available (because during generation time we don't know which build config is effective). But still, the option is defaulted to FALSE initially. So, in this case, we just not able to prevent VS user from shooting himself in the foot for Release build config.

Issue #17, opened 2013-10-12
Problems running with raspbian


cartuchogl:
I've try to run Urho3D in a raspberry pi model b rev2 running raspbian.

Compilation is fine, but when I try to run any a demo script, out is:

``` shell
$ ./Urho3D Data/LuaScripts/01_HelloWorld.lua 
[Sat Oct 12 17:38:59 2013] INFO: Opened log file Urho3D.log
[Sat Oct 12 17:38:59 2013] INFO: Added resource path /home/pi/repositories/Urho3D/Bin/CoreData/
[Sat Oct 12 17:38:59 2013] INFO: Added resource path /home/pi/repositories/Urho3D/Bin/Data/
[Sat Oct 12 17:38:59 2013] ERROR: Could not open window
```

I added a LOGERROR with SDL_GetError after the code that show this message in Source/Engine/Graphics/OpenGL/OGLGraphics.cpp. Now the out show:

``` shell
....
[Sat Oct 12 17:38:59 2013] ERROR: Could not load OpenGL ES/EGL library
```

ldd not show link with libEGL

``` shell
$ ldd Urho3D
    /usr/lib/arm-linux-gnueabihf/libcofi_rpi.so (0xb6fa1000)
    libbcm_host.so => /opt/vc/lib/libbcm_host.so (0xb6f87000)
    libGLESv2.so.2 => /usr/lib/arm-linux-gnueabihf/libGLESv2.so.2 (0xb6f6e000)
    libdl.so.2 => /lib/arm-linux-gnueabihf/libdl.so.2 (0xb6f63000)
    librt.so.1 => /lib/arm-linux-gnueabihf/librt.so.1 (0xb6f54000)
    libpthread.so.0 => /lib/arm-linux-gnueabihf/libpthread.so.0 (0xb6f35000)
    libstdc++.so.6 => /usr/lib/arm-linux-gnueabihf/libstdc++.so.6 (0xb6e68000)
    libm.so.6 => /lib/arm-linux-gnueabihf/libm.so.6 (0xb6df7000)
    libgcc_s.so.1 => /lib/arm-linux-gnueabihf/libgcc_s.so.1 (0xb6dcf000)
    libc.so.6 => /lib/arm-linux-gnueabihf/libc.so.6 (0xb6ca0000)
    libvcos.so => /opt/vc/lib/libvcos.so (0xb6c8d000)
[...]

PR #2310, opened 2018-04-30
Make Urho3D.Containers move-friendly

eugeneko:
I wanted to make this refactoring long time ago...
It might be risky, so I'd like to have this code reviewed by somebody except me. I already fixed three bugs here, so there may be others.

- Implemaneted move semantic for `HashMap`, `HashSet`, `String`, `List` and `Vector`. These containers could be efficently moved via `std::move` now.
- `Vector` use `std::move` under the hood now. So, `Vector` could store heavy types _much_ more efficiently. It also could store non-copyable types like `UniquePtr`.
- `Vector` is 100% safe for self-insertions now, just for sake of world beauty.
- Minor changes...

I did some benchmarks for `Push`ing things in vector, w/ and w/o reserve, with tiny and heavy type (`int` and `std::string`)
I don't really like the result because I can't explain why new vesion of the vector is faster than the old even for small types. It looks like either benchmarking or code error, soo...
http://quick-bench.com/vIRUOIHs6-bgWRL0DWYnnPhC8wk
http://quick-bench.com/7YEZOuEzx5vpMI-rb_PwEXHqSH4

I smoke-tested all the samples, worked fine so far. CI-ing things right now.

@weitjong, maybe even @cadaver ?


weitjong:
This has happened in the past. Sometime Travis CI upgraded their VMs but left some garbage behind. It also does not help if we just keep adding more samples. In order to reclaim the space back, I need to spend some time (which I don't have now) to analyze the disk utilization on the VM. As workaround you can exclude some of the samples from being built and tested. That should buy you some time. See this commit to see how it was done in the past. I believe the magic ingredient is still there. The '#n' is the job number, for your case it is '#17'. List the sample you want to exclude in the square bracket.

https://github.com/urho3d/Urho3D/commit/aa706344ff39709eb55687496a72b35b18360ed0

weitjong:
It looks like I was wrong to blame it on Travis CI side too quickly. I have just rerun one of the recent CI build job from the other test branch and the job 17 did not end up with out of space issue. So, we can rule out it is caused by Travis VM. It could be that we have almost hit the space limit but just not there yet until these changes. Have you also compared the binary size difference between the old and new in both Release and Debug build config, besides the performance comparison?

eugeneko:
@weitjong Just tested it

Item | Debug (Before) | Debug (After) | Release (Before) | Release (After)
--|--|--|--|--
Urho3DPlayer.exe | 22237184 | 22488576 | 8844288 | 8884736
Urho3D.lib | 164144388 | 166693520 | 137379864 | 137687072

1.1% growth in Debug, 0.4% growth in Release for Urho3DPlayer.exe

1.5% growth in Debug, 0.2% growth in Release for Urho3D.lib

weitjong:
I think that is a fair trade-off. When you are ready, you can just ignore this technical build error and move on to merge the commits into a new test branch for full CI tests. I will help to fix/workaround the technical build error there.

weitjong:
The linter does not happy with changes :)

eugeneko:
@weitjong Oh, I got it. Just read about linter hints, will update code.
I don't see other issues with the code tho.

eugeneko:
Well, I fixed these linter issues locally. However, I'm unsure if it's good idea to actually push them.

The idea of linter suggestion is that if we pass comething as `const T&` and then just copy it (e.g. in ctor), it's better to pass it by value and then `std::move` it.

On the other hand, this change make code less clean. Explicit `std::move` usually makes defect injection easier because someone may accidently use moved-away variable. And I don't believe that this change speed up our code even in 0.1%.

So, I vote for shuting down exactly this linter error.

weitjong:
I agree that for this particular check is only a suggestion, not a code error that must be fixed. I am ok if you choose to disable it all together instead of adding NOLINT marker. Linter is just a tool, we (devs) should be the one in control. On the CI side I can only make the tool to return error when it detects any linter errror or warning or suggestion, in order to get the attention of the committer. If suppressing the error by adding NOLINT is too much of work then disable the particular check is the way to go.

eugeneko:
> If suppressing the error by adding NOLINT is too much of work then disable the particular check is the way to go.

Not that much of work, but I think this suggestion is one that I don't want to follow, unless benefit is confirmed.

Source/Urho3D/Container/Str.h
```diff
@@ -155,9 +164,20 @@ class URHO3D_API String
     /// Assign a string.
     String& operator =(const String& rhs)
     {
-        Resize(rhs.length_);
-        CopyChars(buffer_, rhs.buffer_, rhs.length_);
+        if (&rhs != this)
```
rokups: Maybe this should be `assert`? That would be consistent with other operators. Besides there is no reason for user to ever do this anyway.

Source/Urho3D/Container/Vector.h
```diff
@@ -65,14 +70,21 @@ template <class T> class Vector : public VectorBase
...
     Vector(const Vector<T>& vector)
     {
-        *this = vector;
+        DoInsertElements(0, vector.Begin(), vector.End(), CopyTag{});
     }
+
+    /// Move-construct from another vector.
+    Vector(Vector<T> && vector)
+    {
+        Swap(vector);
```
rokups: Maybe also add `assert(&vector != this)`?

Source/Urho3D/Container/Vector.h
```diff
@@ -94,12 +107,20 @@ template <class T> class Vector : public VectorBase
         // In case of self-assignment do nothing
         if (&rhs != this)
```
rokups: Similarly i propose converting this check to `assert`.

Source/Urho3D/Container/Vector.h
```diff
@@ -213,6 +213,14 @@ template <class T> class Vector : public VectorBase
         return Buffer()[index];
     }
 
+    /// Create an element at the end.
+    template <class... Args> T& Emplace(Args&&... args)
+    {
+        T value(std::forward(args)...);
+        Push(std::move(value));
```
rokups: Using `value` variable breaks perfect forwarding i believe. How about this?

> Push(std::forward(args)...);

Source/Urho3D/Container/HashSet.h
```diff
@@ -187,6 +187,13 @@ template <class T> class HashSet : public HashBase
         head_ = tail_ = ReserveNode();
         *this = set;
     }
+
+    /// Move-construct from another hash set.
+    HashSet(HashSet<T> && set) noexcept
+    {
+        Swap(set);
```
rokups: Add `assert(&rhs != this);`

Source/Urho3D/Container/Str.h
```diff
@@ -60,6 +60,15 @@ class URHO3D_API String
         *this = str;
     }
 
+    /// Move-construct from another string.
+    String(String && str) noexcept :
+        length_(0),
+        capacity_(0),
+        buffer_(&endZero)
+    {
+        Swap(str);
```
rokups: Add `assert(&str != this);`

Source/Urho3D/Container/HashMap.h
```diff
@@ -230,6 +230,13 @@ template <class T, class U> class HashMap : public HashBase
         head_ = tail_ = ReserveNode();
         *this = map;
     }
+
+    /// Move-construct from another hash map.
+    HashMap(HashMap<T, U> && map) noexcept
+    {
+        Swap(map);
```
rokups: Add `assert(&map != this);`

eugeneko:
@rokups I updated Vector.Emplace.

Since things are green, I'm going to merge it in the nearest future.

Source/Samples/01_HelloWorld/HelloWorld.cpp
```diff
@@ -37,6 +37,11 @@ URHO3D_DEFINE_APPLICATION_MAIN(HelloWorld)
 HelloWorld::HelloWorld(Context* context) :
     Sample(context)
 {
+    Vector<UniquePtr<int>> vec;
+    vec.Push(MakeUnique<int>(10));
+    vec.Push(MakeUnique<int>(11));
+    vec.Push(MakeUnique<int>(12));
+    vec.Insert(vec.Begin(), MakeUnique<int>(13));
```
pengfei666: What do you mean here?

Source/Urho3D/Container/Str.h
```diff
@@ -60,6 +60,15 @@ class URHO3D_API String
         *this = str;
     }
 
+    /// Move-construct from another string.
+    String(String && str) noexcept :
+        length_(0),
+        capacity_(0),
+        buffer_(&endZero)
+    {
```
pengfei666: Is it necessary to initialize member variable here? as Swap also could assign them correct value.

Source/Samples/01_HelloWorld/HelloWorld.cpp
eugeneko: Oh, thanks, I forgot to revert this chunk of code.
I added it to test compilation on multiple compilers and platforms.

Source/Urho3D/Container/Str.h
eugeneko: `Swap` will _swap_ them, not assign. And `String && str` definetely shan't be filled with uninitialized garbage.

Source/Urho3D/Container/Vector.h
```diff
@@ -192,24 +213,61 @@ template <class T> class Vector : public VectorBase
...
+            new (&Back()) T(std::forward<Args>(args)...);
+        }
+        else
+        {
+            T value(std::forward<Args>(args)...);
+            Push(std::move(value));
+        }
+        return Back();
+    }
+
```
pengfei666: Standard C++ library provides insert/push_back, emplace/emplace_back.

I see Emplace is same as STL emplace_back.

A little confused.

Then where is STL emplace?

pengfei666:
The idea of linter suggestion is that if we pass comething as const T& and then just copy it (e.g. in ctor), it's better to pass it by value and then std::move it.


For this, I vote for using copy and move.

CLANG check also recommend to do that.

Source/Urho3D/Container/Str.h
pengfei666: Makes sense. Thanks.

pengfei666:
Though copy and move only could bring a little boost, that is still considered.

 And, optimizing for rvalue is indeed what you are doing in this PR.

pengfei666:
All in all, following modern move sementic is a good practice.

pengfei666:
BTW, I read your code carefully, give you a thumb up 👍

Source/Urho3D/Container/Vector.h
eugeneko: Ok, I thought about naming it `EmplaceBack`.
Since you mentioned it too, I'll rename this function.
About `emplace`... I don't consider this function so useful and I'm too lazy to implement it, at least in this PR.
It'd be just construct & insert anyway.
BTW, &&-inserts are missing too.
Next time, maybe.

eugeneko:
> For this, I vote for using copy and move.

You know... I'd follow this CLang hint if it was at least a bit usefull or meaningful.
I mean, imporvement in some generic place like container, or maybe some proven bottleneck or like this.
Instead, we have a bunch or random places in code that have literally nothing to do with my changes:
```
/home/travis/build/urho3d/Urho3D/Source/Urho3D/Core/Spline.cpp:50:16
/home/travis/build/urho3d/Urho3D/Source/Urho3D/Graphics/Material.cpp:182:80
/home/travis/build/urho3d/Urho3D/Source/Urho3D/Graphics/ShaderVariation.cpp:34:34
/home/travis/build/urho3d/Urho3D/Source/Urho3D/Graphics/ShaderVariation.cpp:41:51
/home/travis/build/urho3d/Urho3D/Source/Urho3D/Graphics/ShaderVariation.cpp:50:51
/home/travis/build/urho3d/Urho3D/Source/Urho3D/IO/FileSystem.cpp:239:44
/home/travis/build/urho3d/Urho3D/Source/Urho3D/IO/FileSystem.cpp:263:40
/home/travis/build/urho3d/Urho3D/Source/Urho3D/Network/HttpRequest.cpp:39:65
/home/travis/build/urho3d/Urho3D/Source/Urho3D/Network/HttpRequest.cpp:39:96
/home/travis/build/urho3d/Urho3D/Source/Tools/SpritePacker/SpritePacker.cpp:66:16
/home/travis/build/urho3d/Urho3D/Source/Tools/SpritePacker/SpritePacker.cpp:66:37
```

PR #2313, opened 2018-05-13
Editor: more info about models

1vanK:
Hm, simple smoothed cube has 20 vertices... I think problem in blender exporter


weitjong:
Nice!

PR #2320, opened 2018-06-02
Layer and Position.z in 2D games,which is the first?

spwork:
Hello everyone, I recently used `StaticSprite2D` to study the relationship between `SetLayer` and` position.z` and found that `z` has priority over `SetLayer.` Layers only work when `z` is equal. I looked at the unity engine and found that it was the opposite. It was to consider layer first and then consider z. Which kind of order is universal? Should we be consistent with unity?


weitjong:
I am not familiar with 2D subystem. It has lost its maintainer long time ago, so you are welcome to submit more changes you think is beneficial for the general use case. I think this PR is otherwise merge-ready if not because of dirty commit history. Could you kindly squash it into a single commit? Thanks.

PR #2321, opened 2018-06-02
Fixed the 2D rendering order makes the layer priority higher than the Z

spwork:
Hello everyone, I recently used StaticSprite2D to study the relationship between SetLayer andposition.z and found that z has priority over SetLayer. Layers only work when z is equal. I looked at the unity engine and found that it was the opposite. It was to consider layer first and then consider z. Which kind of order is universal? Should we be consistent with unity?


weitjong:
Thanks again.

1vanK:
Nice!

PR #2323, opened 2018-06-03
Add check for degenerate sub-triangles in OcclusionBuffer::DrawTriang…

### LINKED ISSUE #2322: Some Urho3D samples built using WASM throw uncaught runtime exception
The latest CI build for WASM uses the Emscripten v1.38.4.

eugeneko:
…le2D. Fixes #2322.

BTW, @weitjong, do we have anything for debugging Emscripten?
It could provide really readable callstacks if it is asked to.

I added this snippet and it took me ~5min to locate the defect. Unsure if we could use lesser debug levels tho...
```
if (EMSCRIPTEN)
    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g4")
endif ()
```

### DIFF
--- a/Source/Urho3D/Graphics/OcclusionBuffer.cpp
+++ b/Source/Urho3D/Graphics/OcclusionBuffer.cpp
@@ -879,103 +879,120 @@ void OcclusionBuffer::DrawTriangle2D(const Vector3* vertices, bool clockwise, un
     if (!clockwise)
         middleIsRight = !middleIsRight;
 
+    const bool topDegenerate = topY == middleY;
+    const bool bottomDegenerate = middleY == bottomY;
+
     Gradients gradients(vertices);
-    Edge topToMiddle(gradients, vertices[top], vertices[middle], topY);
     Edge topToBottom(gradients, vertices[top], vertices[bottom], topY);
-    Edge middleToBottom(gradients, vertices[middle], vertices[bottom], middleY);
 
     int* bufferData = buffers_[threadIndex].data_;
 
     if (middleIsRight)
     {
         // Top half
-        int* row = bufferData + topY * width_;
-        int* endRow = bufferData + middleY * width_;
-        while (row < endRow)
+        if (!topDegenerate)
         {
-            int invZ = topToBottom.invZ_;
-            int* dest = row + (topToBottom.x_ >> 16u);
-            int* end = row + (topToMiddle.x_ >> 16u);
-            while (dest < end)
+            Edge topToMiddle(gradients, vertices[top], vertices[middle], topY);
+            int* row = bufferData + topY * width_;
+            int* endRow = bufferData + middleY * width_;
+            while (row < endRow)
             {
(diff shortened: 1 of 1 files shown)


weitjong:
BTW, EMCC is based on LLVM/Clang so in theory if you specify the level multiple times then the last option takes effect.

eugeneko:
> Use the Debug build config.

It was predictable answer...
I wanted to confirm that we _do_ setup Web debugging in debug config via `-g?` options. I didn't manage to find it in cmake files.

weitjong:
I just did a quick grep to make sure. Finders keepers, you should learn to do the git grep :)

```
$ git grep "\-g4"
CMake/Modules/UrhoCommon.cmake:                set (CMAKE_EXE_LINKER_FLAGS_DEBUG "${CMAKE_EXE_LINKER_FLAGS_DEBUG} -g4 -s DISABLE_EXCEPTION_CATCHING=1")
CMake/Modules/UrhoCommon.cmake:                set (CMAKE_MODULE_LINKER_FLAGS_DEBUG "${CMAKE_MODULE_LINKER_FLAGS_DEBUG} -g4 -s DISABLE_EXCEPTION_CATCHING=1")
```

Also in Debug build config, the build will generate a source map. Last time I tried I was even able to see the C++ source code while tracing through in the Chrome inspector. I was using JS output then though.

eugeneko:
Ok, let it go in.
I don't think this PR requires thoughtful review since it's just two more `if`s around occlusion code and some non-functional line reordering.

eugeneko:
Houston we have a problem...

@weitjong Any idea why does it fail so strangely?
https://travis-ci.org/urho3d/Urho3D/jobs/387779163
etc

weitjong:
I have explained it before, I think. Don't worry about those lone build failure or build error. Our CI build setup has pushed Travis VM very hard and sometime might bring Travis VM to its knee. If you bother with it then you can simply just click on the restart button to redo the job. What you need to pay attention to is when multiple CI jobs failed at the same time, like what happening now for Web CI due to the commit related to StringHash reverse map.

eugeneko:
> What you need to pay attention to is when multiple CI jobs failed at the same time, like what happening now for Web CI due to the commit related to StringHash reverse map.

Whoops :embarrassed:
I didn't notice it /_-

eugeneko:
Since I managed to run Web builds locally, I'll try to investigate.

weitjong:
I am able to reproduce the error locally. And I think it is due to stale PCH and/or ccache objects. I am doing another build now after clearing all the cache from a new build tree.

weitjong:
Nope. spoke too soon. Still failed.

eugeneko:
I pushed sum fix. HTH. Will see.
https://github.com/urho3d/Urho3D/commit/4862785ab7a07b2623729d9f43735cc60518c9d1

weitjong:
That works for me.

PR #2334, opened 2018-06-17
Re-enable suspending main loop when application is paused on android

### LINKED ISSUE #2328: Race condition when suspending application on Android
Any idea what is rationale behind disabling [SDL_ANDROID_BLOCK_ON_PAUSE](https://github.com/urho3d/Urho3D/blob/master/Source/ThirdParty/SDL/src/video/android/SDL_androidevents.c#L29-L30) in SDL? This creates a race condition when suspending application. Tapping "tasks" button rapidly eventually causes application to glitch out and objects become black or appear as if they were "textured" with uninitialized memory. Even worse side effect is that minimized application still spins main loop at 10 fps which is a battery killer. Reenabling `SDL_ANDROID_BLOCK_ON_PAUSE` i have not noticed any unwanted side effects. Decision for disabling this should probably be reevaluated.

@cadaver / @weitjong either of you recall why was it actually disabled? Comment is rather obscure. I thought `SDL_ANDROID_BLOCK_ON_PAUSE` meant app could no longer be quit by pressing UI button, but that works fine.

rokups:
Fixes #2328.

Previous behaviour allowed main loop to spin (at 10fps, as if window is minimized) while application was in background. This is both bad for battery life and also introduces a race condition where rapidly suspending and resuming application may get it into the state where application is resumed while some or most GPU objects are not recreated.

### DIFF
--- a/Source/ThirdParty/SDL/src/video/android/SDL_androidevents.c
+++ b/Source/ThirdParty/SDL/src/video/android/SDL_androidevents.c
@@ -26,8 +26,7 @@
 #if SDL_VIDEO_DRIVER_ANDROID
 
 /* We're going to do this by default */
-// Urho3D: do not block on pause to allow exiting the application
-//#define SDL_ANDROID_BLOCK_ON_PAUSE  1
+#define SDL_ANDROID_BLOCK_ON_PAUSE  1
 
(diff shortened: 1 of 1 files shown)


weitjong:
Thanks!

PR #2336, opened 2018-06-20
Replace integer constants with enums and use Use FlagSet<Enum> for flags

rokups:
This is a joint effort of @eugeneko and mine to bring more typesafety into Urho3D. Most constants groups were replaced with enums. Method parameters now use enum type instead of plain `int`/`unsigned`. We opted out of using `enum class` to preserve backwards compatibility. Use of `FlagSet<Enum>` allows to combine enum members and form masks as before. We might have missed some parameter/return types so if anyone spots anything - do tell.


eugeneko:
I pushed flagset branch into main repo. It doesn’t have suspicious changes
but still fails to generate Lua bindings for some reason in 50% or builds.

ср, 20 июня 2018 г. в 19:05, Rokas Kupstys <notifications@github.com>:

weitjong:
I cleared the VM build cache and triggered a rebuild. If the result is the same then we can at least rule out it was caused by build cache. If it the result is ok then it was just technical build error caused by the stale cache.

rokups:
@weitjong thanks! Looks like it did a trick.

@eugeneko could you elaborate on "suspicions changes"?

eugeneko:
1) Static cast in generic place may result in random behavior changes
including additional casts and allocations. I was very careful when
implemented these helpers in order to avoid any side effects during
attribute accessing.

2) Using enum instead of int/flagSet in lua bindings looks hacky. It is
shorter and probably work, but I don’t think it’s clean and fair solution.

чт, 21 июня 2018 г. в 9:57, Rokas Kupstys <notifications@github.com>:

rokups:
1) I see. You know this stuff better so you get to deal with attributes then :+1: 
2) Just followed your example in some commit ;) Thanks for sorting it out.

By the way feel free to force-push to this branch if needed.

rokups:
@weitjong could you please clear build cache for https://github.com/urho3d/Urho3D/tree/flagset? Some builds are again failing with similar errors.

rokups:
Weird core dumps again. Could you please clean cache once more and trigger a full rebuild on travis?

weitjong:
Next time you can try to add "[cache clear]" as part of the commit message to see if it helps. I also notice Travis seems to keep trashing our cache lately, but I don't want to jump any conclusion on the cause yet.

rokups:
Please restart travis build. Many jobs failed because of network timeouts.

weitjong:
As communicated earlier, you can add a meta-instruction in the commit message to force the cache to be cleared.

rokups:
Yes i know, except there is nothing to commit and multiple people are using this branch. Due to this making dummy commit only to trigger build or overwriting commits both seem like a bad idea.

rokups:
No, it is not available to me

weitjong:
My bad. Earlier I thought you were hit by the corrupted cache issue again.

eugeneko:
I'm going to merge this PR in the nearest future after final review. If
anyone has objections, it's time to speak up.

2018-07-10 13:14 GMT+03:00 Yao Wei Tjong <notifications@github.com>:

PR #2341, opened 2018-06-23
Improved sample 41_databaseDemo

LordGolias:
Issue: sample 41_databaseDemo requires `URHO3D_DATABASE_ODBC` option to be enable. However, it only uses `SQLite` and therefore should only require `URHO3D_DATABASE_SQLITE` to be compilable.

This PR contains commit 40c5e1f that fixes this.


weitjong:
The database demo should work with either SQLite embedded database or any OCBC-compliance database (tested with MariaDB last time). So, the condition check in the build script is correct, i.e. it will configure this target when either of these two build options is enabled. Still, thanks for the PR.

PR #2342, opened 2018-06-24
GetOSVersion() update for macOS

Florastamine:
Nothing fancy. The patch updated `GetOSVersion()` to recognize newer macOS versions, i.e. High Sierra and Mojave. Also more precise version reporting on some older versions.


weitjong:
Thanks!

Issue #2326, opened 2018-06-07
gcc 8.1 incorrect code


alex-rass-88:
Bug in gcc 8.1. Incorrect code is being compiled. It's like an inside-out image.
![06-bug](https://user-images.githubusercontent.com/7008147/41107292-c0998e66-6a7a-11e8-9294-cd81c40d8aa8.png)

In gcc 7.3 normal:
![06-norm](https://user-images.githubusercontent.com/7008147/41107323-cebc966e-6a7a-11e8-8fad-1a90babb0c4c.png)

How to get workaround the problem I do not know yet. So keep in mind.

weitjong:
Reproduced on my GCC 8.1.1. I have been using Clang 6.0 lately, everything is normal/OK there too.

weitjong:
Didn’t have enough time to conclude anything last night, but at least I was able to narrow the issue down a little bit to compiler optimization problem. As soon as any -O level is given then the emitted code becomes wrong. My gut feeling is, it has something to do with vectorization. If so, it is our code base that needs to adapt to the new GCC 8.x. It does not seem like a compiler bug to me. But like I said, this is just my inconclusive opinion.

alex-rass-88:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80898

alex-rass-88:
In Debug and Release with gcc 8.1 difference worldTransform matrix in camera node.
Problem in Matrix3x4 constructors (Matrix3x4.h lines 172 and 186). If disable SSE code it work normal.

```
    Matrix3x4(const Vector3& translation, const Quaternion& rotation, float scale) noexcept
    {
#ifdef URHO3D_SSE_HIDE
        __m128 t = _mm_set_ps(1.f, translation.z_, translation.y_, translation.x_);
        __m128 q = _mm_loadu_ps(&rotation.w_);
        __m128 s = _mm_set_ps(1.f, scale, scale, scale);
        SetFromTRS(t, q, s);
#else
        SetRotation(rotation.RotationMatrix() * scale);
        SetTranslation(translation);
#endif
    }

    /// Construct from translation, rotation and nonuniform scale.
    Matrix3x4(const Vector3& translation, const Quaternion& rotation, const Vector3& scale) noexcept
    {
#ifdef URHO3D_SSE_HIDE
        __m128 t = _mm_set_ps(1.f, translation.z_, translation.y_, translation.x_);
[...]

PR #2344, opened 2018-07-03
Fix for #2326

alex-rass-88:
Remove '-ffast-math' option.
Remove GCC 8.x check in Matrix3x4.h

-ffast-math dangerous option.


alex-rass-88:
I can test on gcc 7.3 with the option enabled and disabled

alex-rass-88:
tested gcc 7.3.0 on samples 12_PhysicsStressTest and 20_HugeObjectCount. There is no difference in performance.

weitjong:
I suppose the difference depends on how frequent functions from the math library are being called in an actual game application.

alex-rass-88:
'-ffast-math' does not always provide performance gains.

PR #2355, opened 2018-07-29
 Fixed DX11 DSV/RTV mismatch error when rendering UI to render target

SuperWangKai:
In DX11 renderer, when we set render target for UI elements and let them render, the depth-stencil view will NOT be applied before the drawcall, which will lead DX11 Debug Layer error happen (We can use RenderDoc to inspect).

**"48_Hello3DUI"** has the issue.

The solution is set the depth-stencil view before drawcall as well.

default


SuperWangKai:
Ha, that's epic. Fixed!

Issue #2358, opened 2018-08-07
l10n (Localization) JSON format


ArnisLielturks:
At the moment the localization files are with the following structure:
```
{
	"lang":{
		"de":"Sprache: Deutsch"
	},
	"quit":{
		"de":"Verlassen"
	},
	"Press this button":{
		"de":"Drücken Sie diese Taste"
	},
	"title":{
		"de":"Titel"
	}
}

```

which is kinda hard to generate. If we look at the existing services which manage translations, like https://lokalise.co or https://poeditor.com, none of them generate JSON with the following structure.
I have a strong background developing WEB applications and I would suggest to also create support for the translation files wit the following structure:
```
{
	"lang": "Sprache: Deutsch",
	"quit": "Verlassen",
	"Press this button": "Drücken Sie diese Taste",
	"title": "Titel"
}
```
where 1 JSON file only contains 1 language. 

I don't want to replace the existing functionality, I want to extend it to support multiple formats. Any thoughts?

ArnisLielturks:
PR: https://github.com/urho3d/Urho3D/pull/2359

PR #2359, opened 2018-08-07
localization JSON file now supports simple key-value structure

ArnisLielturks:
Issue: https://github.com/urho3d/Urho3D/issues/2358


rokups:
Is this a good idea though? I did very same thing in my little project and i immediately bumped into issue where changing english string required changes in files for every language. That is not very practical.

ArnisLielturks:
The thing is that ideally you wouldn't want to generate those files manually at all. These changes still allow the previous format to be used, but for me it's not an option and I hope that I'm not the only one thinking like that.

ArnisLielturks:
Any other thoughts? 
@rokups @weitjong @eugeneko

rokups:
Last time looking at the changes (and submitted issue) i looked over the fact that this is meant to load autogenerated files. In that case i say it is perfectly fine even if that format sucks for manual editing.

What still looks confusing is that both formats are mashed into single `LoadJSON()` call and depending on existence of `language` parameter this method behaves quite differently. This is not very intuitive. Maybe `LoadJSON()` should be split to two private methods handling each format? I really am not sure. Others should weight in on this one.

weitjong:
In general for smallish changes you don't need to get the approval from any of us. That's the benefit of having the push/write privilege. So, don't feel bad or anything when no one from the core team provides any feedback. If there are major issues, rest assure one of us will shout (or so I hope).

In this particular case, however, I actually don't like to have two different formats. Personally I feel it will cause more confusion to users. We are allowed to break the API, if it needs to.

ArnisLielturks:
Splitted the `LoadJSON` methos up into 2 methods - `LoadSingleLanguageJSON` and `LoadMultipleLanguageJSON`. I would like to keep both formats for now. If everything else related to implementation is ok, you can safely merge this.

rokups:
This broke use of old format. We can no longer use:
```
{
	"lang":{
		"de":"Sprache: Deutsch"
		"lt": "Kalba: Lietuvių",
	},
	"Quit":{
		"de":"Verlassen"
		"lt": "Išeiti",
	},
```
Note that this format supports 3 languages, not two. One is main language which is english, and english phrase is used as a key. Multiple translations are provided for a phrase of main language.

ArnisLielturks:
@rokups Will fix this when I get a chance. It does make sense to just return the translation key when no translation is present.

rokups:
Returning key sounds good. `LoadSingleLanguageJSON()` should also be renamed to something else because it also is supposed to load multiple languages. At this point `LoadJSONFile()` also becomes bad because it does two completely different things based on language parameter and that is not obvious. I should have brought all this up during PR review. I was lazy, sorry :|

ArnisLielturks:
Maybe we should get rid of the `LoadJsonFile` method completely and leave only `LoadSingleLanguageJSON` and `LoadMultipleLanguageJSON` methods.

ArnisLielturks:
@rokups 
I of course don't know how others use the translation mechanisms but all of the projects that I have worked on doesn't have term as a default language if not translation files are provided. All the Translation keys are basically just system texts, and if no translation is set, system texts are displayed on the UI, same goes if specific translation file is missing some sort of keys, you can easily see system texts in the UI. They are usually in all upper-case snail format - `GAME_TITLE`, `SETTINGS_MENU`, `TUTORIAL_TITLE`. This helps in the cases where your proposed format contains some sort of spelling mistakes which leads to rewriting all the translation files to take into acount the newly generated key. 

I did test out this functionality on my sample project using both formats and the only problem  that I saw was large amount of warnings because of the missing translations.

Regarding the method names I would suggest to get rid of the `LoadJsonFile` method. `LoadMultipleLanguageJSON` and `LoadSingleLanguageJSON` should be sufficient enough to handle everything.

rokups:
This is how i expected to use the system, yes. I guess i abused it a little by inserting english phrases instead of system text to have default language translation. Anyhow im just throwing this idea out, should anyone find it useful... :)

> I did test out this functionality on my sample project using both formats and the only problem that I saw was large amount of warnings because of the missing translations.

Did you test using old format with a file that contains multiple languages? I suspect not.

> Regarding the method names I would suggest to get rid of the `LoadJsonFile` method. `LoadMultipleLanguageJSON` and `LoadSingleLanguageJSON` should be sufficient enough to handle everything.

`LoadJsonFile` is actually convenient. Maybe having overloads for `LoadMultipleLanguageJSON` and `LoadSingleLanguageJSON` that take a string resource name would be a good idea.

And of course `LoadMultipleLanguageJSON` and `LoadSingleLanguageJSON` need to be renamed. They both are loading different formats that can contain translations of multiple languages after all.

ArnisLielturks:
@rokups I used both formats for testing. And to make it even more interesting I removed some of the keys in both files. Everything worked just fine.

`LoadSingleLanguageJSON` only loads 1 language, all the translations in the file belongs to 1 language that user has specified when loading the file.

rokups:
> `LoadSingleLanguageJSON` only loads 1 language, all the translations in the file belongs to 1 language that user has specified when loading the file.

This is the issue. That format may contain multiple languages. Why should we load each language from that same file through separate calls if we can load them all at once? That was the old behavior.

ArnisLielturks:
@rokups Not sure what you meant by that. My intention was that the old format could be used via `LoadMultipleLanguageJSON` without specifying language as a parameter, it will be picked up automatically when parsing the file. `LoadSingleLanguageJSON` is the method that will handle the new format and it requires the language parameter to know for which language translations needs to be loaded.

rokups:
ohh damn. looks like i got confused. ill take another look and let you know if anything is actually broken. sorry :|

ArnisLielturks:
@rokups No problem :)

rokups:
I verified that everything works perfectly. Yes this confusion was a spectacular fail on my part.

I still think `LoadJsonFile` is confusing due to reasons already stated. But it is convenient as well. Maybe it should be replaced with overloads of `LoadMultipleLanguageJSON` and `LoadSingleLanguageJSON` that take resource name as parameter? That way we would always know exactly what code is doing, have a benefit of ability to load data from middle of other json data structure or convenience of using resource name to load translations file.

PR #2365, opened 2018-08-21
Log which adapter used for DX11

SuperWangKai:
Useful when we have multiple GPUs.


Source/Urho3D/Graphics/Direct3D11/D3D11Graphics.cpp
```diff
@@ -2168,6 +2168,11 @@ bool Graphics::CreateDevice(int width, int height, int multiSample)
     // (the application will switch manually if it wants to)
     dxgiFactory->MakeWindowAssociation(GetWindowHandle(window_), DXGI_MWA_NO_ALT_ENTER);
 
+	DXGI_ADAPTER_DESC desc;
+	dxgiAdapter->GetDesc(&desc);
+	String adapterDesc(desc.Description);
```
SuperWangKai: WCHAR -> utf8

weitjong:
Thanks for your PR. Since the piece of information is not readily available and need extra steps to prepare, the preparation steps would waste the CPU time if at the end the logging is not enabled or that the effective level is higher than INFO. Could you wrap the preparation steps inside an if statement?

weitjong:
Looks great. Thanks.

PR #2370, opened 2018-08-27
Log which adapter is used on OpenGL & D3D9

Florastamine:
PR #2365 only deals with Direct3D 11 so here's another one which also covers OpenGL.

### DIFF
--- a/Source/Urho3D/Graphics/OpenGL/OGLGraphics.cpp
+++ b/Source/Urho3D/Graphics/OpenGL/OGLGraphics.cpp
@@ -500,8 +500,7 @@ bool Graphics::SetMode(int width, int height, bool fullscreen, bool borderless, 
     CheckFeatureSupport();
 
 #ifdef URHO3D_LOGGING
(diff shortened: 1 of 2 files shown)


Source/Urho3D/Graphics/OpenGL/OGLGraphics.cpp
```diff
@@ -500,6 +500,9 @@ bool Graphics::SetMode(int width, int height, bool fullscreen, bool borderless,
     CheckFeatureSupport();
 
 #ifdef URHO3D_LOGGING
+    String adapterDesc = String((const char *) (glGetString(GL_VENDOR))) + " " + (const char *) (glGetString(GL_RENDERER));
+    URHO3D_LOGINFO("Adapter used " + adapterDesc);
```
weitjong: May be better to use `URHO3D_LOGINFOF()` with "Adapter used %s %s" as the formatter and the two `glGetString()` as paramaters.

Florastamine:
Done. Also I took the opportunity to also provided support for D3D9.

SuperWangKai:
Nice work!

PR #2380, opened 2018-10-01
incoming/outgoing packet counter, last heard time logic added

ArnisLielturks:


Source/Urho3D/Network/Connection.cpp
```diff
@@ -288,15 +290,23 @@ void Connection::SendRemoteEvents()
...
+            GetPacketsInPerSec(),
+            GetPacketsOutPerSec(),
+            GetBytesInPerSec(),
+            GetBytesOutPerSec(),
+            GetLastHeardTime());
         URHO3D_LOGINFO(statsBuffer);
     }
 #endif
 
+    if (packetCounterTimer_.GetMSec(false) > 1000)
```
weitjong: I feel strange to find the logic to be inside the `SendRemoteEvents()`. What if there is no remote events being sent for more than a second?

Source/Urho3D/Network/Connection.cpp
```diff
@@ -1010,10 +1023,10 @@ float Connection::GetRoundTripTime() const
     return 0.0f;
 }
 
-float Connection::GetLastHeardTime() const
+unsigned Connection::GetLastHeardTime() const
 {
-    //TODO
-    return 0.0f;
+    Timer t = lastHeardTimer_;
```
weitjong: Don't think a copy is needed here.

Source/Urho3D/Network/Connection.h
```diff
@@ -200,8 +200,8 @@ class URHO3D_API Connection : public Object
     /// Return the connection's round trip time in milliseconds.
     float GetRoundTripTime() const;
 
-    /// Return the time since last received data from the remote host in milliseconds.
-    float GetLastHeardTime() const;
+    /// Return the time since last received data from the remote host in miliseconds.
```
weitjong: Should be `milliseconds`

Source/Urho3D/Network/Connection.cpp
ArnisLielturks: SendRemoteEvents is always called on network update, the next few lines after this actually check if there are remote events that needs to be sent out, if not, the method exits.

Source/Urho3D/Network/Connection.cpp
ArnisLielturks: Otherwise compiler threw errors when the const method was trying to call `lastHeardTimer_.GetMsec(false)` method.

ArnisLielturks: Possible solution to this might be creating const method on Timer class which would return read only timer value in milliseconds.

weitjong:
I see. Does `const_cast` work here?

ArnisLielturks:
> I see. Does `const_cast` work here? Sent from my iPhone7+

This works `return const_cast<Connection*>(this)->lastHeardTimer_.GetMSec(false);`
But I would say that non-const `GetLastHeardTime` method is a way to go, since by the `const_cast` we basically say that we will ignore the `const` method anyway.

Source/Urho3D/AngelScript/NetworkAPI.cpp
```diff
@@ -74,11 +74,11 @@ static void RegisterConnection(asIScriptEngine* engine)
     engine->RegisterObjectMethod("Connection", "String get_address() const", asMETHOD(Connection, GetAddress), asCALL_THISCALL);
     engine->RegisterObjectMethod("Connection", "uint16 get_port() const", asMETHOD(Connection, GetPort), asCALL_THISCALL);
     engine->RegisterObjectMethod("Connection", "float get_roundTripTime() const", asMETHOD(Connection, GetRoundTripTime), asCALL_THISCALL);
-    engine->RegisterObjectMethod("Connection", "float get_lastHeardTime()", asMETHOD(Connection, GetLastHeardTime), asCALL_THISCALL);
+    engine->RegisterObjectMethod("Connection", "unsigned get_lastHeardTime()", asMETHOD(Connection, GetLastHeardTime), asCALL_THISCALL);
```
weitjong: I believe it should be "uint" here.

weitjong:
> This works return const_cast<Connection*>(this)->lastHeardTimer_.GetMSec(false);
But I would say that non-const GetLastHeardTime method is a way to go, since by the const_cast we basically say that we will ignore the const method anyway.

I am thinking along the line of:

```
unsigned Connection::GetLastHeardTime() const { return const_cast<Timer&>(statsTimer_).GetMSec(false); }
```

PR #2381, opened 2018-10-14
Fixing typo on NavigationMesh API

mvendra:
Correcting a typo on Source/Urho3D/Navigation/NavigationMesh.h 's API. This patch also corrects all usages inside of the urho repo, and it builds. Might break client code for some users, so buyer's beware.


weitjong:
Thanks!

PR #2393, opened 2018-11-24
Added new EmitterTypes

Capital-Asterisk:
* Added SphereVolume, Cylinder, and Ring enum with calculations
 * Modified EditorParticle to incorporate new emitter shapes

The options for emitter shapes is quite small, and I thought I should add to it. Sphere volume emits particles in a random point inside a sphere instead of the surface. Cylinder is a random point in a cylinder, which can be flattened to make a circle that emits in evenly distributed points. Ring emits particles on the edges of a circle, but becomes a tube shape when sized vertically (image below).

image


weitjong:
Thanks for your PR! It is now merged.

PR #2400, opened 2018-12-06
P2P multiplayer support with samples

ArnisLielturks:
I would like to get some feedback on the implementation as I think that the code quality can be improved but not sure what is the best way.


ArnisLielturks:
One more thing, `P2PMultiplayerExtended` sample will be removed in the nearest future and placed in a seperate repository since it's quite big and involves external systems to be set up.

weitjong:
Thanks for your PR. I will try to review it over this weekend. I hope Eugene or others have time to do so too.

Modanung:
@ArnisLielturks Could you formulate your doubts about the current form of the implementation?

ArnisLielturks:
@Modanung Oh boy, where do I begin...

1. Bacwkards comatiblity
When doing the implementation I always kept in mind that the current networking mode - server-client should be untouched and should remain the same. I think I succeeded with that by introducing mode switch for the networking - classic server-client and p2p mode. As I remember I also implemented safety checks to disable network connections with different modes, not sure if it works in 100% of the cases.

2. Replication
Replication is something that becomes increasingly difficult for the p2p mode so that is something that I didn't try to fix or update. https://github.com/ArnisLielturks/Urho3D-P2P-Multiplayer sample uses the old replication mechanism without any modifications, though I believe the approach there is a bit rough and all players might get out of sync pretty quickly. But the good news is that I don't think that the engine itself should deal with that at the moment since it requires heavy modification of the current replication model.

3. Host election
One thing that I noticed when doing the implementation is that if the new peer tries to connect to different peer which is not a host, they would form new p2p session which would lead to unpredictable behaviour. Maybe I was doing something nasty there and didn't enable some sort of switch to disable that, but it seemed like it was a bug in the networking library itself. What can you do to avoid that? When implementing p2p always make sure that new peers are connecting to the host, otherwise you're 🦆 ed. You're also 🦆 ed if there's a host migration going on while you connect. 

In conclusion I believe that the current implementation is quite good. It requires additional servers to be set up but that's what you get when doing p2p application.

IlyaStuurFedorov:
Are there any updates on this?

PR #2402, opened 2018-12-11
Move refcounts into objects

1vanK:


1vanK:
Currently is allowed using WeakPtr for raw pointers when refs_ == 0 (no any SharedPtr) and

```
void RefCounted::AddWeakRef()
{
    assert(refs_ > 0);
    weakRefs_++;
}
```

cause assert when

```
    VariantMap& eventData = GetEventDataMap();
    eventData[P_ELEMENT] = this;
```
↓
```
    /// Assign from a RefCounted pointer. The object will be stored internally in a WeakPtr so that its expiration can be detected safely.
    Variant& operator =(RefCounted* rhs)
    {
        SetType(VAR_PTR);
        value_.weakPtr_ = rhs;
        return *this;
    }
```
↓
```
    /// Assign from a raw pointer.
    WeakPtr<T>& operator =(T* ptr)
    {
        if (ptr_ == ptr)
            return *this;

        ReleaseRef();
        ptr_ = ptr;
        AddRef();

        return *this;
    }
```
so with this method we have problem when raw pointer deleted before WekPtr

PR #2406, opened 2018-12-15
improve FPS for StaticModelGroup 

ab4daa:
When running application like sample 20_HugeObjectCount(toggle object moving and object group),
the bottleneck of URHO3D_PROFILE(ReinsertToOctree) comes from StaticModelGroup::OnWorldBoundingBoxUpdate() in Drawable::GetWorldBoundingBox().

By making it update multi-threaded, I have 20 FPS gain(113->133) for sample 20_HugeObjectCount(toggle object moving and object group).

### DIFF
--- a/Source/Urho3D/Graphics/Octree.cpp
+++ b/Source/Urho3D/Graphics/Octree.cpp
@@ -63,6 +63,20 @@ void UpdateDrawablesWork(const WorkItem* item, unsigned threadIndex)
     }
 }
 
+void UpdateDrawablesBBWork(const WorkItem* item, unsigned threadIndex)
+{
+    Drawable** start = reinterpret_cast<Drawable**>(item->start_);
+    Drawable** end = reinterpret_cast<Drawable**>(item->end_);
+
+    while (start != end)
+    {
+        Drawable* drawable = *start;
+        if (drawable)
+            drawable->GetWorldBoundingBox();
+        ++start;
+    }
+}
+
 inline bool CompareRayQueryResults(const RayQueryResult& lhs, const RayQueryResult& rhs)
 {
     return lhs.distance_ < rhs.distance_;
@@ -442,6 +456,30 @@ void Octree::Update(const FrameInfo& frame)
     if (!drawableUpdates_.Empty())
     {
         URHO3D_PROFILE(ReinsertToOctree);
+        
+        WorkQueue* queue = GetSubsystem<WorkQueue>();
+        int numWorkItems = queue->GetNumThreads() + 1; // Worker threads + main thread
+        int drawablesPerItem = Max((int)(drawableUpdates_.Size() / numWorkItems), 1);
+        PODVector<Drawable*>::Iterator start = drawableUpdates_.Begin();
+        // Create a work item for each thread
+        for (int i = 0; i < numWorkItems; ++i)
+        {
+            SharedPtr<WorkItem> item = queue->GetFreeItem();
+            item->priority_ = M_MAX_UNSIGNED;
+            item->workFunction_ = UpdateDrawablesBBWork;
+            item->aux_ = NULL;
+
+            PODVector<Drawable*>::Iterator end = drawableUpdates_.End();
+            if (i < numWorkItems - 1 && end - start > drawablesPerItem)
+                end = start + drawablesPerItem;
+
+            item->start_ = &(*start);
+            item->end_ = &(*end);
+            queue->AddWorkItem(item);
+
+            start = end;
+        }
+        queue->Complete(M_MAX_UNSIGNED);
 
         for (PODVector<Drawable*>::Iterator i = drawableUpdates_.Begin(); i != drawableUpdates_.End(); ++i)
         {


weitjong:
So far I don't see any issue, except perhaps I would update the Doxygen comment on `Drawable::OnWorldBoundingBoxUpdate()` to indicate that "it may be called from a worker thread".

eugeneko:
Unfortunatelly, this change introduce even more races than we already have.
You are basically do multithreading for this call:
https://github.com/urho3d/Urho3D/blob/17c4a1022d3342b6318fa095661a0ecba6306284/Source/Urho3D/Graphics/StaticModel.cpp#L402
But `GetWorldTransform` is unsafe to call in worker thread if Node is dirty.
We already have some bugs with nested `Drawable`s, and I failed to find nice solution that doesn't have potential performace penalty.
Basically, we have to undirt the nodes in main thread before scheduling MT jobs for updates, and I'm afraid we have to do it more than once.

ab4daa:
Hi, Eugene

If I implement the following (ugly) code in Node.cpp, then call `scene->parallel_update_children_transform();` before the PR code to update each node top-down.
Would it be OK?(this freak still got me about +10 FPS for sample 20)
I understand the loading of each thread would be unbalance, but in application like sample 20, it should be OK.

My application is very similar with sample 20, so I don't want to give up the chance for better performance.
Thanks for your review!


```
void Node::UpdateChildrenWorldTransform() 
{
    Vector<SharedPtr<Node> >::Iterator it;

    UpdateWorldTransform();
    for (it = children_.Begin(); it != children_.End(); ++it)
    {
        (*it)->UpdateChildrenWorldTransform();
    }
}
```
```
static void UpdateNodesWorldTransformWork(const WorkItem* item, unsigned threadIndex)
{
    SharedPtr<Node>* start = reinterpret_cast<SharedPtr<Node>*>(item->start_);
    SharedPtr<Node>* end = reinterpret_cast<SharedPtr<Node>*>(item->end_);

    while (start != end)
    {
        Node* node = *start;
        if (node)
            node->UpdateChildrenWorldTransform();
        ++start;
    }
}
```
```
void Node::parallel_update_children_transform()
{
    WorkQueue* queue = GetSubsystem<WorkQueue>();
    int numWorkItems = queue->GetNumThreads() + 1; // Worker threads + main thread
    int nodesPerItem = Max((int)(children_.Size() / numWorkItems), 1);
    Vector<SharedPtr<Node> >::Iterator start = children_.Begin();
    // Create a work item for each thread
    for (int i = 0; i < numWorkItems; ++i)
    {
        SharedPtr<WorkItem> item = queue->GetFreeItem();
        item->priority_ = M_MAX_UNSIGNED;
        item->workFunction_ = UpdateNodesWorldTransformWork;
        item->aux_ = NULL;

        Vector<SharedPtr<Node> >::Iterator end = children_.End();
        if (i < numWorkItems - 1 && end - start > nodesPerItem)
            end = start + nodesPerItem;

        item->start_ = &(*start);
        item->end_ = &(*end);
        queue->AddWorkItem(item);

        start = end;
    }
    queue->Complete(M_MAX_UNSIGNED);
}
```

ab4daa:
On second thought, this should be done in app side, not in the generic part of engine.
In app I have knowledge of parent nodes of instances, so I can undirty parent nodes first, then parallel update world bounding box of static model group safely.

weitjong:
I just push another commit on top of yours to avoid the race condition pointed out by Eugene. Could you check if this general fix is still useful for you. Thanks.

ab4daa:
Thanks weitjong.
But there is bottleneck in StaticModelGroup::OnWorldBoundingBoxUpdate() because the StaticModelGroup is exactly dirty.
I think skipping dirty StaticModelGroup does not make sense.

Also the optimization only works for StaticModelGroup(as far as I know), and the dependancy of dirty parent node, maybe it should not be put in the generic code.

I personally think the more sanity way to do it is manually doing the following after setting all new transform of nodes(maybe in the end of app update event handler):
  1. undirty parent nodes of StaticModelGroup instance nodes by Node::GetWorldTransform()
  2. parallel StaticModelGroup::GetWorldBoundingBox() like PR code

If you agree, maybe I could try modifying sample 20 a bit to introduce ..the trick?
like: 
1. add some parent nodes for StaticModelGroup
2. add pressing B to toggle parallel update WorldBoundingBox

ab4daa:
I thought the race condition comes from common parent nodes of instance nodes of StaticModelGroup?

If I am wrong please point out:
eg. instance1 of StaticModelGroup1 and instance2 of StaticModelGroup2 have common parent node1, 
when calling Node::GetWorldTransform() in StaticModelGroup::OnWorldBoundingBoxUpdate(), 
both StaticModelGroup1 and StaticModelGroup2  will check dirty/undirty common parent node1 which cause race condition.

Skipping dirty node of StaticModelGroup will not help if I am not wrong

ab4daa:
Thanks for your explaination, weitjong.
My application tends to dirty most of nodes, I guess  I should do it in app side.

PR #2410, opened 2019-01-11
Fix missing bit shift in BatchQueue::SortFrontToBack2Pass(). Thanks t…

cadaver:
Fixes a long-standing bug which would cause all material ID's to be zero in the sorting, making it ineffective in regard to different materials. See https://discourse.urho3d.io/t/the-doubt-in-batchqueue-sortfronttoback2pass/4807


weitjong:
Thanks cadaver. I am out of town right now so just merging this without double checking.

PR #2411, opened 2019-01-12
Make Audio system pause sound thread safe.

Y-way:
Make Audio system pause sound thread safe.


weitjong:
Thanks, it is merged now. I just cherry-picked your last commit.

Y-way:
> Thanks, it is merged now. I just cherry-picked your last commit.

All right.

Issue #2352, opened 2018-07-23
Parent ordering issues in network updates


etherealvisage:
Hi,

There appears to be a bug in the server/client state replication. Nodes marked as 'dirty' are put into a [HashSet defined here](https://github.com/urho3d/Urho3D/blob/b0f2b5a94f567465bfb1f88427e5e2924552a2bb/Source/Urho3D/Network/Connection.h#L288) by [this line](https://github.com/urho3d/Urho3D/blob/b0f2b5a94f567465bfb1f88427e5e2924552a2bb/Source/Urho3D/Network/Connection.cpp#L256). They are then pulled out and processed in whatever order the `HashSet` thinks is appropriate.

The problem with this is that if you have a multi-level hierarchy of `Node`s created in a single network update, and a child gets processed before a parent in this ordering, then it will fail to be added into the scene [on the client side](https://github.com/urho3d/Urho3D/blob/b0f2b5a94f567465bfb1f88427e5e2924552a2bb/Source/Urho3D/Scene/Node.cpp#L1486).

If the child's parent attribute is never marked as dirty at a later time, it will never be inserted into the hierarchy and thus be silently dropped.

Either adding a note to the documentation that only one level of a node hierarchy can be created at a time, or changing this to use some kind of topological ordering would be great. :-)

Thanks!

\- ethereal

ArnisLielturks:
Can we simply do a `nodesToProcess_.Sort();` to sort out the HashSet before processing it? All the children will have larger Node/Component ID than it's parent so this issue should be fixed.

etherealvisage:
Unfortunately not all the children will have a larger ID, since node IDs can wrap (see `Scene::GetFreeNodeID`). That just makes the bug significantly subtler. :-)

LeithKetchell:
I've just patched this issue - however when I forked the project on github and uploaded my changed files, and was just about to issue a Pull Request, I noticed that my local copy of the project is horribly out of date with respect to the main branch on github - we're now using slike networking instead of knet - this means that the download links on the urho3D homepage are pointing at very old archives, and that nobody is curating the issues with respect to wholesale changes in the codebase.

Just incase this issue somehow becomes relevant again, here are the patched files with respect to Issue #2353
[Patched.zip](https://github.com/urho3d/Urho3D/files/2753925/Patched.zip)

eugeneko:
[...]

PR #2412, opened 2019-01-15
Files patched for Issue# 2352 (unordered network replication of node hierarchies)

LeithKetchell:
Connection class was gathering dirty node IDs in an unordered container (HashMap), which causes network Clients to process newly-replicated nodes in an arbitrary order.
Resolved by replacing the HashMap with a Vector (an ordered container).

Please disregard the initial commit as my project files were not up to date.

### DIFF
--- a/Source/Urho3D/Network/Connection.cpp
+++ b/Source/Urho3D/Network/Connection.cpp
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2008-2017 the Urho3D project.
+// Copyright (c) 2008-2019 the Urho3D project.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -38,9 +38,16 @@
 #include "../Scene/SceneEvents.h"
 #include "../Scene/SmoothedTransform.h"
 
-#include <kNet/kNet.h>
+#include <SLikeNet/peerinterface.h>
+#include <SLikeNet/statistics.h>
+
+#ifdef SendMessage
+#undef SendMessage
+#endif
 
 #include "../DebugNew.h"
+
+#include <cstdio>
 
 namespace Urho3D
 {
@@ -60,30 +67,29 @@ PackageUpload::PackageUpload() :
 {
 }
 
-Connection::Connection(Context* context, bool isClient, kNet::SharedPtr<kNet::MessageConnection> connection) :
+Connection::Connection(Context* context, bool isClient, const SLNet::AddressOrGUID& address, SLNet::RakPeerInterface* peer) :
     Object(context),
     timeStamp_(0),
-    connection_(connection),
+    peer_(peer),
     sendMode_(OPSM_NONE),
     isClient_(isClient),
     connectPending_(false),
     sceneLoaded_(false),
-    logStatistics_(false)
+    logStatistics_(false),
+    address_(nullptr)
 {
     sceneState_.connection_ = this;
-
-    // Store address and port now for accurate logging (kNet may already have destroyed the socket on disconnection,
-    // in which case we would log a zero address:port on disconnect)
-    kNet::EndPoint endPoint = connection_->RemoteEndPoint();
-    ///\todo Not IPv6-capable.
-    address_ = Urho3D::ToString("%d.%d.%d.%d", endPoint.ip[0], endPoint.ip[1], endPoint.ip[2], endPoint.ip[3]);
-    port_ = endPoint.port;
+    port_ = address.systemAddress.GetPort();
+    SetAddressOrGUID(address);
 }
 
 Connection::~Connection()
 {
     // Reset scene (remove possible owner references), as this connection is about to be destroyed
-    SetScene(0);
+    SetScene(nullptr);
+
+    delete address_;
+    address_ = nullptr;
 }
 
 void Connection::SendMessage(int msgID, bool reliable, bool inOrder, const VectorBuffer& msg, unsigned contentID)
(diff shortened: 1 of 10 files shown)


eugeneko:
The idea of using Vector isn't actually bad.
However, I'm unsire if it's working properly in current implementation.

https://github.com/urho3d/Urho3D/pull/2412/files#diff-7b9673cfc396212ff55f1c1bee79ab6eL1159

What if Scene::MarkReplicationDirty called twice for the same node? Is it possible?

https://github.com/urho3d/Urho3D/pull/2412/files#diff-55761705d5f88d7b0ec255a02b3d3ae1L1792

It looks similar, but in this line double-adding seems impossible. Is it true? Or node can appear in the dirty list without marking dirty?

https://github.com/urho3d/Urho3D/pull/2412/files#diff-b12ad27bb5ee6ae88c0b4c3766ead908L1136

Is it really ok to have quadratic node processing complexity?
Does this function really have to be called with `unsigned nodeID` parameter instead of index?

LeithKetchell:
Hey, let's talk about it :)

I am assuming a few things, so let me start with my assumptions.

Assuming that we're single-threaded... by the time we process dirty nodes, they can't be re-marked as dirty - that part of the processing has been completed, and so they (should not?) appear twice in the list of dirty nodes... The changes I propose occur late in the piece, where we have our list of dirty nodes, presumably collected in a top-down fashion during evaluation of the scene node hierarchy (so parent nodes should appear prior to child nodes, irrespective of the possibly-recycled IDs).
I may be barking up the wrong tree, which is why this patch really needs to be tested, in the use-case that was proposed by the OP of #2352, which was the serialization of a node hierarchy whose IDs are not representative of the order of the node hierarchy... In other words, just create a few nodes, connect them in a different order than you created them, and see if they replicate correctly.
We know there is an issue for the master branch, let's see if my patch is enough to fix it.
If not, we're a step closer to a full patch, and can stare down the root cause.

ArnisLielturks:
I want these changes to be added to the engine as I have experienced the problems caused without them. But has anyone actually tried to test them with the proposed scenario? https://github.com/urho3d/Urho3D/issues/2352

Source/Urho3D/Network/Connection.cpp
```diff
@@ -146,7 +146,7 @@ void Connection::SendRemoteEvent(Node* node, StringHash eventType, bool inOrder,
         URHO3D_LOGERROR("Sender node is not in the connection's scene, can not send remote node event");
         return;
     }
-    if (!node->IsReplicated())
+    if (node->GetID() >= FIRST_LOCAL_ID)
```
PredatorMF: Can this line be commented or explained ?

Source/Urho3D/Network/Connection.cpp
```diff
@@ -335,16 +325,16 @@ void Connection::SendRemoteEvents()
 
 void Connection::SendPackages()
 {
-    while (!uploads_.Empty())
+    while (!uploads_.Empty() && connection_->NumOutboundMessagesPending() < 1000)
```
PredatorMF: 1000 is a bit magical, what is the reason behind this number ?

Source/Urho3D/Network/Connection.cpp
```diff
@@ -335,16 +325,16 @@ void Connection::SendRemoteEvents()
...
         unsigned char buffer[PACKAGE_FRAGMENT_SIZE];
 
         for (HashMap<StringHash, PackageUpload>::Iterator i = uploads_.Begin(); i != uploads_.End();)
         {
             HashMap<StringHash, PackageUpload>::Iterator current = i++;
             PackageUpload& upload = current->second_;
-            auto fragmentSize =
-                (unsigned)Min((int)(upload.file_->GetSize() - upload.file_->GetPosition()), (int)PACKAGE_FRAGMENT_SIZE);
+            unsigned fragmentSize =
+                Min((upload.file_->GetSize() - upload.file_->GetPosition()), PACKAGE_FRAGMENT_SIZE);
```
PredatorMF: The auto was fine.

Source/Urho3D/Network/Connection.cpp
```diff
@@ -491,7 +470,7 @@ void Connection::ProcessLoadScene(int msgID, MemoryBuffer& msg)
 
     // In case we have joined other scenes in this session, remove first all downloaded package files from the resource system
     // to prevent resource conflicts
-    auto* cache = GetSubsystem<ResourceCache>();
+    ResourceCache* cache = GetSubsystem<ResourceCache>();
```
PredatorMF: Can't this be auto, C++11 is already a requirement and most of the code tries to follow.

Source/Urho3D/Network/Connection.cpp
```diff
@@ -557,7 +536,7 @@ void Connection::ProcessSceneUpdate(int msgID, MemoryBuffer& msg)
 
             // Read initial attributes, then snap the motion smoothing immediately to the end
             node->ReadDeltaUpdate(msg);
-            auto* transform = node->GetComponent<SmoothedTransform>();
+            SmoothedTransform* transform = node->GetComponent<SmoothedTransform>();
```
PredatorMF: Please, auto.

Source/Urho3D/Network/Connection.cpp
```diff
@@ -1245,7 +1218,7 @@ void Connection::ProcessExistingNode(Node* node, NodeReplicationState& nodeState
 
     // Check from the interest management component, if exists, whether should update
     /// \todo Searching for the component is a potential CPU hotspot. It should be cached
-    auto* priority = node->GetComponent<NetworkPriority>();
+    NetworkPriority* priority = node->GetComponent<NetworkPriority>();
```
PredatorMF: auto priority...

Source/Urho3D/Network/Connection.h
```diff
@@ -201,7 +195,7 @@ class URHO3D_API Connection : public Object
     float GetRoundTripTime() const;
 
     /// Return the time since last received data from the remote host in milliseconds.
-    unsigned GetLastHeardTime() const;
+    float GetLastHeardTime() const;
```
PredatorMF: float is not cool for a timestamp or time deltas

eugeneko:
@PredatorMF you are commenting outdated code

PredatorMF:
Its possible, hard to navigate through my phone, I'll take those comments down.

eugeneko:
@SirNate0 Linear search in list would be about 100 times slower than in vector, so I don't really think list is better in this case. List is good in only one case -- when you almost don't work with container as a whole (no loops) but only with individual elements.

Maybe we can do some specialized container combination...

PR #2415, opened 2019-01-25
HTTPS support - civetweb update

ArnisLielturks:
HTTPS support can be built using 2 build options:

`URHO3D_SSL=1` enables HTTPS support using OpenSSL library, by default this option should be enough on unix systems, since OpenSSL is included by default

`./script/cmake_generic.sh -DURHO3D_SSL=1`

`URHO3D_SSL_DYNAMIC=1` can be used to load SSL support libraries dynamically. This will be enabled only if `URHO3D_SSL=1` build option is present 
`script\cmake_vs2017.bat -DURHO3D_SSL=1 -DURHO3D_SSL_DYNAMIC=1`

I don't have a OSX system, so I would appreciate if someone could test this for me.


weitjong:
Thanks for the PR. I see whether I can test it on MacOX later.

weitjong:
I think it still needs more work. I tried both URHO3D_SSL and URHO3D_SSL+URHO3D_SSL_DYNAMIC on my Fedora first, and both got same error.

```
truncating vsnprintf buffer: [,]
[Sat Jan 26 15:13:24 2019] INFO: Initializing SSL
[Sat Jan 26 15:13:24 2019] ERROR: SSL Initialization failed!
[Sat Jan 26 15:13:24 2019] ERROR: HttpRequest error: SSL is not initialized
```

Also, never see the "truncation" message in the log before. I suppose it is something new too in this branch.

Do you plan to make more changes? Do you mind if I make changes into it?

ArnisLielturks:
@weitjong I would appreciate if you could help with your contributions. Otherwise it would take me ages to cover all the platforms and cases.

ArnisLielturks:
@weitjong I actually did not, I used latest commit from their master branch which is not what I initally planned.

weitjong:
I am starting to work on merging the upstream release 1.11 and also incorporating some of the upstream build options into our build. The "CIVETWEB_ENABLE_SSL" will be one of them.

ArnisLielturks:
@weitjong By utilizing the same names for the parameters? It seemed that most of those options are only used in the CMake file and translated to something else when doing the actual build, that's why I replaced that with "URHO3D_" prefixed ones.

weitjong:
At the moment I think it’s better to keep those build options using their own prefixes (or namespace group in cmake-gui), as long as those associated defines are all locally scoped in Civetweb lib. I also believe that http and https are not mutually exclusive. I.e. the new build option would just make https protocol support available or not, without affecting the support for http protocol.

weitjong:
Didn’t have time to work on this for the past few days and the coming week I will be out of town. So, feel free to add more commits to the branch in the meantime.

ArnisLielturks:
@weitjong Ok. Will do.

ArnisLielturks:
@weitjong Got the same problem as you when I built it with fedora docker image and ran the sample in the headless mode. Will try to work out where the problem is comming from.

ArnisLielturks:
Openssl issue related to this: https://github.com/civetweb/civetweb/issues/641

weitjong:
@ArnisLielturks I am back now. What is the current status with the latest commit?

ArnisLielturks:
@weitjong I was able to fix the fedora issue, at least on the docker image that I used. You can check that out (using URHO3D_SSL=1 option). Also I updated the code to use the 1.11 version from the civetweb, but I left the CMake file as it was in my initial implementation.

ArnisLielturks:
Can someone help to test this with OSX systems? It seems that it's the last platform that needs additional testing.

ArnisLielturks:
@weitjong Finally was able to fix the OSX issue, it was related to missing openssl dependencies. That means that all of the platforms are now covered which were tested.

OpenSSL is no longer optimal dependency, since HTTP is deprecated, HTTPS is a must so openssl is needed for the engine but it seems that our dockerized builds doesn't have it installed at the moment. I could help with that but I'm pretty sure that for some platforms I will need an assistance. I see that you have https://github.com/weitjong/dockerized repository, that means I could provide PR's for that in the upcomming weeks.

weitjong:
Well done! You are welcome to contribute to send PR there.

Source/Samples/43_HttpRequestDemo/HttpRequestDemo.cpp
```diff
@@ -109,10 +114,16 @@ void HttpRequestDemo::HandleUpdate(StringHash eventType, VariantMap& eventData)
...
 
+#ifdef URHO3D_SSL
+                JSONValue val = json->GetRoot().Get("ip");
+#else
                 JSONValue val = json->GetRoot().Get("origin");
+#endif
 
-                if (val.IsNull())
-                    text_->SetText("Invalid string.");
+                if (val.IsNull()) {
```
eugeneko: Bracket style

Source/Urho3D/Network/HttpRequest.cpp
```diff
@@ -94,6 +94,8 @@ void HttpRequest::ThreadFunction()
     {
         port = ToInt(host.Substring(portStart + 1));
         host = host.Substring(0, portStart);
+    } else if (protocol.Compare("https", false) >= 0) {
```
eugeneko: Bracket style too

Source/Urho3D/Network/HttpRequest.cpp
```diff
@@ -109,19 +111,33 @@ void HttpRequest::ThreadFunction()
     }
 
     // Initiate the connection. This may block due to DNS query
-    /// \todo SSL mode will not actually work unless Civetweb's SSL mode is initialized with an external SSL DLL
     mg_connection* connection = nullptr;
+
+#if defined(URHO3D_SSL) || defined(URHO3D_SSL_DYNAMIC)
+    if (mg_check_feature(MG_FEATURES_SSL) & MG_FEATURES_SSL) {
```
eugeneko: same, also tabs are misaligned somewhere

ArnisLielturks:
@weitjong enabled HTTPS on the native dockerized build but stumbled upon another problem regarding linking static OpenSSL library. Is there a possibility to make the static openssl part of Urho's library? Is there another library which are linked the same way? Otherwise other projects are required to link openssl themselves when using urho with HTTPS support which is not a great solution.

weitjong:
Statically linking OpenSSL is possible, but I don't think we have ever done that before for other underlying libs (aside from 3rd-party libs that we build ourselves), and also I don't think it should be forced upon to Urho3D library user, should you still decide to go this approach (which means you probably need another "URHO3D_SSL_STATIC" build option to control it). This library is about security, so dynamic linking has the advantage IMHO as they get patched faster by system owners when a newer/fixed version becomes available; if owners don't patch their systems then it also clear it is not Urho3D fault when their system get hacked.

There is way to declare transitive dependency in the modern CMake 3.x. But unfortunately our CMake build script is largely still based on old CMake 2.x. Currently we workaround this kind of dependency problem by using `define_dependency_lib` macro in `UrhoCommon.cmake`. This macro has been earmarked for deprecation, but it will not be removed until all the CMake build scripts are modernized. So, it should be OK for you to modify the macro to take care of your need for now.

ArnisLielturks:
@weitjong that did the trick!

ArnisLielturks:
@weitjong I believe you can retag the latest docker image. Also I believe this PR is done if no one has any objections. Additional documentation will be added by me either in Github's wiki or in the forums.

ArnisLielturks:
@weitjong Can we merge this in it's current state?

weitjong:
I can help to retag the DBE this weekend. Then you can decide when to merge freely.

weitjong:
I have re-tagged the DBE images. Cheers.

PR #2420, opened 2019-02-21
Fix skipped visible elements due to UI layout optimization.

ssinai1:
Fix: UI layout optimization does not take into account layout spaceing.


weitjong:
Thanks for the PR.

PR #2422, opened 2019-02-26
Fixed resource reloading failure caused by load resource as `XMLFile`…

SuperWangKai:
`ResourceCache::GetResource` will keep the resource type the first time it is called.

If we once called 
```angelscript
XMLFile@ xml = cache.GetResource("XMLFile", path);
``` 
or 
```c++
XMLFile* xml = cache->GetResource<XMLFile>(path);
```

even though the xml file is a `Material` e.g., the type will always be `XMLFile`. This causes resource reloading failed since only `XMLFile`'s `BeginLoad` or `EndLoad` will be called, instead of `Material`'s.

However, currently, in the editor, this code path is not called (`useCache` is a false), it is still a bug. And this `ResourceCache` characteristic should be taken seriously.

### DIFF
--- a/bin/Data/Scripts/Editor/EditorResourceBrowser.as
+++ b/bin/Data/Scripts/Editor/EditorResourceBrowser.as
@@ -1334,29 +1334,30 @@ bool GetXmlType(String path, StringHash &out fileType, bool useCache = false)
         return false;
 
     String name;
+    File@ file;
+    XMLFile@ xml;
+
     if (useCache)
     {
-        XMLFile@ xml = cache.GetResource("XMLFile", path);
-        if (xml is null)
+        file = cache.GetFile(path);
+        if (file is null)
             return false;
-
-        name = xml.root.name;
     }
     else
     {
-        File@ file = File();
+        file = File();
         if (!file.Open(path))
             return false;
-
-        if (file.size == 0)
-            return false;
-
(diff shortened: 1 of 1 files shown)


weitjong:
Will review and merge if no one object it.

weitjong:
Got time to really have a closer look on this issue now. Personally I actually don't see any problem with existing code. The "XMLFile" and the "Material" are different resource type in Urho3D class hierarchy. The "XMLFile" is not a "parent" class of all resource types serialized using XML format. So, I failed to see why we need this fix to be applied.

SuperWangKai:
Hi @weitjong 

The code I modified is currently used in ResourceBrowser of Editor. Though it is currently not used due to `bool useCache = false` . The code basically loads a xml file and detect the resource type by reading the root name of the it.

However, in the ResourceCache implementations, a resource is managed by "Resource Group" which uses resource type + resource path to index a certain resource. 

https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/Resource/ResourceCache.cpp#L632

In our case, once a resource path is successfully loaded using XMLFile type, it is a type of XMLFile. We can interpret the resource as Material using `GetResource<Material>` e.g., it is stored as XMLFile internally in the ResourceCache, thus, the reloading will fail. 

Let me reproduce the reloading bug for you when I have the environment later.

PR #2424, opened 2019-03-06
Fix SLikeNet crash on android device.

sssooonnnggg:
On Android device, the SLikeNet library crashes with a **__stack_chk_fail.** 
Because when the variable **l_prefix** is 128, the variable **i** is 16. The code line try to write l_mask[16] made the crash.

I fixed this bug with reference to libuv.

[https://github.com/libuv/libuv/blob/e4087dedf837f415056a45a838f639a3d9dc3ced/src/unix/android-ifaddrs.c](url)


ArnisLielturks:
@sssooonnnggg Thank you for the fix!
`linux_adapter.cpp` and `linux_adapter.h` is our own implementation to support the MinGW and other compilers. So these changes are acceptable.

PR #2429, opened 2019-03-12
The array size is insufficient

ab4daa:
In D3D11Graphics.cpp, d3dBlendEnable[] does not have the same size as d3dSrcBlend[].
d3dBlendEnable[] have different size in D3D9Graphics.cpp, too.


weitjong:
Thanks for the PR. It is merged now.

PR #2431, opened 2019-03-14
Add result support to angelscript function call

pat2nav:
Hi Urho team

It's very usefull to call an angelscript function and get the return of the function

Per exemple in my game, I call this AS function to get player node

```
URHO::Variant Result;
ScriptFile->Execute("Node @ CreatePlayer(Scene @ scene)", Parameters, URHO::VariantType::VAR_PTR, Result);
URHO::Node * Node = (URHO::Node*) Result.GetPtr();
```

So I propose to add 3 new functions to scriptfile with 2 new parameters

**VariantType resultType**: to specify the type of result expected
**Variant & result**: contains the result of function

### DIFF
--- a/Source/Urho3D/AngelScript/ScriptFile.cpp
+++ b/Source/Urho3D/AngelScript/ScriptFile.cpp
@@ -290,14 +290,7 @@ bool ScriptFile::HasEventHandler(Object* sender, StringHash eventType) const
         return false;
 }
 
-bool ScriptFile::Execute(const String& declaration, const VariantVector& parameters, bool unprepare)
-{
-    Variant result;
-    VariantType resultType = VariantType::VAR_NONE;
-    return Execute(declaration, parameters, resultType, result, unprepare);
-}
-
-bool ScriptFile::Execute(const String& declaration, const VariantVector& parameters, VariantType resultType, Variant&result, bool unprepare)
+bool ScriptFile::Execute(const String& declaration, const VariantVector& parameters, Variant* functionReturn, bool unprepare)
 {
     asIScriptFunction* function = GetFunction(declaration);
     if (!function)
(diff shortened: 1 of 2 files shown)


Source/Urho3D/AngelScript/ScriptFile.cpp
```diff
@@ -321,6 +335,78 @@ bool ScriptFile::Execute(asIScriptFunction* function, const VariantVector& param
 
     scriptSystem->IncScriptNestingLevel();
     bool success = context->Execute() >= 0;
```
weitjong: Based on the Angelscript documentation, the script function returns value is only valid when the execution is actually finished (asEXECUTION_FINISHED).

```suggestion
     bool success = context->Execute() == asEXECUTION_FINISHED;
```

Source/Urho3D/AngelScript/ScriptFile.h
```diff
@@ -81,14 +81,24 @@ class URHO3D_API ScriptFile : public Resource, public ScriptEventListener
 
     /// Query for a function by declaration and execute if found.
     bool Execute(const String& declaration, const VariantVector& parameters = Variant::emptyVariantVector, bool unprepare = true);
+    /// Query for a function by declaration and execute and return result if found.
```
weitjong: Is it possible to inspect AS context to find out what is the type of the return value? Passing the return type on each call here seems not that user-friendly.

pat2nav: Ok It seems possible to get return value using asIScriptFunction::GetReturnTypeId()

Source/Urho3D/AngelScript/ScriptFile.cpp
pat2nav: Ok fixed

weitjong:
Just in case you forgot it, you need to push the changes for us to see them.

pat2nav:
Finally I propose to let the bool return for error checking and get script return value as variant pointer
with automatic type detection

`bool Execute(const String& declaration, const VariantVector& parameters = Variant::emptyVariantVector, Variant* functionReturn=nullptr, bool unprepare = true);`

pat2nav:
> 
> 
> Just in case you forgot it, you need to push the changes for us to see them.

Sorry I'm learning how to make github pull request

ArnisLielturks:
Cool idea! This is something that would make my life easier. What about support for LUA to do the same?

pat2nav:
> 
> 
> Cool idea! This is something that would make my life easier. What about support for LUA to do the same?

Sorry I am not familiar with Lua scripting in urho

weitjong:
Since this is not about Urho API binding, I don't think there is a requirement that we need to have feature parity between them.

weitjong:
Thanks again!

PR #2432, opened 2019-03-16
Fixed warnings Linux where GCC under padantic complained "extra ‘;’".

zlum:


weitjong:
Thanks for your PR. It is merged now.

PR #2437, opened 2019-03-28
Add UI custom material rendering

PredatorMF:
Adds material option to BorderImage.
Discussion: https://discourse.urho3d.io/t/custom-ui-mesh-source/4244


ArnisLielturks:
I guess the only thing missing from this is the LUA integration.

ArnisLielturks:
@weitjong If that's ok with you we can merge this. Will add the LUA integrations in the separate PR + maybe we could add samples to show how it works..

weitjong:
Thanks for asking. I didn't comment on it earlier because I haven't got time to test it myself. I have no objection that you merge the PR.

PR #2439, opened 2019-04-03
Fixed a bug where using CScriptDictionary can lead to crash on arm64.

holyfight6:
The script Dictionary will lead to nullptr access on arm64. The macro wrappers don't support asGeneric when using  a ? type(like the ?&out parameter in "void Set(const String &in, const ?&in)").


weitjong:
Thanks for your PR. I didn't have time to actually test it on ARM64, but the changes look ok-ish to me.

PR #2443, opened 2019-04-23
Slikenet build updates

ArnisLielturks:


ArnisLielturks:
Update on this:

There should be additional cmake tests created for the windows platforms when using MinGW. There are 2 different C++ implementations that should be checked for the availability of the `inet_*` methods. Built in `HAVE_INET_NTOP` defines in the CMake doesn't work as expected, so following conditions should be compiled to check the actual function availability

```
check_cxx_source_compiles ("#include <arpa/inet.h>\nint main() {\n    struct sockaddr_in sa;\n    char str[INET_ADDRSTRLEN];\n    inet_pton(AF_INET, \"192.0.2.33\", &(sa.sin_addr));\n    inet_ntop(AF_INET, &(sa.sin_addr), str, INET_ADDRSTRLEN);\n}" INET_FUNCTIONS_EXISTS_1)
```

and 

```
check_cxx_source_compiles ("#include <winsock2.h>\n#include <ws2tcpip.h>\nint main() {\n    struct sockaddr_in sa;\n    char str[INET_ADDRSTRLEN];\n    inet_pton(AF_INET, \"192.0.2.33\", &(sa.sin_addr));\n    inet_ntop(AF_INET, &(sa.sin_addr), str, INET_ADDRSTRLEN);\n}" INET_FUNCTIONS_EXISTS_2)
```

The second code requires `ws2_32 lib` to be added before compiling the sample.

What could be the best solution to check this? Ideally I would like to add the `ws2_32 lib` only for the second test, otherwise it breaks everything else if the library is not found.

weitjong:
Have you tried to use "CMAKE_REQUIRED_LIBRARIES" variable in conjunction with it?

ArnisLielturks:
@weitjong I did, but that's the thing that it breaks other cmake checks in cases when the MinGW doesn't contain this library,

ArnisLielturks:
Maybe there is an easy way how we could set this library only for single cmake check?

weitjong:
This variable can be set and reset back to what its previous value was pre- and post- a specific check. I think you can see examples in the CMake build scripts for SDL.

ArnisLielturks:
@weitjong Thanks, will test that out.

ArnisLielturks:
@weitjong Thanks! Looks like your proposed solution fixed this issue.

Issue #2386, opened 2018-11-03
Analigned read/write on ARM in VectorBuffer and MemoryBuffer


orefkov:
Current version is crashed on ARM when read/write in VectorBuffer and MemoryBuffer if accesed on addresses not aligned on 4 byte. For example, write 1 byte, and next  write 4 bytes. Current code try first read/write by 4 bytes, but on ARM arc it is not always possible. Need replace by simple memcpy.

weitjong:
Contribution to fix this is welcome. Note that personally I don't have any generic ARM board to test the ARM build artifacts.

orefkov:
Ok, I will try contribute it, if I can pass the stringent requirements for the quality of the code design and the placement of spaces.
Note that almost any smartphone inside contains ARM :)

weitjong:
Fair enough. If the issue is reproducible on my Raspberry-Pi too then I will be interested to help to verify your fix.

orefkov:
In my working copy now it looks like this:
```
unsigned MemoryBuffer::Write(const void* data, unsigned size)
{
    if (size + position_ > size_)
        size = size_ - position_;
    if (!size)
        return 0;

    auto* srcPtr = (unsigned char*)data;
    unsigned char* destPtr = &buffer_[position_];
    position_ += size;
	memcpy(destPtr, srcPtr, size);
	/*
    unsigned copySize = size;
    while (copySize >= sizeof(unsigned))
    {
        *((unsigned*)destPtr) = *((unsigned*)srcPtr);
        srcPtr += sizeof(unsigned);
        destPtr += sizeof(unsigned);
        copySize -= sizeof(unsigned);
    }
    if (copySize & sizeof(unsigned short))
    {
        *((unsigned short*)destPtr) = *((unsigned short*)srcPtr);
        srcPtr += sizeof(unsigned short);
        destPtr += sizeof(unsigned short);
    }
    if (copySize & 1u)
        *destPtr = *srcPtr;
	*/
    return size;
}

```
The rest is similar.

eugeneko:
I personally vote for `memcpy`.
Don't see any reason why it's not used there.
I mean, compiler developers surely know how to copy bytes in the fastest way.

weitjong:
I agree with both of you. I am not the original author (Lasse) and I have no insight why it was written like so. From source code history this piece of code seems ancient! So it could be because at the time of its writing, the `memcpy` implementation of certain compiler was sub-optimal.

[...]

PR #2444, opened 2019-05-03
#2386 Unaligned read/write on ARM in VectorBuffer and MemoryBuffer 

elix22:
Fix Android crash .
Unaligned read/write on ARM in VectorBuffer and MemoryBuffer


weitjong:
Close #2386.

PR #2445, opened 2019-05-04
Android crash in ARM/Release configuration while calling MemoryBuffer::Read() or VectorBuffer::Read()

elix22:
Android crash in ARM/Release configuration while calling MemoryBuffer::Read() or VectorBuffer::Read()


weitjong:
Thanks again.

PR #2449, opened 2019-06-13
 If disable Log module, URHO3D_LOGTRACEF(message) will cause parameters mismatch.

lhinuz:


weitjong:
Thanks for your PR. If you could drop the unrelated change then the PR can be merged immediately.

weitjong:
Duplicated by #2450.

PR #2451, opened 2019-06-20
Replace some SharedArrayPtr by std::unique_ptr

1vanK:
std::unique_ptr allows store arrays, so avoid refcounting

### DIFF
--- a/Source/Urho3D/Audio/Audio.cpp
+++ b/Source/Urho3D/Audio/Audio.cpp
@@ -118,7 +118,7 @@ bool Audio::SetMode(int bufferLengthMSec, int mixRate, bool stereo, bool interpo
     fragmentSize_ = Min(NextPowerOfTwo((unsigned)mixRate >> 6u), (unsigned)obtained.samples);
     mixRate_ = obtained.freq;
     interpolation_ = interpolation;
-    clipBuffer_ = new int[stereo ? fragmentSize_ << 1u : fragmentSize_];
+    clipBuffer_.reset(new int[stereo ? fragmentSize_ << 1u : fragmentSize_]);
 
     URHO3D_LOGINFO("Set audio mode " + String(mixRate_) + " Hz " + (stereo_ ? "stereo" : "mono") + " " +
             (interpolation_ ? "interpolated" : ""));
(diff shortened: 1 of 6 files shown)


iSLC:
Urho does have a `unique_ptr` implementation and the distinction between deleting arrays or regular types can be made with a template specialization like:
```cpp
/// Primary template of `DefaultDelete`.
template<class T> struct DefaultDelete {
    using value_type = T; ///< The type given via template parameter.
    /// Default constructor.
    constexpr DefaultDelete() noexcept = default;
    /// Function call operator that forwards the call to delete on the given pointer.
    void operator () (T * ptr) const noexcept {
        static_assert(sizeof(T), "Cannot delete pointer to incomplete type.");
        delete ptr;
    }
};
// Specialization of `DefaultDelete` for arrays.
template<class T> struct DefaultDelete< T[] > {
    using value_type = T; ///< The type given via template parameter.
    /// Default constructor.
    constexpr DefaultDelete() noexcept = default;
    /// Function call operator that forwards the call to delete on the given pointer.
    void operator () (T * ptr) const noexcept {
        static_assert(sizeof(T), "Cannot delete pointer to incomplete type.");
        delete[] ptr;
    }
};
/// Omit specialization for array objects with a compile time length.
template < class T, unsigned N > struct DefaultDelete< T[N] >;
/// Omit specialization for void type.
template < > struct DefaultDelete< void >;
```

I'll see if I can "rewrite" the `UniquePtr` class tonight to make it closer to what `std::unique_ptr` has to offer.

Maybe work on all the smart pointer classes a bit to make them use the standard that the engine is using now.

1vanK:
As long as I remember UniquePtr has been added before the engine moved to C++ 11. Now we have С++11 by default so we can minimise exiting code.

weitjong:
I am not C++ expert like you guys, but personally I like to see the Urho3D code base to be modernized. At work I have the privilege to use modern programming languages too like Kotlin and modern JS. They are a joy to use. I have read good thing about unique_ptr. So, you have my vote for this PR.

Modanung:
Consistency is also worth something. Shouldn't this change be part of a broader decision to use `std`-types where possible? This may include replacing _all_ smart pointers as well as switching from`Urho3D::Vector`s to `std::vector`s and from `Urho3D::String`s to `std::string`s.
Seems to me like this would break a lot of Urho's existing consistency however broad, narrow or structurally implemented.

1vanK:
Current Urho's implementations of smart pointers and strings is very good (and I like much more than std::). But in some rare cases we can minimize the code without loss of functionality and performance. I do not think that this will be a violation of a single style, becoude we already have some std:: things in code (std::move for example)

PR #2458, opened 2019-07-16
Return 0, not NULL, in function returning an int

smortex:
```
[ 12%] Building CXX object Source/ThirdParty/StanHull/CMakeFiles/StanHull.dir/hull.cpp.o
/home/romain/Projects/Urho3D/Source/ThirdParty/StanHull/hull.cpp:2590:28: error: cannot initialize return object of type 'int' with an rvalue of type 'nullptr_t'
        if(verts_count <4) return NULL;
                                  ^~~~
/usr/include/sys/_null.h:37:14: note: expanded from macro 'NULL'
#define NULL    nullptr
                ^~~~~~~
1 error generated.
```

The function `overhull()` is supposed to return an `int`, but `NULL` is defined as a nullptr, which the compiler does not allow you to store as an int.

The fix consist in replacing `NULL` with the raw value `0`.

### DIFF
--- a/Source/ThirdParty/StanHull/hull.cpp
+++ b/Source/ThirdParty/StanHull/hull.cpp
@@ -34,7 +34,7 @@
 		THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 -----------------------------------------------------------------------*/
 
-// Modified by Lasse Oorni for Urho3D
+// Modified by Lasse Oorni and Romain Tartière for Urho3D
 
 // Urho3D: use a namespace to not clash with Urho3D's inbuilt math types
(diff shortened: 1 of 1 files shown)


ArnisLielturks:
Thanks for the PR! @weitjong What's the approach for updating `StanHull` library?

weitjong:
@ArnisLielturks Thanks for asking. It is the same as any other 3rd-party libs, preferably with change marker. You don't need to concern with whether it uses a subtree or not, the change can go directly to master branch/repo regardless. If you are up to it, you can turn one into subtree yourself, if it is not already.

smortex:
@ArnisLielturks I amended the commit to include the requested changes.  Thanks!

PR #2459, opened 2019-07-16
Add missing headers for checking inet_pton() and inet_ntop() usability on FreeBSD

smortex:
```
[ 19%] Building CXX object Source/ThirdParty/SLikeNet/CMakeFiles/SLikeNet.dir/Source/src/BitStream.cpp.o
In file included from /home/romain/Projects/Urho3D/Source/ThirdParty/SLikeNet/Source/src/BitStream.cpp:48:
/home/romain/Projects/Urho3D/Source/ThirdParty/SLikeNet/Source/include/slikenet/linux_adapter.h:17:10: fatal error: 'winsock2.h' file not found
#include "winsock2.h"
         ^~~~~~~~~~~~
1 error generated.
```

This is dues to the way `inet_pton()` and `inet_ntop()` are looked for at the CMake level: Linux only requires `arpa/inet.h` but FreeBSD needs a few more includes as documented in the man page:

```
NAME
     inet_aton, inet_addr, inet_network, inet_ntoa, inet_ntoa_r, inet_ntop,
     inet_pton, inet_makeaddr, inet_lnaof, inet_netof – Internet address
     manipulation routines

LIBRARY
     Standard C Library (libc, -lc)

SYNOPSIS
     #include <sys/types.h>
     #include <sys/socket.h>
     #include <netinet/in.h>
     #include <arpa/inet.h>
[...]
```

Because these headers are missing, the functions are not found and winsock2 is selected.

Since theses headers are available on Linux, I just added them to the check for `inet_pton()` and `inet_ntop()` without conditional macros to avoid cluttering the test.  As a result, CMake now finds the correct functions and does not attempt to use winsock2.

### DIFF
--- a/Source/ThirdParty/SLikeNet/CMakeLists.txt
+++ b/Source/ThirdParty/SLikeNet/CMakeLists.txt
@@ -24,7 +24,7 @@ add_definitions (-D_RAKNET_SUPPORT_Rackspace=0 -D_RAKNET_SUPPORT_PacketizedTCP=0
 add_definitions (-D_RAKNET_SUPPORT_ReadyEvent=0 -D_RAKNET_SUPPORT_MessageFilter=0 -D_RAKNET_SUPPORT_FileListTransfer=0)
 
 # Check specific function availability which may be missing from different MinGW versions
-check_cxx_source_compiles ("#include <arpa/inet.h>\nint main() {\n    struct sockaddr_in sa;\n    char str[INET_ADDRSTRLEN];\n    inet_pton(AF_INET, \"192.0.2.33\", &(sa.sin_addr));\n    inet_ntop(AF_INET, &(sa.sin_addr), str, INET_ADDRSTRLEN);\n}" INET_FUNCTIONS_EXISTS_1)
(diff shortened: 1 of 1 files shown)


ArnisLielturks:
Thanks for the PR! I would like to test this for myself since these checks have given quite a lot of headache in the past. But the changes look fairly innocent and nothing should go wrong.

smortex:
I understand: I was wondering if it would be better or not to check for each header individually and only include the one which are found, but since 1) all where available on the systems I checked and 2) I have no idea about how to do this properly, I decided to to go to the easy way and rely on the developer's feedback to further improve this PR if needed.

If some adjustments are required, feel free to include some directions / keywords to help me find relevant documentation and implement the changes.

Thanks!

ArnisLielturks:
Our CI uses only Windows, OSX and Ubuntu images (if I'm not mistaken), so other OS's are left out from the daily checks. I would check this on Fedora and CentOS too, since I know few of guys are using that. I usually try to build it using docker as it's the most easiest way for me personally to test it.

weitjong:
I am able to reproduce this issue on (new) FreeBSD DBE (cross-compiling from Linux host targeting FreeBSD 12).

smortex:
I saw today that the Travis-CI build was all failing.  It was 13 days old, so I guess it was OK when I submitted the Pull-Request and was re-triggered and failed (Travis complained about the branch not existing).

In order to be 100% sure that the branch was okay, I rebased it on top of a recent master.

ArnisLielturks:
Issues were caused by me, I created a local branch with your changes, that way all of CI supported platforms were picked up. But for some weird reason it marked this PR as failed by our CI tools. Anyway, the test was a success so I think this is merge ready. If any issues appear, will fix them accordingly.

PR #2460, opened 2019-07-16
Make the 4 arguments vsnprintf_s() visible to FreeBSD

smortex:
```
[ 19%] Building CXX object Source/ThirdParty/SLikeNet/CMakeFiles/SLikeNet.dir/Source/src/RakString.cpp.o
/home/romain/Projects/Urho3D/Source/ThirdParty/SLikeNet/Source/src/RakString.cpp:1392:17: error: no matching function for call to 'vsnprintf_s'
        int numChars = vsnprintf_s(stackBuff, 511, str, ap);
                       ^~~~~~~~~~~
/home/romain/Projects/Urho3D/Source/ThirdParty/SLikeNet/Source/include/slikenet/linux_adapter.h:42:5: note: candidate function not viable: requires 5 arguments, but 4 were provided
int vsnprintf_s(char *buffer, size_t sizeOfBuffer, size_t count, const char *format, va_list argptr);
```

A template is conditionally defined but the condition did not include FreeBSD.

### DIFF
--- a/Source/ThirdParty/SLikeNet/Source/include/slikenet/linux_adapter.h
+++ b/Source/ThirdParty/SLikeNet/Source/include/slikenet/linux_adapter.h
@@ -42,7 +42,7 @@ typedef int errno_t;
 int vsnprintf_s(char *buffer, size_t sizeOfBuffer, size_t count, const char *format, va_list argptr);
 #endif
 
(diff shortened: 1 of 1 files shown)


ArnisLielturks:
Thanks for the PR!

PR #2464, opened 2019-07-22
SLikeNet libcat security

josheb:
SLikeNet dependency was updated to the latest version along with Urho3D specific changes (ToHash method and mingw fix)
LIBCAT_SECURITY = 1 in NativeFeatureOverrides.h so the library is built with support for public/private key encryption.
The encryption from the SLikeNet examples has also been added to tools as EncryptionTool so that keys can be generated on the command line.
This does not include updates to the network module itself at this time, but would allow raw access to SLikeNet with encryption support at a low level.


ArnisLielturks:
@josheb Seems like OpenSSL dependency is missing from our CI builds. OpenSSL is also optional dependency for our CivetWeb integration (WIP), check https://github.com/urho3d/Urho3D/blob/civetweb-update/Source/ThirdParty/Civetweb/CMakeLists.txt#L61-L98 how to possibly fix this issue where openssl is used to allow the engine to make HTTPS requests.

josheb:
@ArnisLielturks Thanks for the link, that's very helpful.  My week is off to it's usual torturous start but I'll see what I can do next time I'm able to dig into it.

ArnisLielturks:
@josheb Honestly I don't think it should be optional for the Civetweb but for the networking functionality I think it should be optional, because as some of the guys on the forum mentioned, one might implement their own encryption on networking and some of them might not use it at all, since there is no de-facto way how to deal with it.

ArnisLielturks:
@josheb Update. My Civetweb update branch now makes the openssl dependency mandatory, I hope it will be merged in the upcoming weeks, the functionality is there, we just have to set up the CI/CD environments accordingly just so all of them have the OpenSSL installed. I think that fixes your OpenSSL dependency problem.  One thing that worries me though is the amount of changes in this PR, like `linux_adapter.h` changes, which are added by me to support multiple build systems + multiple defines which check the current OS.

PR #2467, opened 2019-07-26
Tooltip improvements

MystMagus:
In working with ToolTips I've found that it's good to be able to hide them at will for stuff like repositioning, and alternative targets works well for setting a ToolTip for a layout or similar (say a row with a Text and a LineEdit) where normally the objects inside it would eat the input.


MystMagus:
Oops, that included the changes from the pull request I made before. Gotta fix that

PR #2468, opened 2019-07-26
Tooltip improvements

MystMagus:
In working with ToolTips I've found that it's good to be able to hide them at will for stuff like repositioning, and alternative targets works well for setting a ToolTip for a layout or similar (say a row with a Text and a LineEdit) where normally the objects inside it would eat the input.


ArnisLielturks:
Does this mean we could show tooltip while hovering multiple UI elements?

MystMagus:
Mm, do you mean hovering multiple elements at the same time (ie. in a multitouch situation), or showing the same ToolTip for multiple different ones? Cause this is the later. It will always appear in the same place (as it's still the child of one element), but can be made to appear from hovering any element as long as it's added to the alt targets (since the hovering is handled entirely in ToolTips update method).
The bit I said about being useful for layouts was since you can use it with each stacked element as an alt target.

ArnisLielturks:
If there are no objections, will merge this after few hours.

PR #2473, opened 2019-08-09
added missing 'Apply' function for AnimationState

Polynominal:
AnimationState needs to have the Apply function binding for lua. Needed for bone-less animation playback without animation controller. Not sure why it was missing in the first place, I assume it got forgotten.  For reference I followed 'Node animations' instructions in the documentation, which explicitly mentions the method. 

Works as intended in my project.


ArnisLielturks:
Oh, completely forgot: Thanks for your contribution @Polynominal 👍

PR #2474, opened 2019-08-12
added function to make a created custom geometry class into a simple …

codexhound:
Made an auto function in the custom geometry class to create a circle (makes the object into a circle) from the input params. Can be used to create a circle segment also.


Source/Urho3D/Graphics/CustomGeometry.cpp
```diff
@@ -204,6 +204,34 @@ bool CustomGeometry::DrawOcclusion(OcclusionBuffer* buffer)
     return success;
 }
 
+void CustomGeometry::MakeCircle(float radius, int inc, Vector3 normal, float startTheta, float endTheta)
+{
+    Clear();
+	float stepSize = (endTheta - startTheta) / (float)inc;
```
ArnisLielturks: Fix indentation

Source/Urho3D/Graphics/CustomGeometry.h
```diff
@@ -65,6 +65,8 @@ class URHO3D_API CustomGeometry : public Drawable
     unsigned GetNumOccluderTriangles() override;
     /// Draw to occlusion buffer. Return true if did not run out of triangles.
     bool DrawOcclusion(OcclusionBuffer* buffer) override;
+	/// Make the custom geometry into a circle, change start and stop to make a segment instead
+	void MakeCircle(float radius = 1, int inc = 100, Vector3 normal = Vector3(0,1,0), float startTheta = 0, float endTheta = 2*M_PI);
```
ArnisLielturks: Fix indentation

.gitignore
```diff
@@ -111,3 +111,20 @@ configure*
 *.bak
 Thumbs.db
 .directory
+*.db
```
ArnisLielturks: These changes in .gitignore are not needed. Looks like this is caused because of specific project configuration.

Source/Urho3D/Graphics/CustomGeometry.cpp
```diff
@@ -204,6 +204,34 @@ bool CustomGeometry::DrawOcclusion(OcclusionBuffer* buffer)
     return success;
 }
 
+void CustomGeometry::MakeCircle(float radius, int inc, Vector3 normal, float startTheta, float endTheta)
```
ArnisLielturks: can you name `inc` parameter differently to something like `iterations` maybe? Just so it better explains what it does when reading only header files.

codexhound: made those changes

PR #2480, opened 2019-08-16
Update Box2D

1vanK:
Box2D now supports Shape Cast


1vanK:
https://github.com/erincatto/Box2D/blob/master/Testbed/Tests/ShapeCast.h

weitjong:
Thanks!

PR #2482, opened 2019-08-18
Fix compiler warning:

Loki-Astari:
Compiler Warning was:

````
        Resource/JSONValue.h:200:110: warning: reference to stack memory associated with parameter 'defaultValue' returned [-Wreturn-stack-address]
        const String& GetString(String defaultValue = String::EMPTY) const { return IsString() ? *stringValue_ : defaultValue;}
````

Note: This is also a series problem as the parameter is a stack variable and will be destroyed on
the function return. This returning a reference to this value is returning a reference to a value
that has been destroyed.

Changing the parameter to also be a `const String&` does not change the meaning or usage of the
function.

### DIFF
--- a/Source/Urho3D/AngelScript/ResourceAPI.cpp
+++ b/Source/Urho3D/AngelScript/ResourceAPI.cpp
@@ -401,7 +401,7 @@ static void RegisterJSONValue(asIScriptEngine* engine)
     engine->RegisterObjectMethod("JSONValue", "uint GetUInt(uint defaultValue = 0) const", asMETHOD(JSONValue, GetUInt), asCALL_THISCALL);
     engine->RegisterObjectMethod("JSONValue", "float GetFloat(float defaultValue = 0) const", asMETHOD(JSONValue, GetFloat), asCALL_THISCALL);
     engine->RegisterObjectMethod("JSONValue", "double GetDouble(double defaultValue = 0) const", asMETHOD(JSONValue, GetDouble), asCALL_THISCALL);
-    engine->RegisterObjectMethod("JSONValue", "const String& GetString(const String& defaultValue = String(\"\")) const", asMETHOD(JSONValue, GetString), asCALL_THISCALL);
+    engine->RegisterObjectMethod("JSONValue", "const String& GetString(const String& defaultValue) const", asMETHOD(JSONValue, GetStringWithDefault), asCALL_THISCALL);
     engine->RegisterObjectMethod("JSONValue", "Variant GetVariant(Variant defaultValue = Variant()) const", asMETHOD(JSONValue, GetVariant), asCALL_THISCALL);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by iSLC** [comment]
Doesn't the `const` actually introduce the bug? Because you can now return a reference to a value on the stack.

For example:
```cpp
`xyz.GetString(String("abc"));`
```
I would get the reference of `String("abc")` which is on the stack at this point.

But if the `const` is removed then i can't do that and i'm forced to do it like this:
```cpp
String d("abc");
auto & v = xyz.GetString(d);
```
So `v` is a reference to either `d` or the member value. And neither can be on the stack.

Loki-Astari:
Yes you are correct.
Let me re-submit without the const

Loki-Astari:
Now the trouble here is that `String::EMPTY` is a constant value. So we can't just use a reference. 

Possible Solutions:

* Pass defaultValue as const reference.
* Pass defaultValue as reference.
* Pass defaultValue by value.
* Return by value

### Pass defaultValue as const reference:

As pointed out by `iSLC` is not usable as this allows temporary variables, which may have been destroyed.

### Pass defaultValue as reference

The problem here is that `String::EMPTY` has a type `const reference` and thus can not be used as a default value.

### Pass defaultValue by value

This was the original bug. The problem here is that we are returning a reference and if we have a default value then you are returning a reference to a parameter that will be destroyed on function exit.

### Return by value

This will allow the parameter to be passed by constant reference. It has the same affect as returning by const reference. The trouble is that it will require a copy to return the value so it becomes less efficient.

## Conclusion

The only way to solve the problem is to return by value. This will be less efficient but it will be correct.

````
String GetString(String const& defaultValue = String::EMPTY) const { return IsString() ? *stringValue_ : defaultValue;}
````

## Optimization

There are situations where we can still return by `const reference` if we know both alternatives can still be maintained. So in the situation where no defaultValue is provided by the user we know that String::Empty is always available so we can return that or the original value by const reference.

````
String GetString(String const& defaultValue) const { return IsString() ? *stringValue_ : defaultValue;}
const String& GetString() const { return IsString() ? *stringValue_ : String::EMPTY;}
````

**Comment by iSLC** [comment]
Well, since I'm returning a `const String&` then I can remove the const qualifier from `String::EMPTY` with something like `const_cast< String & >(String::EMPTY)` since i know that I don't modify it in the function body and I know that it won't be modified outside it because it regains back the const qualifier.

```cpp
const String& GetString(String& defaultValue = const_cast< String & >(String::EMPTY)) const { return IsString() ? *stringValue_ : defaultValue; }
```

Isn't what why `const_cast` exists? To get around situations like these where you know it to be safe.

Loki-Astari:
@iSLC Looks like I am being extremely slow on this over the lat two days.

Loki-Astari:
@iSLC OK. Just tried that.

Unfortunately that leads to situations where people can not pass "String const&" objects as parameter. So that does not work either.

Loki-Astari:
Can't re-open this.
So I have created a new one.
https://github.com/urho3d/Urho3D/pull/2483

PR #2483, opened 2019-08-19
Fix compiler warning:

Loki-Astari:
Compiler Warning was:

````
    Resource/JSONValue.h:200:110: warning: reference to stack memory associated with parameter 'defaultValue' returned [-Wreturn-stack-address]
    const String& GetString(String defaultValue = String::EMPTY) const { return IsString() ? *stringValue_ : defaultValue;}
````

Now the trouble here is that `String::EMPTY` is a constant value. So we can't just use a reference.

Possible Solutions:

* Pass defaultValue as const reference.
* Pass defaultValue as reference.
* Pass defaultValue by value.
* Pass a type of reference but value keep local copy
* Return by value

### Pass defaultValue as const reference.
As pointed out by `iSLC` is not usable as this allows temporary variables, which may have been destroyed.

### Pass defaultValue as reference.
The problem here is that `String::EMPTY` has a type `const reference` and thus can not be used as a default value.

### Pass defaultValue by value.
This was the original bug. The problem here is that we are returning a reference and if we have a default value then you are returning a reference to a parameter that will be destroyed on function exit.

### Pass a type of reference but value keep local copy
We could potentially keep a local copy in the function .

1. by using a static function variable. We can return a reference to this variable as it will live an appropriate amount of time. The problem here is that the default value only lives until the next call to `GetString()` at which point it will be over-riden. Not all objects will share the same variable so a call to any `GetString()` will change the value. If a user kept a reference to the value this effectively causes the value to change behind the scenes.

2. by using a member variable. Again it only last until the `GetString()` might be slightly longer as the `GetString()` would have to be called on the same object. But this alters the state of the object which long term is undesirable.

### Return by value
This will allow the parameter to be passed by constant reference. It has the same affect as returning by const reference. The trouble is that it will require a copy to return the value so it becomes less efficient.

## Conclusion
The only way to solve the problem is to return by value. This will be less efficient but it will be correct.

````
String GetString(String const& defaultValue = String::EMPTY) const { return IsString() ? *stringValue_ : defaultValue;}
````

## Optimization
There are situations where we can still return by `const reference` if we know both alternatives can still be maintained. So in the situation where no defaultValue is provided by the user we know that String::Empty is always available so we can return that or the original value by const reference.

````
String GetString(String const& defaultValue) const { return IsString() ? *stringValue_ : defaultValue;}
const String& GetString() const { return IsString() ? *stringValue_ : String::EMPTY;}
````

## Why not pass by reference and cast away const on String::EMPTY
Since GetString() returns a const reference why can we simply not use "Pass defaultValue as reference" as defined above but cast away the `const` for the default value.

````
    const String& GetString(String& defaultValue = const_cast< String & >(String::EMPTY)) const { return IsString() ? *stringValue_ : defaultValue; }
````

The problem here is that we can not use temporary object or pass as default values a const String.

````
    const String  myDefault = "a";
    object.GetString(myDefault);        // Fail
    object.GetString(String("Temp"));   // Fail
````

### DIFF
--- a/Source/Urho3D/AngelScript/ResourceAPI.cpp
+++ b/Source/Urho3D/AngelScript/ResourceAPI.cpp
@@ -401,7 +401,7 @@ static void RegisterJSONValue(asIScriptEngine* engine)
     engine->RegisterObjectMethod("JSONValue", "uint GetUInt(uint defaultValue = 0) const", asMETHOD(JSONValue, GetUInt), asCALL_THISCALL);
     engine->RegisterObjectMethod("JSONValue", "float GetFloat(float defaultValue = 0) const", asMETHOD(JSONValue, GetFloat), asCALL_THISCALL);
     engine->RegisterObjectMethod("JSONValue", "double GetDouble(double defaultValue = 0) const", asMETHOD(JSONValue, GetDouble), asCALL_THISCALL);
-    engine->RegisterObjectMethod("JSONValue", "const String& GetString(const String& defaultValue = String(\"\")) const", asMETHOD(JSONValue, GetString), asCALL_THISCALL);
+    engine->RegisterObjectMethod("JSONValue", "String GetStringWithDefault(const String& defaultValue) const", asMETHOD(JSONValue, GetStringWithDefault), asCALL_THISCALL);
     engine->RegisterObjectMethod("JSONValue", "Variant GetVariant(Variant defaultValue = Variant()) const", asMETHOD(JSONValue, GetVariant), asCALL_THISCALL);
     engine->RegisterObjectMethod("JSONValue", "VariantMap GetVariantMap(VariantMap defaultValue = VariantMap()) const", asMETHOD(JSONValue, GetVariantMap), asCALL_THISCALL);
 
--- a/Source/Urho3D/Resource/JSONValue.h
+++ b/Source/Urho3D/Resource/JSONValue.h
@@ -197,7 +197,16 @@ public:
     /// Return double value.
     double GetDouble(double defaultValue = 0.0) const { return IsNumber() ? numberValue_ : defaultValue; }
     /// Return string value.
-    const String& GetString(String defaultValue = String::EMPTY) const { return IsString() ? *stringValue_ : defaultValue;}
+    /// Need two version because.
+    ///     Can not pass by value (as undefined to return a reference)
+    ///     or reference (as String::EMPTY is const)
+    ///     or const reference (as you can pass temporary and undefined to return a reference to this).
+    /// So we need to return by value.
+    /// To make it more efficient we don't provide a default parameter version so we can return by reference.
+    /// We also need a version to be compatable with the script system (GetStringWithDefault()) that has unique name.
+    String GetString(String const& defaultValue) const { return GetStringWithDefault(defaultValue);}
+    const String& GetString() const { return IsString() ? *stringValue_ : String::EMPTY;}
+    String GetStringWithDefault(String const& defaultValue) const { return IsString() ? *stringValue_ : defaultValue;}
     /// Return C string value.
     const char* GetCString(const char* defaultValue = nullptr) const { return IsString() ? stringValue_->CString() : defaultValue;}
     /// Return JSON array value.

### CODE REVIEW

**Comment by weitjong** [comment]
Personally I dislike the latest version. I like the first initially proposed version better. Perhaps as has been done elsewhere in our code base, we should also have the `const char*` variant for the default value and return the `const String&` regardless. Perhaps you guys worry too much[?]

**Comment by Loki-Astari** [comment]
The trouble here is that we tried all these possibilities.

Only the last one works. All the other fail in some way. UB is actually something to be **very** worried about so I don't think we are worrying too much.


## Pass defaultValue as const reference.
````
    const String& GetString(String const& defaultValue = String::EMPTY) const
                          //                             ^^^^^^^
    { return IsString() ? *stringValue_ : defaultValue;}

    ...
    JSONValue   value;
    String const&   string = value.GetString(String("abc"));
    std::cout << string;         // Broken passing a reference to a temporary value that no longer exists.
                                             // the GetString returned a reference to "defaultValue" which was a reference
                                             // to a temporary object that was passed as a parameter.
                                             //
                                             // Definitely UB. Result most likely a crash.
````

## Pass defaultValue as reference.
````
    const String& GetString(String& defaultValue = String::EMPTY) const
    { return IsString() ? *stringValue_ : defaultValue;}

    // This does not compile as `String::EMPTY` is a const value.
    // We could cast away the constness to get around this problem.

    const String& GetString(String& defaultValue = const_cast<String&>(String::EMPTY)) const
    { return IsString() ? *stringValue_ : defaultValue;}

    // This compiles but now you have places in the code that passed a parameter that
    // was a const reference or a temporary object. So the code will fail to compile for all these
    // locations.

````

## Pass defaultValue by value.
This is the current implementation and the problem we are trying to fix.
````
    // Current implementation
    const String& GetString(String defaultValue = String::EMPTY) const
    { return IsString() ? *stringValue_ : defaultValue;}

    ...
    JSONValue   value;
    String const&   string = value.GetString();
    std::cout << string;         // Broken passing a reference to a value that no longer exists.
                                             // the GetString returned a reference to "defaultValue" an automatic
                                             // object local to GetString() that is now out of scope.
                                             //
                                             // Definitely UB. Result most likely a crash.
````


## Pass a type of reference but value keep local copy
````
    const String& GetString(String const& defaultAlternative = String::EMPTY) const
    {
        static String defaultValue = defaultValue;
        return IsString() ? *stringValue_ : defaultValue;
    }

    ...
    JSONValue   value;
    String const&   string = value.GetString(String("abc"));
    std::cout << string; 

    // This works.
    // But the side effect here is that the returned value is only valid till the next call of GetString()
    // Even if I call GetString() on another object (which could ligitametly have a different default.

     JSONValue   someObjectData;
     String const&   someObjectString = someObjectData.GetString(String("An Object Value"));

     std::cout << string << "\n" << someObjectString << "\n;
     // This will print out
     //  An Object Value         -> I would expect this line to be abc
     //  An Object Value
````

## Return by value
````
    String GetString(String const& defaultValue = String::EMPTY) const
    { return IsString() ? *stringValue_ : defaultValue;}

    ...
    JSONValue   value;
    String const   string = value.GetString();
    std::cout << string;         // Works. The only way I could get it to work correctly.
````

I then added a single optimization.
````
    String GetString(String const& defaultValue) const
    { return IsString() ? *stringValue_ : defaultValue;}
    String const& GetString() const
    { return IsString() ? *stringValue_ :  String::EMPTY;}
````

I then made it nice to work with scripting language (which did not like two versions of the named function being available).
````
    String GetString(String const& defaultValue) const 
    { return GetStringWithDefault(defaultValue);}
    const String& GetString() const
    { return IsString() ? *stringValue_ : String::EMPTY;}
    String GetStringWithDefault(String const& defaultValue) const    // scripting uses this.
    { return IsString() ? *stringValue_ : defaultValue;}
````

**Comment by weitjong** [comment]
Tell me again why would anyone do this?

`String const&   string = value.GetString(String("abc"));`

instead of 

`String const&   string = value.GetString("abc");`?

**Comment by Loki-Astari** [comment]
There is no difference between those two.  
The second one:

````
    String const& string = value.GetString("abc");
    // This is equivalent to 
    String const& string = value.GetString(String("abc"));
    // As the class String has a one argument constructor.
    // So the result is that you are building a temporary object.
    // A temporary object can only be passed as a const reference.
````

But are you saying we should add a method:

````
    String GetString(char const* defaultValue) const
    { return IsString() ? *stringValue_ : String(defaultValue);}
````

That's fine. But you will notice that you still need to return by value to make that work correctly.

**Comment by iSLC** [comment]
Things would be so much simpler if the engine had `StringView` and `Span` types 😄

**Comment by weitjong** [comment]
This API method is not new, so if possible I would like to keep its existing signature by returning `const String&`. I am still not convinced the provided reasons warrant for API breaking change, but this is just my opinion. I would rather just add a warning comment in the first variant that takes another `const String&` as default value to read that the reference must be kept valid as it could be returned as is to the caller. For the new variant that I proposed which takes `const char*` as the default value, we are more free to choose how we would like to implement it. However, instead of

```
String GetString(char const* defaultValue) const
```

I would still prefer:

```
const String& GetString(char const* defaultValue) const
```

Probably using a backing `Hashmap<StringHash, String>` of the "seen" default values.

**Comment by Loki-Astari** [comment]
What you are suggesting is a breaking change.

````
    const String& GetString(const char* defaultValue) const
````

Simply changing the interface like this is a breaking change to the interface. Any call that used to pass a `String` object will no longer compile.

This will no longer compile.
```
    String  someFunctionThatGetsADefault();
    ....
    JSONValue   value;
    String const&   string = value.GetString(someFunctionThatGetsADefault());
````

This will also now fail to compile.
````
   String  someMemberOrLocalVariable;
    ....
    JSONValue   value;
    String const&   string = value.GetString(someMemberOrLocalVariable);
````

By making this change you are no longer supporting an interface that used to exist. There are lots of ways of passing a default value that were not `chat const*` that will no longer be supported.

On the other hand what I suggested is **NOT** a breaking change. It is a backward compatible change. It also supports the simple addition of your suggest new interface without introducing any breaks.

**Comment by eugeneko** [comment]
Having just these two guys will solve everything:
```
const String& GetString() const;
String GetString(const String& defaultValue) const;
```
And yes, current API is error-prone and basically UB. Luckily nobody uses it.

**Comment by weitjong** [comment]
Ops, I think I forgot in Urho the `Urho3D::String` constructor can take `const char*` as single parameter (non-explicit). Basically auto convert the latter to `String`. In that case if I still want this new variant then it can only take this signature

```
String GetString(const char* defaultValue) const
```

Which is not much different than

```
String GetString(const String& defaultValue) const
```

and worse, these two cannot coexist due to the same String auto-conversion issue (Now I regret  a little why not heeding the linter suggestion to make it becomes “explicit” constructor with single parameter).

**Comment by Loki-Astari** [comment]
Is this project even maintained.
Should I just go find another library that is being used?

**Comment by weitjong** [comment]
I am sorry to say we are running in "headless" mode and thing is not moving as fast as it used to be at the moment.

**Comment by Modanung** [comment]
@Loki-Astari You're assistance is appreciated, as we are indeed shorthanded.  
What I think @eugeneko meant to say was that the function he mentioned has not caused enough trouble for end-users for its instability to be noticed. It cannot be said the engine unused. Come visit the [forums](https://discourse.urho3d.io/) to see what people are working on.

**Comment by SirNate0** [comment]
Can we use rvalue references to pick out when we have a temporary object instead of a normal reference? Something like
```
    const String& GetString() const
    {
        return IsString() ? *stringValue_ : String::EMPTY;
    }
    //String& GetString(String& defaultValue)
    //{
    //     return IsString() ? *stringValue_ : defaultValue;
    //}
    // To preserve the return being const I propose either the addition of the above method (which would allow const and not-const returns) or the (unnecessary) addition of a const to the value return below
    String GetString(String&& defaultValue) const
    {
        return IsString() ? *stringValue_ : std::move(defaultValue);
    }
    const String& GetString(const String& defaultValue) const
    {
        return IsString() ? *stringValue_ : defaultValue;
    }
```
I'm not seeing any warnings with all of these calls (tested in another class as I didn't want to spend the time rebuilding the whole library).
```
    static void TestDefaults() {
        Test tester("Hello");
        tester.GetString();
        tester.GetString("Hi");
        tester.GetString(tester.GetString());
        tester.GetString(String("iH"));
        tester.GetString({"What's up?"});
        String s("Moving");
        tester.GetString(tester.GetString(std::move(s)));
    }
```
I'm by no means a C++ expert, so if I missed something don't hesitate to point it out.

PR #2485, opened 2019-08-20
Fixed the depth test behavior of debug renderer sphere sectors.

SirNate0:
The `AddLine` calls were missing the `depthTest` flag, so it was impossible to actually get a sphere sector without depth test enabled.

Also added more details to the documentation for that function. Perhaps the documentation could be clarified differently, particularly regarding the angle. I initially expected the angle to range from 0 to 180 and be measured from the axis to the edge of the sector, but instead it ranges from 0 to 360 measuring the edge to the other edge of the sector.


weitjong:
Thanks!

PR #2486, opened 2019-08-22
Borderimage disabledimageoffset

MystMagus:
Moved the Disable Image Offset parameter from Button to BorderImage.

Implemented special behavior for CheckBox, but more behaviors could be added that would make sense. I'm especially thinking of classes that have sub-elements, like ScrollView for example. I'm unsure if in cases like that we want to make it so that objects like this disable/enable the sub-elements that they create themselves when their state is changed, or if we'd rather leave that to SetEnabledRecursive (for ScrollView specifically I also wonder if maybe the content element should also be enabled/disabled when the view is, but I would say that is a separate thing).

I've also made local changes to the default ui texture and such for the editor/samples to include disabled images for the general element types, that I can include if you think it makes sense.

* also, still not too experienced with pull requests and forks and such, so dunno if I'm doing this right still


weitjong:
First of all, probably you need to squash all the commits and also observe the spacing convention used in Urho3D code base, .e.g `if` keyword should be followed by a space before opening bracket.

MystMagus:
> First of all, probably you need to squash all the commits and also observe the spacing convention used in Urho3D code base, .e.g `if` keyword should be followed by a space before opening bracket.

Right, forgot about checking on some of the conventions, sorry. I'll fix it later.

As for the other thing, I'm not quite sure how to do that (and how things should ideally look afterwards). Can you explain a little more or provide a link to more info?

weitjong:
https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History.

You need to be comfortable to work with git commands using CLI though. If you have difficulty then you can leave the rebase/squash operation to us.

MystMagus:
That would be great, learning from example I find is often the best way. Could you post the commands that you use as well? And don't worry about it being CLI, I'm comfortable with it.

codexhound:
MystMagus:
Heres how I did it using the second method, its 3 lines in the command prompt, just make sure not to reset too many commits backward
https://gist.github.com/patik/b8a9dc5cd356f9f6f980

weitjong:
Thanks for the effort, but the commit history is wrong now and all the commits still not squashed into one with your name as author/committer. However, don’t worry about it.

weitjong:
Earlier I was replying on an iPad, thinking the message from a same person.

p.s. Note to self: never comment on anything while on toilet break.

weitjong:
Thanks. commit 63ef7312bffb047ea95d1c9a7b272df5476bb1cb

PR #2487, opened 2019-08-23
Constexpr iterators

iSLC:


Source/Urho3D/Urho3DConfig.h
```diff
@@ -0,0 +1,95 @@
...
+// Macros with a strong guarantee of what c++ version is available.
+#define URHO_CPP11_OR_LATER (URHO_CPLUSPLUS >= 201103L)
+#define URHO_CPP14_OR_LATER (URHO_CPLUSPLUS >= 201402L)
+#define URHO_CPP17_OR_LATER (URHO_CPLUSPLUS >= 201703L)
+
+// Macros for specific c++ versions
+#if (URHO_CPLUSPLUS >= 201103L)
+	// Least match
+	#define URHO_CPP11(x) x
+	#define URHO_CPP14(x)
```
1vanK: tabs

iSLC:
My bad. Will fix in a few hours when I get home.

Thank you for the heads-up.

iSLC:
I've chosen to make one last change. Which is to create a `Base` for container that is not directly related to code from `Container` directory.

The reason is that I want to avoid a situation where new code will be thrown in there simply because there is no other place for it.

I remember seeing this in the Turso3D from cadaver's WIP engine. And it made sense. For example, there are a few more things in Container that don't really belong there. Such as `Swap`, `RefCounted`, `FlagSet`, `ForEach` etc. These headers were thrown in there simply because there was no other place. But they are not directly related or just limited to the code from the `Container` directory.

I hope this doesn't intrude to much but I wanted to make room for future code and changes.

weitjong:
Thanks. If there is no objection from others then it can be merged shortly.

eugeneko:
@iSLC, okay, I have question. Why did you introduce nearly-empty folder "Base" when we already have folder "Core"? I don't see clean separation between these two.
You could argue that "Core" is about engine core functionality. I'd reply that it's not, not now.
I'm not against introducing new folders, as long as they have clean responsibility and inambigous separation from existing folders. "Base" has neither, for now.

eugeneko:
And one more thing. Can you prove that `URHO_STRICMP` and `URHO_STRNICMP` are valid for all supported platforms?

iSLC:
Well, I would ask to define Platform first. But I'll try to confirm tonight whether they're supported on the compilers and libraries that the engine is supposed to compile on.

If I can't confirm that or the function is located on a different header for some implementation. I will either remove them or provide a generic implementation on such configuration.

Or we can debate and pick which route provides the most desired result.

Right now they're leftover that remained from something I was working on before making this pull request. So you could say they're not that necessary atm.

iSLC:
As for the `Core` vs `Base`. The reason I made the `Base` folder was to move some things outside the `Containers` that (_like I said_) are used in conjunction with containers but should not really be limited to that or have to get them from the `Containers` folder.

I guess i was going for [something like this](https://github.com/cadaver/turso3d/tree/master/Turso3D/Base). Where the actual containers are left in the `Containers` folder but things like swap, iterator, sort, algorithms, hashing and even some things from the `Math` folder such as `min/max`.

As for the `Core` folder. To be honest. I don't really see dependency tree somewhere that suggests who and what should depend on who. The logic I was trying to follow was that `Container` should not depend on things from `Core`  because I'd like to consider `Core`  as the core component(s) of the engine from which every other component derives or depends on. And `Core` then depends on `Container` because it uses containers like everything else. However, the code that doesn't really fit in `Container` but is used by both `Core` and `Container` should be somewhere that doesn't break that dependency "rule".

To put it differently. `Container` depends on code that isn't really a component but isn't strictly limited to containers (thus `Base`). `Core` depends on `Container` and some of the code that is used in conjunction with `Container` but not strictly limited to that (again, `Base`). `Container` cannot depend on `Core`. And everything else depends on `Core` and inherently `Container` and `Base`.

Currently at work. So I don't know if I can make much sense right now. Hopefully I can describe this choice in more detail. And whether I should do another pull request to move them in the locations you suggest.

Let me know what you decide. So I can undo any undesired changes I made.

eugeneko:
@iSLC Thanks for the answer, it is starting to make sense for me now.
So, the `Base` has no dependecies, `Containers` is for containers and depends on `Base` and `Core`.

I don't oppose this idea on its own, but I still question its implementation.

First, I'd say that RandomAccessIterator is kinda _related_ to containers. Moreover, it should be nested class of Vector or just plain pointer. So it doesn't really fit the idea of `Base` folder.

Second, `LowerBound` and `UpperBound`... I've added them several years ago and I keep regretting it.
There's no place for manual implementation of standard algorythms in year 2019. I think the best way to go is remove this stuff and just make RandomAccessIterator compatible with standard algos.

Third, I agree that in perfect world we should have moved things like `Random`, `FlagSet`, `Hash` etc into `Base`. Renamed `Str.h` to `String.h` too. Maybe even renamed `Containers` into `Base` as a whole (like Turso) because I cannot really tell whether `Pair` or `String` or `Allocator` should be in `Base` or `Container`.

However, this breaks compatibility. If we want to go this way, let's break and fix _all the files_.
Moving them one by one really sucks from the perspective of user (like me).

1vanK:
@iSLC 

It seems code
```
#if URHO_CPP11_OR_LATER
    // Least match
    #define URHO_CPP11(x) x
    #define URHO_CPP14(x)
    #define URHO_CPP17(x)
    // Exact match
    #define URHO_CPP11_ONLY(x) x
    #define URHO_CPP14_ONLY(x)
    #define URHO_CPP17_ONLY(x)
#elif URHO_CPP14_OR_LATER
    // Least match
    #define URHO_CPP11(x) x
    #define URHO_CPP14(x) x
    #define URHO_CPP17(x)
    // Exact match
    #define URHO_CPP11_ONLY(x)
    #define URHO_CPP14_ONLY(x) x
    #define URHO_CPP17_ONLY(x)
#elif URHO_CPP17_OR_LATER
    // Least match
    #define URHO_CPP11(x) x
    #define URHO_CPP14(x) x
    #define URHO_CPP17(x) x
    // Exact match
    #define URHO_CPP11_ONLY(x)
    #define URHO_CPP14_ONLY(x)
    #define URHO_CPP17_ONLY(x) x
#else
    #error Compiler does not support required c++ standard. How did you get here?
#endif
```

never worked. For example for C++17 the first block `URHO_CPP11_OR_LATER` is used instead `URHO_CPP17_OR_LATER`. This blocks should be in reverse order.

PR #2490, opened 2019-08-25
Mapped UI Elements

codexhound:
Function Mapped UI Elements:
-Easily Install Key Mappings to dedicated handler
-Install functions for handler usage
-Easily Install Element Mappings to dedicated handler
-Install functions for handler usage

-Dedicated Handlers will not fire when the UIElement is disabled

Easy operations on function mapped elements, useful for windows where you wish to disable only the event mapped elements

Example Usage:
https://github.com/codexhound/Mapped-UI-Example

### DIFF
--- a/Source/Urho3D/UI/UIElement.cpp
+++ b/Source/Urho3D/UI/UIElement.cpp
@@ -33,6 +33,8 @@
 #include "../UI/UI.h"
 #include "../UI/UIElement.h"
 #include "../UI/UIEvents.h"
+#include "../UI/Text.h"
+#include "../Input/InputEvents.h"
 
 #include "../DebugNew.h"
 
@@ -92,11 +94,17 @@ static bool CompareUIElements(const UIElement* lhs, const UIElement* rhs)
 
 XPathQuery UIElement::styleXPathQuery_("/elements/element[@type=$typeName]", "typeName:String");
 
-UIElement::UIElement(Context* context) :
+UIElement::UIElement(Context* context, UIElement * parent) :
     Animatable(context),
     pivot_(std::numeric_limits<float>::max(), std::numeric_limits<float>::max())
 {
+
+    if (parent != nullptr) {
+        parent->AddChild(this);
+    }
     SetEnabled(false);
+
+    SubscribeToEvent(Urho3D::E_KEYDOWN, URHO3D_HANDLER(UIElement, HandleKeyPressed));
 }
 
 UIElement::~UIElement()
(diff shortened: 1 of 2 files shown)


eugeneko:
Could you please elaborate the purpose of mapped elements?

Source/Urho3D/UI/UIElement.cpp
```diff
@@ -2241,4 +2249,404 @@ void UIElement::SetRenderTexture(Texture2D* texture)
...
+    if (child != nullptr) {
+        succ = true;
+        child->SetVisibilityMapped(en, true, recursive);
+    }
+    return succ;
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////
+
+void UIElement::HandleItemChanged(Urho3D::StringHash eventType, Urho3D::VariantMap& eventData)
```
eugeneko: Where is this guy used?

rokups:
Not to sound mean, but this PR is completely unsuitable for merging.

Most importantly - looking at the code we have no idea what it does and what it is meant for. This is a huge red flag.

* Code style violations are all over the place. Minor, but important.
* Strings are passed by copy.
* `std::map` is used. It is again a violation of existing practices and it is terribly inefficient while at that.
* `std::shared_ptr` in use. Engine does not use smart pointers from stdlib.
* Use of raw pointers instead of smart pointers is all over the place.
* `name == String()` is not how you check for empty string.

All in all, even if idea behind this PR is good - implementation is very bad. Do not get discouraged though. Please let us know what you tried to achieve so we can figure out best way to do that. Then we work from there.

rokups:
We would still appreciate if you elaborated on what you tried to achieve here.

codexhound:
Allow doing operations on specific child ui elements. Say you only want to
disable all the mapped buttons, but nothing else, maybe you only want to
make those elements visible. The point was to allow a defined method for
more controlled operations on children elements. Doing an operation
recursively on all the children is not what you want in most cases.

Also the explicit event handlers (ItemChanged and KeyPress) will not fire
if the parent element (mapped element holder) is not enabled. For example
if a key was mapped via addMappedKey, it will not fire if the ui element
were disabled.

Maybe there is a way of doing this at the actual event level, so that would
be a better option to solve the same thing but Idk. or maybe I misread and
this is already the case.

rokups:
> Say you only want to disable all the mapped buttons, but nothing else, maybe you only want to
make those elements visible.

This definitely sounds useful. Though i am inclined to think that recursing into all child elements is a right way to achieve this. These actions would happen very rarely and UI hierarchies are not that deep to choke a frame. You can use tags to group certain UI elements and later access them by calling `GetChildrenWithTag()`.

> addMappedKey

I missed this bit. It is a second feature in a single PR. Next time please submit a PR with only one feature. We may have different opinions about different features, no need for one feature to kill entire PR :]

Now about feature itself.. I am not sure it is warranted. You still have to implement a function that is passed to `addMappedKey()`. Class that implements said function should also subscribe to key events and handle them. I suspect you aimed at implementing something along the lines of shortcut keys. This would be fine, except shortcut keys are not a generic UI concept and thus implementation should probably live in `Button`. This also does not implement any kind of combining multiple keys that are usually used for shortcuts which further hints me that this is a feature more suited to your specific application rather than a general solution.

Do not get discouraged. The only way to improve is to make mistakes :]

codexhound:
Always useful to learn more about the available functions in Urho. Agreed,
if this were to be integrated, it would be best to use the existing
GetChildrenWithTag() method.

Why would a shortcut key implementation live in 'Button'? I agree it's not
suited to UI, it's actually even more generic and could be moved even to
Object. That way any application class could use it.

Note: one of the purposes of this was to auto 'disable' subscribed events
including key events when the ui element was disabled. Having the dedicated
event filters allowed this because they checked isEnabled()

rokups:
> Why would a shortcut key implementation live in 'Button'?

Key shortcuts make sense for UI elements that can be pressed. `Button` is exactly that. Nobody is opening combo-boxes with key shortcuts. For element like `Text` it makes even less sense.

> Note: one of the purposes of this was to auto 'disable' subscribed events
including key events when the ui element was disabled.

Another way would be using `Object::SetBlockEvents()`. It is not automatic, but probably isnt a big deal to call extra function when you are disabling UI element. Besides disabling something like button should stop sending events like `E_PRESSED`. I am surprised that i can not find logic disabling these events, but i can not believe `E_PRESSED` would still be sent for disabled buttons. Makes no sense \o/

eugeneko:
@rokups 

>  I am surprised that i can not find logic disabling these events, but i can not believe E_PRESSED would still be sent for disabled buttons.

`UI::GetElementAt` disables them

@codexhound 

> Maybe there is a way of doing this at the actual event level, so that would
be a better option to solve the same thing but Idk. or maybe I misread and
this is already the case

We already have something similar in `Menu.cpp`. That implementation actually has more features than one in this PR. The best way to go is probably generalize code and move shortcuts support from `Menu` to `Button`

codexhound:
Key shortcuts make sense for UI elements that can be pressed. Button is
exactly that. Nobody is opening combo-boxes with key shortcuts. For element
like Text it makes even less sense.
      Text, yes it doesn't make sense, but even then you never know what
people want to have a shortcut for. Why over specify something.

rokups:
> you never know what people want to have a shortcut for

We have a pretty good idea based on existing UI systems in place. This is not UI research project, that is why we are sticking to established paradigms.

Issue #2469, opened 2019-08-03
Documentation improvement: document enums in doxygen


faesong:
**Why:**

https://urho3d.github.io/documentation/1.7.1/class_urho3_d_1_1_geometry.html#aeb6ac4fb32bebb05e54fac9a1d44da69

For example, say you go and look at _Geometry::SetDrawRange_ method. There is argument of _PrimitiveType_ type, which is enum defining whether it's a _triangle list_, _line list_, etc. But the list of possible values is not available through documentation -- doxygen is ignoring all those enums and they show up like if they were c++ builtin types with no documentation.

**How**
http://www.doxygen.nl/manual/docblocks.html#structuralcommands

So, briefly, we have to put, for example, ``// \file GraphicsDefs.h`` into each header file  in order for it to be included into Doxygen documentation.

I hope I got the _How_ part right, as for now I'm not able to test it quick enough...

In a week or so I might be able to do the pull request (can't promise though), unless somebody else jumps in, or maybe I got it all wrong...

Anyway, thanks for great engine btw. Pleasure to work with.

faesong:
Probably shouldn't close this until everyone is happy about this: https://github.com/urho3d/Urho3D/pull/2492#issuecomment-525800619

PR #2492, opened 2019-08-27
Make doxygen catch global enums

faesong:
Already posted issue about it https://github.com/urho3d/Urho3D/issues/2469 (tldr: global enums are skipped by doxygen unless you put \file element into corresponding header or source file).

The reason for it is so that nobody has to grep through documentation in order to figure out which values can given enum have. Yay.

### DIFF
--- a/Source/Urho3D/AngelScript/Script.h
+++ b/Source/Urho3D/AngelScript/Script.h
@@ -20,6 +20,8 @@
 // THE SOFTWARE.
 //
 
+/// \file
+
 #pragma once
 
 #include "../Core/Mutex.h"
--- a/Source/Urho3D/AngelScript/ScriptInstance.h
+++ b/Source/Urho3D/AngelScript/ScriptInstance.h
@@ -20,6 +20,8 @@
 // THE SOFTWARE.
 //
 
+/// \file
+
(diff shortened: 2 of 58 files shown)


weitjong:
Thanks!

weitjong:
It worked, though it has also caused other unrelated defect. See this newly generated  page which was not exists at all before this change.

https://urho3d.github.io/documentation/HEAD/_graphics_defs_8h.html#a81232b006e10e8bdacec055d6d4f637d

faesong:
Yes, but from what it seems like, there's no other way. Maybe I should've put more effort into researching that...

Though on the other hand it's maybe alright as it is: the enum documentation _has to_ stay somewhere. And if enum is not declared inside the class, then it's totally logical to be documented on a separate page.

Another possible fix for all this is to declare all the enums inside a class -- either in a corresponding class, or just make a new empty class... I don't know how does doxygen behave with c++11 "enum class", but that one is also an option.

To be honest, after so much time spent figuring it out, I still don't understand why couldn't doxygen just work with global enums...

weitjong:
We can keep this until we can figure out a better way. Thanks again for the effort.

PR #2494, opened 2019-08-28
Event handler based on type-erasure.

iSLC:
Removes the need for virtual methods in the event handler. I don't expect this to be merged due to the loss of `EventHandler11Impl`. But I still decided to make it available if the need is there.

Long story short: It uses lambdas to forward event calls instead of virtual methods. I'll detail the pros and cons bellow.

Cons:
- Things like `EventHandler11Impl` are no longer possible since `EventHandler` no longer has any virtual methods. So you can't use lambdas like your used to.

(Possible) Pros:
- The `EventHandler` structure is basically a POD now. It can be copied moved etc. Which opens other possibilities to how do you want to store it. Maybe like vector for cache friendliness. Might require extra work to avoid issues with un+subscribing during events but it is possible (_did it in a quite fast signals-and-slots implementation a while back_).

I won't masquerade this as being faster or something. Because I don't have a significant application to properly profile this. But it can be used to achieve a faster event system.

### DIFF
--- a/Source/Urho3D/AngelScript/ScriptFile.cpp
+++ b/Source/Urho3D/AngelScript/ScriptFile.cpp
@@ -692,13 +692,13 @@ void ScriptFile::AddEventHandlerInternal(Object* sender, StringHash eventType, c
 
     if (!sender)
     {
-        i->second_->SubscribeToEvent(eventType, new EventHandlerImpl<ScriptEventInvoker>
-            (i->second_, &ScriptEventInvoker::HandleScriptEvent, (void*)function));
+        i->second_->SubscribeToEvent(eventType, new EventHandler
+            (i->second_, EventHandler::MemberFwd<ScriptEventInvoker, &ScriptEventInvoker::HandleScriptEvent>(), (void*)function));
     }
     else
     {
-        i->second_->SubscribeToEvent(sender, eventType, new EventHandlerImpl<ScriptEventInvoker>
-            (i->second_, &ScriptEventInvoker::HandleScriptEvent, (void*)function));
+        i->second_->SubscribeToEvent(sender, eventType, new EventHandler
+            (i->second_, EventHandler::MemberFwd<ScriptEventInvoker, &ScriptEventInvoker::HandleScriptEvent>(), (void*)function));
     }
 }
 
--- a/Source/Urho3D/AngelScript/ScriptFile.cpp
+++ b/Source/Urho3D/AngelScript/ScriptFile.cpp
@@ -693,12 +693,12 @@ void ScriptFile::AddEventHandlerInternal(Object* sender, StringHash eventType, c
     if (!sender)
     {
         i->second_->SubscribeToEvent(eventType, new EventHandler
-            (i->second_, EventHandler::MemberFwd<ScriptEventInvoker, &ScriptEventInvoker::HandleScriptEvent>(), (void*)function));
+            (i->second_, EventHandler::Forwarder<ScriptEventInvoker, &ScriptEventInvoker::HandleScriptEvent>(), (void*)function));
     }
     else
     {
         i->second_->SubscribeToEvent(sender, eventType, new EventHandler
-            (i->second_, EventHandler::MemberFwd<ScriptEventInvoker, &ScriptEventInvoker::HandleScriptEvent>(), (void*)function));
+            (i->second_, EventHandler::Forwarder<ScriptEventInvoker, &ScriptEventInvoker::HandleScriptEvent>(), (void*)function));
     }
 }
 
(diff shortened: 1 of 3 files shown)


1vanK:
> Because I don't have a significant application to properly profile this.

I agree that this is a big problem for the engine at the moment. I also have several completed small games (some of them are in my repository). But none of them can become a serious benchmark. The only major project on Urho3D that I tested - Hellbraker. And it has good enough performance on my system (> 100 fps). Some parts of game written on Angel Script, so rewriting it to c++ would noticeably increase perfomance. But this is a commercial game and we don’t have access to the full source code for testing purposes.

eugeneko:
> I don't expect this to be merged due to the loss of EventHandler11Impl. But I still decided to make it available if the need is there.

Pull Request is, by definition, request from contributor to merge a piece of code into main repo.
PR is not a tool for code sharing.
When you create PR without intention of it being merged... That looks weird.

iSLC:
I guess I expressed my self a bit incorrectly there. I guess what I was trying to say is that the change does come with a few inconveniences which make it reasonable to deny the PR. And that I am aware it is a long shot. And whether this is is merged or not. I'm OK with it. I wanted to at least give the it the chance to be refused rather than assume it will be.

The intention to merge this is there. But the authority isn't. Nor the desire to push my agenda into this repository.

My bad.

eugeneko:
Oh, I totally forgot about this PR.

> There would be no virtuals

@rokups Calling `std::function` _is_ virtual function itself.

I'm going to close this PR because it is basically featurecut and step back to C++98.

Lambda event handler is the main use case for me and IMO most of engine handlers should be implemented this way too.

It's true that in this approach we will get one memory lookups instead of two, but this tiny difference doesn't justify featurecut for me.

rokups:
> @rokups Calling std::function is virtual function itself.

Meh sucks. We need a non-virtual alternative then.

eugeneko:
@rokups We don't have much choise.

You can get one memory indirection and one function pointer call if you do function pointer trick like @iSLC or EnTT. In this case you should forget about lambda capture list.

Or you can get two memory indirections and one function pointer call if you do std::functon or any alternative. This way capture lists are supported.

rokups:
I am using this in another project (from hugoam):
```cpp
#include <iostream>
#include <type_traits>
using namespace std;

template <typename T>
class function;

template <typename Return, typename... Args>
class function<Return(Args...)>
{
public:
    function() {}

    template <class T, typename = typename std::enable_if<std::is_invocable<T, Args...>::value>::type>
    function(T functor)
    {
        m_func = [](const void* user, Args... args) -> Return
        {
            const T& func = *static_cast<const T*>(user);
            return func(static_cast<Args&&>(args)...);
        };

        m_dtor = [](void* user)
        {
            T& func = *static_cast<T*>(user);
            func.~T();
        };

        static_assert(sizeof(T) <= sizeof(m_storage));
        new(m_storage) T(std::move(functor));
    }

    ~function()
    {
        if(m_dtor) m_dtor(m_storage);
    }

    Return operator()(Args... args) const
    {
        return m_func(m_storage, static_cast<Args&&>(args)...);
    }

    explicit operator bool() { return m_func != nullptr; }

    using Func = Return(*)(const void*, Args...); Func m_func = nullptr;
    using Dtor = void(*)(void*); Dtor m_dtor = nullptr;
    void* m_storage[8];
};

struct Complex
{
    int state;
    Complex()
    {
        state = 123;
        cout << "ctor\n";
    }
    Complex(const Complex& other)
    {
        state = other.state;
        cout << "copy\n";
    }
    ~Complex()
    {
        cout << "dtor\n";
    }
};

int main() {
    function<void()> fn;
    {
	   Complex a;
	   fn = [a]{
            cout << a.state << "\n";
	   };
    }
	fn();
	return 0;
}
```
No virtuals. Supports capture list.

eugeneko:
@rokups I just realized that the storage here may actually be dynamic, so it would be full-functional std::function.
But... does it really worth it :trollface: ? I'd just use `std::function`.
If you call virtual function rarely, there's no problem.
If you call virtual function often, then you have a problem in your code, not in the function call.

rokups:
What part of `std::function` is virtual? I use bunch of lambdas for events and events are bread and butter of urho3d. So as it now we have two virtual calls for every event (one from `EventHandler11Impl` and one from `std::function`)? That does sound like it would be worth it to have `Urho3D::Function`.

eugeneko:
> What part of std::function is virtual?

Type-erased call, at least.

> That does sound like it would be worth it to have Urho3D::Function

Why? We can just get rid of all this pre-C++-11 garbage and store `std::function` everywhere.

rokups:
Well you said `std::function`. I have no idea how or where `std::function` calls virtuals so if im talking nonsense please dont mind me :)

PR #2476, opened 2019-08-14
More custom geometry

codexhound:
-Added shape creation functions in custom geometry
-Added shape protrusion
-Added circle generation
-Added circle graph generation
-Added sphere generation
-Added square generation
-Added mesh generation from vector of points

Math
-Added vector rotation
-Added degrees to radians and radians to degrees
-Added average function

PR #2496, opened 2019-09-04
Cleanup Math code.

eugeneko:
Okay, I'm doing cleanup here for PR #2476.
I don't want to be "that jerk who just discards other peoples' work", so I'm going to reason my decisions and I'm also giving you guys time to discuss these changes before merge if you have good arguments against this PR. No comments for a week == everybody agrees.

1) `enum AngleType` and related logic is removed completely.

Radians are not used in Urho API. Like, never.
https://urho3d.github.io/documentation/HEAD/_coding_conventions.html
This functionality is against code conventions, so it is to be removed.
`ToRadians` and `ToDegrees` are good on their own, so I kept them.

2) `Matrix3::MakeRotationMatrix` is removed.

There are _a lot_ of issues with this function.

- All the work with Euler angles in Urho is done via Quaternion, so this function is basically duplicating functionality.
- It is chunk of boilerplate math code.
- It violates code conventions and uses the opposite rotation direction comparing to other functions in Urho.
- It has Z*Y*X rotation order and Quaternion has Y*Z*X.
I don't say that one order is more right than another, but having two different orders for similar API is confusing.
- It just doesn't fit the rest of API and stick like a sore thumb.

So, all in all this function looks redundant, overcomplicated and misfit.
I would suggest using Quaternion(x, y, z) if order of rotation is acceptable.
If not, it is possible to write local helper OR extend Quaternion ctors in the same way I did in rbfx:
https://github.com/rokups/rbfx/blob/master/Source/Urho3D/Math/Color.h#L102

3) `Average`... refactored.

It shouldn't even compile because it references `Vector` that is never included.
I tried both to make it more generic and to fix this dependency.
And added check for zero, ofc.
It still looks shitty for me because of type deduction, but at least it looks like a propper c++ algorithm.

---

I honestly don't know what to do with `CustomGeomety`  part.
This is useful functionality that I personally need from time to time, so I don't want to outright revert it.
However, I have more issues to this code than I can fix in few evenings.

@ArnisLielturks, @rokups, @codexhound

### DIFF
--- a/Source/Urho3D/Graphics/CustomGeometry.cpp
+++ b/Source/Urho3D/Graphics/CustomGeometry.cpp
@@ -294,7 +294,7 @@ void CustomGeometry::MakeShape(const Vector<Vector3>& pointList , bool connectTa
             DefineVertex(current);
             DefineVertex(next);
         }
-        
+
     }
     Commit();
 }
@@ -311,7 +311,7 @@ void CustomGeometry::FillShape(const Vector<Vector3>& shapeList, bool connectTai
         BeginGeometry(usedGeomNum, PrimitiveType::TRIANGLE_STRIP);
         auto centerPoint = Vector3(0, 0, 0);
         if (connectTail) {
-            auto centerPoint = Urho3D::Average(shapeList);
+            auto centerPoint = Average(shapeList.Begin(), shapeList.End());
         }
         Vector<Vector3> vertices(3);
         Vector3 current;
@@ -334,7 +334,7 @@ void CustomGeometry::FillShape(const Vector<Vector3>& shapeList, bool connectTai
 
                 vertices = { centerPoint, current, next };
 
-                normal = Urho3D::Average(vertices);
+                normal = Average(vertices.Begin(), vertices.End());
                 normal.Normalize();
                 normal.Orthogonalize(normal);
                 DefineVertex(vertices.At(0));
@@ -371,10 +371,10 @@ void CustomGeometry::MakeSphere(float radius, size_t iterations)
     CreateQuadsFromBuffer(m_xyPoints, iterations, iterations, true);
 }
 
-void CustomGeometry::ProtrudeShape(const Vector<Vector3>& mShapeList, 
+void CustomGeometry::ProtrudeShape(const Vector<Vector3>& mShapeList,
     const Vector<Vector3>& mPointList, bool connectTail)
 {
-    Vector3 centerPoint = Urho3D::Average<Vector3>(mShapeList);
+    Vector3 centerPoint = Average(mShapeList.Begin(), mShapeList.End());
     Vector3 pointCurrent;
     Vector3 shapeCurrent;
     Vector3 shapePointVec;
@@ -392,7 +392,7 @@ void CustomGeometry::ProtrudeShape(const Vector<Vector3>& mShapeList,
         bufferCount++;
     }
 
-    
+
     int count = 0;
     while (pointIter != mPointList.End()) {
         shapeIter = mLastShapePos.Begin();
@@ -428,7 +428,7 @@ void CustomGeometry::CreateQuadsFromBuffer(const Vector<Vector3>& pointList, siz
     else {
         SetNumGeometries(1);
     }
-    
+
     //Create the quads from the buffer
     BeginGeometry(0, Urho3D::PrimitiveType::TRIANGLE_STRIP);
     for (size_t i = 0; i < zIterations; i++) {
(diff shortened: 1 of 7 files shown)


1vanK:
> No comments for a week == everybody agrees.

I totally agree with this practice. At the moment, some PRs just hang for months (y Box2D update for example). And people simply lose motivation to make new PRs. After the merger, there is a much greater chance that someone will find error (just something will stop working for someone). In the end, we can always roll back any PR if it is bad.

codexhound:
Sorry, didn't see this.  Looks good to me.

rokups:
> I don't want to be "that jerk who just discards other peoples' work"

You should see how bkaradzic handles pull requests in bgfx :trollface: Seriously now, listing reasons is good. This is the only way to make them learn 👍 

> CustomGeomety

* `CustomGeometry::GetCircleShape` and `CustomGeometry::GetSquareShape` are instance methods while they could be static. They could be returning data to user-supplied array+size as well.
* `CustomGeometry::MakeCircleGraph` - what is this useful for?
* `CustomGeometry::MakeShape` and `CustomGeometry::FillShape` are useful. I am unsure if they should be merged or not. After all it is making a shape. Filled or hollow is just a parameter.
* `CustomGeometry::MakeCircle`, `CustomGeometry::MakeSphere` and `CustomGeometry::MakeSquare` should be complemented with cone/cylinder/dome/plane/pyramid/torus. This would allow dropping .mdl files for these basic shapes. Besides it is useful to have ability to create these shapes with some configuration. For example creating cone and pyramid could be same function as pyramid is a cone where base is a rectangle. All of this will be greatly useful when i port csg implementation from godot.
* `CustomGeometry::ProtrudeShape` 👍 

I am not sure about using `Vector<>` for parameters. It is safer, yes, but it robs us of avoiding memory allocation when defining shapes whose points we already have generated. Be it simple shape that we can manually stuff points to an array or a shape that was deserialized from file and we have an array of points already.

Also these helpers seem they might be misplaced. Or maybe not. Calling these helpers allows mutating/overwriting contents of current `CustomGeometry` instance. Is this a desired property?

codexhound:
> 
> 
> > I don't want to be "that jerk who just discards other peoples' work"
> 
> You should see how bkaradzic handles pull requests in bgfx :trollface: Seriously now, listing reasons is good. This is the only way to make them learn 👍
> 
> > CustomGeomety
> 
>     * `CustomGeometry::GetCircleShape` and `CustomGeometry::GetSquareShape` are instance methods while they could be static. They could be returning data to user-supplied array+size as well.
These are static methods
> 
>     * `CustomGeometry::MakeCircleGraph` - what is this useful for?
I admit this can only be used for a very specific instance, to make a 3D graph. It's still useful though I'm not sure whether it still belongs here.
>     * `CustomGeometry::MakeShape` and `CustomGeometry::FillShape` are useful. I am unsure if they should be merged or not. After all it is making a shape. Filled or hollow is just a parameter.
They do two different things. One essentially connects points using the line method, and one draws triangles and fills using the center of a bunch of points so I don't think they could be easily integrated. Experience tells me it's better to not try to put too much functionality into one function anyways though.
> 
>     * `CustomGeometry::MakeCircle`, `CustomGeometry::MakeSphere` and `CustomGeometry::MakeSquare` should be complemented with cone/cylinder/dome/plane/pyramid/torus. This would allow dropping .mdl files for these basic shapes. Besides it is useful to have ability to create these shapes with some configuration. For example creating cone and pyramid could be same function as pyramid is a cone where base is a rectangle. All of this will be greatly useful when i port csg implementation from godot. Agreed, that's what I was going for originally, just too much work for the first pass.
> 
>     * `CustomGeometry::ProtrudeShape` 👍
> 
> 
> I am not sure about using `Vector<>` for parameters. It is safer, yes, but it robs us of avoiding memory allocation when defining shapes whose points we already have generated. Be it simple shape that we can manually stuff points to an array or a shape that was deserialized from file and we have an array of points already.
> Agreed, probably better to make the input more "dynamic"
> Also these helpers seem they might be misplaced. Or maybe not. Calling these helpers allows mutating/overwriting contents of current `CustomGeometry` instance. Is this a desired property? True, originally it was meant to operate on a blank instance generating some basic mesh, although other suggestions for usage and implementation are good. Currently once the basic shape has been generated any other operations would have to be done completely manually as before.

eugeneko:
I had an idea for a while... To create something like `ModelBuilder`. Utility class that lets user easily construct geometry from code. I even had some prototypes of this class.

And this new code in `CustomGeometry` is something that really fits this idea. But this code is misplaced now: `CustomGeometry` is not responsible for building stuff, it is responsible for rendering.

PR #2499, opened 2019-09-05
Merge update to SDL 2.0.10

JTippetts:
From weitjong's fork.


weitjong:
At least the first commit should be included. There could be others too that should just stay not being merged to upstream as it is specific to my fork.

weitjong:
Once the CI cleared this dev branch https://github.com/urho3d/Urho3D/tree/upstream-upgrade-sdl then it will be merged into master. Any breakage will be fixed in the master branch.

Issue #2501, opened 2019-09-06
Improve Materials Documentation


SirNate0:
The [Materials](https://urho3d.github.io/documentation/HEAD/_materials.html) documentation page is a wonderful resource, but presently it does not specify what the default values will be. I don't mind fixing this myself over the next few days, but I was wondering what people thought would be the best way to specify it? Some options I've thought of:

1. A separate XML section with values populated with the default material values
2. Comments within the existing section `<cull value="cw|ccw|none" /><!--default ccw-->`
3. A secondary attribute `<cull value="cw|ccw|none" default="ccw" />` and perhaps a note that the default attributes specify the default for the value, they are not required in the material file itself
4. Something along the lines of `<cull value="cw|ccw|none = ccw" />`
5. Minimal effort: just adding a reference to Material::ResetToDefaults() and having people look in the code (bonus: the documentation won't get out of date).

My preferred approaches are 1, which then provides a concrete example of a material definition in the documentation, and 3, which places the information next to the relevant part of the specification while avoiding looking cleaner/clearer (in my opinion) than 4. Downsides of 1 - it basically duplicates the documentation that is there presently. Downsides of 3 - it could lead to some confusion about a default attribute being available to set the default for the engine.

For those interested, default values are set in `Material::ResetToDefaults()`

weitjong:
IMHO you can try to use the HTML tags (e.g. ```<strong>``` or ```<span class='default'>``` to enclose  the text representing the default value of an attribute. The CSS could be adjusted to style how it will looks afterward. A meta program/script to auto update the tag if/when the code default value actually change will be just a bonus.

SirNate0:
While I certainly like the idea of using formatting to display it instead, I ended up not taking that approach because of the increased difficulty in maintaining the documentation that would impose and because I felt there wouldn't be a good way to do that for the options that aren't enumerated (the technique name, for example, rather than the cull value).

PR #2502, opened 2019-09-07
Included default material definition in the documentation.

SirNate0:
My initial solution to #2501

I added a Docs/CreateDefaultMaterial.as script to save the default material values and included the default material in the documentation. Note that this script is not called anywhere by the build system, the Docs/DefaultMaterial.xml will require manual updates (the script is there to make those easier). 

I'm open to other approaches, for example we could leave out the Docs/DefaultMaterial.xml file and just past the contents in a \code block in Docs/Reference.dox instead. Depending on the version of doxygen we're using, this might be better: for my local version (1.8.11) the \include command seems to erroneously add line numbers to the block, which make copying and pasting the contents difficult and clutters it up unnecessarily for such a short file, as well as simply looking different than the code block above it. We could also go with some of the other proposed options from the issue, but I felt this was one of the simplest to implement and easiest to maintain approaches.

### DIFF
--- a/Docs/CreateDefaultMaterial.as
+++ b/Docs/CreateDefaultMaterial.as
@@ -1,0 +1,7 @@
+void Start()
+{
+Material@ mat = Material();
+mat.Save("DefaultMat.xml");
+Print("Saved default material to DefaultMat.xml");
+engine.Exit();
+}
--- a/Docs/DefaultMaterial.xml
+++ b/Docs/DefaultMaterial.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0"?>
+<material>
+	<technique name="Techniques/NoTexture.xml" quality="0" loddistance="0" />
+	<parameter name="UOffset" value="1 0 0 0" />
+	<parameter name="VOffset" value="0 1 0 0" />
+	<parameter name="MatDiffColor" value="1 1 1 1" />
+	<parameter name="MatEmissiveColor" value="0 0 0" />
+	<parameter name="MatEnvMapColor" value="1 1 1" />
+	<parameter name="MatSpecColor" value="0 0 0 1" />
+	<parameter name="Roughness" value="0.5" />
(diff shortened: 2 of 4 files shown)


ArnisLielturks:
@weitjong What are your thoughts about this? I think it would be better if we would add the default material directly in the documentation and avoid any unnecessary files in the build tree.

weitjong:
I am not in favor of this PR simply because it requires manual updates. If we merge this as is then we would need a person to maintain this XML file long term and I don't believe anyone would like to do that. To me, the only viable way is to do it right the first time by having a new automation logic added somewhere in the CI that updates the default values in one form or another in our documentation (see my comment in #2501). If you notice almost all of our online documentation are kept up-to-date automatically by our CI, with a few exception like adding the new contributors to the list. I believe I have created an issue for that because I am getting lazy to make a commit just for doing this mundane task again and again. So, I vote against this PR.

ArnisLielturks:
Other option might be just adding this info in the Github wiki, since everyone can edit it and keep it up to date.

ArnisLielturks:
@SirNate0 could you move this documentation to the Github wiki please?

1vanK:
We really need this spam from github-actions? I'm tired of cleaning the notification list.

weitjong:
> We really need this spam from github-actions? I'm tired of cleaning the notification list.

If you are receiving the notifications for every issues/PRs that are being marked (before they get swept), even for those that you have not participated with then it is because you have opted yourself to be a watcher for this repository. You can opt-out to be watcher or setting up an auto-filter for your inbox for the time being until the bot has done its work. I have tried to limit the rate to not more than 10 issues/PRs per day. If you do that, however, you will miss the chance to keep the issue/PR that may of interest to you alive.

SirNate0:
> @SirNate0 could you move this documentation to the Github wiki please?

@ArnisLielturks done. Sorry it took so long.

PR #2503, opened 2019-09-07
Fixed console styling to account for the addition of LOG_TRACE

SirNate0:
LOG_TRACE offset the log levels by 1 but the console styling did not add a
trace styling, so warnings were printed with error styling, etc.

I've not tested this much, but if you just open up the console in one of the samples it seems to print as expected. (To get an error you can run an invalid command through the FileSystem and to get regular text you can run echo hello or something like that)

I went with a sort of yellow-gray for the trace text styling. Feel free to change it as you see fit, I was more concerned with getting errors and warnings to be the right color than choosing the best color for the trace text, so I just went with something that seemed maybe reasonable (I haven't actually seen it as text, so that remains to be confirmed).

### DIFF
--- a/Source/Urho3D/Engine/Console.cpp
+++ b/Source/Urho3D/Engine/Console.cpp
@@ -51,6 +51,7 @@ static const int DEFAULT_HISTORY_SIZE = 16;
 
 const char* logStyles[] =
 {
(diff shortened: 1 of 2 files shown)


weitjong:
Thanks!

PR #2505, opened 2019-09-15
Added check for setting cursor to current

Modanung:
This change prevents the following code from resulting in a segmentation fault:

```
UI* ui{ GetSubsystem<UI>() };
ui->SetCursor(new Cursor(context_));
ui->SetCursor(ui->GetCursor());
```

Thanks to @TIG3R99 for [locating](https://discourse.urho3d.io/t/changing-cursors/5581) this bug.


Modanung:
Beyond _line 158_ it's just line endings that changed, btw.

weitjong:
Thanks!

PR #2506, opened 2019-09-19
fix bug for SpritePacker in Tools

error408:
If the forth parameter is "packerInfos.Size()" in "stbrp_init_target", I cannot pack three image(1920 * 1080、355 * 2 and 2 * 774) into a big image（2048 * 2048）.
In nothing/stb/stb_rect_pack.h comments, note that to guarantee best results, either:
1. make sure 'num_nodes' >= 'width'
or  2. call stbrp_allow_out_of_mem() defined below with 'allow_out_of_mem = 1'

See it in [](https://github.com/nothings/stb/blob/master/stb_rect_pack.h) line 124 to 143

:( bad on English


error408:
And "stbrp_pack_rects" will return success or fail

weitjong:
Thanks!

PR #2510, opened 2019-09-22
DebugRenderer additions / bugfix

LeithKetchell:


Modanung:
I think it would make sense to split these modifications up into multiple PRs. The bug fix may need little discussion, as opposed to the other changes.

weitjong:
This seems to be a duplicate PR from #2412 without showing any sign of improvement.

weitjong:
When I have time I will also probably install a bot to auto close all the stale PR and issues.

PR #2511, opened 2019-09-26
Fix HiresTimer overflow bug.

sssooonnnggg:
When the application is running for a long time in a browser environment,  emscripten_get_now()*1000.0 may larger than 2^32-1, so HiresTick() may return wrong value, which causes timeStep_ is always 0.0f.


weitjong:
Thanks!

PR #2512, opened 2019-10-01
Suppress internal warnings of the library

reconndev:
I have lots of warnings enabled in my code such as below and the library reported on its internals which made it difficult to read what is actually needed.
    -Wall
    -Wextra
    -Weffc++
    -Wpedantic

This change fixes it entirely.


weitjong:
The proposed changes and the PR description are different. Also, I don’t believe suppressing all warnings is a right thing to do. We may do that for a good  reason only, e.g when then log entries would exceed the limits of CI, and in that case, we only set it just for CI and not for general.

reconndev:
What I did was to put the library headers into "-isystem" compiler flag so that the internal warnings are not displayed. I guess this is not a good idea to totally wipe them out but a suggestion to be able to trigger those.

PR #2513, opened 2019-10-11
Added a Matrix3x4 overload for Node::SetWorldTransform()

SirNate0:
Added a Matrix3x4 overload for Node::SetWorldTransform(), removed some 'both's that referred to three things from the comments.


weitjong:
Thanks!

Issue #2414, opened 2019-01-22
IOS PVRTC Texture data size error


Yuki001:
Hi, 
When I use PVRTC texture in IOS, I get a GL_INVALID_VALUE error in glCompressedTexImage2D function call.
I found the data size of PVRTC texture is wrong , PVRTC1_4BPP 's data size should be the same of ETC1.

OGLTexture.cpp : line 238 
function Texture::GetRowDataSize
```
    case COMPRESSED_RGB_PVRTC_4BPPV1_IMG:
    case COMPRESSED_RGBA_PVRTC_4BPPV1_IMG:
       return (unsigned)(width * 4 + 7) >> 3u;

    case COMPRESSED_RGB_PVRTC_2BPPV1_IMG:
    case COMPRESSED_RGBA_PVRTC_2BPPV1_IMG:
       return (unsigned)(width * 2 + 7) >> 3u;
```
should be changed to 
```
    case COMPRESSED_RGB_PVRTC_4BPPV1_IMG:
    case COMPRESSED_RGBA_PVRTC_4BPPV1_IMG:
        return ((unsigned)(width + 3) >> 2u) * 8; // = ETC1 

    case COMPRESSED_RGB_PVRTC_2BPPV1_IMG:
    case COMPRESSED_RGBA_PVRTC_2BPPV1_IMG:
        return ((unsigned)(width + 3) >> 2u) * 4;
```

This change makes my PVRTC texture load successful.

weitjong:
I have not used this before, so I could be wrong. However, according to this spec then our current implementation seems to be correct. https://www.khronos.org/registry/OpenGL/extensions/IMG/IMG_texture_compression_pvrtc.txt

Yuki001:
According to Texture.cpp : Line 150
```
unsigned Texture::GetDataSize(int width, int height) const
{
    if (IsCompressed())
        return GetRowDataSize(width) * ((height + 3) >> 2u);
    else
        return GetRowDataSize(width) * height;
}
```
The full data size for a compress image is GetRowDataSize(width) * height / 4, so GetRowDataSize(width) for compress image = 4 * real row size 

According to the spec : 
Image Size = ( max(width, 8) * max(height, 8) * 4 + 7) / 8
Real Row Size =  ( max(width, 8) * 4 + 7) / 8
GetRowDataSize(width) = ( max(width, 8) * 4 + 7) / 8 * 4 ≈ width * 2 

ETC1 's data size is somewhat correct.
[...]

PR #2514, opened 2019-10-14
Fix wrong pvrtc data size.

sssooonnnggg:
This caused pvr textures can't display on iOS devices. #2414

### DIFF
--- a/Source/Urho3D/Graphics/Graphics.h
+++ b/Source/Urho3D/Graphics/Graphics.h
@@ -679,6 +679,8 @@ private:
     bool dxtTextureSupport_{};
     /// ETC1 format support flag.
     bool etcTextureSupport_{};
+	/// ETC2 format support flag.
+	bool etc2TextureSupport_{};
     /// PVRTC formats support flag.
     bool pvrtcTextureSupport_{};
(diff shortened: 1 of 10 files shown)


eugeneko:
I cannot really check this fix now.
However, I recommend to use parentheses -- op priority around bit shifts is tricky sometimes.

rokups:
Hey @sssooonnnggg could you please provide a sample texture that failed to load on iOS and now works after fix is applied?

Also you should not mix multiple things in a single PR. etc2 support should be submitted in another PR. Also mind formatting there ;)

eugeneko:
The initial fix cannot work.
It makes me suspicious -- how did you test your fix?

I also wonder if it's possible to use simpler math for `PVRTC`, similar to `DXT` formats. I'm not an expert, tho.

sssooonnnggg:
My bad， I will close this PR and create new PRs.

rokups:
You did not have to close it. You may clean up your branch and force-push it.

sssooonnnggg:
@rokups I didn' realize that I had mix multiple things on the master branch. I have created two separate branches to fix this. Thans a lot . ：）

PR #2515, opened 2019-10-14
Fix wrong pvrtc data size.

sssooonnnggg:
This caused pvr textures can't display on iOS devices.

Texture for testing :  [https://tmpfiles.org/dl/28949/bj.pvr](url)

### DIFF
--- a/Source/Urho3D/Graphics/OpenGL/OGLTexture.cpp
+++ b/Source/Urho3D/Graphics/OpenGL/OGLTexture.cpp
@@ -237,11 +237,11 @@ unsigned Texture::GetRowDataSize(int width) const
 
     case COMPRESSED_RGB_PVRTC_4BPPV1_IMG:
     case COMPRESSED_RGBA_PVRTC_4BPPV1_IMG:
-        return (unsigned)(width * 4 + 7) >> 3u;
+        return ((unsigned)(width + 3) >> 2u) * 8;
 
     case COMPRESSED_RGB_PVRTC_2BPPV1_IMG:
     case COMPRESSED_RGBA_PVRTC_2BPPV1_IMG:
-        return (unsigned)(width * 2 + 7) >> 3u;
+        return ((unsigned)(width + 7) >> 3u) * 8;
 
     default:
         return 0;


eugeneko:
My question stands -- are you sure this code is correct?
Let's say we have 1x1 texture. It should be 8 bytes (1 block) for both formats, if I understand things right.
But `((unsigned)(width * 4 + 7) >> 3u) * 4` is `(4+7)/8*4 = 11/8*4 = 4`
and `((unsigned)(width * 2 + 7) >> 3u) * 4` is `(2+7)/8*4 = 9/8*4 = 4`

sssooonnnggg:
Hi, this fix just multiplied the origin result by 4 .
When I run my application on iOS device, the pvr texture can't display properly.
Then I add some logs at this location (OGLTexture2D.cpp, line 166):

```c++
if (wholeLevel)
    // log the result of GetDataSize(width, height)
    glCompressedTexImage2D(target_, level, format, width, height, 0, GetDataSize(width, height), data);
```
However, the ouput is 1/4 of the correct size which I got from some Webgl test code.

Webgl sample code : 
```javascript
    const gl = this.gl;

    let ext = gl.getExtension('WEBGL_compressed_texture_pvrtc');

    const texture = gl.createTexture();
    gl.activeTexture(gl.TEXTURE0);
    gl.bindTexture(gl.TEXTURE_2D, texture);
    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.NEAREST);
    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.NEAREST);


    const PVR_HEADER_LENGTH = 13; // The header length in 32 bit ints.
    const header = new Int32Array(this.textureData, 0, PVR_HEADER_LENGTH);

    const PVR_HEADER_MAGIC = 0;
    const PVR_HEADER_FORMAT = 2;
    const PVR_HEADER_HEIGHT = 6;
    const PVR_HEADER_WIDTH = 7;
    const PVR_HEADER_MIPMAPCOUNT = 11;
    const PVR_HEADER_METADATA = 12;

    const format = header[PVR_HEADER_FORMAT];
    // const internalFormat = PVR_TO_FORMAT[format] || -1;

    const width = header[PVR_HEADER_WIDTH];
    const height = header[PVR_HEADER_HEIGHT];
    const levels = header[PVR_HEADER_MIPMAPCOUNT];
    const dataOffset = header[PVR_HEADER_METADATA] + 52;
    console.log(dataOffset);
    const pvrtcData = new Uint8Array(this.textureData, dataOffset);
    console.log(pvrtcData.length);
    gl.compressedTexImage2D(gl.TEXTURE_2D, 0, ext.COMPRESSED_RGBA_PVRTC_4BPPV1_IMG, 256, 256, 0, new Uint8Array(pvrtcData));
```

Then I got : `pvrtcData.length` = `GetDataSize(width, height)` * 4 .

Maybe the origin implementation `GetRawDataSize` was totally wrong, but this fix works for iOS device when loading pvr(`PVRTC_4BPP`).

Sorry for my poor English. :)

eugeneko:
> Maybe the origin implementation GetRawDataSize was totally wrong

I _think_ this statement is correct and PVRTC should be implemented in the same way as DXT:

`((unsigned)(width + 3) >> 2u) * 8` for 4bpp, and
`((unsigned)(width + 7) >> 3u) * 8` for 2bpp

This is all I got from format specs. I'd like to hear out others before doing anything.

eugeneko:
@sssooonnnggg Can you please try formulas from my previous comment? I don't have environment to test them, but they should work in all cases?

sssooonnnggg:
@eugeneko 

I think the formulas have the same problem. When width = 2, ((unsigned)(width + 3) >> 2u) * 8 = 8, and ((unsigned)(width + 7) >> 3u) * 8 =  8.

After reading pvr format specs, I think the formulas should be quiet simple :

`width * 2`  for 4bpp, and
`width` for 2bpp.

Because pvr texure requires the width should be power of 2, so width can only be 2, 4, 8, 16, etc.

4bpp means 4 bits per pixel, by other means 1/2 byte per pixel. Consider the code below:
```c++
unsigned Texture::GetDataSize(int width, int height) const
{
    if (IsCompressed())
        return GetRowDataSize(width) * ((height + 3) >> 2u);
    else
        return GetRowDataSize(width) * height;
}
```

If texture is compressed, the result is divided by 4, so we should return `width * 2` at `GetRawDataSize` for `4bpp`, and return `width`  for `2bpp`.

eugeneko:
> I think the formulas have the same problem. When width = 2, ((unsigned)(width + 3) >> 2u) * 8 = 8, and ((unsigned)(width + 7) >> 3u) * 8 = 8.

Yeah, these numers looks valid, what's the problem?

> After reading pvr format specs, I think the formulas should be quiet simple

These formulas just cannot work, because 1x1, 2x2 and 4x4 textures (lowest MIPs) must have the same row size -- 8 bytes (1 block). And your formulas are obviously linear.

sssooonnnggg:
Yes, you're right, I misunderstood specs.

I have tested formulas which you provided on iOS devices, and it woked properly.

Thanks for your explaination.

eugeneko:
Thanks, I'll merge it as soon as CI completes.

Issue #2181, opened 2017-11-21
OpenGL ES 3 Support


sofmz:
It would be nice if Urho3D supports OpenGL ES 3, there are couple of features that are very useful. One example is the standardised ETC2 compressed texture. Urho users that are targeting for mobile platforms now need to work with different compressed texture format and different restrictions for Android and iOS, supporting ES3 could make the ETC2 possible to be used. 

Most mobile devices nowadays support ES3 , in fact most mainstream mobile devices that are released in past 2-3 years support OpenGL ES3.

weitjong:
Sorry for the fat finger.

rokups:
You got me worried there for a moment

PR #2516, opened 2019-10-14
Add ETC2 textures support.

sssooonnnggg:
#2181 

Add ETC2 textures support for mobile devices and mobile browsers (WebGL).


eugeneko:
I'm ok with this PR.
But commits will be squashed before merge.

sssooonnnggg:
OK, commits are squashed.

Source/Urho3D/Resource/DecompressETC2.h
```diff
@@ -0,0 +1,14 @@
+
```
weitjong: The header and implementation files are missing the Urho3D copyright notice. Also, just for consistency sake, I think all the newly added functions should be declared using `Urho3D` namespace.

Source/Urho3D/Resource/Image.cpp
```diff
@@ -704,6 +725,16 @@ bool Image::BeginLoad(Deserializer& source)
...
+        case 22:
+            compressedFormat_ = CF_ETC2_RGB;
+            components_ = 3;
+            break;
+
+        case 23:
+            compressedFormat_ = CF_ETC2_RGBA;
+            components_ = 4;
+            break;
+
```
weitjong: I actually wanted to ask why there is duplication here but forgot to ask it in the earlier review. Care to enlighten us? Perhaps you might as well explain the reason as code comment, if they are intentional.

1vanK:
Why did you create a new file DecompressETC2.h? Why not just add code to Decompress.h

1vanK:
We already have ETC supputs. I don’t know about this format, but are you sure that they have nothing in common? It looks like you're re-writing functions to parse all data blocks (Sorry if I'm wrong, I glanced at the code briefly)

sssooonnnggg:
@1vanK The old decompressETC code can't support the ETC2(diffrerent from ETC1) format(with or without alpha channel),and is also copied from another library.

Decompressetc2.cpp code is from libctx, which is not a part of libktx's interface but some kind of its implementation. I think it is not necessary to add the full library to our code base.Maybe need more license file?

Source/Urho3D/Resource/Image.cpp
sssooonnnggg: @weitjong  Both .ktx files and .pvr files can support etc2 image format, their difference is just the header.

weitjong:
IANAL but since 1vanK has already brought the thing up then it needs to be looked into now. The part that states "it is in unmodified form" is problematic. I think moving that source file in its original name unmodified to the 3rd-party libs is the right way to go. We can easily add the original license file there like all our 3rd-party deps.

Source/Urho3D/Resource/Image.cpp
weitjong: I see. If you could add a short comment line to say which is which that would be great.

sssooonnnggg:
OK, I will move that file to.rhe 3rd-party libs.

rokups:
Hey @sssooonnnggg this is a great PR. Absolutely love it that you added decompression code which i did not bother with.  I noticed that gist you linked also has ETC1 decompression. If you are including that maybe it would be worthwhile to also make ETC1 decompression use your included lib and delete current ETC1 decompression code?

sssooonnnggg:
@rokups Good suggestion, I will test new decompression code on ETC1 format and then delete current ETC1 decompression code. :)

sssooonnnggg:
I have add ETCPACK as a 3rd-party lib and use it to decompress both ETC1/ETC2 texture, it works properly now.

Source/Urho3D/Resource/Decompress.cpp
```diff
@@ -269,187 +278,7 @@ void DecompressImageDXT(unsigned char* rgba, const void* blocks, int width, int
...
-                    }
-                }
-            }
-
-            // advance
-            sourceBlock += bytesPerBlock;
-        }
-    }
-}
+// PVRTC decompression based on the Oolong Engine, modified for Urho3D
```
eugeneko: Is this comment relevant?

sssooonnnggg: It just changed comment
`// ETC and PVRTC decompression based on the Oolong Engine, modified for Urho3D`
To
`// PVRTC decompression based on the Oolong Engine, modified for Urho3D`

Source/CMakeLists.txt
```diff
@@ -68,7 +68,7 @@ if (URHO3D_LIB_TYPE STREQUAL SHARED AND NOT WIN32 AND NOT ANDROID AND NOT IOS AN
 endif ()
 
 # Add targets
-foreach (TARGET FreeType LZ4 PugiXml rapidjson SDL StanHull STB)
+foreach (TARGET FreeType LZ4 PugiXml rapidjson SDL StanHull STB ETCPACK)
```
weitjong: Keep the list sorted, please.

Source/ThirdParty/ETCPACK/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+# Define target name
+set (TARGET_NAME ETCPACK)
+
+# Define source files
+set (SOURCE_FILES "source/etcdec.cxx")
```
weitjong: For consistency with the rest of the CMake build script, remove the unnecessary double quotes.

weitjong:
Also, need to append the EXTPACK license to the https://github.com/urho3d/Urho3D/blob/master/Source/ThirdParty/LICENSES

weitjong:
Thanks,

PR #2518, opened 2019-10-16
Optimize shader file read.

Y-way:


eugeneko:
What's the performace impact, e.g. with standard Urho shaders?

iSLC:
It doesn't optimize much. Just implements single line comments. At least that's what I see from the changes. I'm on phone right now so I can't tell much.

eugeneko:
@iSLC Manual parsing of #include directives already breaks line numbers, so I'm not that worried about it.

I'm mostly interested about "optimization" part, since the PR is named this way.
If you say "optimize", then there should be measurements to prove that you are actually optimizing, unless change is simple and obvious (like copy elision or caching).

Y-way:
> @iSLC Manual parsing of #include directives already breaks line numbers, so I'm not that worried about it.
> 
> I'm mostly interested about "optimization" part, since the PR is named this way.
> @Y-way If you say "optimize", then there should be measurements to prove that you are actually optimizing, unless change is simple and obvious (like copy elision or caching).

My describe mistake.I mean that reading shader file with this code will prevent some mistake while comment lines and '#include 'lines have space character at begining.

bin/CoreData/Shaders/GLSL/AutoExposure.glsl
```diff
@@ -26,17 +26,18 @@ float GatherAvgLum(sampler2D texSampler, vec2 texCoord, vec2 texelSize)
     return lumAvg / 4.0;
 }
 #endif
-
-void VS()
+#ifdef COMPILEVS
+void main()
```
rokups: Seems to me backwards compatibility is broken here for no good reason.

eugeneko:
@Y-way Am I right that this PR doest these two things?:

1) Support spaces before `#include`

2) Avoid errors in commented lines.

Can you please provide an example of such error? When it happens?

I think I know what are you talking about, but I'm not sure.

Y-way:
> @Y-way Am I right that this PR doest these two things?:
> 
> 1. Support spaces before `#include`
> 2. Avoid errors in commented lines.
> 
> Can you please provide an example of such error? When it happens?
> 
> I think I know what are you talking about, but I'm not sure.

Yes, you're right.
https://github.com/urho3d/Urho3D/blob/68635536db18277495f0bc27fdfcd19b24fb7261/Source/Urho3D/Graphics/Shader.cpp#L183
If there are blank characters at the begining of '#include', it will record errors in log file like below.

ERROR: Failed to compile vertex shader LitSolid(DIRLIGHT NORMALMAP PERPIXEL):
0(8) : error C7529: OpenGL does not allow #include directives
0(9) : error C0109: Can't open include file "Uniforms.glsl"
0(134) : error C1503: undefined variable "cUOffset"
0(134) : error C1503: undefined variable "cUOffset"
0(134) : error C1503: undefined variable "cVOffset"
0(134) : error C1503: undefined variable "cVOffset"
0(138) : error C1503: undefined variable "cViewProj"
0(143) : error C1503: undefined variable "cClipPlane"
0(149) : error C1503: undefined variable "cZone"
0(153) : error C1503: undefined variable "cDepthMode"
0(280) : error C1503: undefined variable "cViewInv"
0(280) : error C1503: undefined variable "cViewInv"
0(280) : error C1503: undefined variable "cViewInv"
0(281) : error C1503: undefined variable "cViewInv"
0(281) : error C1503: undefined variable "cViewInv"
....

rokups:
> Support spaces before #include

`#include` statement mimics c preprocessor which does not support spaces before include. Shaders should not support that either. A better error reporting would be preferable.

eugeneko:
@Y-way I'm mostly curious why do you remove comments. Does it have functional meaning?

Y-way:
> @Y-way I'm mostly curious why do you remove comments. Does it have functional meaning?

IMO, If we deal it at CPU, it will not do again at GPU, and removing comments  will reduce memory.

nem0:
@rokups https://gcc.gnu.org/onlinedocs/gcc-2.95.3/cpp_1.html whitespace is allowed before #

Y-way:
I did a test for loading shader.
The below picture show the memory need between changing before and now.
The left side is before and right side is now.
Loading speed as before, but, the memory needed reduce nearly half.

Shader

iSLC:
Memory doesn't seem to be a problem. As indicated by the fact that it took less time to process when there was more memory allocated.

Try allocating memory in advance. I suspect you'll win at both memory savings and processing time.

eugeneko:
@Y-way Just curious -- how did you measure "memory" column?

Y-way:
@iSLC Thank you for your suggestion.Let me think more.

Y-way:
@eugeneko What your mean is how do i get the value of memory?
https://github.com/urho3d/Urho3D/blob/9c6e6b12e19b0a2546b1f026a48049b6f4627d3c/Source/Urho3D/Resource/Resource.h#L92

Modanung:
Doesn't this application discrimination create incompatibilities of liberties?

> THE LICENSEE ACKNOWLEDGES AND ACCEPTS THAT THE SOFTWARE (I) IS **NOT
LICENSED FOR**; (II) IS NOT DESIGNED FOR OR INTENDED FOR; AND (III) **_MAY
NOT BE USED FOR_**; ANY MISSION CRITICAL APPLICATIONS SUCH AS, BUT NOT
LIMITED TO OPERATION OF NUCLEAR OR HEALTHCARE COMPUTER SYSTEMS AND/OR
NETWORKS, AIRCRAFT OR TRAIN CONTROL AND/OR COMMUNICATION SYSTEMS OR
ANY OTHER COMPUTER SYSTEMS AND/OR NETWORKS OR CONTROL AND/OR
COMMUNICATION SYSTEMS ALL IN WHICH CASE THE FAILURE OF THE SOFTWARE
COULD LEAD TO DEATH, PERSONAL INJURY, OR SEVERE PHYSICAL, MATERIAL OR
ENVIRONMENTAL DAMAGE. LICENSEE'S RIGHTS UNDER THIS LICENSE WILL
TERMINATE AUTOMATICALLY AND IMMEDIATELY WITHOUT NOTICE IF LICENSEE
FAILS TO COMPLY WITH THIS PARAGRAPH.

Issue #2517, opened 2019-10-16
There is a problem in TileMap module


Y-way:
The renderer could not properly display tiles that are flipped and rotated.

https://github.com/urho3d/Urho3D/blob/68635536db18277495f0bc27fdfcd19b24fb7261/Source/Urho3D/Urho2D/TileMapLayer2D.cpp#L360
https://github.com/urho3d/Urho3D/blob/68635536db18277495f0bc27fdfcd19b24fb7261/Source/Urho3D/Urho2D/TileMapLayer2D.cpp#L389
If the tile-block set swapxy true, StaticSprite2D maybe render incorrectness.
https://github.com/urho3d/Urho3D/blob/68635536db18277495f0bc27fdfcd19b24fb7261/Source/Urho3D/Urho2D/StaticSprite2D.cpp#L295-L337

[TMX-Map format tile-flipping](https://docs.mapeditor.org/en/stable/reference/tmx-map-format/#tile-flipping)
"When rendering a tile, the order of operation matters. The diagonal flip (x/y axis swap) is done first, followed by the horizontal and vertical flips."

![map1](https://user-images.githubusercontent.com/4865384/66905633-ca0a0580-f038-11e9-8267-548de45364cf.png)

[TileMapTest.zip](https://github.com/urho3d/Urho3D/files/3733482/TileMapTest.zip)

PR #2519, opened 2019-10-16
Fix tilemap display problem

Y-way:
The renderer could not properly display tiles that are flipped and rotated.
[#2517](https://github.com/urho3d/Urho3D/issues/2517)


eugeneko:
Why do we need duplicate conditions for `flipX_`, `flipY_` and `swapXY_`?
This code was unclear even before these changes, now it's even more obscure.

Are you sure it cannot be simpler? Lemme think about this code for a while.

Y-way:
Sorry,I haven`t found any other better way to make it simpler.

Y-way:
Make `StaticSprite2D: : UpdateSourceBatches()` logic clearer

Modanung:
I reordered things a bit, what do you think?
https://github.com/Y-way/Urho3D/pull/1

Modanung:
When testing the tile flipping and saving a TMX with the latest version of Tiled I get:
`ERROR: Invalid version`

Modanung:
The map displays, but the flipping needs some work.

Y-way:
Can this pr be merged?

PR #2522, opened 2019-10-17
Added rotate and flip attributes to BorderImage

MystMagus:


Source/Urho3D/UI/BorderImage.cpp
```diff
@@ -42,6 +42,9 @@ BorderImage::BorderImage(Context* context) :
     imageBorder_(IntRect::ZERO),
     hoverOffset_(IntVector2::ZERO),
     disabledOffset_(IntVector2::ZERO),
+    rotate_(0),
```
Modanung: `bool` initialized with numeral.

PR #2524, opened 2019-10-17
Lineedit filter

MystMagus:
Added the ability to set a character filter to line edits.

### DIFF
--- a/Source/Urho3D/Container/Str.cpp
+++ b/Source/Urho3D/Container/Str.cpp
@@ -892,6 +892,54 @@ String String::SubstringUTF8(unsigned pos, unsigned length) const
     return ret;
 }
 
+unsigned String::FindUTF8(unsigned c, unsigned startPos) const
+{
+    for (unsigned i = startPos; i < LengthUTF8(); ++i)
+    {
+        if (AtUTF8(i) == c)
+            return i;
+    }
+
+    return NPOS;
+}
+
+unsigned String::FindUTF8(const String& str, unsigned startPos) const
+{
+    if (str.Empty() || str.LengthUTF8() > LengthUTF8())
+        return NPOS;
+
+    unsigned first = str.AtUTF8(0);
+    for (unsigned i = startPos; i <= LengthUTF8() - str.LengthUTF8(); ++i)
+    {
+        unsigned c = AtUTF8(i);
+        if (c == first)
+        {
+            unsigned skip = NPOS;
+            bool found = true;
+            for (unsigned j = 1; j < str.LengthUTF8(); ++j)
+            {
+                c = AtUTF8(i + j);
+                char d = str.AtUTF8(j);
+
+                if (skip == NPOS && c == first)
+                    skip = i + j - 1;
+
+                if (c != d)
+                {
+                    found = false;
+                    if (skip != NPOS)
+                        i = skip;
+                    break;
+                }
+            }
+            if (found)
+                return i;
+        }
+    }
+
+    return NPOS;
+}
+
 void String::EncodeUTF8(char*& dest, unsigned unicodeChar)
 {
     if (unicodeChar < 0x80)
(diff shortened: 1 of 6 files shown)


Source/Urho3D/Container/Str.h
```diff
@@ -409,10 +409,10 @@ class URHO3D_API String
     /// Join substrings with a 'glue' string.
     void Join(const Vector<String>& subStrings, const String& glue);
     /// Return index to the first occurrence of a string, or NPOS if not found.
-    unsigned Find(const String& str, unsigned startPos = 0, bool caseSensitive = true) const;
-    /// Return index to the first occurrence of a character, or NPOS if not found.
     unsigned Find(char c, unsigned startPos = 0, bool caseSensitive = true) const;
     /// Return index to the last occurrence of a string, or NPOS if not found.
+    unsigned Find(const String& str, unsigned startPos = 0, bool caseSensitive = true) const;
+    /// Return index to the first occurrence of a character, or NPOS if not found.
```
weitjong: Looks like the code and comment are messed up here.

Source/Urho3D/Container/Str.cpp
```diff
@@ -892,6 +892,54 @@ String String::SubstringUTF8(unsigned pos, unsigned length) const
...
+    {
+        unsigned c = AtUTF8(i);
+        if (c == first)
+        {
+            unsigned skip = NPOS;
+            bool found = true;
+            for (unsigned j = 1; j < str.LengthUTF8(); ++j)
+            {
+                c = AtUTF8(i + j);
+                char d = str.AtUTF8(j);
```
weitjong: Should be `unsigned`?

Source/Urho3D/UI/LineEdit.cpp
```diff
@@ -511,6 +515,24 @@ void LineEdit::SetCursorBlinkRate(float rate)
 void LineEdit::SetMaxLength(unsigned length)
 {
     maxLength_ = length;
+    if(line_.Length() > maxLength_)
```
weitjong: spacing

Source/Urho3D/UI/LineEdit.cpp
```diff
@@ -511,6 +515,24 @@ void LineEdit::SetCursorBlinkRate(float rate)
...
+        UpdateCursor();
+    }
+}
+
+void LineEdit::SetTextFilter(const String& filter)
+{
+    filter_ = filter;
+    if(FilterText(line_))
+    {
+        if(cursorPosition_ > line_.LengthUTF8())
```
weitjong: spacing

Source/Urho3D/UI/LineEdit.cpp
```diff
@@ -601,20 +623,40 @@ void LineEdit::UpdateCursor()
...
     SDL_SetTextInputRect(&rect);
 
+    // Restart blinking
+    cursorBlinkTimer_ = 0.0f;
+
     // Scroll if necessary
     int sx = -GetChildOffset().x_;
     int left = clipBorder_.left_;
     int right = GetWidth() - clipBorder_.left_ - clipBorder_.right_ - cursor_->GetWidth();
+    if(right < left)
```
weitjong: spacing

MystMagus:
Yes, you are right, it is more of a character filter (which is why I wrote that in the request description). This is since there isn't a regexp function in the engine (as far as I know at least, do correct me if I'm mistaken). Course, even if there were it would probably take quite a bit more processing than this (and for larger chunks of text on change I assume, though I am not too familiar with the inner workings of the stuff).

Isn't it only filtering the text that is added in all places? I realize it's not a good idea to filter all of it all the time too, did I miss one?

As for the use case, I've used it for numbers. Course, that's something that can quite easily be handled manually with events as well (not to mention that this won't allow you to make sure a float is properly formatted or anything like that), so I guess this might not be enough to put in really (or not the filter part at least, if the search thing looks good then that should be fine I guess).

(also sorry about the spacing, I keep forgetting it cause it's weird to me)

SirNate0:
What about calling it a CharacterFilter instead of a TextFilter?

PR #2526, opened 2019-10-19
Supports dynamic loading of charcodes

Y-way:
Sometimes we need to load characters dynamically, especially CJK characters. because they contain huge characters

### DIFF
--- a/Source/Urho3D/UI/FontFaceFreeType.cpp
+++ b/Source/Urho3D/UI/FontFaceFreeType.cpp
@@ -196,26 +196,18 @@ bool FontFaceFreeType::Load(const unsigned char* fontData, unsigned fontDataSize
 
     int textureWidth = maxTextureSize;
     int textureHeight = maxTextureSize;
-    hasMutableGlyph_ = false;
-
+    
     SharedPtr<Image> image(new Image(font_->GetContext()));
     image->SetSize(textureWidth, textureHeight, 1);
     unsigned char* imageData = image->GetData();
     memset(imageData, 0, (size_t)image->GetWidth() * image->GetHeight());
     allocator_.Reset(FONT_TEXTURE_MIN_SIZE, FONT_TEXTURE_MIN_SIZE, textureWidth, textureHeight);
-
-    for (unsigned i = 0; i < charCodes.Size(); ++i)
-    {
-        unsigned charCode = charCodes[i];
-        if (charCode == 0)
-            continue;
-
-        if (!LoadCharGlyph(charCode, image))
-        {
-            hasMutableGlyph_ = true;
-            break;
-        }
-    }
+    
+    const unsigned char code[] = " !\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~";
+    for (unsigned i = 0; i < sizeof(code); ++i)
+        LoadCharGlyph(code[i], image);
+    
+    hasMutableGlyph_ = true;
 
     SharedPtr<Texture2D> texture = LoadFaceTexture(image);
     if (!texture)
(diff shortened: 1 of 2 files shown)


eugeneko:
There is `hasMutableGlyph_` that is unconditionally true now.
It may affect realtime performace of text rendering.

weitjong:
Not really sure I should comment in this PR or not as it is beyond my expertise. But I think we need a better glyph caching mechanism for CJK than what we have currently. Having said that, by turning off the cache unconditionally may not be a good answer. Especially if the change would also impact non-CJK font users.

Y-way:
In my opinion, this effect is limited.This PR doesn't change the font cache actually, it just makes the characters cache as needed.

weitjong:
I see. I have looked at the code closer now. So, the `hasMutableGlyph_` flag will just cause the `GetGlyph()` being called repeatedly for each character to be rendered by Text class, which is not ideal if it happened unconditionally for all cases. The loaded glyph still being cached as you pointed out, however in a separate texture per character[?] now.

Just an idea. Instead of setting that flag unconditionally, could it be initialized as per what we have now but we find another way to somehow just load a predefined list of char codes from library user, not hardcoding it like you have proposed. For elementary Chinese app, one may preload 3000 common chars for example. If there is no predefined list then library should still attempt to load all the glyphs as per now. i.e no impact to non-CJK font users.

Y-way:
Good idea.
I think we can define pre-cache charaters in a xml file which defining font`s parameters.

weitjong:
Thanks for the effort. I like the new implementation of the idea. However, just curious why don’t you use hashmap of unsigned to Boolean. I also think it would be better to leave the two non-CJK fonts out of this PR. Perhaps we can add a new free Chinese font from Google Fonts to serve as an example. We could actually use a Chinese font in HelloGUI or somewhere to showcase the IME feature. And lastly, may be we should add a few liners in the online docs describing the “precache” element tag.

Y-way:
@weitjong Thanks for your reply. I have done what you said. 
And I'm worried about my description because my English is not good. 
Would you like to help me for improving the example? I`m not sure which Google Fonts are free.

weitjong:
If there are no other objections then I will merge this PR over the weekend. Probably I will make a few minor cosmetic changes. About the Google Fonts I thought (IANAL) they are all “free to use”?

Y-way:
Ok. Thanks.

Source/Urho3D/UI/Font.h
```diff
@@ -109,6 +112,8 @@ class URHO3D_API Font : public Resource
     FontType fontType_;
     /// Signed distance field font flag.
     bool sdfFont_;
+    /// Pre-Cache charaters
+    HashMap<unsigned, bool> precacheChars_;
```
eugeneko: I don't get when this `HashMap` contains `false`...
Maybe `HashSet` will do?

bin/Data/Fonts/Anonymous Pro.xml
```diff
@@ -0,0 +1,9 @@
+<font>
+<absolute x="0" y="0" />
+<absoluteoffset x="0" y="0" />
+<scaled x="0" y="0" />
+<scaledoffset x="0" y="0" />
+<precache charset="utf8">
+<![CDATA[ !\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~]]>
+</precache>
+</font>
```
eugeneko: Can you please add some indentation and newline at the end of file, in both fonts?

eugeneko:
If I get the code right, `Font` loads _all_ characters when there's no precache string. Is it intended?

weitjong:
Yes, that is intended. I will not merge the two xml files in the PR to ensure the existing fonts are loaded/cached as per before. Instead I will attempt to add a new Chinese font with a precache setting, assuming we are ok to use one of the Google Fonts this way.

eugeneko:
> Yes, that is intended.

Don't you think that it looks a bit confusing?
If I explicitly specify empty pre-cache string, Font will cache all the symbols instead of none.
It is worth mentioning in docs.

Perfect solution would be extra flag to turn on/off whole feature (like `load_on_demand` defaulted to false), but I'm fine with both options.

Source/Urho3D/UI/Font.h
Y-way: Thanks. You`re right.I will change it.:)

PR #2527, opened 2019-10-21
Angelscript fix

orefkov:
Fixed some errors in AngelScript.
1. Fixed wrong binding for bool Sphere::opEquals - was declared as bool& Sphere::opEquals.
2. In Anroid ARM64 using Dictionary in scripts led segviolation, because autowrapper for generic calling convention incorrect work with functions with '?&' parameters. Autowrapper for this methods was replaced by manual implementation.


eugeneko:
It seems there are a lot of unrelated changes...

orefkov:
Sorry, need only 3 files from last commit.

PR #2531, opened 2019-10-24
Add base pass to pbr emissive technique

talarcon:


ArnisLielturks:
Without knowing anything about PBR's implementation, this looks good to go based on other material defs like https://github.com/urho3d/Urho3D/blob/9e38d803888982ee7e459a7756a21ee051ef81b8/bin/CoreData/Techniques/PBR/PBRDiffNormal.xml

eugeneko:
> Without knowing anything about PBR's implementation, this looks good to go

Unless there is confirmed reproducible bug in PBR and this PR is confirmed fix for said bug, I vote against it. Especially considering that there is no description in the PR.

PBR implementation is full of hacks and tricky details, there's no place for random changes.

talarcon:
I didn't added any description because i think it was obvious, my bad.
Beyond how PBR is implemented, it could be wrong or not, but it's there.
This looks like a someone forgot to add a line in the technique, without it some geometries are overlapping each other.

eugeneko:
> This looks like a someone forgot to add a line in the technique

Or have deliberately removed said line.

> without it some geometries are overlapping each other.

Does emission works as intended with proposed fix?

talarcon:
> Or have deliberately removed said line.

I first ask in the forum for anyone who knows why this doesn't contains base pass. It's no warranty but maybe somebody knows.
I've made a minimum test. It's a sphere with a material using PBRDiffNormalEmissive.xml, and a plane using DiamonPlate.xml material and a directional light.
with base pass
pbrdiffnormalemissive-base
without base pass
pbrdiffnormalemissive-nobase

also i've tried with PBRDiffNormal.xml technique, because i've the same result as Emissive with the car's model and wasn't sure if EMISSIVEMAP was used. but it's seems it work fine.
pbrdiffnormal

eugeneko:
I tested PBR a bit.
Apparently, our samples don't use this technique.
And it seems (for me) that this is actually a bug, not some deliberately
made decision.

чт, 19 мар. 2020 г. в 17:51, talarcon <notifications@github.com>:

weitjong:
@eugeneko Since you have tested it and OK with this PR, can you can help to merge it. Thanks.

PR #2532, opened 2019-10-25
Macro to be included in RefCounted derived classes for efficient RTTI

Y-way:


iSLC:
If you wanted to make this efficient, you would've use some constexpr. But more importantly, this duplicates functionality found in the Object type. Which can get quite confusing. Probably also introduces a vtable to some types that don't have it or need it. And also doesn't fully replace RTTI because you can't do dynamic casts with it. Which I think are used in some places in the engine. So basically we now have one propper RTTI implementation that comes with the compilers. Another RTTI for dealing with Object types and one more for dealing with reference counted types.

I'm on phone right now but can definitely see a little debate on the usefulness of this feature.

eugeneko:
1) There is now duplicate functionality in Object and RefCounted.
Moreover, it is _partially_ duplicated. RefCounted doesn't have neither `TypeInfo` nor base class info.
Context can create `Object`-s by type, but cannot create `RefCounted`-s by type.

2) Can you confirm that `ClassID` is the same on Windows when you use DLL? I have a suspicion that in this case `ClassID`-s of same class will be different in `Urho3D.dll` and in user exe.

3) This PR is breaking user code and I cannot find any good excuse for it.

Can you explain your use case so we can look at this PR with some context?
What kind of task are you solving?

eugeneko:
@Y-way nah, that's fine.
If you will decide to PR similar functionality later, let's discuss the best way to do it.
Also keep in mind that ClassID may not work in DLL.

Y-way:
Thanks.@eugeneko 
 
IMO,The ClassID would be unique in the same DLL.Because it is stored at static data segment.

eugeneko:
> IMO,The ClassID would be unique in the same DLL.Because it is stored at static data segment.

If you call `GetClassID` from DLL, you will get static variable from DLL data segment.
But if you then call `GetClassID` from exe, you will get _second_ static variable from exe data segment.

We already have one **unfixed** bug with similar root cause:
https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/Core/Object.cpp#L64
`IsTypeOf` sometimes doesn't work with DLL because pointers are different for same type.

Y-way:
> If you call `GetClassID` from DLL, you will get static variable from DLL data segment.
> But if you then call `GetClassID` from exe, you will get _second_ static variable from exe data segment.
> 
> We already have one **unfixed** bug with similar root cause:
> https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/Core/Object.cpp#L64
> `IsTypeOf` sometimes doesn't work with DLL because pointers are different for same type.

Thank you very much! I see. :)

Y-way:
> We already have one **unfixed** bug with similar root cause:
> https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/Core/Object.cpp#L64
> `IsTypeOf` sometimes doesn't work with DLL because pointers are different for same type.

@eugeneko 
I have checked the code. I think it is wrong.
Maybe change `if (current == typeInfo)` to `if (current->GetType() == typeInfo->GetType())` is right.

Do you think so?

eugeneko:
@Y-way Yes, I'm sure there should be `GetType()`-s

PR #2533, opened 2019-10-26
Fixing IsTypeOf() result error.

Y-way:


Source/Urho3D/Core/Object.cpp
```diff
@@ -58,10 +58,15 @@ bool TypeInfo::IsTypeOf(StringHash type) const
...
+    if (typeInfo == nullptr)
+    {
+        return false;
+    }
+    
     const TypeInfo* current = this;
     while (current)
     {
-        if (current == typeInfo)
+        if (current->GetType() == typeInfo->GetType())
```
weitjong: Perhaps `current == typeInfo || current->GetType() == typeInfo->GetType()`

Source/Urho3D/Core/Object.cpp
```diff
@@ -58,10 +58,15 @@ bool TypeInfo::IsTypeOf(StringHash type) const
 
 bool TypeInfo::IsTypeOf(const TypeInfo* typeInfo) const
 {
+    if (typeInfo == nullptr)
+    {
+        return false;
+    }
```
weitjong: Usually for one liner we don't put it inside a code block.

Y-way:
IMO, `IsTypeOf(StringHash type)` is not safe yet, because the different strings maybe have the same StringHash, although the probability is extremely low.

weitjong:
If you cannot trust StringHash in this comparison then you might as well just don't use it and just compare the string directly.

eugeneko:
> because the different strings maybe have the same StringHash

@Y-way if you accidentally get such classes in your code, Urho as whole will fall apart.
You can assume that StringHash-es for type names never collide.
Fallback to string comparison make this function almost useless because you are basically re-implementing standard `dynamic_cast`.

Y-way:
@weitjong @eugeneko Yes, you are right. So we did not discuss how to improve it.Right.
There is no problem at present.

weitjong:
Actually I believe both Eugene and me are basically saying the same thing. To me, the String comparison following the StringHash comparison looks a little superfluous and I don't recall I ever see it being done elsewhere in the code. If the current StringHash could not be trusted then we need to fix the StringHash, it can be trusted as far as I know.

PR #2534, opened 2019-10-26
On OpenGL ES 2.0. npot textures don't support mipmaps.

kakashidinho:
I'm not sure if this pull request is useful. The 2d platform demo displays wrong graphics on WebGL & OpenGL ES 2.0 devices due to the use of non-power-of-two (npot) textures with mipmaps filter and repeat wrap mode.
By default, standard GLES 2.0 will display black color when using npot texture unless the min filter is set to linear or nearest, and wrap mode is set to clamp to edge.

### DIFF
--- a/Source/Urho3D/Graphics/OpenGL/OGLTexture2D.cpp
+++ b/Source/Urho3D/Graphics/OpenGL/OGLTexture2D.cpp
@@ -479,7 +479,14 @@ bool Texture2D::Create()
     }
 
     levels_ = CheckMaxLevels(width_, height_, requestedLevels_);
-#ifndef GL_ES_VERSION_2_0
+#ifdef GL_ES_VERSION_2_0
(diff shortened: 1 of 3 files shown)


southdy:
IsPowerOfTwo function in MathDef.h

kakashidinho:
Sorry, I didn't have time to finish this PR recently. The original purpose of this PR is just to inform Urho3D devs that npot textures are not always fully supported by GLES2 gpus out there. So there is a potential bug that might affect the games on certain devices.
I will try to make change to this PR if I have time in future.
FYI, I'm contributing Metal backend implementation to ANGLE project. This PR happened due to some one reported that when he tried to use ANGLE's metal backend with Urho3d, the npot textures were rendered black. And that was because of limited support of standard GLES2 by default. 
However, recently I have added full support for npot textures to Metal backend by implementing GL_OES_texture_npot extension, so there is little incentive (from my side) to continue this PR atm.

PR #2538, opened 2019-10-30
Added an extensible interaction mechanism between the C++ and Android parts

orefkov:
Added an extensible interaction mechanism between the C++ part of the library and the
Android part of the application. It includes the ability to send commands with
parameters from the C++ part from game's SDL thread, which are processed by methods in
the main UI thread of the main activity.
Android part can send notifications from UI thread with the result of processing commands
or events that have occurred in app to the main SDL thread of the game, in which it is
sent to subscribers via SendEvent from Input subsystem.
Added binding to AngelScript.
In launcher MainActivity, added methods are command handlers: show a toast, open url,
share text, request some system info and battery status.
Added example of an AngelScript using this functionality.

### DIFF
--- a/Source/Urho3D/Android/Android.cpp
+++ b/Source/Urho3D/Android/Android.cpp
@@ -1,0 +1,96 @@
+//
+// Copyright (c) 2008-2019 the Urho3D project.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+#ifdef __ANDROID__
+#include "../Precompiled.h"
+#include "../Input/Input.h"
+#include "../Resource/JSONFile.h"
+#include "../IO/Log.h"
+#include "Android.h"
+#include "AndroidEvents.h"
+#include <jni.h>
+#include <SDL/SDL_events.h>
+
+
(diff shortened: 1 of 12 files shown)


Source/Urho3D/Android/Android.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+void PostCommandToAndroid(const JSONFile& data)
+{
+    JNIEnv *mEnv = Android_JNI_GetEnv();
+    if (!mEnv)
+    {
+        URHO3D_LOGERROR("No mEnv in PostCommandToAndroid");
+        return;
+    }
+    jclass mainClass = mEnv->FindClass("com/github/urho3d/UrhoActivity");
```
rokups: These could be cached globally or variables could be static. Classes are not changing therefore there is no point in doing reflection lookup every time command is posted.

rokups:
I suppose this is nice. Java's native methods worked fine last time i needed to do this, but such method eliminates need for user to deal with jni which is a plus. What i do not particularly like is "android" part. This functionality could be platform-agnostic. We could call it `SendToMobile()` and then we can implement backend for iOS. Although `SendToMobile()` does kind of suck tbh. Maybe `SendToPlatform()` would make more sense? Though it would be noop on desktop platforms. I am not sure.

orefkov:
> I suppose this is nice. Java's native methods worked fine last time i needed to do this, but such method eliminates need for user to deal with jni which is a plus. What i do not particularly like is "android" part. This functionality could be platform-agnostic. We could call it `SendToMobile()` and then we can implement backend for iOS. Although `SendToMobile()` does kind of suck tbh. Maybe `SendToPlatform()` would make more sense? Though it would be noop on desktop platforms. I am not sure.

For starters, I don’t use IOS, so I don’t know how everything works there, how to develop programs for it, I don’t have the opportunity to test it. In general, the mechanism was developed by me mainly for use in scripts. If you make a game in C++, then organizing interaction with the platform is not so difficult for you. I am making Android games using Urho3DPlayer + scripts in AngelScript. And so that it was possible to get from scripts access to different functionality that is in the android part (AdMob, IAP, some sys info, battery and so on), without getting into the C ++ part every time - this mechanism was created. With it, you can do game logic in scripts, and platform things in java (at least for android, I don’t know how in iOS it is done).
Yes, it is good choice call it "SendToPlatform" and "PlatformNotyfy". I will do it asap.

rokups:
> For starters, I don’t use IOS, so I don’t know how everything works there, how to develop programs for it, I don’t have the opportunity to test it.

It is ok if iOS part is missing. All i am saying is that we should consider other platforms.

> scripts

Yep it makes sense. This would be a useful addition. :+1:

SirNate0:
It looks pretty useful!
I was wondering, though, why the argument for PostCommandToPlatform is a JSONFile and not either a string or a JSONValue? Also, will the stuff that's added checking battery charging and such require additional permissions for the app?

orefkov:
The JSON format was chosen as a universal format for passing parameters to the called method, for a uniform call to an unlimited set of methods. When you need new functionality on the java side, you simply create a “void Name (JSONObject params)” method in your class that extends the MainActivity and you can immediately call it from a script or C++ by setting the attribute "method" to “Name” in the object and putting it there the rest of the parameters.
JSONFile is chosen because it has a function to convert the object to a string, and if I create it myself in PostCommandToPlatform, I have to somehow initialize and store the Context and everything that is connected with it, which I don’t want to do so as not to complicate the implementation.
Battery info does not require additional permissins, but any other functionality that you add yourself may require additional permissions. For example, if you will use it for IAP or advertising.

ArnisLielturks:
This looks very useful, although I have to say that there is too much sample code in the Android files just for demo purposes. I would say that there should be 2 methods max which would show the 2 way communication - platform-to-engine and engine-to-platform.

Issue #2542, opened 2019-11-12
Bug in bloom and bloomhdr shaders post 1.7 release


PsychoCircuitry:
I had noticed that sometime after the 1.7 release the bloomhdr shader started looking awful, pixelated and not very “bloomy”. I spent some time looking into it today and discovered the cause.

The bloom shader uses the passed in uniform variables for inverse size and offset. Documentation for renderpath says these uniforms are case sensitive to the render target name. Well apparently the cases don’t match between the render targets and uniforms, so what was happening was just stacking pixelated textures with no blur.

The fix is simple, either capitalize the render target names in the render path xml or change the uniform definitions to lowercase.

I’m not sure what the preferred approach is as render targets seem to always be defined with lowercase names and uniforms always start c-uppercase.

This issue only occurred post 1.7 release.

The ldr bloom shader is also affected by this same issue.

weitjong:
I prefer changing the render target names to match, but I am fine either ways. Could you help us submit a PR to fix this?

PsychoCircuitry:
Submitted a PR. I changed the render target names, as it's one fix for both shader languages. Sorry for 3 commits to the bloom.xml file, still new to github.

weitjong:
Thanks for your PR. We have other sample post-process beside bloom which I believe could also have the same issue[?], if not corrected. Now I actually wonder why it happened in the first place. In the master branch (after 1.7.1) we have changed the StringHash class to be case-sensitive in calculating the hash value. But how it relates to this bug is still unclear to me, I haven't got time to debug this yet.

PsychoCircuitry:
Yeah, it looks like autoexposure and blur have the same issues as well. I will work on that soon.

PR #2546, opened 2019-11-18
Bloom and BloomHDR PostProcess Fix

PsychoCircuitry:
Updated the render target names so the uniforms will work correctly in the shaders.

### DIFF
--- a/bin/Data/PostProcess/Bloom.xml
+++ b/bin/Data/PostProcess/Bloom.xml
@@ -1,6 +1,6 @@
 <renderpath>
-    <rendertarget name="blurv" tag="Bloom" sizedivisor="4 4" format="rgb" filter="true" />
-    <rendertarget name="blurh" tag="Bloom" sizedivisor="4 4" format="rgb" filter="true" />
+    <rendertarget name="Blurv" tag="Bloom" sizedivisor="4 4" format="rgb" filter="true" />
+    <rendertarget name="Blurh" tag="Bloom" sizedivisor="4 4" format="rgb" filter="true" />
     <command type="quad" tag="Bloom" vs="Bloom" ps="Bloom" psdefines="BRIGHT" output="blurv">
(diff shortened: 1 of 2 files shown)


ArnisLielturks:
@Modanung @eugeneko Any comments about the PR? Looks good for me. Also tested it and it works as expected.

eugeneko:
Nope, looks fine for me.

weitjong:
Please hold on for this one. See #2542

ArnisLielturks:
@PsychoCircuitry are you planning to add fixes for other post processes too in this PR?

ArnisLielturks:
@weitjong I would suggest merging this PR and add additional fixes in a separate PR.

weitjong:
I actually afraid we/I have introduced a bug somewhere that causes all these. So, I would rather fix the root cause, if it is indeed the case. Having said that, I really don't have the time debug it myself.

ArnisLielturks:
@weitjong no problem. I have been looking at the graphics pipeline lately and noticed couple more  bugs regarding the shaders so I will pick this up.

ArnisLielturks:
@weitjong @Modanung This issue is caused by https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/Graphics/View.cpp#L1866-L1869
since now we have case sensitive stringhashes. 
We could find how many shaders are impacted by searching `c[rtname]InvSize` and `c[rtname]Offsets` usage in shaders

ArnisLielturks:
So in the case of Bloom shaders the blur would actually blur out region of 0 radius, since the uniforms would not be initialized with the actual values.

Modanung:
Wouldn't it also be better to use `static_cast<float>(...)` instead of C-style `(float)`?

ArnisLielturks:
Blur and auto exposure fixes provided in a seperate commit.

PR #2549, opened 2019-11-23
Move Context subsystem removal to Engine class

AthosArantes:
Context should not need to know about subsystems.
Also reduces the number of GetSubsystem() calls.
Does not work on non-blocking main function.

### DIFF
--- a/Source/Urho3D/Core/Context.cpp
+++ b/Source/Urho3D/Core/Context.cpp
@@ -134,14 +134,6 @@ Context::Context() :
 
 Context::~Context()
 {
-    // Remove subsystems that use SDL in reverse order of construction, so that Graphics can shut down SDL last
-    /// \todo Context should not need to know about subsystems
-    RemoveSubsystem("Audio");
-    RemoveSubsystem("UI");
-    RemoveSubsystem("Input");
-    RemoveSubsystem("Renderer");
-    RemoveSubsystem("Graphics");
-
     subsystems_.Clear();
     factories_.Clear();
 
--- a/Source/Urho3D/Engine/Application.cpp
+++ b/Source/Urho3D/Engine/Application.cpp
@@ -54,7 +54,7 @@ Application::Application(Context* context) :
     engineParameters_ = Engine::ParseParameters(GetArguments());
 
     // Create the Engine, but do not initialize it yet. Subsystems except Graphics & Renderer are registered at this point
-    engine_ = new Engine(context);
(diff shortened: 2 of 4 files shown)


weitjong:
Thanks for your PR but personally I don't think this is the right direction. The proposed changes make the `Engine` class knows too much about what subsystems are available.

1vanK:
There are different points of view. Context is a storage for global objects in order to not to place them in global scope. You pass a context pointer to a function so that it can access these global objects. And this storage can keep any data (including subsystems). From another point of view, the subsystem is a part of the engine and this PR makes sense.

Modanung:
I liked the [smaller change](https://discourse.urho3d.io/t/subsystem-removal-in-context-destructor/5728) you proposed initially better.

AthosArantes:
@Modanung Just adding RemoveSubsystem calls in the Engine destructor isn't enough because there's no guarantee it'll be the first subsystem to be destroyed.

SirNate0:
Firstly, I have no problem with the engine knowing about all the other subsystems since it is already the one that creates them. However, if there concern is about having all of these different private members, instead of storing a bunch of specific shared ptrs it should be possible to just store a Vector<SharedPtr<Object>> and then ensure those are destroyed in the right order (backwards, if I'm not mistaken).

If we want a more general approach I also like Lasse's suggestion from the forum. Something like a
PODVector<StringHash> of types in the Context that is appended when a subsystem is added, and then we iterate backwards through it and remove every one that is found rather than clearing the subsystems hashmap. If we don't want to worry about the order for most of them, we could have a separate RegisterOrderedSubsystem that adds the type to the vector to be removed in a specific order (though I think that is not needed, since time for the app to close seems like a fairly minor issue).

Also, some of the spacing is wrong, I think (tabs instead of spaces). Further, only the 5 subsystems that require the specific order need to be stored by the engine, and the rest can be removed as we already did, I'm pretty sure. Lastly, you need to actually clear the shared ptrs in Engine in the correct order, not just remove the subsystems (you may have structured the class so that it happens correctly, but I feel it would be better to be explicit so no one refactors the class later and breaks it).

Modanung:
@cadaver's [suggestion](https://discourse.urho3d.io/t/subsystem-removal-in-context-destructor/5728/11):

> If you wanted to guarantee subsystem removal order, you could use a different data structure for that in addition to the lookup map, ie. a vector in Context, that the subsystems are also added to. And then removed in a reverse order.

PR #2553, opened 2019-12-04
Fix ETC1/ETC2 decompression error.

ssinai1:
Fixed:
RGBA ETC2 textures alpha channel decompression error.
ETC1/ETC2 decompression error (and heap overwrite) when width or height of the image (or image level) is not a multiple of 4.

### DIFF
--- a/Source/Urho3D/Engine/Engine.cpp
+++ b/Source/Urho3D/Engine/Engine.cpp
@@ -87,9 +87,6 @@ typedef struct _CrtMemBlockHeader
 } _CrtMemBlockHeader;
 #endif
 
-// Implemented by ETCPACK
-extern void setupAlphaTable();
-
 namespace Urho3D
 {
 
@@ -218,9 +215,6 @@ bool Engine::Initialize(const VariantMap& parameters)
         URHO3D_LOGINFOF("Created %u worker thread%s", numThreads, numThreads > 1 ? "s" : "");
     }
 #endif
-
-    // ETCPACK initialization
-    setupAlphaTable();
 
     // Add resource paths
     if (!InitializeResourceCache(parameters, false))
--- a/Source/Urho3D/Engine/Engine.cpp
+++ b/Source/Urho3D/Engine/Engine.cpp
@@ -87,6 +87,9 @@ typedef struct _CrtMemBlockHeader
 } _CrtMemBlockHeader;
 #endif
 
+// Implemented by ETCPACK
+extern void setupAlphaTable();
+
 namespace Urho3D
 {
 
@@ -215,6 +218,9 @@ bool Engine::Initialize(const VariantMap& parameters)
         URHO3D_LOGINFOF("Created %u worker thread%s", numThreads, numThreads > 1 ? "s" : "");
     }
 #endif
+
+    // ETCPACK initialization
+    setupAlphaTable();
 
     // Add resource paths
     if (!InitializeResourceCache(parameters, false))
--- a/Source/Urho3D/Graphics/OpenGL/OGLTexture2D.cpp
+++ b/Source/Urho3D/Graphics/OpenGL/OGLTexture2D.cpp
@@ -287,10 +287,7 @@ bool Texture2D::SetData(Image* image, bool useAlpha)
             }
             else
             {
-                // ETCPACK needs padding
-                int w = ((level.width_ + 3) / 4) * 4;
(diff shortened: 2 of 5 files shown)


weitjong:
Correct me if I am wrong but it looks like you are assuming that if a texture is compressed then it it ETC1/ETC2 and needs padding[?]. Also, it is unclear to me whether it is safe to call the setupAlphaTable unconditionally.

ssinai1:
This padding has no effect on non ETC1/ETC2 compressed textures (the temporary buffer is a bit larger).
It seems that ETCPACK library cannot be switched off, so the setupAlphaTable() function will always be linked.
It is a child call of the main function, so the needed globals will be already initialized.

ssinai1:
Maybe more elegant solution would be :

// in Source/Urho3D/Resource/Image.cpp :
unsigned CompressedLevel::GetDecompressionBufferSize()
{
    if (!data_) return 0;
    int w, h;
    switch (format_)
    {
    case CF_ETC1:
    case CF_ETC2_RGB:
	case CF_ETC2_RGBA:
		w = ((width_ + 3) / 4) * 4;
		h = ((height_ + 3) / 4) * 4;
    default:
		w = width_;
		h = height_;
	}
	return w * h * 4;
}

// and using in OGLTexture*.cpp files
...
auto* rgbaData = new unsigned char[level.GetDecompressionBufferSize()];
...

eugeneko:
> This padding has no effect on non ETC1/ETC2 compressed textures (the temporary buffer is a bit larger).

The problem is that you violate the contract of the function `Decompress`:
```
/// Decompress to RGBA. The destination buffer required is width * height * 4 bytes. Return true if successful.
```

Actually, I have better idea.
Decompression is already performed block by block.
Therefore, you need neither separate pass for re-shuffling decompressed image nor you need padded output buffer.
You can just decompress each block into small local array of 16 bytes (or 32? I don't remember) and then copy relevant part into output buffer.

ssinai1:
That's a good idea.

weitjong:
About the alpha table initialization -- since the Engine can be started in headless mode, do you think it is better to move that initialization to Graphics subsystem instead?

ssinai1:
But what if someone wants to write a command line texture converter?

weitjong:
Then he/she doesn’t need Urho3D.

eugeneko:
> since the Engine can be started in headless mode, do you think it is better to move that initialization to Graphics subsystem instead?

@weitjong Changes in this PR belong to `Resource` subsystem and have nothing to do with `Graphics` subsystem. I don't really understand why we should break `Resource` subsystem when `Graphics` gets disabled.

weitjong:
You are right. Moving it to Graphics subsystem also doesn’t seem correct. I don’t know, I just feel that initializing the alpha table in the Engine class is out of place to me. It seems to be a redundant call for those who don’t use it and that’s what bothering me.

eugeneko:
There are two options that I find legit.

One, this call may be done in `ResourceCache` ctor because it's the closese thing to Resource subsystem.
Second, this call may be done lazily right in `Decompress.cpp`, using either `std::once_flag` or static member variable trick:
```
static const bool placeholder = []() { setupAlphaTable(); return true; }();
```
AFAIK this should be 100% legit as long as we use this alpha table only after this line.

weitjong:
I wonder if we should just wrap this function as one of Urho3D API and let user call it if it is needed.

ssinai1:
In the SpritePacker tool there is usage of the Image class without initializing the Engine.
Context is a shared object, so what about moving to there.

eugeneko:
We usually try to move code from Context to more specific places, not wise
versa.

I honestly think that lazy initialization is the tool for solving exactly
this issue.

Manual calls are not really good because they will have to be published in
script API (already pain in the ass) and user will have no way to know that
they should call it. User will just try to read pixels from Image and get
garbage w/o any assert or message or whatever.

пт, 6 дек. 2019 г. в 20:54, ssinai1 <notifications@github.com>:

ssinai1:
I note that setupAlphaTable is not seems to be thread safe. (to place it in Decompress)

eugeneko:
That’s why I have written how to call it in thread safe manned few comments
above...

пт, 6 дек. 2019 г. в 21:09, ssinai1 <notifications@github.com>:

ssinai1:
I vote for static initialization (c++ version >= c++11).

weitjong:
Thanks!

Issue #2550, opened 2019-11-24
Upgrade Bullet to 2.89


dsakurai:
On my Mac, I get this error due to the Bullet library being old:
~~~
Source/ThirdParty/Bullet/src/LinearMath/btVector3.h:343:7: error: argument value 10880 is outside the valid range [0, 255]
      [-Wargument-outside-range]
                y = bt_splat_ps(y, 0x80);
                    ^~~~~~~~~~~~~~~~~~~~
/opt/projects/programs/urho3d/Urho3D-vcs/Source/ThirdParty/Bullet/src/LinearMath/btVector3.h:47:32: note: expanded from macro 'bt_splat_ps'
#define bt_splat_ps( _a, _i )  bt_pshufd_ps((_a), BT_SHUFFLE(_i,_i,_i,_i) )
~~~

If you have a look at the Bullet 3 repo, they fixed it by modifying this line in `LinearMath/btVector3.h`:
~~~
#define BT_SHUFFLE(x, y, z, w) (((w) << 6 | (z) << 4 | (y) << 2 | (x)) & 0xff)
~~~
https://github.com/bulletphysics/bullet3/blob/master/src/LinearMath/btVector3.h

I think Urho3D should upgrade Bullet3. Do you agree?

weitjong:
Thanks for letting us know. Yes, we should upgrade. The real question is, who will have time to do it first. Once upgraded, we can remove these lines. https://github.com/urho3d/Urho3D/blob/master/CMake/Modules/UrhoCommon.cmake#L733-L738

weitjong:
I have upgraded Bullet from 2.86.1 to 2.89 (still testing it in my fork repo). To my surprise the Bullet code base actually has also IK support. Somehow the last person who contributed the Bullet upgrade work from version 2.83.6 to 2.86.1 has not done a great job. Now I am in a dilemma because we already have another standalone IK library contributed by @TheComet. On one hand I want to quickly get my Bullet upgrade work merge to Urho3D upstream master branch without spending more time (nor that I have the expertise) to integrate the IK from Bullet. On the other hand I am afraid we are missing something important if we keep ignoring the IK code from Bullet, with all due respect the IK library that we currently have should be less mature than the one from Bullet. Any thought?

TheComet:
If bullet has a better IK solver then by all means, use that. Using worse code out of courtesy towards me doesn't help Urho3D.

I will take a closer look at Bullet's solver and get back to you with pros/cons.

[...]

PR #2554, opened 2019-12-04
Fix clang argument outside range error

smortex:
When building with a recent version of CLang, the following error is
reported:

> error: argument value 10880 is outside the valid range [0, 255]

The error has been fixed in the upstream project here:
https://github.com/bulletphysics/bullet3/pull/2232

### DIFF
--- a/Source/ThirdParty/Bullet/src/LinearMath/btVector3.h
+++ b/Source/ThirdParty/Bullet/src/LinearMath/btVector3.h
@@ -40,7 +40,7 @@ subject to the following restrictions:
 #endif
 
 
-#define BT_SHUFFLE(x,y,z,w) ((w)<<6 | (z)<<4 | (y)<<2 | (x))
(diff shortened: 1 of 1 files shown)


weitjong:
Related to #2550.

smortex:
@weitjong ah, you are right!  I was quickly fixing this issue that was new for me in FreeBSD before adding Urho3D to the ports tree (I had a working port some months ago that I forgot to commit, and it was not building anymore):
https://www.freshports.org/graphics/urho3d/

weitjong:
After upgrading Bullet to 2.89 the workaround for Clang in `UrhoCommon.cmake` is not needed anymore. Also, the proposed local fix in the Bullet library is also not needed once the Bullet upgrade branch (currently still testing it in my fork) is merged to Urho3D master branch. I would like close this PR and track the activity in the #2550.

PR #2560, opened 2019-12-08
Network Changes

igorsegallafa:
- Now you can send messages through urho networking using message ID as unsigned int, instead of byte (only urho messages, SLiketNet messages keep the same thing).
- Is possible to set the max connections of Urho server, on **Network::StartServer** function. Before was allowed only 128 connections, and you couldn't change it.


weitjong:
@ArnisLielturks Do you have time to review this PR?

weitjong:
@ArnisLielturks Thanks for your time.
@igorsegallafa I think the only things miss out in this PR is to update the script API binding accordingly for the `StartServer()` method. Look for "StartServer" in Source/Urho3D/AngelScript/NetworkAPI.cpp and Source/Urho3D/LuaScript/pkgs/Network/Network.pkg

igorsegallafa:
Commited.

weitjong:
Thanks!

Issue #2050, opened 2017-07-30
Vertical Progress Bar wrong direction


QBkGames:
The ProgressBar in vertical mode scales from top to bottom. It would be a lot more useful in practice if it scaled from bottom to the top.
Also, when setting the range, it should also clamp the value to prevent the situation when the range is made smaller than its current value.

Suggestions:
Have the text be able to show 2 modes: Percent (the way it is now) and Ratio, which would show "[value] / [range]", this would be useful for example:
 - Inventory:  "25 / 50 kg"
 - HP: "250 / 600 HP", etc
Which also means that you would be able to set the unit to something other than just "%".

weitjong:
Already closed in #2561.

PR #2561, opened 2019-12-08
Fix Progress Bar direction #2050

igorsegallafa:
Progress bar in vertical mode was scaling from top to bottom, the correct way on this mode is from bottom to top.


eugeneko:
I agree, I'd expect ProgressBar going from bottom to top too.

I don't like these mystic `Clamp`-s on the lines 151 and 158 since they could just be `Min` and `Max`, but I don't insist on changin it.

igorsegallafa:
I agree with you,  but I tried to keep the same code structure, for this I didn’t changed it.

weitjong:
Thanks!

Issue #2563, opened 2019-12-24
Bug with PODVector::Push(*this)


SirNate0:
Possibly related to #1556, PODVectors's `void Push(const PODVector<T>& vector)` fails if you pass it *this as it resizes the vector, and then uses the other vector's `size_` to copy the buffer, but the other vector (i.e. this vector)'s `size_` was also just doubled, resulting in potentially overlapping buffers (and probably generally incorrect results).

The relevant piece of the valgrind output:
```
==30471== Source and destination overlap in memcpy(0x1c464d88, 0x1c464d60, 80)
==30471==    at 0x4C32513: memcpy@@GLIBC_2.14 (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==30471==    by 0x797F58: Urho3D::PODVector<future::VoxelLine::LengthDir>::CopyElements(future::VoxelLine::LengthDir*, future::VoxelLine::LengthDir const*, unsigned int) (Vector.h:1210)
==30471==    by 0x7E368F: Urho3D::PODVector<future::VoxelLine::LengthDir>::Push(Urho3D::PODVector<future::VoxelLine::LengthDir> const&) (Vector.h:875)
==30471==    by 0x7E2A2E: Urho3D::PODVector<future::VoxelLine::LengthDir>::operator+=(Urho3D::PODVector<future::VoxelLine::LengthDir> const&) (Vector.h:778)
```

eugeneko:
It's not just related. It's duplicate, it was fixed before and the fix was reverted due to performace reasons.

PR #2564, opened 2019-12-24
Add check for Pushing *this to PODVector

SirNate0:
My first attempt at a solution to #2563. I'm certainly open to other options if anyone can think of a better approach to this. I'm pretty sure the issues described in #1556 may still affect this when pushing a single item from the old vector.

Do you think it might be better to go with a more complete change like was done with PR #1561?

### DIFF
--- a/Source/Urho3D/Container/Vector.h
+++ b/Source/Urho3D/Container/Vector.h
@@ -850,9 +850,11 @@ public:
     /// Add another vector at the end.
     void Push(const PODVector<T>& vector)
     {
-        unsigned oldSize = size_;
-        Resize(size_ + vector.size_);
-        CopyElements(Buffer() + oldSize, vector.Buffer(), (this == &vector) ? oldSize : vector.size_);
+        // Obtain the size before resizing, in case the other vector is another reference to this vector
+        unsigned thisSize = size_;
+        unsigned vectorSize = vector.size_;
+        Resize(thisSize + vectorSize);
+        CopyElements(Buffer() + thisSize, vector.Buffer(), vectorSize);
     }
 
     /// Remove the last element.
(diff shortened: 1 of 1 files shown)


Source/Urho3D/Container/Vector.h
```diff
@@ -852,7 +852,7 @@ template <class T> class PODVector : public VectorBase
     {
         unsigned oldSize = size_;
         Resize(size_ + vector.size_);
-        CopyElements(Buffer() + oldSize, vector.Buffer(), vector.size_);
+        CopyElements(Buffer() + oldSize, vector.Buffer(), (this == &vector) ? oldSize : vector.size_);
```
weitjong: How about:
```
// Obtain the size before resizing, in case the other vector is another reference to this vector
unsigned thisSize = size_;
unsigned vectorSize = vector.size_;
Resize(thisSize + vectorSize);
CopyElements(Buffer() + thisSize, vector.Buffer(), vectorSize);
```

eugeneko:
It doesn't fix the issue with possible re-allocation.
Just consider PODVector is unsafe for self-insertion and never do it.
I fixed this scenario of `Vector` and `PODVector` long time ago, but my fix for `PODVector` was rolled back by Lasse due to performance reasons.

weitjong:
I have looked at the code earlier and believe even when buffer is reallocated, both of the above code should still work.

eugeneko:
@weitjong Oh, right, I've mistaken this scenario with _selt-insertion_ via iterators that is unsafe:
https://github.com/urho3d/Urho3D/blob/a72ac4049e7e193d90428b787f314ef4c29c9358/Source/Urho3D/Container/Vector.h#L912

Source/Urho3D/Container/Vector.h
eugeneko: ++, this one looks cleaner.

SirNate0: Changed it to that version, I agree, it's cleaner.

weitjong:
Thanks!

PR #2566, opened 2019-12-27
Send Mouse Clicks on Event

igorsegallafa:
Get the clicks from SDL_MouseButtonEvent and send it on Urho Event (E_MOUSEBUTTONDOWN and E_MOUSEBUTTONUP).

This is useful for detect double click (or more) without use UI Event.

### DIFF
--- a/Source/Urho3D/Input/Input.cpp
+++ b/Source/Urho3D/Input/Input.cpp
@@ -1646,9 +1646,9 @@ void Input::ResetState()
     ResetTouches();
 
     // Use SetMouseButton() to reset the state so that mouse events will be sent properly
-    SetMouseButton(MOUSEB_LEFT, false);
-    SetMouseButton(MOUSEB_RIGHT, false);
-    SetMouseButton(MOUSEB_MIDDLE, false);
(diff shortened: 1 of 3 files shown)


eugeneko:
Oh, so SDL has double-click support?

I wonder how it works with out double-click detection.
Let's say I did two clicks. How SDL sends it? First event for first click and then second event for double click? Or it waits for some time and only then sends double click event?

igorsegallafa:
Ye, its possible for >= SDL 2.0.2. This works like you said, one event for each click, and the click counting in evt.button.clicks.

Here’s the documentation about it:
https://wiki.libsdl.org/SDL_MouseButtonEvent

ArnisLielturks:
@eugeneko any comments about the implementation?

eugeneko:
Fairly straightforward, don't see any issues there.

weitjong:
I see there is at least one API change. If so, the corresponding scripting API binding must be updated as well.

ArnisLielturks:
@weitjong those are private methods which are not exposed to scripting.

weitjong:
Ah ok. I was in office and just quickly glanced though the changes. My bad.

PR #2570, opened 2020-01-04
Improve StringHash Saving

SirNate0:
Since we have the optional ability to compile support for a reverse StringHash lookup, I added the ability to save the string version rather than just the hash for VariantMaps and events in ValueAnimations. Without URHO3D_HASH_DEBUG enabled, reading from these strings is still supported, but saving will simply be done by the hash value instead.

As part of this I also added support for reading a VariantMap from JSON with arbitrary string keys, they simply have to be prefixed with $ to distinguish, for example "BAD" as a string from 0xBAD as a hexadecimal number. I'm open to a different marker besides $, I just like it as I believe it would be allowed in Javascript, it's easy to type, and it would generally be reasonably clear that it's a marker and not part of the string, plus $ looks a lot like S for String.

### DIFF
--- a/Source/Urho3D/Scene/ValueAnimation.cpp
+++ b/Source/Urho3D/Scene/ValueAnimation.cpp
@@ -145,7 +145,10 @@ bool ValueAnimation::SaveXML(XMLElement& dest) const
         const VAnimEventFrame& eventFrame = eventFrames_[i];
         XMLElement eventFrameElem = dest.CreateChild("eventframe");
         eventFrameElem.SetFloat("time", eventFrame.time_);
-        eventFrameElem.SetUInt("eventtype", eventFrame.eventType_.Value());
+        if (StringHash::GetGlobalStringHashRegister() && !eventFrame.eventType_.Reverse().Empty())
+            eventFrameElem.SetAttribute("eventname",eventFrame.eventType_.Reverse());
+        else
+            eventFrameElem.SetUInt("eventtype", eventFrame.eventType_.Value());
         eventFrameElem.CreateChild("eventdata").SetVariantMap(eventFrame.eventData_);
     }
 
@@ -219,7 +222,10 @@ bool ValueAnimation::SaveJSON(JSONValue& dest) const
         const VAnimEventFrame& eventFrame = eventFrames_[i];
         JSONValue eventFrameVal;
         eventFrameVal.Set("time", eventFrame.time_);
-        eventFrameVal.Set("eventtype", eventFrame.eventType_.Value());
+        if (StringHash::GetGlobalStringHashRegister() && !eventFrame.eventType_.Reverse().Empty())
+            eventFrameVal.Set("eventname",eventFrame.eventType_.Reverse());
+        else
+            eventFrameVal.Set("eventtype", eventFrame.eventType_.Value());
         JSONValue eventDataVal;
         eventDataVal.SetVariantMap(eventFrame.eventData_,GetContext());
         eventFrameVal.Set("eventdata", eventDataVal);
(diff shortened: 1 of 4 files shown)


Source/Urho3D/Scene/ValueAnimation.cpp
```diff
@@ -141,7 +145,10 @@ bool ValueAnimation::SaveXML(XMLElement& dest) const
         const VAnimEventFrame& eventFrame = eventFrames_[i];
         XMLElement eventFrameElem = dest.CreateChild("eventframe");
         eventFrameElem.SetFloat("time", eventFrame.time_);
-        eventFrameElem.SetUInt("eventtype", eventFrame.eventType_.Value());
+        if (StringHash::GetGlobalStringHashRegister() && !eventFrame.eventType_.Reverse().Empty())
+            eventFrameElem.SetAttribute("eventname",eventFrame.eventType_.Reverse());
```
ArnisLielturks: space missing after `("eventname",`

Source/Urho3D/Scene/ValueAnimation.cpp
```diff
@@ -198,7 +209,7 @@ bool ValueAnimation::SaveJSON(JSONValue& dest) const
         JSONValue keyFrameVal;
         keyFrameVal.Set("time", keyFrame.time_);
         JSONValue valueVal;
-        valueVal.SetVariant(keyFrame.value_);
+        valueVal.SetVariant(keyFrame.value_,GetContext());
```
ArnisLielturks: space missing after `(keyFrame.value_,`

Source/Urho3D/Scene/ValueAnimation.cpp
```diff
@@ -211,9 +222,12 @@ bool ValueAnimation::SaveJSON(JSONValue& dest) const
         const VAnimEventFrame& eventFrame = eventFrames_[i];
         JSONValue eventFrameVal;
         eventFrameVal.Set("time", eventFrame.time_);
-        eventFrameVal.Set("eventtype", eventFrame.eventType_.Value());
+        if (StringHash::GetGlobalStringHashRegister() && !eventFrame.eventType_.Reverse().Empty())
+            eventFrameVal.Set("eventname",eventFrame.eventType_.Reverse());
```
ArnisLielturks: space missing after `("eventname",`

Source/Urho3D/Scene/ValueAnimation.cpp
```diff
@@ -211,9 +222,12 @@ bool ValueAnimation::SaveJSON(JSONValue& dest) const
...
         JSONValue eventFrameVal;
         eventFrameVal.Set("time", eventFrame.time_);
-        eventFrameVal.Set("eventtype", eventFrame.eventType_.Value());
+        if (StringHash::GetGlobalStringHashRegister() && !eventFrame.eventType_.Reverse().Empty())
+            eventFrameVal.Set("eventname",eventFrame.eventType_.Reverse());
+        else
+            eventFrameVal.Set("eventtype", eventFrame.eventType_.Value());
         JSONValue eventDataVal;
-        eventDataVal.SetVariantMap(eventFrame.eventData_);
+        eventDataVal.SetVariantMap(eventFrame.eventData_,GetContext());
```
ArnisLielturks: space missing after `(eventFrame.eventData_,`

ArnisLielturks:
@SirNate0  So if I understand this correctly, this affects only the reading from JSON and XML files?

eugeneko:
I think I need to comment on this PR.

I added `StringHash` reversal functionality for debug purposes only and I actually expect users to disable it before shipping. It's called `URHO3D_HASH_DEBUG` for that reason.
I never profiled it and I think it slows down some segements of the code.

Therefore, I recommend to treat `URHO3D_HASH_DEBUG` somewhat like debug C++ build.
Sure, you can do nice debug-only things like logging, but they shouldn't affect functionality.

Having said that, I don't defend hashes in the saved text formats. It's awful.

PR #2573, opened 2020-01-25
Added support for MCST Elbrus 2000 (e2k) CPU architecture

r-a-sattarov:
Added MCST Elbrus 2000 (e2k) CPU architecture initial support.
This is VLIW/EPIC architecture, like Intel Itanium.

About this architecture:

The following thigs have been done for e2k arch support:
- strip UTF-8 BOM from some files for compability with MCST lcc compiler ver. < 1.24
- added initial support in CMake configuration
- added initial support to AngelScript (ver. 2.33.0)


weitjong:
Thanks for your PR.

Personally I don't like files with BOM as well. Perhaps we should separate the commits for the BOM removal from this PR. In case the PR does not make it, I still want them to be removed. Also, so that we can have better focus in reviewing changes for this PR. In general I am OK with this PR as long as we do not require to declare the support for it officially anywhere in our documentation. It also meas we will not setup any CI for this platform.

I would like to know if these commits are really what needed to get Urho3D compiles there or they are just an initial commits and more are expected to come later.

r-a-sattarov:
A strip UTF-8 BOM from some files is needed for compatibility with MCST eLbrus C Compiler (lcc) version 1.23 (and lower), because this compiler does not understand such files and because of this, compilation is not possible (a change in this behavior is expected in MCST lcc version 1.24). Therefore, these changes are an important part of PR.

Changes in CMake scripts are needed for configuration and compilation, and changes in AngelScript are needed for Editor to work.

All changes in this PR are needed to configure, compile and run Editor (and built-in tests). And there are enough of them for the initial support of e2k architecture.

Configuration and compilation logs on e2k platform:
- [Urho3D_cmake_config_e2k.log](https://github.com/urho3d/Urho3D/files/4113247/Urho3D_cmake_config_e2k.log)
- [Urho3D_compile_e2k.log](https://github.com/urho3d/Urho3D/files/4113249/Urho3D_compile_e2k.log)

weitjong:
Ok. In that case I have no issue to merge this PR as a whole.

Source/ThirdParty/AngelScript/include/angelscript.h
```diff
@@ -40,8 +40,8 @@
 #ifndef ANGELSCRIPT_H
 #define ANGELSCRIPT_H
```
weitjong: This header is from third-party. If you modified it then following our existing convention I would recommend you put your name in the “Modified for Urho3D by” line at the top of the file.

Source/ThirdParty/AngelScript/source/as_config.h
```diff
@@ -178,6 +178,14 @@
 // Oracle Solaris Studio (previously known as Sun CC compiler)
 // __SUNPRO_CC is defined
 
+// Local (or Little) C Compiler
```
weitjong: Additionally, newly local modifications for Urho3D must be clearly marked with a code comment, e.g.  // URHO3D - reason for modification

Source/ThirdParty/AngelScript/source/as_config.h
```diff
@@ -228,6 +236,9 @@
 // AS_SPARC
 // Define this for SPARC CPU family
 
+// AS_E2K
```
weitjong: Same applies here. This is so that the next person doing the 3rd-party lib upgrade would not accidentally overwrite them.

CMake/Modules/UrhoCommon.cmake
```diff
@@ -606,8 +606,14 @@ else ()
             set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${ARM_CFLAGS}")
         else ()
             if (NOT XCODE AND NOT WEB)
-                set (CMAKE_C_FLAGS "-mtune=generic ${CMAKE_C_FLAGS}")
-                set (CMAKE_CXX_FLAGS "-mtune=generic ${CMAKE_CXX_FLAGS}")
+                if (E2K)
+                    # MCST lcc compiler use =native option instead =generic
```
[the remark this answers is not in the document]
r-a-sattarov: MCST lcc compiler also use "mtune", "march" and "mcpu" options. And all this options use value "native" instead "generic".

r-a-sattarov:
Thank you for your comments! 👍 I made required changes to code.
If all is well, please merge this PR.

CMake/Modules/UrhoCommon.cmake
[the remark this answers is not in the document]
weitjong: What I want to point out was that we have this URHO3D_DEPLOYMENT_TARGET build option. The option value is defaulted to “native” to correspond to the default compiler’s behavior when not given specifically any march compiler flag. That this, the compiler actually does not accept “native” as the value for march.

If I understand you correctly then your compiler ONLY understand “native” and nothing else. If that is the case then the URHO3D_DEPLOYMENT_TARGET loses its meaning when targeting Elbrus. Or to put it in another words, the option can be conditionally removed for Elbrus.

weitjong:
I am traveling at the moment. I will merge your PR when I am back.

CMake/Modules/UrhoCommon.cmake
```diff
@@ -606,8 +606,14 @@ else ()
             set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${ARM_CFLAGS}")
         else ()
             if (NOT XCODE AND NOT WEB)
```
weitjong: On second thought, if my understanding is correct then I think we can revert back your proposed change below and just change the condition on this line to have a new check “AND NOT E2K” at the back. That should make thing aligned, at least to me.

r-a-sattarov: Yes, that can be done. But in any case, in code we will need a condition with checking e2k architecture (for using option -mtune=native).

r-a-sattarov:
I will close this PR and re-open it again to reduce the number of commits 😄

weitjong:
You should actually just force push and keep the original PR number. Anyway, after reading your previous comment the lcc compiler only accept “native” value for the mtune and march, I thought/think when the mtune compiler flag is not explicitly given then it would not cause any problem as it would be defaulted to native. Can you try without given any flags at all in your build?

r-a-sattarov:
By default (without my changes in code of UrhoCommon.cmake) flags will be this:
--  CFLAGS: -mtune=generic -msse3 -pthread -fdiagnostics-color=auto
Option "-mtune=generic" will cause an error in MCST lcc compiler.

With my changes (in Common.cmake) flags will be this:
--  CFLAGS: -mtune=native -msse3 -pthread -fdiagnostics-color=auto
With option "-mtune=native" everything will be ok.

Flag "-march=native" does not appear among other compilation flags, because on e2k condition on line 617 (URHO3D_SSE AND NOT XCODE AND NOT WEB) does not activate.
If change condition on line 617 to (URHO3D_SSE OR E2K AND NOT XCODE AND NOT WEB), then compilation flags will become this:
--  CFLAGS: -mtune=native -march=native -msse3 -pthread -fdiagnostics-color=auto

But MCST lcc compiler has such a feature: if -mtune, -march and / or -mcpu options are available on command line at same time, only -mtune option is used, and -march and / or -mcpu options are ignored. So I thought that a change in a line 617 is not required (because option "-march=native" will not be used).

That is why I added an additional condition (line 612) for checking e2k architecture to use the "-mtune=native" option (instead "-mtune=generic" by default).

weitjong:
I think you have misunderstood me. Anyway, please try the change I asked in your new PR.

PR #2575, opened 2020-01-26
Fixed warning: enumeral and non-enumeral type…

zlum:
…in conditional expression [-Wextra]

Urho3D/Input/Input.h:106

 int GetHatPosition(unsigned index) const { return index < hats_.Size() ? hats_[index] : HAT_CENTER; }

gcc (Debian 8.3.0-6) 8.3.0


weitjong:
Thanks!

PR #2576, opened 2020-01-28
Added support for MCST Elbrus 2000 (e2k) CPU architecture

r-a-sattarov:
Added MCST Elbrus 2000 (e2k) CPU architecture initial support.
This is VLIW/EPIC architecture, like Intel Itanium.

About this architecture:

The following thigs have been done for e2k arch support:

- strip UTF-8 BOM from some files for compability with MCST lcc compiler ver. < 1.24
- added initial support in CMake configuration
- added initial support to AngelScript (ver. 2.33.0)

It is re-opened PR (previous was https://github.com/urho3d/Urho3D/pull/2573)

Configuration and compilation logs on e2k platform:
- [Urho3D_cmake_config_e2k.log](https://github.com/urho3d/Urho3D/files/4124289/Urho3D_cmake_config_e2k.log)
- [Urho3D_compile_e2k.log](https://github.com/urho3d/Urho3D/files/4124276/Urho3D_compile_e2k.log)


CMake/Modules/UrhoCommon.cmake
```diff
@@ -605,10 +605,15 @@ else ()
...
-            if (NOT XCODE AND NOT WEB)
+            if (NOT XCODE AND NOT WEB AND NOT E2K)
                 set (CMAKE_C_FLAGS "-mtune=generic ${CMAKE_C_FLAGS}")
                 set (CMAKE_CXX_FLAGS "-mtune=generic ${CMAKE_CXX_FLAGS}")
             endif ()
+            if (E2K)
+                # MCST lcc compiler use =native option instead =generic
+                set (CMAKE_C_FLAGS "-mtune=native ${CMAKE_C_FLAGS}")
+                set (CMAKE_CXX_FLAGS "-mtune=native ${CMAKE_CXX_FLAGS}")
+            endif()
```
weitjong: Try remove these lines and see if everything still works. Thanks.

r-a-sattarov: In this case (after remove these lines), optimization flags in CMake configuration will be as follows:
-- CFLAGS: -msse3 -pthread -fdiagnostics-color=auto
(without "-mtune=native")

With this options compilation is done without any problems, everything still works.

weitjong: Perfect. That’s what I expected. Can you push that commit. Thanks!

r-a-sattarov: Need to remove these lines? In this case, for e2k architecture, "-mtune=native" option will be lost.

weitjong: Isn’t all compiler emit native instruction sets by default?

weitjong: Assuming “cc” is an alias to your compiler, if you would do a simple build like so, “cc HelloWorld.c”, would it generate a native binary?

r-a-sattarov: Yes, in that case it generate a binary file.

By default, MCST lcc compiler uses  "-march=elbrus-vX" option, where X is version of processor command system (by default, this option takes value corresponding to model of the processor on which the compiler is currently running).
For example, for the Elbrus-4C (4 core, 750MHz) processor option will be "-march=elbrus-v3", and for the Elbrus-8C (8 core, 1300 MHz) processor option will be "-march=elbrus-v4".

r-a-sattarov: I wanted, by default, to include additional optimizations ("-mtune=native" option) for the processor on which compilation will go. But probably it’s better not to do that 😄

r-a-sattarov: I decided to recreate PR (in which I remove these lines).
Sorry for my perfectionism 😄 Many thanks!

PR #2581, opened 2020-02-15
Upgrade to support spine runtime 3.7 and above

royalstream:
I updated **AnimatedSprite2D.cpp** to support Spine version 3.7 and 3.8

In a nutshell:

- **flipX/Y** became **scaleX/Y**
Followed instructions at http://esotericsoftware.com/forum/Spine-Unity-3-6-to-3-7-Upgrade-Guide-10446

- **spRegionAttachment_computeWorldVertices** gets two additional parameters and **spVertexAttachment_computeWorldVertices** gets replaced by **spMeshAttachment_computeWorldVertice**
Followed instructions at http://esotericsoftware.com/forum/Upgrading-QML-plugin-to-latest-c-runtime-9419

- **SP_ATTACHMENT_SKINNED_MESH** goes away.
Followed instructions at http://zh.esotericsoftware.com/forum/Updating-to-latest-spine-c-runtime-6720

- Other structural changes that were easy to follow (e.g. **slot->a/r/g/b** is **slot->color.a/r/g/b**)

I also created a sample cloning **33_Urho2DSpriterAnimation** but since the runtime is not included with Urho3D it's probably not a good idea to include it. I created a separate repo for it, and included the runtime files for your convenience: https://github.com/royalstream/Urho3D-spine37up

### DIFF
--- a/Source/Urho3D/Urho2D/AnimatedSprite2D.cpp
+++ b/Source/Urho3D/Urho2D/AnimatedSprite2D.cpp
@@ -306,6 +306,8 @@ void AnimatedSprite2D::UpdateSpineAnimation(float timeStep)
     worldBoundingBoxDirty_ = true;
 }
 
+// This enum used to be defined in spine/RegionAttachment.h but it got moved inside RegionAttachment.c so it's no longer accessible.
+// It's required because AnimatedSprite2D::UpdateSourceBatchesSpine() references its values (SP_VERTEX_*)
 typedef enum {
 	SP_VERTEX_X1 = 0, SP_VERTEX_Y1, SP_VERTEX_X2, SP_VERTEX_Y2, SP_VERTEX_X3, SP_VERTEX_Y3, SP_VERTEX_X4, SP_VERTEX_Y4
 } spVertexIndex;
--- a/Source/Urho3D/Urho2D/AnimatedSprite2D.cpp
+++ b/Source/Urho3D/Urho2D/AnimatedSprite2D.cpp
@@ -119,8 +119,8 @@ void AnimatedSprite2D::SetAnimationSet(AnimationSet2D* animationSet)
 
         // Create skeleton
         skeleton_ = spSkeleton_create(skeletonData);
-        skeleton_->flipX = flipX_;
-        skeleton_->flipY = flipY_;
+        skeleton_->scaleX = flipX_ ? -1 : 1;
+        skeleton_->scaleY = flipY_ ? -1 : 1;
 
         if (skeleton_->data->skinsCount > 0)
         {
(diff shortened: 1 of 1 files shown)


Source/Urho3D/Urho2D/AnimatedSprite2D.cpp
```diff
@@ -306,6 +306,10 @@ void AnimatedSprite2D::UpdateSpineAnimation(float timeStep)
     worldBoundingBoxDirty_ = true;
 }
 
+typedef enum {
```
weitjong: When is this enum used?

royalstream: This enum was part of the spine runtime, in older versions it was inside **RegionAttachment.h**
It's no longer there so, all the references to its values (e.g. **slotVertices[SP_VERTEX_X1]**) won't compile if we don't define it manually.

What happened is that it got moved into the source file (**RegionAttachment.c**) so it's no longer accessible from the header, and its values now have different names.

```c++
typedef enum { 
BLX = 0, BLY, ULX, ULY, URX, URY, BRX, BRY 
} spVertexIndex;
```

Alternatively we could use these new names to match theirs, in that case the code should look like this:

```c++
vertices[0].position_ = worldTransform * Vector3(slotVertices[BLX], slotVertices[BLY]);
vertices[1].position_ = worldTransform * Vector3(slotVertices[ULX], slotVertices[ULY]);
vertices[2].position_ = worldTransform * Vector3(slotVertices[URX], slotVertices[URY]);
vertices[3].position_ = worldTransform * Vector3(slotVertices[BRX], slotVertices[BRY]);

vertices[0].color_ = color;
vertices[1].color_ = color;
vertices[2].color_ = color;
vertices[3].color_ = color;

vertices[0].uv_ = Vector2(region->uvs[BLX], region->uvs[BLY]);
vertices[1].uv_ = Vector2(region->uvs[ULX], region->uvs[ULY]);
vertices[2].uv_ = Vector2(region->uvs[URX], region->uvs[URY]);
vertices[3].uv_ = Vector2(region->uvs[BRX], region->uvs[BRY]);
```

I kept the old names to minimize the amount of changes.

royalstream: Needless to say, this only comes into play if there's a spine runtime at **Source/ThirdParty/spine** because all that code is inside a `#ifdef URHO3D_SPINE` ... `#endif` section.

weitjong: I see. In that case it would be great if you can add one or two liner of comment in the code to explain where it is being used. I have no other issues with this PR. Thanks.

royalstream: Done. Thanks!

weitjong:
Thanks!

PR #2582, opened 2020-02-15
Allow to make custom class inherited from Drawable2D

pat2nav:
Move the AS Drawable2D template definion from source to header


weitjong:
Just in case you didn't receive the notification from CI, your proposed change has caused GCC and MinGW build error. Let us know if you don't have host system with GCC to fix this.

ArnisLielturks:
Looks good to me

weitjong:
Thanks!

PR #2583, opened 2020-02-16
Fix a memory leak in the Urho3D changes to SDL.

rainerdeyke:
The commit speaks for itself.  There was a memory leak in the changes to SDL by Urho3D.  I fixed it.

### DIFF
--- a/Source/ThirdParty/SDL/src/video/SDL_video.c
+++ b/Source/ThirdParty/SDL/src/video/SDL_video.c
@@ -19,7 +19,7 @@
   3. This notice may not be removed or altered from any source distribution.
 */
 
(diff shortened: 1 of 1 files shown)


weitjong:
Thanks for your PR. Is that change available in the latest master branch of upstream SDL? If it is not, so it is your own proposed local change then it should be clearly marked with “Modified by _you_ for Urho3D”. See how it is done for other local changes in 3rd-party libs.

rainerdeyke:
The patch only modifies code that is already labelled as being for Urho3D, so the only thing missing was my name, which I have now added.

weitjong:
Thanks! It is merged in 469b30b605b0c3cef703997b864aa1df18fb1ca8

PR #2584, opened 2020-02-17
PR template

ArnisLielturks:
Template for all submitted PR's which contains a checklist for  things to do in the PR before it could be merged. It would also work as a reminder for the author and reviewers what should be taken into account when reviewing it.

This PR is open for discussion what things should be added/removed from the list and if we need such functionality at all.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -10,9 +10,10 @@
 
 ### To-do
 
+- [ ] I have familiarized myself with [Contribution Checklist](https://urho3d.github.io/documentation/HEAD/_contribution_checklist.html)
 - [ ] Code is formatted according to the guidlines [Coding conventions](https://urho3d.github.io/documentation/HEAD/_coding_conventions.html)
-- [ ] Angel Script bindings include my changes (if there are any) [AS bindings](https://github.com/urho3d/Urho3D/tree/master/Source/Urho3D/AngelScript)
-- [ ] LUA bindings include my changes (if there are any) [LUA bindings](https://github.com/urho3d/Urho3D/tree/master/Source/Urho3D/LuaScript)
+- [ ] Angel Script bindings are up-to-date with my changes [AS bindings](https://github.com/urho3d/Urho3D/tree/master/Source/Urho3D/AngelScript)
+- [ ] LUA bindings are up-to-date with my changes [LUA bindings](https://github.com/urho3d/Urho3D/tree/master/Source/Urho3D/LuaScript)
 - [ ] I have tested my changes and I'm happy with the quality of it
 
 ### Platforms used for development/testing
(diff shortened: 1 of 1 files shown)


1vanK:
I highly recommend not oblige contributors to add bindings if they don't want to. It stops many. The contributors is only obliged to correct the bindings when changing existed functionality, so that after his PR the project still to be compiled. Those interested in the bindings will always be able to add them on their own (this has happened more than once).

.github/pull_request_template.md
```diff
@@ -0,0 +1,27 @@
...
+
+- [ ] Bug
+- [ ] Feature
+- [ ] Improvement
+- [ ] Documentation
+
+### To-do
+
+- [ ] Code is formatted according to the guidlines [Coding conventions](https://urho3d.github.io/documentation/HEAD/_coding_conventions.html)
+- [ ] Angel Script bindings include my changes (if there are any) [AS bindings](https://github.com/urho3d/Urho3D/tree/master/Source/Urho3D/LuaScript)
```
1vanK: AS links to Lua

.github/pull_request_template.md
```diff
@@ -0,0 +1,27 @@
...
+- [ ] LUA bindings include my my changes (if there are any) [LUA bindings](https://github.com/urho3d/Urho3D/tree/master/Source/Urho3D/AngelScript)
```
1vanK: Lua links to AS

1vanK: "include my my"

ArnisLielturks:
@1vanK I agree that we should not force the authors to update the bindings. Are you proposing that we remove the bindings block from the template or just state that it's not obligatory? The sole purpose of this PR is to make the review process faster and a bit cleaner for all the future PR's.

weitjong:
May be add a link to https://urho3d.github.io/documentation/HEAD/_contribution_checklist.html

Until the auto-binding tool is available, the requestor may get other help to get the binding done. The merge can wait until then.

ArnisLielturks:
@Modanung had a great idea that we could add the issue templates too.

They could also contain links to the forum and github's wiki page to avoid duplicate issues + automatic tagging.

weitjong:
BTW, you can add "[ci skip]" anywhere in the commit message to skip the CI build for non code change commits. Saving Gaia.

1vanK:
Currently we use Doxygen for generation of docs. Doxygen fully parses C++ sources, so has full information about classes. Is it possible to use it information to generate bindings?

http://www.doxygen.nl/manual/features.html
> Generates structured XML output for parsed sources, which can be used by external tools.

Modanung:
@1vanK I took your suggestion to the [forums](https://discourse.urho3d.io/t/script-binding-automation/5921/42).

PR #2587, opened 2020-02-19
Moving away from C-style casts where possible

Modanung:
If no objections arise, I will continue with these replacements.

### DIFF
--- a/Source/Urho3D/AngelScript/Addons.cpp
+++ b/Source/Urho3D/AngelScript/Addons.cpp
@@ -1320,8 +1320,8 @@ void CScriptArray::Sort(asUINT startAt, asUINT count, bool asc)
         return;
     }
 
-    const int start{ startAt };
-    const int end{ startAt + count };
+    const int start{ static_cast<int>(startAt) };
+    const int end{ static_cast<int>(startAt + count) };
 
     // Check if we could access invalid item while sorting
     if( start >= static_cast<int>(buffer->numElements) || end > static_cast<int>(buffer->numElements) )
--- a/Source/Urho3D/AngelScript/Addons.cpp
+++ b/Source/Urho3D/AngelScript/Addons.cpp
@@ -585,7 +585,7 @@ void CScriptArray::Resize(asUINT numElements)
     if( !CheckMaxSize(numElements) )
         return;
 
-    Resize((int)numElements - (int)buffer->numElements, (asUINT)-1);
+    Resize(static_cast<int>(numElements) - static_cast<int>(buffer->numElements), -1u);
 }
 
 // Internal
@@ -593,8 +593,9 @@ void CScriptArray::Resize(int delta, asUINT at)
 {
     if( delta < 0 )
     {
-        if( -delta > (int)buffer->numElements )
-            delta = -(int)buffer->numElements;
+        if( -delta > static_cast<int>(buffer->numElements))
+            delta = -static_cast<int>(buffer->numElements);
+
         if( at > buffer->numElements + delta )
             at = buffer->numElements + delta;
     }
@@ -912,7 +913,7 @@ bool CScriptArray::Less(const void *a, const void *b, bool asc, asIScriptContext
 
             if( r == asEXECUTION_FINISHED )
             {
-                return (int)ctx->GetReturnDWord() < 0;
+                return static_cast<int>(ctx->GetReturnDWord()) < 0;
             }
         }
     }
@@ -1071,7 +1072,7 @@ bool CScriptArray::Equals(const void *a, const void *b, asIScriptContext *ctx, S
             r = ctx->Execute();
 
             if( r == asEXECUTION_FINISHED )
-                return (int)ctx->GetReturnDWord() == 0;
+                return static_cast<int>(ctx->GetReturnDWord()) == 0;
 
             return false;
         }
@@ -1188,7 +1189,7 @@ int CScriptArray::Find(asUINT startAt, void *value) const
     }
 
     // Find the matching element
-    int ret = -1;
+    int ret{ -1 };
     asUINT size = GetSize();
 
     for( asUINT i = startAt; i < size; i++ )
@@ -1196,7 +1197,7 @@ int CScriptArray::Find(asUINT startAt, void *value) const
         // value passed by reference
         if( Equals(At(i), value, cmpContext, cache) )
         {
-            ret = (int)i;
+            ret = static_cast<int>(i);
             break;
         }
     }
@@ -1319,11 +1320,11 @@ void CScriptArray::Sort(asUINT startAt, asUINT count, bool asc)
         return;
     }
 
-    int start = startAt;
-    int end = startAt + count;
+    const int start{ startAt };
+    const int end{ startAt + count };
 
     // Check if we could access invalid item while sorting
-    if( start >= (int)buffer->numElements || end > (int)buffer->numElements )
+    if( start >= static_cast<int>(buffer->numElements) || end > static_cast<int>(buffer->numElements) )
     {
         asIScriptContext *ctx = asGetActiveContext();
 
--- a/Source/Urho3D/AngelScript/Script.cpp
+++ b/Source/Urho3D/AngelScript/Script.cpp
@@ -328,9 +328,9 @@ const char **Script::GetEnumValues(int asTypeID)
     enumValues_[asTypeID].Resize(count + 1);
     for (unsigned i = 0; i < count; ++i)
     {
-        int val = -1;
-        const char* name = type->GetEnumValueByIndex(i,&val);
-        if ((unsigned)val >= count)// use unsigned for val so negative values will be flagged as invalid
+        int val{ -1 };
+        const char* name = type->GetEnumValueByIndex(i, &val);
+        if (static_cast<unsigned>(val) >= count)// use unsigned for val so negative values will be flagged as invalid
         {
             URHO3D_LOGDEBUGF("Could not register enum attribute names for type %d."
                       "%s has value of %d, which is outside of the range [0,%d) for a 0-based enum.",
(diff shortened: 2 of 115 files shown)


weitjong:
This kind of refactoring does not give any tangible improvement. I may be wrong. Also, if it is not enforceable then future contributors may again send in code with C-style cast. Something you might want to think about. Your intention is good though, so I am not stopping you. Just make sure you have a good IDE to guide you through to avoid wrong refactor result due to it is a mundane and tedious work (more suitable for a custom Clang refactoring tool).

Modanung:
Doesn't a `static_cast` increase safety and reduce overhead?
Also I have encountered several cases of redundant casting.

eugeneko:
I have conflicted feelings about these changes.

On the one hand, I encourage people to always use explicit C++ casts instead of C-casts for many reasons I don't want to list now.
On the other hand, stirring the whole codebase in order to make non-functional adjustments is huge pain in the ass for current and future PRs and merges.

I personally don't stir the code unless I make functional change somewhere around.
However, sometimes it's better to get rid of the some issue all at once in whole codebase...

Yeah, I'm conflicted.
Just curious, how did you do this refactoring? Manually or with some tools?

Modanung:
> I encourage people to always use explicit C++ casts instead of C-casts for many reasons I don't want to list now.

Please _do_ share your reasoning, for obvious reasons. :slightly_smiling_face:

> Just curious, how did you do this refactoring? Manually or with some tools?

I'm using `grep` and QtCreator.

eugeneko:
> I'm using grep and QtCreator.

How did you check if the cast is redundant? By manually looking at types in IDE?

> Please do share your reasoning, for obvious reasons

I think my reasons are obvious.
C++ casts are more specific, 3 specialized casts instead of one generic.
C++ casts are easier to manage and revise, because they are keyworded,
C++ casts are more explicit and it's easier to see when you are doing something dirty or probably illegal.

Modanung:
By judging each statement.
As an example:
`(float)someInt / (float)otherInt`
is equal to:
`static_cast<float>(someInt) / otherInt`

weitjong:
That’s exactly what I am afraid. As C-style cast may not be exactly doing just a static_cast, manually converting all the occurrences to static_cast without properly analyzing the code original intention may actually introduce bug.

Earlier my comment was referring to improvement in term of runtime performance.

Modanung:
I understand they are not the same. As such I'm focusing on `(float)`, `(int)` and `(unsigned)`, leaving pointer casting untouched.

eugeneko:
I also want to point out that it's bad practice to mix unrelated changes into PR.
I agree with casts conversion from C-style to C++-style since it's commonly accepted good practice, and I can understand when you fix outright code standard violations here and there.

However, I disagree with unreasoned non-functional code stirring.
Every changed line should have good reason to be changed, and "looking nice" is not good reason.
Especially "column-like" alignment of nearby lines, because this change is actually _harmful_ for code maintainability and merge-ability.

Modanung:
> ...this change is actually harmful for code maintainability...

@eugeneko Are you familiar with multi-line cursors?

eugeneko:
> Are you familiar with multi-line cursors?

How is it related to the subject?
When multiple lines are aligned in column, change in one line will likely cause change in the rest of lines.
The more lines of code I have to change in order to make some specific change, the worse code maintainability and merge-ability is.

Modanung:
Multi-line cursors have no use in unaligned code.

1vanK:
This is a good change, but it will be difficult to check. I sympathize with the one who will merge it)

eugeneko:
> Multi-line cursors have no use in unaligned code

Strange, I use them just fine for months, in unaligned code.
E.g. VSCode can jump to next similar snippet regardless of its alignment via Ctrl+D.
If your IDE doesn't support it, maybe you need to change IDE.

Columns introduce unnecessary merge conflicts and are not automatically maintained when you do replace or rename in IDE, which makes them questionable or even harmful, if the code is actually changed by different people.

eugeneko:
Also, column-izing for sake of column-izing can go long way...
I prefer not to start walking this path, ever.

https://github.com/OpenXRay/xray-16/blob/xd_dev/src/xrGame/ActorFollowers.h
image

Modanung:
@eugeneko I agree there's extremes that one should avoid, but generally alignment helps spot variation and anomalies. Code injecting proprietary IDEs are not a sane option, but rather an affront to human rights that any respectable developer ought to boycott.

eugeneko:
> I agree there's extremes that one should avoid, but generally alignment helps spot variation and anomalies

This is true. This is also too weak reason to change existing code.
You can write new chunk of code and columnize it as you wish, and I won't object.
Non-functional changes in existing unrelated code is just a form of hand-waving that causes more issues than it solves. "Looks better" is subjective criteria, therefore it cannot be the only reason to change code.

> Code injecting proprietary IDEs are not a sane option

You are talking nonsense. IDE is _text editor_, how on earth can it be code-injecting?
At least, I have never seen it injecting text snippets into my text files.

eugeneko:
I have a feeling that this PR is getting more and more off-topic...
Can you avoid pushing your personal perferences under agenda of useful code quality change?
This PR is called "Moving away from C-style casts", but in fact it is not.
It has multiple controversal changes that has nothing to do with moving from C-style casts.

weitjong:
To make the review easier the code formatting change should be kept as minimal as possible.

Personally, I also do not want to waste too much time in formatting code. As the code get changes in future, it is a hassle to make it looks aligned to some artificial column. I have once configured my CLion IDE code style formatter to be in line with Urho3D code convention. A single combo keys press would format the code automatically for me. What we have now in our code base was largely the result of one global formatting action I did long time ago.

Modanung:
I will redo unpushed changes and refrain from making any further style adjustments except for when I also modify the casting, as I hope/assume you don't mind changes like these:
```diff
- return IntRect((int)(r1.left_ * s + r2.left_ * t), (int)(r1.top_ * s + r2.top_ * t), (int)(r1.right_ * s + r2.right_ * t),
-     (int)(r1.bottom_ * s + r2.bottom_ * t));
+ return IntRect(static_cast<int>(r1.left_   * s + r2.left_   * t),
+                static_cast<int>(r1.top_    * s + r2.top_    * t),
+                static_cast<int>(r1.right_  * s + r2.right_  * t),
+                static_cast<int>(r1.bottom_ * s + r2.bottom_ * t));
```

eugeneko:
As I said, I don't argue with changes when you actually change something meaningful in code, like in your example. I don't like touching the code only for sake of touching.

I have a suggestion for you, tho. It's up to you whether to follow it or not.
When you make columns, you should consider show typical code auto-formatting will change them.
Columns in your example are likely to be ruined by typical auto-formatting tool.
However, columns like this will be preserved because they follow clear rules:
```
return {
    static_cast<int>(r1.left_ * s + r2.left_ * t),
    static_cast<int>(r1.top_ * s + r2.top_ * t),
    static_cast<int>(r1.right_ * s + r2.right_ * t),
    static_cast<int>(r1.bottom_ * s + r2.bottom_ * t)
};
```
And they still look quite good -- much better than original code.
In other words, don't fight with auto-formatting when you make columns, but work together with it.

Modanung:
You're right, someone might refactor _top_ to _upmost_ and _bottom_ to _lower_. Note taken.
I don't see how it would fight auto-formatting when its indentations are preserved, though.

eugeneko:
> I don't see how it would fight auto-formatting when its indentations are preserved, though

There are two separate aspects, and I can explain why I personally follow such rules.

1) Spaces inside statement, i.e. `static_cast<int>(r1.top_    * s + r2.top_    * t)`.
I use agressive space removal, which is default in Visual Studio and may exist in other IDEs. This basically means that **all** spaces within statement are collapsed into single space. It really helps to kill stray spaces or insert missing spaces. However, it instantly ruins any column alignment within statement.
And there's no compromise. I either have to disable this tool altogether and introduce formatting 

2) Non-default indentation, i.e this:
```
return IntRect(static_cast<int>(r1.left_   * s + r2.left_   * t),
               static_cast<int>(r1.top_    * s + r2.top_    * t)
```
This indentation is very fragile. Let's say I want to replace IntRect(...)-ctor with new { ... }-ctor.
Or, generally speaking, I rename function/class/variable/anything.
Such indentation will become invalid as soon as I change length of the name.
I see a lot of places in my day job project with broken indentation because of such formatting.
Someone aligned function parameters with `(`, then class got renamed in refactoring, and now we have completely wrecked formatting.
On the other hand, 4 spaces will always be 4 spaces, they don't need any maintenance and will never become broken.

In the end, I have sturdy formatting that is easily enforced by automated tools, doesn't work against IDE and is less likely to get ruined in the future.

This is my personal opinion on formatting, but hopefully my reasoning is not completely subjective.

Modanung:
> This is my personal opinion on formatting, but hopefully my reasoning is not completely subjective.

It does make more sense now. Thanks for the explanation.

Modanung:
This is becoming pretty big. Maybe it would be better if I split this up into alphabetical bundles?

eugeneko:
@Modanung I thought about this too.
If you push small PRs, let's say, one per folder, it would be much easier to review them one by one.

Modanung:
I learned something about regular expressions which will allow me to be more thorough, second time round. Any types I'm missing?
```
reset; grep -Pr '(?!sizeof)\(((unsigned|float|double|int|char|short|long|bool|size_t)\s?)+\)(?!"|;|,| const)' .
```

eugeneko:
Be careful with regexes...
C-style casts may look like `int(a)`, `(int)a`, `(float)(int)a`, `float(int(a))`, `(float)((int)a)`, `(int)(const void*)a` and so on...
That's part of the reason why I don't like them.  Too hard to find.

PR #2586, opened 2020-02-18
FileWatcher enhanced. (Currently only available for Windows)

SuperWangKai:
#2421

Issue #2421, opened 2019-02-22
FileWatcher does not watch folder


SuperWangKai:
Directory watching can be very useful for tools. Currently, the folder watching is not fully functional, which means we cannot implement Editor asset browser refresh.

SuperWangKai:
And also, the file watching does not know file creation or file deletion (#2299).

SuperWangKai:
The implementations in Windows/MacOS/Linux are not consistent.

Windows:
https://github.com/urho3d/Urho3D/blob/6dfb25159f9b6020a0c5273d26e797820a8fbb4b/Source/Urho3D/IO/FileWatcher.cpp#L250
https://github.com/urho3d/Urho3D/blob/6dfb25159f9b6020a0c5273d26e797820a8fbb4b/Source/Urho3D/IO/FileWatcher.cpp#L262

Linux:

https://github.com/urho3d/Urho3D/blob/6dfb25159f9b6020a0c5273d26e797820a8fbb4b/Source/Urho3D/IO/FileWatcher.cpp#L116
https://github.com/urho3d/Urho3D/blob/6dfb25159f9b6020a0c5273d26e797820a8fbb4b/Source/Urho3D/IO/FileWatcher.cpp#L298

MacOS:
https://github.com/urho3d/Urho3D/blob/6dfb25159f9b6020a0c5273d26e797820a8fbb4b/Source/Urho3D/IO/FileWatcher.cpp#L314
https://github.com/urho3d/Urho3D/blob/6dfb25159f9b6020a0c5273d26e797820a8fbb4b/Source/Urho3D/IO/MacFileWatcher.m#L136

SuperWangKai:
I don't have other platform for testing on hand. According to the implementations, I think the event support table should be like this -

|       | File Modified | File Created | File Deleted | File Moved | Directory events |
| :---:  | :---:  | :---:  | :---:  | :---:  | :---:  |
| Windows | Y | N | N | Y | N |
| Linux | Y | N | N | Y | N |
| MacOS | Y | Y | Y | Y | N |

weitjong:
The implementation for each platform you mentioned was done by different person at different time. As I recall it, it was only Windows-only feature before I added the Mac-OS one and later another guy contributed the Linux one. And we all minimally just implemented what the FileWatcher class advertises itself for, “Watches a directory and its subdirectories for files being modified.” The additional events are just bonus or consequences from each API. Contribution is welcome to make all the extra standardize.

[...]

PR #2589, opened 2020-02-22
(DISCARDED) FileWatcher extended with minimum modification. (Currently only implemented for Windows)

SuperWangKai:
Extended `FileWatcher` with minimum modification to current `FileWatcher` and resource-auto-reload function. Only Windows implementation has been done.

### DIFF
--- a/Source/Urho3D/IO/FileWatcher.cpp
+++ b/Source/Urho3D/IO/FileWatcher.cpp
@@ -53,7 +53,8 @@ FileWatcher::FileWatcher(Context* context) :
     Object(context),
     fileSystem_(GetSubsystem<FileSystem>()),
     delay_(1.0f),
-    watchSubDirs_(false)
+    watchSubDirs_(false),
+    fullWatch_(false)
 {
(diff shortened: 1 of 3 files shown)


SuperWangKai:
I'm offering an alternative solution comparing to https://github.com/urho3d/Urho3D/pull/2586

The modification to the `FileWatcher` is the same, but in this Pull Request, neither `ResourceCache` nor any scripting interfaces are affected.

Instead of re-using functions from `ResourceCache`, Addtional `FileWatcher` objects can be used to add extra listening ability on target directories.

One use case would be folder monitoring in an Editor, by adding extra `FileWatcher`s, we can have our own path monitoring capability while keeping the original Resource Auto-Reloading untouched.

The only question is I'm not sure if multiple file watching inside same process workable for other platforms (Linux, Mac), on Windows, it just works great.

This is simple sample code works along with this Pull Request, and you can use it to validate.
https://gist.github.com/SuperWangKai/0948b13d408d608d7d3a2128c7aad093

SuperWangKai:
#2421

PR #2592, opened 2020-02-25
Spring cleaning - UI

Modanung:
@weitjong, @eugeneko How does this look?


eugeneko:
If you want my opinion, I find `{ }`-ctors harder to read than plain ` = `. I _do_ use `{ }`-ctors, but sparingly.
However, I know of people who prefer to write `{ }` everywhere due to narrowing conversion checks. Is this the reason why you like it?

This is the part where I hate C++. It offers three almost equivalent ways to make an object with different syntax.

Modanung:
Apart from the narrowing conversion check, the C++ core guidelines advise to use them, which the Urho3D coding conventions point to.

Honestly I agree with the readability issue, and - in other projects - I tend to write:
`Type name{ value };`
Instead of:
`Type name{value};`

I could apply the same here. In some cases I think it looks better without spaces, but I want avoid complex rules for this while still being consistent.

Modanung:
What I also like about using `{}` instead of `()` and `=` is that (almost) all remaining `=` will be assignments, not initializations.
When initializing `()` and `=` have been used somewhat arbitrarily interchangeable. `{}` increases both consistency and safety.

Modanung:
@eugeneko What do you think of that in terms of readability?

eugeneko:
> Honestly I agree with the readability issue, and - in other projects - I tend to write: Type name{ value };
When I do use this style of ctors, I also favor two extra spaces inside brackets.

> the C++ core guidelines advise to use them, which the Urho3D coding conventions point to.
> What I also like about using {} instead of () and = is that (almost) all remaining = will be assignments, not initializations

On the other hand, majority of popular languages use `=`-syntax for initialization and assignment. `{ }` is purely C++ quirk.
I personally use `{ }` when I _construct_ an object, and use `=` when I take existing object or numeric constant.
But I can undertand "almost always `{ }`" policy too.

I wonder if we can enable narrowing conversion warning for `=` assignment... Compilers must have one, for sure.

Modanung:
Also, initializers will be _very_ easy to find with spaces.

Modanung:
The main downside of `{}` might be that it does not work well with `auto`.

Modanung:
Generally I consider the use of `auto` a risky lazy luxury that makes code harder to read and should be avoided, especially in samples. It starts to look like `var` when applied excessively.

Modanung:
I will refrain from pushing further changes until these are merged.

1vanK:
I agree, `float x = 1;` looks much cleaner than `float x{1};`

Modanung:
How about `float{ x };`?

Did you look at the [CppCoreGuidelines](https://github.com/isocpp/CppCoreGuidelines)?
The Urho3D [coding conventions documentation](https://urho3d.github.io/documentation/HEAD/_coding_conventions.html) states that:
> It's recommended to follow C++ Core Guidelines

eugeneko:
> The Urho3D coding conventions documentation states that:
> It's recommended to follow C++ Core Guidelines

Yes. It means that you are encouraged (but not forced) to write your new code according to CCG.
It is not an excuse to change existing working code.

rokups:
One does not simply make style decisions without consulting project maintainers even. Spaces within `{ }` should have been discussed. Also braces for single line statements, which imho are extra noise with no value added.

> I will refrain from pushing further changes until these are merged.

And that sounds like an ultimatum.

Make a clang-format config, lets evaluate it's output and lets apply it to the whole codebase at once. All this manual labor is time wasted. Should you lose interest and half of it is merged already it would leave project in inconsistent state.

Source/Urho3D/UI/DropDownList.cpp
```diff
@@ -314,16 +330,18 @@ bool DropDownList::FilterPopupImplicitAttributes(XMLElement& dest) const
     return true;
 }
 
-void DropDownList::HandleItemClicked(StringHash eventType, VariantMap& eventData)
+void DropDownList::HandleItemClicked(StringHash /*eventType*/, VariantMap& /*eventData*/)
```
rokups: Macro for unused arguments should be preferred. Function signature refactoring feature of IDEs is broken otherwise so it is a bad practice.

1vanK:
Yes, what if everyone will start to create PRы: "I like another of coding style, that's why I passed everything here, hello guys"

Modanung:
RTFM, people.

Source/Urho3D/UI/DropDownList.cpp
eugeneko: [offtopic]
I'm ashamed to admit that I do such things too, sometimes.
When I write empty no-body function, it's really tempting to just comment out parameters.

1vanK:
I look examples:
```
for (int i = 0; i < COUNT; ++i)
        av[i] = i;
```

1vanK:
```
void f3()
{
    int arr[COUNT];
    int i = 0;
    for (auto& e : arr)
         e = i++;
}
```

Modanung:
```
for (int i{ 0 }; i < COUNT; ++i)
        av[i] = i;
```
-----
```
void f3()
{
    int arr[COUNT];
    int i{ 0 };
    for (auto& e : arr)
         e = i++;
}
```

1vanK:
This examples from Core Guide

Modanung:
It seems they were not followed before they existed.

1vanK:
```
for (int i = 0; i < 100; ++i) {   // GOOD: i var is visible only inside the loop
    // ...
}
```

Modanung:
> ### [ES.23: Prefer the {}-initializer syntax](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-list)

1vanK:
That sounds reasonable

1vanK:
It is strange that there are no examples in which they themselves follow their recommendations.

```
const widget x = [&]{
    widget val;                                // assume that widget has a default constructor
    for (auto i = 2; i <= N; ++i) {            // this could be some
        val += some_obj.do_something_with(i);  // arbitrarily long code
    }                                          // needed to initialize x
    return val;
}();
```

Should be
```
const widget x { [&]{
    widget val;                                // assume that widget has a default constructor
    for (auto i {2}; i <= N; ++i) {            // this could be some
        val += some_obj.do_something_with(i);  // arbitrarily long code
    }                                          // needed to initialize x
    return val;
}()};
```
yes?

Source/Urho3D/UI/FontFace.h
```diff
@@ -101,7 +97,7 @@ class URHO3D_API FontFace : public RefCounted
     SharedPtr<Texture2D> LoadFaceTexture(const SharedPtr<Image>& image);
 
     /// Parent font.
-    Font* font_{};
+    Font* font_;
```
rokups: `{}` initializes pointer to `nullptr` while lack of it leaves pointer uninitialized. Afaik we prefer initializing fields within class body when writing new code. This should be `Font* font_ = nullptr;` really.

Source/Urho3D/UI/DropDownList.cpp
Modanung: I adopted this method from Urho's code base. We could switch to using a macro and replace all `'/\*\w+\*/,'` - if we decide to do so - later.

Source/Urho3D/UI/FontFace.h
```diff
@@ -75,21 +75,17 @@ class URHO3D_API FontFace : public RefCounted
...
     bool IsDataLost() const;
+    /// Return if font face uses mutable glyphs.
+    virtual bool HasMutableGlyphs() const { return false; }
 
     /// Return point size.
     float GetPointSize() const { return pointSize_; }
-
     /// Return row height.
     float GetRowHeight() const { return rowHeight_; }
-
```
rokups: This is good. Those extra empty lines have no reason to exist :+1:

Source/Urho3D/UI/FontFace.h
Modanung: There is only one `FontFace` constructor, _it_ initializes `font_`.

1vanK: With `font_{}` inited twice

Initiating in class body not always possible. Will some of members be initialized in constructor, and other in class body?

Modanung:
@1vanK I agree, and I wonder how _they_ would respond to a PR.

eugeneko:
IIRC it was me who added mention of CppCoreGuidelines in Urho codestyle.
CCG is good paper offering a lot of advice.
Everyone is _encouraged_ to follow it when they write their code.

It _does not mean_ that `int i = 0` initialization is forbidden in Urho codebase.
Therefore, since `int i = 0` initialization is _allowed_ in Urho codebase, you don't have legitimate reason to change it, unless you make substantial code refactoring in same place (e.g. in the same function).

Modanung:
I'm cleaning up. With its increased type-safety `{}` ties in to the switch to `static_cast`s.

Modanung:
Part of the result is changes like these:
**Before**
```
const IntRect& uvBorder = (imageBorder_ == IntRect::ZERO) ? border_ : imageBorder_;
int x = GetIndentWidth();
IntVector2 size = GetSize();
size.x_ -= x;
IntVector2 innerSize(
    Max(size.x_ - border_.left_ - border_.right_, 0),
    Max(size.y_ - border_.top_ - border_.bottom_, 0));
IntVector2 innerUvSize(
    Max(imageRect_.right_ - imageRect_.left_ - uvBorder.left_ - uvBorder.right_, 0),
    Max(imageRect_.bottom_ - imageRect_.top_ - uvBorder.top_ - uvBorder.bottom_, 0));

IntVector2 uvTopLeft(imageRect_.left_, imageRect_.top_);
uvTopLeft += offset;
```

**After**
```
const int x{ GetIndentWidth() };
const IntVector2 size{ GetWidth() - x, GetHeight() };
const IntVector2 innerSize{ Max(size.x_ - border_.left_ - border_.right_, 0),
                            Max(size.y_ - border_.top_ - border_.bottom_, 0) };

const IntRect& uvBorder{ imageBorder_ == IntRect::ZERO ? border_ : imageBorder_ };
const IntVector2 uvTopLeft{ imageRect_.Min() + offset };
const IntVector2 innerUvSize{ Max(imageRect_.right_ - imageRect_.left_ - uvBorder.left_ - uvBorder.right_, 0),
                              Max(imageRect_.bottom_ - imageRect_.top_ - uvBorder.top_ - uvBorder.bottom_, 0) };
```

eugeneko:
> I'm cleaning up.

Every time any line of code is changes, it makes things _worse_ by default.
Every changed line of code is an obstacle for merges, reverts, git history browsing and all other things people usually do with repositories.
Therefore, every changed line of code reduces overall PR quality.
Do you consider these factors when you esitmate how benefical code cleanup is?

PS: It's funny that "after" don't fit the screen anymore.
> Part of the result is changes like these:
image

Modanung:
I considered the alternative.

Modanung:
```
const int x{ GetIndentWidth() };
const IntVector2 size{ GetWidth() - x, GetHeight() };
const IntVector2 innerSize{ Max(size.x_ - border_.left_ - border_.right_, 0),
                            Max(size.y_ - border_.top_ - border_.bottom_, 0) };

const IntRect& uvBorder{ imageBorder_ == IntRect::ZERO ? border_ : imageBorder_ };
const IntVector2 uvTopLeft{ imageRect_.Min() + offset };
const int uvW{ Max(imageRect_.right_ - imageRect_.left_ - uvBorder.left_ - uvBorder.right_, 0) };
const int uvH{ Max(imageRect_.bottom_ - imageRect_.top_ - uvBorder.top_ - uvBorder.bottom_, 0) };
const IntVector2 innerUvSize{ uvW, uvH };
```
Better?
The longest lines within the method all got shorter, and this is now shorter than those.

Modanung:
> Every time any line of code is changes, it makes things worse by default.

_Nonsense!_

eugeneko:
> Nonsense!

I'm waiting for arguments. I have explained mine.
Do you imply that changes in this PR don't affect readability of git history?
Do you imply that changes in this PR will not create merge conflicts for any existing forks of Urho?

A lot of people use their custom forks. I have mine, and I don't mean rbfx. I know that many other people fork Urho to make changes suitable for their projects.

Modanung:
Would you rather see an unmergable clean fork?

eugeneko:
The very fact of line change:

- creates false positive in git blame
- creates entry in commit tree that must be reviewed when a person is working with code
- introduces potential merge conflicts with all user forks (and Urho has almost a thousand of them)

Therefore, any line change in inherently harmful. This is undisputable fact for any project with more than one person involved.

The question we should disucss is whether proposed changes overweigh this harm or not.

Modanung:
> and Urho has almost a thousand of them
Yes, including the one requesting this PR.

eugeneko:
> How can any merge - or change even - happen with that logic applied?

By adding up two factors, of course. Inherent harm and proposed benefit from the change.
Feature updates and bug fixes _often_ tip scales to "positive" side.
Not always, as you can see from PR history.

And these arguments is not something I made up for Urho right now. That's how all more or less popular open source projects work, as far as I'm aware.

Modanung:
I don't think this PR holds any conflicts with other open PRs.

eugeneko:
> I don't think this PR holds any conflicts with other open PRs.

I'm not talking about conflicts with open PRs.
I'm talking about confilts with all branches of all forks of Urho.
When people decide to rebase their forks, they will have to deal with said confilcts.
I have maybe several _dozen_ of old incomplete feature branches in my _Urho_ fork.

Modanung:
@ArnisLielturks Do you have an opinion on all this?

> I have maybe several dozen of old incomplete feature branches in my Urho fork.

How many conflict with this PR?

eugeneko:
> How many conflict with this PR?

I have no idea. This question is pointless to ask becaude it's impossible to know in general.
There are thousand of forks, sometimes with local unpublished branches. Unless you check _all_ of them, you cannot know how many real conflics you create.

That's why _any_ line change shall be justified.
The bigger change is, the stronger justification is needed.

Why do you think I don't PR EASTL, even if I consider it benefical?
Becasue I don't have enough justification for such conflict-prone change.
I'm also lazy, but that's only a part of the reason.

Source/Urho3D/UI/FontFace.h
Modanung: In Urho, I believe member initialization almost always happens in the constructor's head with the exception of those belonging to `struct`s.

ArnisLielturks:
Although clean code is a way to go I would vote against the massive refactoring of things just because of it. As others already mentioned, it could potentially cause a lot of troubles to the already existing forks and we tend to take backwards compatibility very seriously. I would be all in if we would refactor the code and add additional functionality and/or improvements to them.

Modanung:
Could you please all look beyond the sheer number of lines?

Maybe even ignore the - inherently messy - diff and open these files side by side in an IDE.

Modanung:
Consider for instance this - otherwise eternal - piece of shode:
```diff
-void OnClickBegin
-    (const IntVector2& position, const IntVector2& screenPosition, int button, int buttons, int qualifiers, Cursor* cursor) override;
+void OnClickBegin(const IntVector2& position, const IntVector2& screenPosition,
+                  int button, int buttons, int qualifiers, Cursor* cursor) override;
/// React to mouse doubleclick.
-void OnDoubleClick
-    (const IntVector2& position, const IntVector2& screenPosition, int button, int buttons, int qualifiers, Cursor* cursor) override;
+void OnDoubleClick(const IntVector2& position, const IntVector2& screenPosition,
+                   int button, int buttons, int qualifiers, Cursor* cursor) override;
/// React to mouse drag begin.
-void
-    OnDragBegin(const IntVector2& position, const IntVector2& screenPosition, int buttons, int qualifiers, Cursor* cursor) override;
+void OnDragBegin(const IntVector2& position, const IntVector2& screenPosition,
+                 int buttons, int qualifiers, Cursor* cursor) override;
/// React to mouse drag motion.
-void OnDragMove
-    (const IntVector2& position, const IntVector2& screenPosition, const IntVector2& deltaPos, int buttons, int qualifiers,
-        Cursor* cursor) override;
+void OnDragMove(const IntVector2& position, const IntVector2& screenPosition,
+                const IntVector2& deltaPos, int buttons, int qualifiers, Cursor* cursor) override;

**Comment by Modanung** [comment]
@eugeneko At what number of forks should no line of code be changed ever again?

**Comment by Modanung** [comment]
Feel free to start marking files as viewed.

**Comment by Modanung** [comment]
There is one exception where I modified a function signature:
```diff
UIElement.h
-void GetBatchesWithOffset(IntVector2& offset, PODVector<UIBatch>& batches, PODVector<float>& vertexData, IntRect currentScissor);
+void GetBatchesWithOffset(const IntVector2& offset, PODVector<UIBatch>& batches, PODVector<float>& vertexData, IntRect currentScissor);
```

Modanung:
If we would switch to C++17, we could `{}` while adhering to the following rule of the CoCoDox:
> Use `auto` for verbose, unknown or redundant types.

eugeneko:
> At what number of forks should no line of code be changed ever again?

I don't remember setting up such limit.

> Could you please all look beyond the sheer number of lines?

When you intorduce defect somewhere in this PR, it would be hell to find it _because_ of the sheer number of lines. The more lines of code you send for PR, the worse review quality is, the more defects are merged unnoticed. The sheer number of lines is the simplest criteria of PR quality. I don't know other criterias so far. Resulting code quality is _irrelevant_ if PR quality is poor.

Source/Urho3D/UI/DropDownList.cpp
Modanung: Maybe we could put the [lambda wrap](https://github.com/urho3d/Urho3D/pull/2593) on the menu as a generalized ignore function?

Modanung:
> I don't remember setting up such limit.

You mentioned an inverse correlation between the number of extant forks and acceptable PR size.

eugeneko:
I don't want to continue this line of discussion, because I'm already walking on the edge of biased personal opinion and this is not right thing to do.

---

To summarize this discussion, neither me, nor @ArnisLielturks, nor @weitjong are ready to take the responsibility to review and merge this PR in its current state (correct me if I'm assuming wrong).
The reasons are unclear scope of this PR, no direct benefits (no features added or bugs fixed) and huge amout of affected lines of code (that directly affects review effort and quality, and also further code maintainability).

Exception can be made for automatic non-functional code changes, like clang-tidy sweep that happened several years ago.

Modanung:
You're over estimating the review effort and conclude a lack of benefits without review.

eugeneko:
> You mentioned an inverse correlation between the number of extant forks and acceptable PR size.

Just to finalize this line of discussion.. Nope. Number of forks doesn't really matter.
_If_ open source project have forks and encourage forks, said project is better be conservative with changes in codebase.

> You're over estimating the review effort and conclude a lack of benefits without review.

I did try to review this PR and I think that the code, in general, looks better after.
It just doesn't justify (in my opinion) the damage to _codebase_ this PR inflicts (false positive blames and possible merge conflicts).

Modanung:
```diff
- auto sliderLength = (int)Max((float)GetWidth() / (range_ + 1.0f), (float)(border.left_ + border.right_));
- 
- if (knob_->IsFixedWidth())
-     sliderLength = knob_->GetWidth();
- 
- float sliderPos = (float)(GetWidth() - sliderLength) * value_ / range_;

+ const int sliderLength{ knob_->IsFixedWidth() ? knob_->GetWidth()
+                                               : Max(static_cast<int>(GetWidth() / (range_ + 1.0f)),
+                                                     border.left_ + border.right_) };
+ 
+ const float sliderPos{ (GetWidth() - sliderLength) * value_ / range_ };
```

Modanung:
So I should cut it up further still? Or do we just not care?

weitjong:
Personally I think you should not stop. I truly believe your intention is good. The points in the core guidelines are good too. But, if the end result is raising eyebrows like this, you should take that as clue.

Modanung:
@weitjong Thank you. I intentionally stopped before going any further to keep merging feasible, and have only made tiny adjustments to already modified files since.

weitjong:
Sorry for the earlier typo :)  Did not catch the error in time due to my eyes problem.

Modanung:
I get the feeling eyebrows are raised by the fact people are new to the way it is supposed to be. This means that all code review has been wrong,

eugeneko:
> So should I cut it up further still? Or do we just not care about code quality?

Codebase quality is more than just quality of code in some specific snapshot.
I don't believe that you actually considered other aspects beyond the code itself.

Modanung:
@eugeneko How and when had you imagined these changes could happen, if ever?

Top to bottom, all 19 files have been `{}`-ified, rid of C-style casts, had a full `const` checkup and some markup corrected. Often several of these changes happened in the same place. In two places a lambda function was added, also in line with [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es28-use-lambdas-for-complex-initialization-especially-of-const-variables).

eugeneko:
> How and when had you imagined these changes could happen, if ever?

In all projects I'm aware of, incremental upgrade is adopted.
Meaning that non-functional changes happen around functional changes.
When a person rewrite or significantly change some existing function/class/module, said person may write new version of code in better style and change reasonable amount of code around this place.

Leeway depends. For example, if a person adds new class members, it may be acceptable to move all member intiailization of this class from cpp to header.

The benefits of such approach are that

- git history and blame stay relevant, they don't contain _too much_ irrelevant entities.

- scope of changes is throttled to manageable amount. I think we all agree that "code cleanup" may go indefinetely long. One can commit hundreds of thousands lines of code "cleaning up" something.

- due to throttled scope code reviews are simpler and chances of breaking something are lower.

richelbilderbeek:
I am happy I saw this discussion, as I consider**ed** contributing to Urho3D. 

In the repos of all my teams, we have a simple rule: if a 'cleaning-up' Pull Request does not break the Continuous Integration build, it will be accepted. This shows openness towards contributions from outside. It makes people feel happy after contributing. 

The discussion above showcases a discouragement from such contributions. I would not blame the contributor for this Pull Request being his last. This does not look like the fun experience a PR should be.

I hope the Urho3D community figures out the rules for contributions, for example, as described in a `CONTRIBUTIONS.md` file. A rule like 'if a Pull Request does not break the build, it will be accepted, unless it breaks with the goals of Urho3D, which are [...]'.

weitjong:
@richelbilderbeek  You forgot the fact that the OP is asking for our opinion and we gave our opinion.

Also, we reserve the right to reject a PR at our discretion, no matter what the Robot says in our CI.

Source/Urho3D/UI/LineEdit.cpp
```diff
@@ -599,10 +599,11 @@ void LineEdit::UpdateText()
...
     {
-        String echoText{};
-
-        for (unsigned i{ 0 }; i < utf8Length; ++i)
-            echoText.AppendUTF8(echoCharacter_);
+        const String echoText{ [&]{
+                String text{};
+                for (unsigned i{ 0 }; i < utf8Length; ++i)
+                    text.AppendUTF8(echoCharacter_);
+                return text; }() };
```
rokups: Why? Is this a trolling attempt?

Modanung:
```diff
+ int horizontalThickness{ 1 };
+ int verticalThickness{ 1 };
+ 
+ if (parent_) switch (parent_->layoutMode_)
+ {
+ case LM_HORIZONTAL: verticalThickness += 2;
+     break;
+ case LM_VERTICAL: horizontalThickness += 2;
+     break;
+ default:
+     break;
+ }


- int horizontalThickness = 1;
- int verticalThickness = 1;
- if (parent_)
- {
-     switch (parent_->layoutMode_)
-     {
-     case LM_HORIZONTAL:
-         verticalThickness += 2;
-         break;
- 
-     case LM_VERTICAL:
-         horizontalThickness += 2;
-         break;
- 
-     default:
-         break;
-     }
- }
```

Source/Urho3D/UI/LineEdit.cpp
Modanung: Are you overlooking the `const String echoText{ [&]{` part?

JTippetts:
I think this whole effort is pointless, to be honest. It doesn't fix anything, it's just code churn for the sake of appearing to make forward progress. I'm with @eugeneko on being careful about screwing up blame, by only making these kinds of cosmetic changes when fixing actual issues.

Source/Urho3D/UI/LineEdit.cpp
rokups: What does that achieve besides hurting code readability and debug build performance?

Modanung:
Switching to `{}` is not purely cosmetic. From the core guidelines:
> Prefer {}. The rules for {} initialization are _simpler, more general, less ambiguous, and safer_ than for other forms of initialization.

Moving away from C-style casts is _not_ purely cosmetic. From Stack Overflow:
> C++ style casts are checked by the compiler. C style casts aren't and can fail at runtime

Does anyone suggest the use of `const` is purely cosmetic? These all have technical reasons and tie into each other.

Modanung:
I will be sure to make a separate PR for each file in the future.

Source/Urho3D/UI/LineEdit.cpp
Modanung: From the CppCoreGuidelines:
> 28: Use lambdas for complex initialization, especially of const variables

Some place else:
> Const is a powerful tool for writing safer code, and it can help compiler optimizations. You should use it as much as possible...

rokups:
Forcing `const` into every corner for the sake of just having `const` like here
```
        const String echoText{ [&]{
                String text{};
                for (unsigned i{ 0 }; i < utf8Length; ++i)
                    text.AppendUTF8(echoCharacter_);
                return text; }() };
```
That ^ is purely cosmetic. And looking bad while at it. Clearly that string is not `const` in given context, but making it `const` gives no advantages either. Compiler does not even optimize code that much based on `const` qualifier.

Code quality is severely reduced by this PR and there are no advantages to merging it, despite what code guidelines say. Code is simple, clean and functioning. After this PR code becomes messy, hard to follow and potentially broken. I appreciate the initiative, but very much dislike stubborn rejection of what everyone else is saying.

Modanung:
From the CppCoreGuidelines:
> [**ES.28: Use lambdas for complex initialization, especially of const variables**](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es28-use-lambdas-for-complex-initialization-especially-of-const-variables)

> [**Con.1: By default, make objects immutable**](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#con1-by-default-make-objects-immutable)

eugeneko:
The purpose of CppCoreGuidelines is to encourage people to write clean and readable code.
If code becomes more messy and less readable, then you are misusing these guidelines.

In this last example, you made the code twice more complicated with _literally zero_ benefit, violating code style and code consitency on the way.

This is basically the opposite of what CCG are aimed for.

> We do not suffer the delusion that every one of these rules can be effectively applied to every code base. Upgrading old systems is hard.

> Consider these rules ideals for new code, opportunities to exploit when working on older code, and try to approximate these ideals as closely as feasible

> **This is not a guide on how to convert old C++ code to more modern code**.

> The rules are not precise to the point where a person (or machine) can follow them blindly

Modanung:
I must admit I have some doubts about the lambdas, but _that's_ what a review is for.

Source/Urho3D/UI/FontFace.h
nem0: Indeed, they are in the way of those very informative comments

1vanK:
Ok, I read the posts and I have one question. Everyone says that the change in style seems to be acceptable, but does not bring a new feature. You actually open the Pandora's box. Now will you accept a code that will be stylistically different from the rest of the code base?

Modanung:
So how many people exactly are contributing while oblivious to the coding conventions?

PR #2595, opened 2020-02-29
Replaced C-style casts, enforced CCG ES.23 & Con.4 and removed redundant brackets

Modanung:


eugeneko:
I don't see how it is different from previous PR.
The only difference is that only 5 functions are affected instead of hundreds.
The idea behind the changes is the same.

eugeneko:
For a record, I will write here formal answer.

1) From Urho coding conventions:

> It's recommended to follow C++ Core Guidelines (except items that contradict Urho legacy, e.g. ES.107, Enum.3, Enum.5)

It literally means that legacy coding conventions of Urho codebase take precedence over C++ Core Guidelines. Including, but not limited to, ES.107, Enum.3, Enum.5.

In the file `Color.cpp` there are 23 intances of `type var = val;` and 0 instances of `type var{ val }`.
I have checked several random files and they have similar statistics.
Therefore, syntax `type var{ val }` contradics legacy coding conventions in pretty much the same way as aforementioned item ES.107 "Don’t use unsigned for subscripts, prefer gsl::index".

`unsigned` may be worse than `ptrdiff_t`, but we already use it everywhere, so it shall stay this way.
`type var = val` may be worse than `type var{ val }`, but we already use it everywhere, so it shall stay this way.

2) By trying to enforce the item ES.23: "Prefer the {}-initializer syntax", this Pull Request directly violates item ES.11: "Use auto to avoid redundant repetition of type names":

```diff
- const auto max = static_cast<double>(M_MAX_UNSIGNED);
+ const double max{ static_cast<double>(M_MAX_UNSIGNED) };
```

I take it as a sign that author doesn't understand C++ Core Guidelines enough to effectively apply them.

---

@weitjong, @ArnisLielturks to avoid repeating this discussion over and over again in the future, I _suggest_ to immediatelly close any further Pull Requests that attempt to restyle existing code.
This statement doesn't apply to Pull Requests with _new_ code written in different style.

I admit that I'm quite passive maintainer, so the final decision is up to you.

Modanung:
I removed instances of ES.23 enforcement where it conflicted with ES.11 and will follow this logic until we switch to C++17.

Modanung:
@eugeneko Does this include ignoring [ES.49](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es49-if-you-must-use-a-cast-use-a-named-cast) because of "legacy" C-style casting, in your view?

eugeneko:
If you collect similar statistics, you will see that C++ casts are widely
used in Urho codebase already. Moreover, they are unavoidable, due to
dynamic cast.

In case of casts, it would make sense to use only C++ casts in new/changed
code and _maybe_ get rid of c-style casts in existing code. I grew more
reluctant to the “maybe” part of this statement in recent days.

вс, 1 марта 2020 г. в 17:45, 魔大农 <notifications@github.com>:

Modanung:
Enforcement of [ES.23](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-list), [ES.49](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-casts-named), [Con.1](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rconst-immutable) and [Con.4](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#con4-use-const-to-define-objects-with-values-that-do-not-change-after-construction) overlaps in motivation and implication.

eugeneko:
I agree. These items do overlap.
It does not mean we should apply them all to Urho codebase.
It also does not mean that you can solely decide the course of action.

Also, one more question.
You have replaced several variable declarations to the new style.
What’s your end goal? To have *all* variable declarations in Urho replaced,
without exception? To replace only a fraction and stop? Something else?


вс, 1 марта 2020 г. в 18:25, 魔大农 <notifications@github.com>:

Modanung:
> I agree. These items do overlap. It does not mean we should apply them all to Urho codebase.

Our code conventions do.

> It also does not mean that you can solely decide the course of action.

Neither can you, that's why we have conventions.

> You have replaced several variable declarations to the new style.

Most of which were missing a `const` as well.

eugeneko:
> Our code conventions do.

Our code conventions do not contain the rule "mindlessly follow everything mentioned in CCG without exception". Even CCG itself doesn't contain such rule.

> Neither can you, that's why we have conventions

Even if we ignore a topic of `{}` vs `=` for a second, you have violated said conventions in both this and previous PR. It's an evidence that you don't understand said conventions well enough to implement/enforce them properly.

Modanung:
> It's an evidence that you don't understand said conventions well enough to implement/enforce them properly.

Better and better.

PR #2597, opened 2020-03-02
Changed required TMX version from 1.0 to 1.*

Modanung:
The latest version of Tiled saves TMX files of version 1.2, Urho currently only accepts files that have version 1.0. To support the latest while maintaining backwards compatibility the version validation has been loosened to `StartsWith( "1.")`. It seems to work well enough at first glance, including for flipped and rotated tiles.

Also enforced ES.23 & Con.4 within the same function.

### DIFF
--- a/Source/Urho3D/Urho2D/TmxFile2D.cpp
+++ b/Source/Urho3D/Urho2D/TmxFile2D.cpp
@@ -453,15 +453,17 @@ bool TmxFile2D::EndLoad()
     if (!loadXMLFile_)
         return false;
 
-    XMLElement rootElem = loadXMLFile_->GetRoot("map");
-    String version = rootElem.GetAttribute("version");
-    if (version != "1.0")
+    const XMLElement rootElem{ loadXMLFile_->GetRoot("map") };
+    const String version{ rootElem.GetAttribute("version") };
+
+    if (!version.StartsWith("1."))
     {
         URHO3D_LOGERROR("Invalid version");
         return false;
     }
 
-    String orientation = rootElem.GetAttribute("orientation");
+    const String orientation{ rootElem.GetAttribute("orientation") };
+
     if (orientation == "orthogonal")
         info_.orientation_ = O_ORTHOGONAL;
     else if (orientation == "isometric")
@@ -481,16 +483,20 @@ bool TmxFile2D::EndLoad()
     info_.tileWidth_ = rootElem.GetFloat("tilewidth") * PIXEL_SIZE;
     info_.tileHeight_ = rootElem.GetFloat("tileheight") * PIXEL_SIZE;
 
-    for (unsigned i = 0; i < layers_.Size(); ++i)
+    for (unsigned i{ 0 }; i < layers_.Size(); ++i)
         delete layers_[i];
+
     layers_.Clear();
 
-    for (XMLElement childElement = rootElem.GetChild(); childElement; childElement = childElement.GetNext())
+    for (XMLElement childElement{ rootElem.GetChild() }; childElement; childElement = childElement.GetNext())
     {
-        bool ret = true;
-        String name = childElement.GetName();
+        bool ret{ true };
+        const String name{ childElement.GetName() };
+
         if (name == "tileset")
+        {
             ret = LoadTileSet(childElement);
+        }
         else if (name == "layer")
         {
             auto* tileLayer = new TmxTileLayer2D(this);
--- a/Source/Urho3D/Urho2D/TmxFile2D.cpp
+++ b/Source/Urho3D/Urho2D/TmxFile2D.cpp
@@ -375,7 +375,14 @@ Sprite2D* TmxImageLayer2D::GetSprite() const
 
 TmxFile2D::TmxFile2D(Context* context) :
     Resource(context),
-    edgeOffset_(0.f)
+    loadXMLFile_{ nullptr },
+    tsxXMLFiles_{},
+    info_{},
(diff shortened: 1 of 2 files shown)


Modanung:
According to the [TMX changelog](https://doc.mapeditor.org/en/stable/reference/tmx-changelog/) only new features were added, so no problems are to be expected.

weitjong:
As usual I would request the PR for fixes to not to have unrelated changes to make the review easier.

Modanung:
We should have a very clear definition of "unrelated" changes as:
1. Most of our codebase follows old conventions
2. People seem expect this to changegradually along with other changes
3. Style-only changes are forbidden

Stay logical

```diff
- if (version != "1.0")
+ if (!version.StartsWith("1."))
```

eugeneko:
> Most of our codebase follows old conventions

When Urho coding conventions were updated several years ago, we discussed what aspects of C++11 and CCG should be adopted and enforced in our codebase. For example, `override`, `nullptr`, `= default`, and some other aspects. Then, @weitjong enforced these aspects.
Since then, most of our codebase follows _current_ Urho coding conventions.
Whether said conventions match items of CCG or not is completely irrelevant.

Modanung:
@eugeneko Earlier you made the erroneous argument `{}` would break consistency. As can be seen in the changes in this PR, it is already used here and there (not by my doing).
```diff
- TileMapInfo2D info_{};
```

Modanung:
On the matter of `~T() override = default` the CCG (among other sources) has this to say:
> #### [C.dtor: Destructors](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#cdtor-destructors)
> "Does this class need a destructor?" is a surprisingly powerful design question. For most classes the answer is "no" either because the class holds no resources or because destruction is handled by the rule of zero; that is, its members can take care of themselves as concerns destruction.

> #### [C.30: Define a destructor if a class needs an explicit action at object destruction](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c30-define-a-destructor-if-a-class-needs-an-explicit-action-at-object-destruction)
> If the default destructor is needed, _but its generation has been suppressed_ (e.g., by defining a move constructor), use =default.

What reason do we have to (silently) handle different?

Modanung:
Alternate version:
https://github.com/urho3d/Urho3D/pull/2600

eugeneko:
> Please write down everything you said and simulate the path of a contributor

You seem to miss the point.
Repository maintainers are the people who decide what changes are applied and in what form.

@weitjong clearly don't want to waste his time on this discussion (and I understand him), but he asked to avoid changes that are not directly related to the subject of PR.

@ArnisLielturks is trying to stay neutral but he doesn't support restyling code for sake of restyling too.

Same goes for me, except I actually _tried_ to explain you technical reasons behind such decision.

I agree that we may need cleaner contribution guidelines and probably less ambigous coding conventions that don't leave the room for misinterpreting. However, people didn't have any issues with it so far.

---

> Earlier you made the erroneous argument {} would break the consistency that - in your view - defines our coding conventions. As can be seen in the changes in this PR, it is already used here and there (not by my doing).

`{}` is used for member initialization in Urho in many places.
It is _not_ used instead of `=` in variable initializations.
Therefore, using `{}` in _variable_ declarations instead of `=` breaks consistency of existing variable declarations. _Especially_ considering that `{}` cannot be used with `auto` and Urho is unlikely to adopt C++17 in the nearest future. Argument is valid.

Modanung:
Attaining clarity is not a waste of time. However pitiful this discussion may be.

> Urho is unlikely to adopt C++17 in the nearest future.

How come?

Modanung:
> Therefore, using {} in variable declarations instead of = breaks consistency of existing variable declarations.

Or one could say applying `{}` to variable declarations _increases_ consistency.

Modanung:
...and this...
>```
>        TrailPoint startPoint{previousPosition_, forwardMotion};
>        TrailPoint nextPoint{worldPosition, forwardMotion};
>```

>```
> Rect calcEffectiveBorder(const IntRect& border, const Vector2& drawSize)
> {
>     Vector2 min{Clamp(border.left_ * PIXEL_SIZE, 0.0f, drawSize.x_),
>                 Clamp(border.bottom_ * PIXEL_SIZE, 0.0f, drawSize.y_)};
>     return Rect{min, {Clamp(border.right_ * PIXEL_SIZE, 0.0f, drawSize.x_ - min.x_),
>                       Clamp(border.top_ * PIXEL_SIZE, 0.0f, drawSize.y_ - min.y_)} /* max*/ };
> }
>```

>```
>         SLNet::RakNetStatistics stats{};
>...
>         SLNet::RakNetStatistics stats{};
>...
>         SLNet::RakNetStatistics stats{};
>```

eugeneko:
> If "style corrections" cannot make up an entire PR and cannot be slipped into others... then how does this gradual change you mentioned earlier come to pass without generating inconsistency?

Gradual change is when you write _new_ code and apply conventions there.
Not when you artificially make PR 20 times bigger than it needs to be.

> ...and what's this?
> ...and this...
> ...and this...

Literally _zero_ of places you menitoned can be written using `=` with same clarity.
Can you find at least one place like `int value{ otherValue }`, with both `value` and `otherValue` of same type?

eugeneko:
Functional changes are merged in https://github.com/urho3d/Urho3D/commit/2345377c768d6a36fe76de200fdd518d2aca486f

PR #2600, opened 2020-03-03
Changed required TMX version from 1.0 to 1.* [ Light ]

Modanung:
Here's a _light_ version of my [earlier pull request](https://github.com/urho3d/Urho3D/pull/2597).

### DIFF
--- a/Source/Urho3D/Urho2D/TmxFile2D.cpp
+++ b/Source/Urho3D/Urho2D/TmxFile2D.cpp
@@ -375,7 +375,8 @@ Sprite2D* TmxImageLayer2D::GetSprite() const
 
 TmxFile2D::TmxFile2D(Context* context) :
     Resource(context),
-    edgeOffset_(0.f)
+    info_{},
+    edgeOffset_{ 0.f }
 {
 }
 
@@ -453,13 +454,10 @@ bool TmxFile2D::EndLoad()
     if (!loadXMLFile_)
         return false;
 
-    XMLElement rootElem = loadXMLFile_->GetRoot("map");
-    String version = rootElem.GetAttribute("version");
-    if (version != "1.0")
-    {
-        URHO3D_LOGERROR("Invalid version");
+    const XMLElement rootElem{ loadXMLFile_->GetRoot("map") };
+
+    if (!IsCorrectVersion(rootElem))
         return false;
-    }
 
     String orientation = rootElem.GetAttribute("orientation");
     if (orientation == "orthogonal")
@@ -525,6 +523,17 @@ bool TmxFile2D::EndLoad()
     loadXMLFile_.Reset();
     tsxXMLFiles_.Clear();
     return true;
+}
+
+bool TmxFile2D::IsCorrectVersion(const XMLElement& element) const
+{
+    const String version{ element.GetAttribute("version") };
+
+    if (version.StartsWith("1."))
+        return true;
+
+    URHO3D_LOGERROR("Invalid TMX version");
+    return false;
 }
 
 bool TmxFile2D::SetInfo(Orientation2D orientation, int width, int height, float tileWidth, float tileHeight)
--- a/Source/Urho3D/Urho2D/TmxFile2D.h
+++ b/Source/Urho3D/Urho2D/TmxFile2D.h
@@ -43,22 +43,16 @@ public:
 
     /// Return tmx file.
     TmxFile2D* GetTmxFile() const;
-
     /// Return type.
     TileMapLayerType2D GetType() const { return type_; }
-
     /// Return name.
     const String& GetName() const { return name_; }
-
     /// Return width.
     int GetWidth() const { return width_; }
-
     /// Return height.
     int GetHeight() const { return height_; }
-
     /// Return is visible.
     bool IsVisible() const { return visible_; }
-
     /// Return has property (use for script).
     bool HasProperty(const String& name) const;
     /// Return property value (use for script).
@@ -111,13 +105,11 @@ public:
 
     /// Load from XML element.
     bool Load(const XMLElement& element, const TileMapInfo2D& info);
-
     /// Store object.
     void StoreObject(const XMLElement& objectElem, const SharedPtr<TileMapObject2D>& object, const TileMapInfo2D& info, bool isTile = false);
 
     /// Return number of objects.
     unsigned GetNumObjects() const { return objects_.Size(); }
-
     /// Return tile map object at index.
     TileMapObject2D* GetObject(unsigned index) const;
 
@@ -137,10 +129,8 @@ public:
 
     /// Return position.
     const Vector2& GetPosition() const { return position_; }
-
     /// Return source.
     const String& GetSource() const { return source_; }
-
     /// Return sprite.
     Sprite2D* GetSprite() const;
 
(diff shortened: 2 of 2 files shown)


Source/Urho3D/Urho2D/TmxFile2D.cpp
```diff
@@ -527,6 +525,17 @@ bool TmxFile2D::EndLoad()
...
 }
 
+bool TmxFile2D::IsCorrectVersion(const XMLElement& element) const
+{
+    const String version{ element.GetAttribute("version") };
+
+    if (version.StartsWith("1."))
+        return true;
+
+    URHO3D_LOGERROR("Invalid TMX version");
```
SirNate0: I would suggest printing the invalid version to the log as well, e.g. 
`URHO3D_LOGERROR("Invalid TMX version: " + version);

eugeneko: That's good point.
I'll merge functional part of this PR later, I can add this.

I have no intention to merge this PR as is, because (1) it is political PR aimed to re-style existing codebase without any prior discussion and/or request, (2) it's about 20 times bigger than it needs to be and (3) it actually goes _against_ Urho coding conventions and, surprisingly, some items of C++ Code Guidelines -- only for sake of changing more lines of code.

Modanung:
@eugeneko By now it should be clear enough the conventions are not clear enough, instead of blasting someone as being on some "politcal" crusade. I have forked because of this nonsense, do not expect me to modify this PR.

eugeneko:
> instead of blasting me as being on some "politcal" crusade

I have probably worded it poorly.
For me, "convert Urho codebase into different style" _is_ political goal.
Our past migration to C++11 was also political change (by my criterias).

> after making the outrageous claim I would "never make any signification changes to the engine".

That's called "extrapolation". It's not me who makes this claim, it's math.
Of course, it's up to you whether this mathematical prediction will stay true or will be proven wrong.

> My intent is benign

I have never thought that you are evil, you know.

I also know people who are doing harm with 100% benign intent in mind, for variety of reasons.

> Also, please be more specific about the CCG items this goes against, vague claims do not inform

For a person who attempts to enforce CCG on existing codebase, you ask this question surprisingly often.

eugeneko:
Functional changes are merged in https://github.com/urho3d/Urho3D/commit/2345377c768d6a36fe76de200fdd518d2aca486f

Modanung:
> For me, "convert Urho codebase into different style" is political goal.

Which, in my view, is exactly not what I'm doing here.

eugeneko:
> I was trying to follow the conventions, which are obviously being ignored

It is _people_ who decide _what_ conventions are followed and _how_.
And people have stated their opinion in comments.

I really don't want to repeat previous discusions because it is emotionally taxing...
So I will wrap it with this statement:

You don't have to acknowledge opinion of maintainers when you write code.
Maintainers don't have to merge changes that they don't consider beneficial for the project.

Modanung:
You are once more contradicting your earlier statements claiming conventions should _not_ be defined by personal opinion. It basically comes down to the conventions not being documented properly. Instead they appear to be an occult set of arbitrary rules, that can only be learned about through an emotionally taxing process. This in itself is a harmful state of affairs that should not be considered solved by excluding contributors pointing out this deplorable fact.

eugeneko:
> conventions should not be defined by personal opinion

Conventions that cover every single aspect of coding would be too long and complicated to be practically usable. That's why we write down the most basic rules and use common sense for the rest.

It is common sense to write in the similar style as the code around. When in Rome, do as the Romans do.

Is it common sense to keep PR on topic and include only relevant changes. If you cannot give straight answer on the question "how this changed line helps users use the project", you probably don't need this line in the PR.

So far, the majority of contributors doesn't have issues with following this common sense.

Modanung:
Almost all code that is around _predates_ the latest changes in conventions.

> “The various modes of worship which prevailed in the Roman world were all considered by the people as equally true; by the philosopher as equally false; and by the magistrate as equally useful.” ― Edward Gibbon, The Decline and Fall of the Roman Empire

eugeneko:
> Why were they changed with no intent to follow them?

Urho community have never agreed to follow CppCoreGuidelines.
Therefore, CppCoreGuidelines have never been a part of Urho coding conventions.

Only a fraction of CCG items has been discussed and accepted.
All these items are listed directly in Urho coding conventions.
I believe that _these_ items are followed. Am I wrong?

eugeneko:
It's funny that we _did_ have the item

> Keep contributions consistent with existing code unless wide scale refactoring is performed.

And I removed it, probably to encourage people using new C++11 features in their PRs, like inplace member initialization and auto and whatever.

Modanung:
> It's recommended to follow C++ Core Guidelines (except items that contradict Urho legacy, e.g. ES.107, Enum.3, Enum.5)

I see only a few exceptions that are mentioned as not-to-follow. Meaning the other items should all be followed unless decided otherwise, and then they should be added to the list of exceptions that are not to be followed. It still seems to me like you are stretching the definition of "legacy" to the point of _ignore_ instead of _apply_ all contradictions in existing code.

eugeneko:
> Meaning the other items should all be followed

_I_ wrote this phase. So you cannot really argue with me about what it means.

Modanung:
Can you _explain_ what it means without rendering it meaningless?

SirNate0:
Firstly, I am of the opinion that contributions to Urho are too difficult to get to happen. That said, I do think that style changes should be made in separate PRs or locally around changes that are being made. The later may encompass this PR. Personally, I feel it's a bit of a stretch since this PR is effectively a 1 line change: `!=` becomes `! ... StartsWith`.

In terms of CCG, I personally don't like them, but I do feel the contribution guidelines are slightly ambiguous. I do come down a bit more on Eugene's side. I think "except items that contradict Urho legacy" would be the guiding principle in which to follow, and the listed ones are not meant to be an exhaustive list. If they were supposed to be exhaustive, "e.g." should be replaced with "i.e." or "namely".

Regarding the actual content, I do have a couple of questions:
1. Why were all the blank lines removed? I don't really care either way, I just want to know the reasoning behind which remained and which were removed.
2. Why was `void SetSpriteTextureEdgeOffset(float offset);` removed?
3. A comment this time: `TmxFile2D::IsCorrectVersion` should not have the log line, probably, and the function should probably just take the string version attribute rather than the element. Though if you want to leave it that way I have no problem with it, it just seems like a better practice to not have the test write to the log.

Though I see you've closed the PR, so feel free to ignore all that.

Modanung:
Non-exhaustive equals vague, in my opinion.

1. It is _common_ for them to be absent, and blank lines to indicate grouping.
2. `SetSpriteTextureEdgeOffset(float offset)` was _moved_ to be with the other setters.
3. I would have welcomed a discussion regarding the form of this function.

@eugeneko closed the PR because he merged his own version of it.

eugeneko:
> Can you explain what it means without rendering it meaningless?

If I expand missed or implied parts...

"It's is recommended (but not required) to follow C++ Core Guidelines (in the new or changed code), as long as it doesn't contradict Urho coding conventions and the code stays consistent with the rest of Urho codebase."

---

By the way, @SirNate0 have just won metaphorical prise. When I said that this PR contradicts CCG, I meant F.1 and F.2

Modanung:
@eugeneko Doesn't the newly introduced version check function _enforce_ F.2?

eugeneko:
> The codebase itself should never form the foundation of code conventions.

This is your personal opinion.

In all practical cases known to me, changes in coding conventions for existing codebase are _driven_ by the current state of this codebase in first place.

Modanung:
This is not some opinion, it is undeniable logic.
_Inspiring_ conventions on existing code is not the same as conventiones _referring_ to existing code.

eugeneko:
> It's like the rules of a sport being defined by the behaviour of the players

Laws of the nature are defined by the nature, not the opposite.
Laws of the country are defined by the culture of the people in this country (mostly), not the opposite.
Rules of the natural language are defined by how majority of people speak (mostly), not the opposite.
Coding conventions are defined by how majority of the code looks like (mostly), not the opposite.

If you don't argee with that... Well, people are free to think whatever they want.
I will not try to change your mind anymore.
If you don't like current Urho convetions, you are free to continue not liking them.
You not liking them is not good enough reason to change them.

Modanung:
Think @eugeneko, _think_. Don't just try to win an argument.

eugeneko:
You are right that existing codebase doesn't completely _define_ conventions.
We have examples of going against "consistency" in the recent update of conventions.
On the other hand, conventions that severely contradict the existing codebase and cannot be enforced bring harmful and confusing inconsistency in the codebase.

Discussing conventions, we search the balance between first and second.
And we _must_ look at existing codebase while doing it.

---

Anyway, all of it doesn't really matter now.
Urho coding conventions are what they are.
I can re-word that item in the way that better represents my initial intention.
If you don't like Urho coding conventions, you may try to change them, but it cannot be done in PR.

Modanung:
I never suggested any changes to the conventions.

eugeneko:
> I'm suggested they should be adhered to

You suggested to adhere to _misinterpretation_, tho.

But I admit that it was my fault -- I added too vague rule into conventions back then.

PR #2602, opened 2020-03-09
Refactoring of window management in Graphics

eugeneko:
So, here is my... fourth?.. attempt to make things better, and I _think_ this attempt is decent enough for PR.

Goals of this PR:

1) Fix several bugs in `Graphics::SetMode` transitions.

2) Make `ToggleFullscreen` behavior customizable.

3) Make default `ToggleFullscreen` behavior less useless. `ToggleFullscreen` doesn't violate display resolution by default anymore, and I find this behaviour much more appropriate.

4) To implement (2) and (3) I had to extract some copy-paste in Graphics.

I also added a sample for testing purposes, if anyone wants to try this PR. Sample is optional part of PR and may be reverted.

Commit messages contain intended functional changes. Please let me know if you spot any behavioral changes not listed here.

@ArnisLielturks, @rokups, maybe @weitjong?

Known bugs I left unfixed:

- DX9 and DX11 backends cannot turn on/off resizable flag after intialization

---

Intended changes:

- Semantics of multi sample level on DX9 backend has changed.
It represents actual hardware more instead of caller's desire.
Behavior of DX9, DX11 and OpenGL backend is synchronized now.

- ~~Secondary mode uses automatic window size~~.
Window will be toggled to desktop-resolution fullscreen.
Fullscreen will be toggled to default-size window.

- Desktop-size borderless window is preferred
to hardware fullscreen by default.

- Fix unwanted window reposition to (0, 0) during
windowed-to-fullscreen-to-windowed mode transition.

- Fix unwanted display resolution change during
fullscreen-to-windowed mode transition on DX9 and DX11.

### DIFF
--- a/Source/Samples/54_WindowSettingsDemo/CMakeLists.txt
+++ b/Source/Samples/54_WindowSettingsDemo/CMakeLists.txt
@@ -1,0 +1,33 @@
+#
+# Copyright (c) 2008-2020 the Urho3D project.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+# Define target name
+set (TARGET_NAME 54_WindowSettingsDemo)
+
+# Define source files
+define_source_files (EXTRA_H_FILES ${COMMON_SAMPLE_H_FILES})
+
+# Setup target with resource copying
+setup_main_executable ()
+
+# Setup test cases
+setup_test ()
--- a/Source/Samples/54_WindowSettingsDemo/WindowSettingsDemo.cpp
+++ b/Source/Samples/54_WindowSettingsDemo/WindowSettingsDemo.cpp
@@ -1,0 +1,314 @@
+//
+// Copyright (c) 2008-2020 the Urho3D project.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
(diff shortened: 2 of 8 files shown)


1vanK:
Currently we use `auto* xxx =` for pointers.
```auto* planeObject = planeNode->CreateComponent<StaticModel>();```

It this PR eugeneko uses `auto cache = GetSubsystem<ResourceCache>();`.
Should we standardize the style for this?

**Comment by rokups** [comment]
@eugeneko argued that `auto*` is pointless. I still think `auto*` is a good idea because:
* it is consistent with `auto&`
* it does give a visual hint that value is a pointer

Second point is especially important considering we need a different access operator (is it a right name) for references and pointers. Extra star does give a hint required to type a correct operator without double-checking.

**Comment by 1vanK** [comment]
In fact, I still feel the chair underneath me heating up when I see this

```auto* ui = GetSubsystem<UI>();```

https://github.com/urho3d/Urho3D/blob/master/Source/Samples/10_RenderToTexture/RenderToTexture.cpp#L219

**Comment by rokups** [comment]
Could you elaborate? What is wrong with it?

**Comment by 1vanK** [comment]
Because `UI* ui = GetSubsystem<UI>();` shorter and clearer

**Comment by rokups** [comment]
I agree. As of late i am trying to get rid of auto in my code as much as possible.

**Comment by eugeneko** [comment]
It is not required to use `auto*`, so I just went for the simplest option. It may be reconsidered.

> Second point is especially important considering we need a different access operator (is it a right name) for references and pointers. Extra star does give a hint required to type a correct operator without double-checking.

And I think this is almost pointless because plain `auto` can be a pointer too.
If you have smart pointer or iterator, you have to use plain `auto` anyway, even tho you have to use `->` access.

`auto*` is supported in the language _out of necessity_ in order to enable syntax like:
```
auto a = 10, *b = &a;
```
Because `*` is a part of variable name, not a part of the type.
Unless you actually use the syntax above, I don't see substantial benefit from `auto*`.
What's the point of special syntax for _some_ pointers, especially one that cannot be enforced?

**Comment by rokups** [comment]
Point is a helpful convention. We use many conventions that can not be 
enforced. We can also misuse things just like we could use plain auto for 
pointers. That does not mean we can not have a convention if it is useful.

**Comment by pengfei666** [comment]
I like `auto ui = GetSubsystem<UI>();` personally.

The reason is here, as we have used the class type `UI` as a template argument, so `UI* ui =` is redundant. Like `auto cache = GetSubsystem<ResourceCache>` is more simplified than `Resource_cache* cache = GetSubsystem<ResourceCache>` 

Similarily, `auto t = static_cast<int>(tt)` instead of `int t = static_cast<int>(tt)`. Actually, it will be reported an error if you are using the second format when you use clang tidy check. 

If you can't see the return type from the statement, I prefer adding a type name before the variable definition. 

Another thing, I also like `auto` instead of `auto*`. I don't know why. Just it

**Comment by eugeneko** [comment]
I used `auto*` in new sample for sake of consitency.
I also accidently pushed it into master branch *whistles*...
Let's say this PR was manually merged.

**Comment by eugeneko** [comment]
@rokups So, how exactly the `*` star helps here? I'd say it causes more confusion than it prevents:
```cpp
for (int i = 0; i <= 4; ++i)
{
  auto text = MakeShared<Text>(context_);
  text->SetText(i == 0 ? "No MSAA" : ToString("MSAA x%d", 1 << i));
  text->SetMinWidth(CeilToInt(text->GetRowWidth(0) + 10));
  multiSampleControl_->AddItem(text);
  text->SetStyleAuto();
}

auto* applyButton = window_->CreateChild<Button>("Apply");
applyButton->SetLayout(LM_HORIZONTAL, 6, IntRect(6, 6, 6, 6));
applyButton->SetStyleAuto();
```

rokups:
`*` hints that `applyButton` is a pointer, not a reference, not a value. On next line you know immediately that you are supposed to write `->` and not `.` when using `applyButton`.

---

But this PR is not about style. We should stop this offtopic.

Why is this PR closed?

eugeneko:
> Why is this PR closed?

I accidentally pushed all relevant commits in the master.

weitjong:
Can you provide the Script API binding as well for those new methods in `Urho3D::Graphics` class?

eugeneko:
If I could make simple one-line bindings, I would have done it.
However, I don't exactly know how to bind structs, especially in AS.
I've just looked at it, and it doesn't look simple.
It's definetely not something I can do properly in 5 mins.
And I don't really want to spend more than 5 mins on bindings unless I'm sure they will be used.
I'd postpone this task until demand.

weitjong:
OK. I will keep this PR open until then.

PR #2608, opened 2020-03-23
Emscripten shell

ArnisLielturks:
Current emscripten fullscreen functionality is broken, this is an attempt to fix it by using custom HTML shell and one additional JS to C++ function which tells the Engine current canvas size. The sample now uses all the available viewport in the browser to display the content and the canvas/renderer resolution is automatically updated on resize.

@eugeneko @weitjong @Modanung @rokups and everyone else interested in this.

**Bugs:**
- [x] Mouse tracking and grabbing when switching browser tabs
- [x] Wrong resolution detected in the browser
- [x] Loading app in background tab doesn't detect input when tab is activated
- [ ] Pointer lock is lost when canvas resizing happens, user must click on canvas to re-enable it
- [x] UI cursor issues when resolution is changed
- [x] UI cursor position wrong in 125% DPI screens


Source/Samples/01_HelloWorld/CMakeLists.txt
```diff
@@ -26,6 +26,11 @@ set (TARGET_NAME 01_HelloWorld)
 # Define source files
 define_source_files (EXTRA_H_FILES ${COMMON_SAMPLE_H_FILES})
 
+if (WEB)
+    # Custom emscripten template
+    define_resource_dirs (HTML_SHELL ${CMAKE_SOURCE_DIR}/web/application.html )
+endif()
+
```
eugeneko: Any chance to get this thing extracted into common place?
Maybe `setup_main_executable`?
I wonder if @weitjong has good ideas on the subject.

Source/Urho3D/Engine/Engine.cpp
```diff
@@ -63,8 +63,25 @@
...
+{
+    using namespace Urho3D;
+    URHO3D_LOGINFOF("JSCanvasSize: %d x %d", width, height);
+    appContext->GetSubsystem<Graphics>()->SetMode(width, height);
+    UI* ui = appContext->GetSubsystem<UI>();
+    ui->SetScale(scale);
+}
+
+using namespace emscripten;
+EMSCRIPTEN_BINDINGS(Module) {
```
eugeneko: Brace style

Source/Urho3D/Engine/Engine.cpp
```diff
@@ -63,8 +63,25 @@
...
-#if defined(__EMSCRIPTEN__) && defined(URHO3D_TESTING)
+#ifdef __EMSCRIPTEN__
+#include "../Graphics/GraphicsEvents.h"
 #include <emscripten/emscripten.h>
+#include <emscripten/bind.h>
+static const Urho3D::Context *appContext;
+
+static void JSCanvasSize(int width, int height, bool fullscreen, float scale)
+{
+    using namespace Urho3D;
```
eugeneko: Let's add error checks for null `appContext` and `ui`, just in case.
This function _technically_ can be called at any point of app lifetime.
And log `fullscreen` and `scale`, too.

web/application.html
```diff
@@ -0,0 +1,209 @@
...
+        var canvasWidth = (screen.availWidth - (window.outerWidth - window.innerWidth)) * devicePixelRatio;
+        var canvasHeight = (screen.availHeight - (window.outerHeight - window.innerHeight)) * devicePixelRatio;
+
+        // Detect fullscreen change and resize canvas resolution accordingly
+        function fullscreenHandler(event) {
+            if (Module['JSCanvasSize']) {
+                if (isFullScreen()) {
+                    Module.JSCanvasSize(screen.width * devicePixelRatio, screen.height * devicePixelRatio, true, devicePixelRatio);
+                } else {
+                    if (document.referrer.includes('itch.io')) {
```
eugeneko: This is oddly specific hack. Why is it needed?
It is worth documenting in comments, too.

web/application.html
```diff
@@ -0,0 +1,209 @@
...
+        document.addEventListener('fullscreenchange', fullscreenHandler, false);
+        document.addEventListener('mozfullscreenchange', fullscreenHandler, false);
+        document.addEventListener('webkitfullscreenchange', fullscreenHandler, false);
+        document.addEventListener('MSFullscreenChange', fullscreenHandler, false);
+
+        // When window size has changed, resize the canvas accordingly
+        window.addEventListener('resize', function(evt) {
+            canvasWidth = (screen.availWidth - (window.outerWidth - window.innerWidth)) * devicePixelRatio;
+            canvasHeight = (screen.availHeight - (window.outerHeight - window.innerHeight)) * devicePixelRatio;
+            fullscreenHandler(evt);
```
eugeneko: `fullscreenHandler` name isn't really appropriate if it's called on resize too.
Maybe rename it? `viewportResizeHandler`, or whatever you like.

web/application.html
```diff
@@ -0,0 +1,209 @@
...
+        });
+
+        canvasElement.addEventListener('click', function (evt) {
+            canvasElement.focus();
+            Module.JSMouseFocus();
```
eugeneko: Whose function is `JSMouseFocus`?
`JSCanvasSize` was user defined in C++. I don't see `JSMouseFocus` anywhere in the code.

web/application.html
```diff
@@ -0,0 +1,209 @@
...
+            print: function (text) {
+                console.log(text);
+            },
+
+            printErr: function(text) {
+                console.error(text);
+            },
+
+            // Urho3D called method which tells the canvas the current renderer resolution
+            SetRendererSize: function(width, height) {
```
eugeneko: How can it be called?

web/application.html
```diff
@@ -0,0 +1,209 @@
...
+                this.canvas.style.height = (canvasWidth / Module.rendererSize.aspectRatio) + 'px';
+            },
+
+            // Retrieve the current status of the application
+            setStatus: function(text) {
+                if (text === 'Running...') {
+                    ready();
+                }
+                if (Module.setStatus.interval) clearInterval(Module.setStatus.interval);
+                var m = text.match(/([^(]+)\((\d+(\.\d+)?)\/(\d+)\)/);
```
eugeneko: I don't know JS best practices, but formatting data to the string to decode it later feels hacky.
Why function param cannot be used instead?
Is it fixed-signature function or something?

eugeneko:
General issue: can you share/publish one sample that captures mouse cursor (e.g. Sample 04) and one sample that doesn't (e.g. HelloGUI sample) so we can try it live?

web/application.html
ArnisLielturks: This part is actually comes from one of the emscripten shells, haven't change it so far.

Source/Samples/01_HelloWorld/CMakeLists.txt
ArnisLielturks: Found a better place https://github.com/urho3d/Urho3D/blob/master/CMake/Modules/UrhoCommon.cmake#L1109-L1115

Source/Urho3D/Engine/Engine.cpp
ArnisLielturks: Seems like this macro doesn't like a case without braces

Source/Urho3D/Engine/Engine.cpp
ArnisLielturks: will do

web/application.html
ArnisLielturks: removed

web/application.html
ArnisLielturks: `viewportResizeHandler ` it is

web/application.html
ArnisLielturks: removed

web/application.html
ArnisLielturks: updated missing parts

ArnisLielturks:
> General issue: can you share/publish one sample that captures mouse cursor (e.g. Sample 04) and one sample that doesn't (e.g. HelloGUI sample) so we can try it live?

I sure can, will provide the links tomorrow. Since I was able to fix the html shell definition in the UrhoCommon file will probably upload all the samples that we have.

ArnisLielturks:
@eugeneko or today, here's the link: http://67.205.130.249/

Modanung:
Switching tabs seems to break mouse grabbing.

rokups:
HelloGUI sample loads slightly blurry (see text). Pressing full-screen button makes text crisp. However if i open inspector - text also becomes crisp. Is it possible that there is an issue with detection of canvas size somehow? A rounding error maybe? No idea. Tested on firefox/linux.

ArnisLielturks:
@rokups @Modanung thanks for the feedback, added both bugs in the description of the PR to track the current progress of it.

I guess we should also test this against the current emscripten functionality, since I believe there are more bugs in the original emscripten code just so this PR doesn't go out of scope.

Modanung:
I did, the bug I mentioned seems new.

eugeneko:
1) Both windowed and fullscreen are blurry, resolution is 1/1.25 of what it should be (I have 125% DPI scale on my PC)

2) When going FS and even after exiting it, both keyboard and mouse input are locked.
Hard to tell if it's new bug or old becase there were no fullscreen...

Chrome 80.0.3987.132 x64, Win10

ArnisLielturks:
@eugeneko There is a workaround:
Type this in developers console when canvas is black or not visible
```
Module.canvas.width = 1024;
Module.canvas.height = 768;
```

Modanung:
Did you know html supports svg as an embedded element? It _would_ be possible to have some clickable icon instead of a button with text without having extra files, _if_ that's not considered "bikeshedding" of course.

ArnisLielturks:
@Modanung I know it does, I was thinking of maybe implementing something similar of what itch.io uses. And no, it's not bikeshedding, it's an obvious UI improvement of the engine. PR's with the icon are welcome. 👍

ArnisLielturks:
@rokups Can you check the updated canvas size detection?
@Modanung the mouse grab issue also should be resolved

http://67.205.130.249/

bin/shell.html
```diff
@@ -117,15 +120,18 @@
             return !!(document.fullscreenElement || document.webkitFullscreenElement || document.mozFullScreenElement || document.msFullscreenElement);
         }
 
-        // Emscripten triggered ready event when the app is actually started
+        // App is ready to launch, make canvas and fulscreen button visible
```
Modanung: fulscreen lacks an L

Modanung:
And indeed the mouse grabbing works! :+1:

ArnisLielturks:
There is another issue, when the app is loaded in the inactive tab, the input won't work. Will look into that now.

eugeneko:
@ArnisLielturks I can confirm that both my issues are fixed in the current published version.
Fullscreen is pixel-perfect and input works.

However, quasi-fullscreen via F11 is not pixel-perfect (slightly blurry) and there's white gap at the bottom.
I don't think it's _that_ important to get F11-fullscreen working now, but it's worth checking IMO.
May be some bug missed.

rokups:
My previous issue is fixed. It is pixel-perfect in F11 mode even. What browser are you using @eugeneko?

@ArnisLielturks resizing browser window does not update canvas size, instead canvas is stretched (firefox).

Modanung:
@rokups 
> Chrome 80.0.3987.132 x64, Win10

eugeneko:
> It is pixel-perfect in F11 mode even

Did you check it with image editor?

ArnisLielturks:
@rokups The window resize event now works correctly, basically just added a delay to the actual resize function. Since the event is called before the resizing has been finished and the app has no idea about the new resolution, it basically just took the old resolution instead.

@eugeneko I think this also fixes the F11 fullscreen issue. 

Plus I think I finally managed to find a workaround of the background loaded tabs. To test it simply open one of the samples, go to another tab for a few seconds till the app is started and then go back to the sample tab. I think this is the last major bug that I know of so far.

ArnisLielturks:
There is one thing that I don't quite like about my implementation: https://github.com/urho3d/Urho3D/pull/2608/files#diff-49bca172f0639e313df34ce2741de7e0

Not really sure what could be a proper place where these handlers could be added. Maybe `Application` class? 
I was actually hoping to get this working together with this https://github.com/urho3d/Urho3D/pull/2538 since the web is just another platform that should be handled separately without making too much mess in the base classes.

rokups:
Looks like it deals with graphics stuff. Why not put it in graphics subsystem? You wouldnt even need to subscribe to E_SCREENMODE then, just do resize before event is sent. Should be safer too, user would receive event after all of moving parts regarding screen mode switching have completed.

ArnisLielturks:
> Looks like it deals with graphics stuff. Why not put it in graphics subsystem? You wouldnt even need to subscribe to E_SCREENMODE then, just do resize before event is sent. Should be safer too, user would receive event after all of moving parts regarding screen mode switching have completed.

Not a bad idea, thanks!

rokups:
Samples still not updated? All of them show me a black screen.

rokups:
When sample loads it wont update resolution for some reason. Default resolution is stretched over entire screen until i trigger any resize event. Then it resets to a crisp image. That delay between resize and update is bit awkward, but definitely better than blurred resolution.

ArnisLielturks:
@rokups oops, my bad. Probably forgot to commit something, that's what I get when trying to work on multiple envs. Looking into that

eugeneko:
Current version starts blurry until I do something to resize it.
Sometimes it resizes to the right size.

Chrome 80.0.3987.149 x64, Win10.

Bug: When window is resized by timer, input focus is lost but mouse is still invisible.

ArnisLielturks:
Please ignore the current samples, I have messed them up, they doesn't reflect the latest changes, it's something inbetween. I have 2 environments, I do commits on 1 machine, but do the compiling stuff on another. There is slight mismatch of the codebase on both of them.

ArnisLielturks:
@eugeneko @rokups Samples updated. Sorry for disturbing, hopefully no new issues pops up. 🤞

eugeneko:
Window is always pixel-perfect now.

New bug spotted.
When the tab is loaded in the background, mouse is hidden after resize by timer.

STR:

1) Middle-click `02_HelloGUI`

2) Wait for loading

3) Open the tab

4) Wait for auto resize

5) Click the content

Mouse is hidden even tho it shall not be hidden in this sample.

Focus is still lost after auto-resize. I have to click to re-enable input.
This may really be confusing since it looks like game freeze or input loss.

ArnisLielturks:
I guess the only bug that remains is the pointer lock loss when canvas resizing happens, user has to click on canvas to re-enable it. Other bugs seem to be resolved.

rokups:
Any idea if it is possible to make window not react to mouse move events if sample uses pointer lock? Currently it is rather awkward that camera rotates when we move mouse over window of samples like SkeletalAnimation. They are obviously not made to interact with camera that way.

ArnisLielturks:
@rokups I couldn't find a flag or a variable in emscriptens JS code that would indicate if the sample should grab the mouse or not. If we just ignore that we might implement another method that would expose the mouse state to JS just like it is with the renderer resolution now. I might add that functionality a bit later, preferably on another PR just so we can merge this and fix the broken web port.

eugeneko:
I agree, it's better to extract input tweaking into separate PR.

ArnisLielturks:
Noticed one more thing, the UI cursor becomes hidden (or even removed?) when resolution is changed. Any ideas what could cause that?

eugeneko:
UI cursor is worthless in Web anyway. It's so wobble and laggy that I'm literally getting sick looking at it.
Why do we hide native cursor in Sample 08 tho? We don't do it in e.g. Sample 02.

ArnisLielturks:
Recreating the UI cursor after resolution change has no effect. UI cursor remains invisible.

ArnisLielturks:
The UI cursor issue should be fixed. Also fixed another bug in the UI cursor functionality, which didn't took into account UI scale, so when UI scale is set to something != 1.0, the cursor position detection was wrong.

Modanung:
I think it would make sense if switching to fullscreen would simultaneously grab the cursor instead of requiring an extra click.

ArnisLielturks:
@Modanung At the moment not really sure if it's possible. That might be fixed in the near future with the approach mentioned here: 

> @rokups I couldn't find a flag or a variable in emscriptens JS code that would indicate if the sample should grab the mouse or not. If we just ignore that we might implement another method that would expose the mouse state to JS just like it is with the renderer resolution now. I might add that functionality a bit later, preferably on another PR just so we can merge this and fix the broken web port.

SirNate0:
@ArnisLielturks I love this PR. I am a little disappointed to see the web console display go, but this is a much, much more usable interface for an actual game. Only 2 issues I'm noticing (Chrome on Ubuntu 16.04 LTS). First, when resizing the window on, for example, http://67.205.130.249/bin/05_AnimatingScene.html, after the pointer has been grabbed & hidden by the sample, the mouse remains hidden but moving the mouse no longer moves the camera. I'm assuming this is what you're referring to by "Pointer lock is lost when canvas resizing happens, user must click on canvas to re-enable it" as after I click, all is fixed. Other than that, there is the delay in resizing the canvas after the window has been resized, which seems perfectly reasonable to me (it's like a second, which seems acceptable). A weird thing happens if I resize the window using the borders though -- for about that second, the canvas just stretches, but then it starts flickering between black and white and a resized version of the canvas as I resize the window.

ArnisLielturks:
> @ArnisLielturks I love this PR. I am a little disappointed to see the web console display go, but this is a much, much more usable interface for an actual game. Only 2 issues I'm noticing (Chrome on Ubuntu 16.04 LTS). First, when resizing the window on, for example, http://67.205.130.249/bin/05_AnimatingScene.html, after the pointer has been grabbed & hidden by the sample, the mouse remains hidden but moving the mouse no longer moves the camera. I'm assuming this is what you're referring to by "Pointer lock is lost when canvas resizing happens, user must click on canvas to re-enable it" as after I click, all is fixed. Other than that, there is the delay in resizing the canvas after the window has been resized, which seems perfectly reasonable to me (it's like a second, which seems acceptable). A weird thing happens if I resize the window using the borders though -- for about that second, the canvas just stretches, but then it starts flickering between black and white and a resized version of the canvas as I resize the window.

Delays are there because specific resize events are called before the resizing is actually finished, it helps to detect the new viewport size correctly. This timer may be decreased to maybe 500ms, but it will still be there. Regarding the console output, everything is redirected to browsers debugger console by using either `console.log` or `console.error`

eugeneko:
@ArnisLielturks Do you have current version available for testing?

ArnisLielturks:
@eugeneko all the latest changes should be available here http://67.205.130.249/ but probably required to clear the browser cache first.

eugeneko:
For the record, one more bug:

In the Sample 08:

1) Capture mouse so the native cursor is visible

2) Press RMB so the rotation mode is enabled and the cursor is hidden

3) Press Esc _quickly several times_

4) Release RMB

5) Native cursor appears, but UI cursor is forever gone

I'm not sure that this bug is in the scope of current inspection, but it's worth recording.

SirNate0:
Update to what @eugeneko found:
Also with Sample 8. If you just press escape twice without clicking in between (so you show the cursor, at which point you have the UI cursor lagging behind the OS cursor, and then you press escape again) and the UI cursor disappears. Resizing the window seems to bring it back, however. Potentially this is expected and unrelated to what Eugene found, but either way I agree that this PR should not wait to fix that issue (the user can always use a custom shell to revert back to the old behavior by using the old html file if it is a pressing issue for anyone).

eugeneko:
@ArnisLielturks One more defect, this one is more serious.
On my laptop with 125% DPI, UI cursor image is rendered 1.25 times further from (0, 0) than it shall be.
However, the clicks are processed in the right place (where hardware cursor pointing)

eugeneko:
Also, I have hard-to-reproduce issue with background loading of Sample 02 on both PCs.
If it is first-time load (with downloading) and the tab is in the background, the input is lost.

Source/Urho3D/UI/UI.cpp
```diff
@@ -1797,7 +1797,7 @@ void UI::HandleMouseMove(StringHash eventType, VariantMap& eventData)
         if (!input->IsMouseVisible())
         {
             if (!input->IsMouseLocked())
-                cursor_->SetPosition(IntVector2(eventData[P_X].GetInt(), eventData[P_Y].GetInt()));
+                cursor_->SetPosition(IntVector2(eventData[P_X].GetInt(), eventData[P_Y].GetInt()) / GetScale());
```
eugeneko: Huh. I suppose here is the bug, that could have been chaught by asserts in https://github.com/urho3d/Urho3D/pull/2520

ArnisLielturks:
@eugeneko Thanks for your time. Will look into those issues. Regarding the background tab loading I have a code that replaces default emscripten app launching. It starts the app only when the tab becomes visible, maybe it's wise to introduce these changes in this PR. Also there is a bug in the default behaviour, since the same WASM object is actually downloaded twice. Single sample app weights about 20MB, so by downloading it twice the initial load becomes twice as slow.

Source/Urho3D/UI/UI.cpp
ArnisLielturks: `int / float` instead of `IntVector / float` fixed it.

ArnisLielturks:
125% DPI cursor issue has been resolved.

Other bugs that were mentioned here will not be fixed in this PR. With that being said I believe the work on this PR is done. You can still of course do a review on this and let me know about any code related issues that you notice, will fix them as quickly as I can.

eugeneko:
Don't see any new bugs in samples. Will review code a bit later.

eugeneko:
@ArnisLielturks didn't want to bother you with irrelevant issues because it was simpler to fix them myself.

I fixed code style/alignment issues in the new code, got rid of one non-initialized variable because they give me the creeps, and simplified scale handling in UI because it looked a bit too cumbersome for me.

If you have any comments on my latest commit, say so.
If not, I'm fine with merging it (with squash, because commit history is quite dirty now).

I didn't really review JS-part because I have almost zero exp with JS...

ArnisLielturks:
@eugeneko totally fine by me and thanks again for your effort. Let's wait for the CI to pass then do the merge

CMake/Modules/UrhoCommon.cmake
```diff
@@ -1106,13 +1107,13 @@ macro (add_html_shell)
...
         else ()
             # Create Urho3D custom HTML shell that also embeds our own project logo
-            if (NOT EXISTS ${CMAKE_BINARY_DIR}/Source/shell.html)
+            if (NOT EXISTS ${CMAKE_SOURCE_DIR}/bin/shell.html)
                 file (READ ${EMSCRIPTEN_ROOT_PATH}/src/shell.html HTML_SHELL)
                 string (REPLACE "<!doctype html>" "<!-- This is a generated file. DO NOT EDIT!-->\n\n<!doctype html>" HTML_SHELL "${HTML_SHELL}")     # Stringify to preserve semicolons
                 string (REPLACE "<body>" "<body>\n<script>document.body.innerHTML=document.body.innerHTML.replace(/^#!.*\\n/, '');</script>\n<a href=\"https://urho3d.github.io\" title=\"Urho3D Homepage\"><img src=\"https://urho3d.github.io/assets/images/logo.png\" alt=\"link to https://urho3d.github.io\" height=\"80\" width=\"160\" /></a>\n" HTML_SHELL "${HTML_SHELL}")
-                file (WRITE ${CMAKE_BINARY_DIR}/Source/shell.html "${HTML_SHELL}")
+                file (WRITE ${CMAKE_SOURCE_DIR}/bin/shell.html "${HTML_SHELL}")
             endif ()
```
weitjong, line 1115: If the html shell is provided out-of-the-box now in our source tree, instead of being generated on-the-fly in our build tree, then these lines can be removed all together.

ArnisLielturks: The idea here was to leave this just so it's easy to go back to legacy template by simly removing the new one.

weitjong: I see but if that is the case then I find it a little offending because the script tries to write into "source tree" directly and user will end up in dirty source tree afterward.

ArnisLielturks: Makes sense, will update the PR

ArnisLielturks: updated with different approach

ArnisLielturks:
@weitjong could you check the latest updates?

weitjong:
Thanks for the updates. It still means someone would have to make the source tree dirty by intentionally deleting the provided shell file in order to get this fallback generated shell file, which I don't think we need to keep. Having said that, I will not stop you from merging this PR because of this small issue.

weitjong:
@ArnisLielturks Will you provide the script API binding?

Source/Urho3D/Graphics/OpenGL/OGLGraphics.cpp
```diff
@@ -53,6 +53,12 @@
 #endif
 
 #ifdef __EMSCRIPTEN__
+#include "../Input/Input.h"
```
rokups, line 56: These relative includes are incorrect and work only by chance. It should be `../../Input/Input.h`.

PR #2609, opened 2020-03-23
CMake config for BindingGenerator

1vanK:
I need help with CMake config. I use
```
COMMAND BindingGenerator "${CMAKE_BINARY_DIR}/Source/Tools/BindingGenerator/generated/xml" "${CMAKE_BINARY_DIR}/Source/Tools/BindingGenerator/generated"
```
to launch BindingGenerator (input dir with xmls and output dir for generated files). BindingGenerator is launched but recieve single argument wih path to executable. What am I doing wrong?


1vanK:
I found the problem

PR #2610, opened 2020-03-25
Update pugixml

1vanK:
Pugixml is active used in bindings generator so I update this lib.

If I understand correctly this modification in old version just duplicates default copy constructor
```
	// Urho3D - copy constructor from another xpath_node
	PUGI__FN xpath_node::xpath_node(const xpath_node& rhs) : _node(rhs._node), _attribute(rhs._attribute)
	{
	}
```
so I did not transfer this to the new version.

### DIFF
--- a/Docs/Urho3D.dox
+++ b/Docs/Urho3D.dox
@@ -251,7 +251,7 @@ Urho3D uses the following third-party libraries:
 - Mustache 1.0 (https://mustache.github.io, https://github.com/kainjow/Mustache)
 - nanodbc 2.12.4+ (https://lexicalunit.github.io/nanodbc)
 - Open Asset Import Library 4.1.0 (http://assimp.sourceforge.net)
-- pugixml 1.7 (http://pugixml.org)
+- pugixml 1.10+ (http://pugixml.org)
 - rapidjson 1.1.0 (https://github.com/miloyip/rapidjson)
 - Recast/Detour (https://github.com/recastnavigation/recastnavigation)
 - SDL 2.0.10+ (https://www.libsdl.org)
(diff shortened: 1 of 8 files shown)


eugeneko:
> If I understand correctly this modification in old version just duplicates default copy constructor

Does this class have default copy ctor, tho?

On the other hand... If CI end up green, maybe we don't need it at all.

1vanK:
> Does this class have default copy ctor, tho?

Not sure I understood the question right. Any class has a default copy constructor if it is not deleted manually.

eugeneko:
> Any class has a default copy constructor if it is not deleted manually.

This statement is false.
Any class has a default copy constructor if all the requirements for default copy constructor are met.
Including, but not limited to, "not deleted manually".

I don't really remember all the rules... Let's wait for CI.

1vanK:
https://en.cppreference.com/w/cpp/language/copy_constructor
```
If no user-defined copy constructors are provided for a class type (struct, class, or union), the compiler will always declare a copy constructor as a non-explicit inline public member of its class.
```

iSLC:
What @eugeneko probably means is that even if one of the member variables does not meet the [copy-constructibe](https://en.cppreference.com/w/cpp/types/is_copy_constructible) requirements. The compiler cannot provide one (_obviously_).

eugeneko:
> the compiler will always declare a copy constructor

Which may or may not be deleted, by the way.
This statement doesn't guarantee that you will get actually working copy ctor.

Also, compilers don't always follow the standard.
I wonder if there's some interference from other ctors of `xpath_node`, maybe on some specific compilers.

1vanK:
Can this be merged?

eugeneko:
I don't see any issues here. @ArnisLielturks, what do you think?

weitjong:
If you go through the git log history, you can see some else has done exactly the same thing while upgrading the version from 1.5 to 1.7 (ec3cbc37b630d111f9f96ab26987e62ae31d876e), and later the copy constructor fix was reinstated again (ea3f05c1c5375f83eaa831443b93396252ae83ff). Unfortunately I cannot recall specifically why I needed to make it explicit then. So, may be will merge your PR as it is now and see how.

Issue #2555, opened 2019-12-04
Mobile Web Sample Screen Resolution


SirNate0:
On an Essential PH-1, at least, the window size ends up significantly too high as resolution on the web samples. From the log, `INFO: Set screen mode 3072x2304 windowed monitor 0`, which is certainly to high considering the screen is actually 2560x1312. At least for mobile, I think whatever code determines the web build window size needs to be revised.

Also, I'm pretty sure this is not just limited to Android+Chrome:

>    BTW, the sample rendered so small on actual 
>    iPhone device. The sample is hardly usable 
>    because it is not using responsive design.

_Originally posted by @weitjong in https://github.com/urho3d/Urho3D/issues/2479#issuecomment-521486531_

eugeneko:
I have somewhat similar issue. It seems that Web sample uses physical size of the screen instead of browser client area. Since I'm not having browser fullscreen, I end up with x2 bigger engine window than expected.

SirNate0:
This might be relevant, we may need to use the same behavior we do on Mac/iOS:

> Ar­guably to sim­pli­fy port­ing, the Em­scripten SDL em­u­la­tion re­cal­cu­lates all in­put event co­or­di­nates to match frame­buffer pix­els. The ac­tu­al DPI scal­ing (or de­vice pix­el ra­tio) is then be­ing ex­posed through dpiS­cal­ing(), mak­ing it be­have the same as Lin­ux, Win­dows and An­droid on high-DPI screens. In con­trast, HTM­L5 APIs be­have like mac­OS / iOS and Plat­form::Em­scripte­nAp­pli­ca­tion fol­lows that be­hav­ior — frame­buf­fer­Size() thus match­es de­vice pix­els while win­dow­Size() (to which all events are re­lat­ed) is small­er on HiD­PI sys­tems. For more in­for­ma­tion, check out the [DPI aware­ness docs](https://doc.magnum.graphics/magnum/classMagnum_1_1Platform_1_1EmscriptenApplication.html#Platform-EmscriptenApplication-dpi).

[https://blog.magnum.graphics/announcements/new-emscripten-application-implementation/#dpi-awareness-revisited](https://blog.magnum.graphics/announcements/new-emscripten-application-implementation/#dpi-awareness-revisited)

SirNate0:
@eugeneko I ran your web samples (https://eugeneko.github.io/) for your light mapper and noticed this problem seemed not to be there. Did you figure out the cause/a workaround?

[...]

PR #2611, opened 2020-03-28
Fix mobile web sample resolution

SirNate0:
Fix mobile web sample resolution by setting EP_HIGH_DPI to false on Web platform (Issue #2555).


ArnisLielturks:
I have also done this in this PR https://github.com/urho3d/Urho3D/pull/2608 but not because of the high resolution concern. The high dpi was turned on by default on all browsers, now it actually checks that in browser and scales up the resolution accordingly.

SirNate0:
I'm going to close this then in favor of the other PR. If needed it can be reopened later.

PR #2612, opened 2020-03-29
Add support for animated tilesets (tmx files)

pflat:
These changes allow the use of animated tiles in tmx maps.
I tested it by updating the map of sample 36 to contain some animated tiles.
I don't know if the way I choose to implement this fits with the codebase (I tried to). If it does, I'll check AS and LUA bindings. I'm dabbling with those bindings, so I'll only commit to that code if this one is ok.

### DIFF
--- a/Source/Urho3D/Urho2D/TileMap2D.cpp
+++ b/Source/Urho3D/Urho2D/TileMap2D.cpp
@@ -27,6 +27,7 @@
 #include "../Resource/ResourceCache.h"
 #include "../Scene/Node.h"
 #include "../Scene/Scene.h"
+#include "../Scene/SceneEvents.h"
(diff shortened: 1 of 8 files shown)


eugeneko:
Apparently it causes crashes in two samples.

pflat:
I believe I got the problem solved. Waiting for the tests to complete.

weitjong:
Although I have not tested it, I believe it is merge-ready. I would have merge it if the script API binding is also provided.

pflat:
The bindings for TmxFile2D were incomplete, both in LUA and AS. I  added bindings for some methods, but the ones still remaining left me uncertain of how to do it.
Also for AngelScript, some classes had incorrect binds for the method GetProperty (they were "pointing" to HasProperty). I took the change correct it.

PR #2619, opened 2020-04-04
Path Class

SirNate0:
Rather than using strings everywhere that are almost exclusively treated as paths, I am proposing we switch to a separate class for paths and for strings. At present, this can't be considered finished (especially since the Script bindings have not been added for the new class), but I would estimate that it is most of the way to being complete at the moment. I haven't tested it on platforms other than Linux and Web at present, so we'll see the results from the CI build for that. Probably some more work remains to be done with Serializables especially, even if it compiles and runs fine at present (implicit conversion from String to the rescue).

This has been discussed some on the forum already, though that thread has more of a focus on specifically allowing relative paths as opposed to switching to a Path class: https://discourse.urho3d.io/t/adding-relative-paths-to-resource-loading/5911/42

Some design choices should probably be considered before this is merged, including (though not limited to):
1. Should Resource and the like have `const Path& GetName() const` or should we stick with what I have done at the present with two methods `const String& GetName() const` and `const Path& GetNamePath() const`, though internally it is just stored as a `Path`.
2. How much functionality of String should be exposed for the Path class? At present it's basically just Empty, Clear, Length, Compare, and an optional CString (the last to ease the transition). What about Contains, Find[Last], etc.?
3. Should JSONValue have a completely separate Path type?
4. Should we switch to a GLOB and/or REGEX based system for FileSystem::CheckAccess? What about speed considerations? Should we support multiple options, possibly with a build flag and/or (constexpr) global variable?
5. Should we wait until the automatic binding generator is finished to work on the script stuff? (Not a design decision, but I still vote yes for this).

With that said, I know it's a long set of changes, but I would very much appreciate feedback from others.

### DIFF
--- a/Source/Samples/Sample.inl
+++ b/Source/Samples/Sample.inl
@@ -344,7 +344,7 @@ void Sample::HandleKeyDown(StringHash /*eventType*/, VariantMap& eventData)
             Image screenshot(context_);
             graphics->TakeScreenShot(screenshot);
             // Here we save in the Data folder with date and time appended
-            screenshot.SavePNG(GetSubsystem<FileSystem>()->GetProgramDir() + "Data/Screenshot_" +
+			screenshot.SavePNG(GetSubsystem<FileSystem>()->GetProgramDir() + "Data/Screenshot_" +
                 Time::GetTimeStamp().Replaced(':', '_').Replaced('.', '_').Replaced(' ', '_') + ".png");
         }
     }
--- a/Source/Urho3D/AngelScript/Script.h
+++ b/Source/Urho3D/AngelScript/Script.h
@@ -74,7 +74,7 @@ public:
     /// Set whether to execute engine console commands as script code.
     void SetExecuteConsoleCommands(bool enable);
     /// Print the whole script API (all registered classes, methods and properties) to the log. No-ops when URHO3D_LOGGING not defined.
-    void DumpAPI(DumpMode mode = DOXYGEN, Path sourceTree = String::EMPTY);
+    void DumpAPI(DumpMode mode = DOXYGEN, Path sourceTree = Path::EMPTY);
     /// Log a message from the script engine.
     void MessageCallback(const asSMessageInfo* msg);
     /// Handle a script exception.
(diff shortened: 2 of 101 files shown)


SirNate0:
Forgot the URHO3D_EXTRAS, I will fix those and re-upload later.

Source/Samples/Sample.inl
```diff
@@ -344,7 +344,7 @@ void Sample::HandleKeyDown(StringHash /*eventType*/, VariantMap& eventData)
             Image screenshot(context_);
             graphics->TakeScreenShot(screenshot);
             // Here we save in the Data folder with date and time appended
-            screenshot.SavePNG(GetSubsystem<FileSystem>()->GetProgramDir() + "Data/Screenshot_" +
+			screenshot.SavePNG(GetSubsystem<FileSystem>()->GetProgramDir() + "Data/Screenshot_" +
```
KatekovAnton, line 347: Why padding is so big? Better to remove such trash from PRs

1vanK:
I still don't understand why need to put a string with a path inside another class.

Even in C# Path just contains collection of static functions to manipulate with strings

https://docs.microsoft.com/en-us/dotnet/api/system.io.path?view=net-5.0

Source/Urho3D/AngelScript/Script.h
```diff
@@ -74,7 +74,7 @@ class URHO3D_API Script : public Object
     /// Set whether to execute engine console commands as script code.
     void SetExecuteConsoleCommands(bool enable);
     /// Print the whole script API (all registered classes, methods and properties) to the log. No-ops when URHO3D_LOGGING not defined.
-    void DumpAPI(DumpMode mode = DOXYGEN, const String& sourceTree = String::EMPTY);
+    void DumpAPI(DumpMode mode = DOXYGEN, Path sourceTree = Path::EMPTY);
```
1vanK, line 77: const Path& sourceTree

Source/Urho3D/AngelScript/ScriptAPIDump.cpp
```diff
@@ -190,7 +190,7 @@ void Script::OutputAPIRow(DumpMode mode, const String& row, bool removeReference
     }
 }
 
-void Script::DumpAPI(DumpMode mode, const String& sourceTree)
+void Script::DumpAPI(DumpMode mode, Path sourceTree)
```
1vanK, line 193: const Path& sourceTree

1vanK:
The code didn't get any easier with the new class. You just added conversions from strings and back to strings.

PR #2620, opened 2020-04-08
Autobinding enums for AngelScript

1vanK:
I need feedback from @weitjong about building system. I'm not sure I made right changes in CMake configuration files. I just made it work somehow.

I have version with automatic generation classes and functions, but code still dirty, so I PR minimal version which binds only enums to get tips about the build system.

But this minimal version may also be useful, as it fixes some detected problems with existing bindings (MouseButton incomplete, TileObjectType2D instead TileMapObjectType2D, etc)

With URHO3D_GENERATEBINDINGS=1:
1) Call doxygen to generate XMLs
2) Build https://github.com/1vanK/Urho3D/tree/master/Source/Tools/BindingGenerator
3) Generate https://github.com/1vanK/Urho3D/blob/master/Source/Urho3D/AngelScript/Generated.cpp
4) Build engine

Some notes:
1) Automatically binds all enums from Urho3D namespace
2) Ignored header files need to specify in https://github.com/1vanK/Urho3D/blob/master/Source/Tools/BindingGenerator/Tuning.cpp
3) Single ignored enums are marked in source code: https://github.com/1vanK/Urho3D/blob/master/Source/Urho3D/Graphics/GraphicsDefs.h#L116
4) Manual written bindings which require special treatment: https://github.com/1vanK/Urho3D/blob/master/Source/Urho3D/AngelScript/Manual.cpp

Problems:
1) Currently XMLs and bindings is regenerated every time when engine is compiled. Not sure I can fix it. How force doxygen regenerate xmls only when engine source code is changed? I do not consider this a big problem, since in most cases users does not need to enable URHO3D_GENERATEBINDINGS at all. Generated.cpp already pregenerated and can be automatically generated by server side.
2) Currently server does not send commit with changed Generated.cpp to repo.
3) I do not touched existed URHO3D_BINDINGS option

EDIT:
Not sure `static void GetXmlFiles(string dirPath, vector<string>& result)` works under linux. I tested only Win version


weitjong:
Cool! I am currently at work. I will try to review at the soonest.

SirNate0:
I have a question/suggestion for this: would it be possible to partially build Urho without the script bindings (perhaps with just empty function definitions for the script binding so the rest of the source doesn't change), then use that to build the binding-generator tool, which then generates the bindings and only then is the full version of the Urho library (with script bindings) compiled? It seems to me that a lot of the code is basically reinventing what Urho already has support for using the standard library instead, which I assume is because Urho won't have been built to be able to use its File or XML or String classes at that point, and if that would work the duplicate features might be avoided.

1vanK:
I would also like to use Urho3D::String instead of std::strings, but it requires a considerable amount of CMake magic and I couldn’t do it.

weitjong:
I am still rushing work for my day job and haven’t got time to review your code yet, but based on the comments above I believe you can use reuse the “mini Urho” build approach. See https://github.com/urho3d/Urho3D/blob/master/Source/Tools/PackageTool/CMakeLists.txt for the inspiration. I will review the code by this week.

eugeneko:
Just a random thought: I suggest to extract ignore tagging into separate file instead of littering engine headers. Some Urho headers already huge and have poor readability.
It's not an issue now, but with function bindings it may become much more severe.

1vanK:
> It's not an issue now, but with function bindings it may become much more severe.

No, also only single line before some functions

eugeneko:
> No, also only single line before some functions

This is literally 50% increase in size for some places in the code.
Also, it somewhat violates incapsulation -- when bindings are scattered all over the engine headers.
Currently, bindings stay in their folders. Would be nice if they continue doing so.

On the other hand, it may be simpler to maintain bindings when they are directly in the header...
As long as this line is simple, it may be ok to keep it. I dunno.

I mostly curious how you will handle functions.

1vanK:
>This is literally 50% increase in size for some places in the code.

I have dirty code on my HDD with classes bindings. An additional comment line does not impair code readability. We already have one or more lines with comments for each function.

> I mostly curious how you will handle functions.

Same way - mark BIND_IGNORE and write manually https://github.com/1vanK/Urho3D/blob/master/Source/Urho3D/AngelScript/Manual.cpp

SirNate0:
Would it be possible to switch to a preprocessor define instead of a separate comment line? Something like `URHO3D_NOSCRIPT` or similar, possibly in combination with Doxygen's PREDEFINED flag when using ENABLE_PREPROCESSING. Then it looks a lot more similar to the existing `URHO3D_API` we have on global functions anyways. I feel that would be a bit more readable than the extra comment on potentially a large number of functions.

1vanK:
Do you mean
```
    /// Returns the active algorithm
    /// @xmlonly BIND_IGNORE @endxmlonly
    Algorithm GetAlgorithm() const;
```
to
```
    /// Returns the active algorithm
    BIND_GNORE Algorithm GetAlgorithm() const;
```

SirNate0:
Yes, exactly. I think something like
```
    /// Returns the active algorithm
    BIND_IGNORE Algorithm GetAlgorithm() const;
```
is a bit more readable. Others may disagree, though.

1vanK:
What if I need several marks?
```
    /// Returns the active algorithm
    AS_IGNORE LUA_IGNORE JS_IGNORE NOT_PROPERTY Algorithm GetAlgorithm() const;
```
```
    /// Returns the active algorithm
    /// @xmlonly AS_IGNORE LUA_IGNORE JS_IGNORE NOT_PROPERTY @endxmlonly
    Algorithm GetAlgorithm() const;
```

1vanK:
Ops I accidentally closed PR xD

SirNate0:
It should work fine even if they're on the line before, I'm pretty sure. Like
```
    /// Returns the active algorithm
    AS_IGNORE LUA_IGNORE JS_IGNORE NOT_PROPERTY 
    Algorithm GetAlgorithm() const;
```
My personal opinion is that language-specific ones should be done in a separate file associated with the binding generator, and the in-header description should reflect what should generally be exposed to any script subsystem. For example, you propose a possible JS binding addition in your example code (whether yours or someone else's in the future), but why should that have to be written in the Urho code. Ideally, large portions of the binding generator should be reusable for other languages, but if we require adding special defines to the source for each generator it becomes much harder to add any other scripting language.

I'm also of the opinion that turning functions into properties should be not-default behavior and you should have to specify that it should be turned into a property, or that the binding code should figure it out automatically (GetVar and SetVar both exist and have matching parameter/return types, therefore we can turn it into a property).

So, as I see it, there should only be two or three markers in the source, and only ever one at a time (as marking it as ignored means we don't care if it's a property, and marking it as a property should imply it's not ignored (or it is implicitly ignored and handled separately).

1vanK:
My vision is that a user should do a minimum of actions when writing new code.

90% of functions whose names begin with Get and Set are converted to properties when bindings.

Putting marks in a separate file also makes it difficult to perception. I see the function and immediately understand how it is binds. I do not need to look for its metadata in another file.

Also single instrument can generate bindings for all scripting languages. I see no reason to care about any other tools.

SirNate0:
Turning things into properties by default is okay with me, but what about languages that don't have properties (Lua, maybe, I don't use it)? Then the binding information in the source is misleading. As I said, though, if you want that to be the default that's fine with me.

To some extent I disagree about the perceptability issue - when looking at the Graphics class I don't really care how it binds, when I care about using the scripting language I would like some sort of documentation to look at like we have for AngelScript already. Otherwise we should just remove the @xmlonly directives and just put the binding information in the normal documentation as well (something like a "Bindings: NO_AS, NO_LUA, NO_JS" line instead of the xml specific one). Then we can tell by looking at the documentation and not just the source. I will grant that in source stuff does help avoid issues with having to update bindings though, as you'd have to go and check each binding code for the given function signature. That said, it's probably pretty simply to ensure that the requested function signature actually exists in the class, so hopefully the build would fail on any special bindings and it would be apparent when compiling anyways.

> Also single instrument can generate bindings for all scripting languages. I see no reason to care about any other tools.

I think JS is actually a good example here. If we only care about this one tool, then either this one tool (and this Urho) must adopt JavaScript, or someone has to fork Urho to support it. I think the #define BIND _IGNORE is a better option for supporting community script bindings (as if nothing else a simple text replacement when reading the files into another tool would hopefully suffice and be a bit less complicated than finding the Doxygen comment ones, but it's pretty comparable either way, just slightly more work to do the text replacement the other way). If you prefer the Doxygen comments that's fine with me. If we do go that route, I think we need to establish what the guidelines for changing those will be, though. Will we accept a PR whose only purpose is to add a bunch of NO_JS tags to the documentation so that someone's JavaScript bindings will work (i.e. without Urho adopting JS itself)? One other thing - if the Doxygen comments will now largely be multi line, will we start breaking some of the really long single line comments into multiple lines, and perhaps have more complete descriptions of some of the more complicated functions in the engine?

1vanK:
In any case BIND_IGNORE will be used if the function is not specific to a particular language. I also see no reason to take care of forks that don’t return the code to the engine. Forks can write their own generator, store metadata in an external file, and do whatever.

eugeneko:
> I also see no reason to take care of forks that don’t return the code to the engine

It's quite fun policy. First, for 10 years, the engine encourages people to fork and tweak it, so the forking is the primary use case for many users. And then "nah, no need to take care of all these forks".

Also, I wonder where you put forks that return code to the engine but the engine doesn't take this code.

---

Back to the topic, I think that comment annotations are still better than code annotations because it can be done together with splitting these awful 300-character-long comments into readable chunks.

1vanK:
> It's quite fun policy. First, for 10 years, the engine encourages people to fork and tweak it, so the forking is the primary use case for many users. And then "nah, no need to take care of all these forks".

I think it was never some kind of taboo. And we have repeatedly violated backward compatibility. P.S. Most recently, you offered to change containers for compatibility with one single fork. Doesn't that violate backward compatibility?

eugeneko:
> Most recently, you offered to change containers for compatibility with one single fork

I offered to change containers because I'm sick of inferior hand-written containers outdated by decade. And I'm not alone. Containers are the _sole_ reason why I don't use Urho as is anymore.

I'm totally fine if someone will go and improve Urho containers so they become as usable as standard ones. Sadly, it appears that I'm the latest person who tried to do it.

---

Damn, I'm rambling again. Sorry, containers are just my weak spot.

SirNate0:
> In any case BIND_IGNORE will be used if functions is not specific to a particular language.

Sounds good.

>  I also see no reason to take care of forks that don’t return the code to the engine. Forks can write their own generator, store metadata in an external file, and do whatever.

Also sounds maybe tolerable. I'm more inclined to agree with Eugene's comment now that I see it, though. But, let me pose a couple of scenarios that make me concerned that this is not the best choice:
1. What about the case of, for example, Python. At least a few users are interested in it, me included. That said, I don't think Urho should support it directly - why should Urho have an entire copy of CPython in the ThirdParty folder so that a handful of people can use Python, and then do we go with Python 2.x or Python 3.x? That said, why should those users not have their bindings supported in the Urho codebase just because we shouldn't have to have a copy of CPython to build Urho.
2. What about the case of users offering to contribute but their contributions being rejected? Let's say C#, as a hypothetical example, just because I think there are a few people who would strongly oppose C# being included. Should C# specific binding info be allowed in the general Urho codebase even though C# was rejected by the engine. Or on the other side, why should changes to the binding info that doesn't change anything else be prevented just because people don't like C#? The user was perfectly willing to submit a PR and give back to the engine, but the maintainers rejected the PR because of whatever reason, but should their job of creating bindings be made more difficult or require them to keep a fork, even though they did "return code to the engine"?

That said, I don't see this so much as a backwards compatibility issue so much as seeking to be forward compatible with what people may wish to do in the future.

1vanK:
Just add PythonGenerator.cpp to BindingGenerator in your fork.

SirNate0:
> If you need ignore specific function that not ignored in other script languages you can do it in this PythonGenerator.cpp (certainly there will be very few such exceptions)

If this is the case why don't we just do that for AS and Lua as well (for the specific ones - I'm still on board with a generic BIND_IGNORE)? To a fair extent how I feel about this is dependent on how many times this happens in the engine - if it's just a handful of functions, it matters little to me. If it's every function that uses a Vector that's a bit different though. Do you have any idea how prevalent the AS/Lua-specific restrictions will be?

> Back to the topic, I think that comment annotations are still better than code annotations because it can be done together with splitting these awful 300-character-long comments into readable chunks.

This also makes me favor the comment annotations. I still slightly favor including them in the general documentation and not just the xml generated ones, though.

1vanK:
> If this is the case why don't we just do that for AS and Lua as well (for the specific ones - I'm still on board with a generic BIND_IGNORE)? To a fair extent how I feel about this is dependent on how many times this happens in the engine - if it's just a handful of functions, it matters little to me. If it's every function that uses a Vector that's a bit different though. Do you have any idea how prevalent the AS/Lua-specific restrictions will be?

I never said that there AS_IGNORE LUA_IGNORE will exist. I have not touched LUA bindings yet and use only BIND_IGNORE currently. I just said that it’s possible to add any marks to comments and get them in your generator with single function call.

eugeneko:
By the way, are we restricted to all caps define-like ignore tag?
Maybe it’s possible to use some user-friendly readable tag and include it
into main docs?

чт, 9 апр. 2020 г. в 23:10, 1vanK <notifications@github.com>:

1vanK:
> By the way, are we restricted to all caps define-like ignore tag?
Maybe it’s possible to use some user-friendly readable tag and include it
into main docs?

It just string. Can be in any letter case

SirNate0:
@1vanK fair enough.

I'm also in support of a more user-friendly format. Perhaps a custom Doxygen alias that could be configured to be machine-friendly for the generator but reader-friendly for the documentation. Something like (using http://www.doxygen.nl/manual/custcmd.html)
* for the documentation
`ALIASES += nobindings="\par There are no bindings for this function:\n" `
* for the generator
` ALIASES += nobindings="NO_BINDINGS" `

1vanK:
I think we need to leave stylistic questions to an end when the main issues are resolved. In any case, my code requires a serious spell checking, because my English is very bad xD

weitjong:
I see a few issues when trying to build it on Linux with the new option enabled (thanks for leaving the `URHO3D_BINDINGS` alone). First of all, the BindingGenerator tool did not build successfully. I have added a dev branch with CI so you can see the compilation error yourself. If you are OK, I may make small build script changes on top this dev branch later on. I haven't really completed reviewing your PR, in fact I just barely started due to the above error.

weitjong:
More fix is needed. Let me know if you wish to have direct write access to the dev branch. I can grant you the push privilege.

1vanK:
yes, pls

weitjong:
> Manual written bindings which require special treatment: https://github.com/1vanK/Urho3D/blob/master/Source/Urho3D/AngelScript/Manual.cpp

I still don't understand why. Could you elaborate a little bit here.

> Problems:
1) Currently XMLs and bindings is regenerated every time when engine is compiled. Not sure I can fix it. How force doxygen regenerate xmls only when engine source code is changed? I do not consider this a big problem, since in most cases users does not need to enable URHO3D_GENERATEBINDINGS at all. Generated.cpp already pregenerated and can be automatically generated by server side.
2) Currently server does not send commit with changed Generated.cpp to repo.
3) I do not touched existed URHO3D_BINDINGS option

The first two are fixable. 3) is not a problem (at least for now).

1vanK:
Generator automatic bindings all enums frum Urho3D namespace, but
```
engine->RegisterEnum("IKAlgorithm");
    engine->RegisterEnumValue("IKAlgorithm", "ONE_BONE", IKSolver::ONE_BONE);
    engine->RegisterEnumValue("IKAlgorithm", "TWO_BONE", IKSolver::TWO_BONE);
    engine->RegisterEnumValue("IKAlgorithm", "FABRIK", IKSolver::FABRIK);

    engine->RegisterEnum("IKFeature");
    engine->RegisterEnumValue("IKFeature", "JOINT_ROTATIONS", IKSolver::JOINT_ROTATIONS);
    engine->RegisterEnumValue("IKFeature", "TARGET_ROTATIONS", IKSolver::TARGET_ROTATIONS);
    engine->RegisterEnumValue("IKFeature", "UPDATE_ORIGINAL_POSE", IKSolver::UPDATE_ORIGINAL_POSE);
    engine->RegisterEnumValue("IKFeature", "UPDATE_ACTIVE_POSE", IKSolver::UPDATE_ACTIVE_POSE);
    engine->RegisterEnumValue("IKFeature", "USE_ORIGINAL_POSE", IKSolver::USE_ORIGINAL_POSE);
    engine->RegisterEnumValue("IKFeature", "CONSTRAINTS", IKSolver::CONSTRAINTS);
    engine->RegisterEnumValue("IKFeature", "AUTO_SOLVE", IKSolver::AUTO_SOLVE);
```
is sigle enum not from Urho3D namespace, but used as usual enum from scripts

`enum TextureUnit` is different for desctop and not desctop

```
enum TextureUnit
{
    TU_DIFFUSE = 0,
    TU_ALBEDOBUFFER = 0,
    TU_NORMAL = 1,
    TU_NORMALBUFFER = 1,
    TU_SPECULAR = 2,
    TU_EMISSIVE = 3,
    TU_ENVIRONMENT = 4,
#ifdef DESKTOP_GRAPHICS
    TU_VOLUMEMAP = 5,
    TU_CUSTOM1 = 6,
    TU_CUSTOM2 = 7,
    TU_LIGHTRAMP = 8,
    TU_LIGHTSHAPE = 9,
    TU_SHADOWMAP = 10,
    TU_FACESELECT = 11,
    TU_INDIRECTION = 12,
    TU_DEPTHBUFFER = 13,
    TU_LIGHTBUFFER = 14,
    TU_ZONE = 15,
    MAX_MATERIAL_TEXTURE_UNITS = 8,
    MAX_TEXTURE_UNITS = 16
#else
    TU_LIGHTRAMP = 5,
    TU_LIGHTSHAPE = 6,
    TU_SHADOWMAP = 7,
    MAX_MATERIAL_TEXTURE_UNITS = 5,
    MAX_TEXTURE_UNITS = 8
#endif
};
```

weitjong:
> yes, pls

You should receive an invitation to become core developer by now.

1vanK:
Yes, invitation received

1vanK:
I successfully compiled branch https://github.com/urho3d/Urho3D/tree/generate-bindings with MinGW64 on my PC. I do not known why Editor is not runned in tests: https://travis-ci.org/github/urho3d/Urho3D/jobs/673542418

weitjong:
Thanks for the effort. Those test run failed means there is/are errors in the AS script binding. It seems it only manifested itself when OpenGL is used. Probably the `Generated.cpp` was generated by you on Windows platform and inadvertently baked in Graphics API specific binding into that file. I am just guessing though.

1vanK:
I use only OpenGL

weitjong:
Whatever the root cause, the run test failed indicates there is AS script binding error.

BTW, one of the reason why I stopped my own experiment with Clang-tool was also it would generate Graphics API specific output, among others.

weitjong:
I am fixing the build script as I go while reviewing your PR. I have to take a number of breaks to rest my old eyes. Hopefully I could solve the problems you mentioned in the PR description above. So, you can just concentrate on the actual work in this dev branch.

Source/Urho3D/CMakeLists.txt
```diff
@@ -146,9 +146,37 @@ endif ()
...
+                               #${CMAKE_SOURCE_DIR}/Source/Urho3D/Lua/Generated.cpp
+                               #${CMAKE_SOURCE_DIR}/Source/Urho3D/JavaScript/Generated.cpp
+                               #${CMAKE_SOURCE_DIR}/Source/Urho3D/CSharp/Generated.cpp
+                               fake_output_to_force_run_every_time
+                        COMMAND ${CMAKE_COMMAND} -E echo "Generating XMLs with AST to ${XML_AST_DIR}"
+                        COMMAND ${DOXYGEN_EXECUTABLE} "${CMAKE_BINARY_DIR}/Source/Tools/BindingGenerator/generated/Doxyfile"
+                        COMMAND BindingGenerator "${XML_AST_DIR}" "${CMAKE_SOURCE_DIR}"
+                        DEPENDS BindingGenerator ${CMAKE_BINARY_DIR}/Source/Tools/BindingGenerator/generated/Doxyfile
+                        COMMENT "Generating bindings to script languages")
+
```
weitjong, line 173: I see the BindingGenerator to be a host-tool. So it should be built using native/host compiler toolchain regardless.

1vanK:
Thanks @SirNate0 for the advice. I replaced BIND_IGNORE by two aliases:

1) @nobind means that for enum there is no binding at all
https://github.com/urho3d/Urho3D/blob/generate-bindings/Source/Urho3D/Graphics/GraphicsDefs.h#L54
2) @manualbind means that bindings for enum are created manually
https://github.com/urho3d/Urho3D/blob/generate-bindings/Source/Urho3D/Graphics/GraphicsDefs.h#L327

Currently difference between them is cosmetic: both convered to BIND_IGNORE for BindGenerator and to empty string for DOC.

1vanK:
I see something strange in the log https://travis-ci.org/github/urho3d/Urho3D/jobs/673542421

```
Generating AS bindings
Scanning dependencies of target Urho3D
[ 53%] Building CXX object Source/Urho3D/CMakeFiles/Urho3D.dir/AngelScript/Addons.cpp.o
[ 53%] Building CXX object Source/Urho3D/CMakeFiles/Urho3D.dir/AngelScript/AudioAPI.cpp.o
[ 53%] Building CXX object Source/Urho3D/CMakeFiles/Urho3D.dir/AngelScript/CoreAPI.cpp.o
[ 53%] Building CXX object Source/Urho3D/CMakeFiles/Urho3D.dir/AngelScript/DatabaseAPI.cpp.o
[ 53%] Building CXX object Source/Urho3D/CMakeFiles/Urho3D.dir/AngelScript/GraphicsAPI.cpp.o
[ 53%] Building CXX object Source/Urho3D/CMakeFiles/Urho3D.dir/AngelScript/EngineAPI.cpp.o
[ 53%] Generating bindings to script languages
Generating XMLs with AST to /home/travis/build/urho3d/Urho3D/build/ci/Source/Tools/BindingGenerator/generated/xml
```

output should be another order
```
[ 53%] Generating bindings to script languages
Generating XMLs with AST to /home/travis/build/urho3d/Urho3D/build/ci/Source/Tools/BindingGenerator/generated/xml
Generating AS bindings
```

This build has correct order https://travis-ci.org/github/urho3d/Urho3D/jobs/673542422
```
[ 52%] Generating bindings to script languages
Generating XMLs with AST to /home/travis/build/urho3d/Urho3D/build/ci/Source/Tools/BindingGenerator/generated/xml
Generating AS bindings
```

weitjong:
I see the problem with the build ordering too in my review, but I didn’t think it is the root cause of the CI build error because I thought the Genrated.cpp is already checked into the source tree in the first place. Then last night I notice you actually instruct the generator to write directly into the source tree, which is bad but could probably explain why CI test run failed. Give me time to fix the problem. I think I have reviewed and understood enough of your original intention in the proposed PR.

1vanK:
My first variant was to write the generated file to CMAKE_BIN_DIR, but I changed it to CMAKE_SRC_DIR. When Generated.cpp in external dir, using `#include "../blabala"` inside Generated.cpp is a mistake. I should use `#include <../blabala>` or I recive bunsch of errors with conflicted headers.

weitjong:
I know exactly what you are saying. Our CI has a special mode where it could update the source tree as well. So far only in this special mode the build script modifies the source tree and even got the changes committed to the remote repository as part the CI. I have tried a few failed attempts to make your PR works and fit into the existing workflow. Sorry for being slow these days.

weitjong:
Can I expect that your proposed BindingGenerator can generate bindings based on any XML input files, as long as it conforms to Doxygen's XML output?

1vanK:
What difference will other files have?

weitjong:
There is currently no other tool that could generate the same XML input files. I am just thinking ahead a little bit on whether I can reuse the BindingGenerator tool with input produced by other Clang-based tools.

1vanK:
I experimented with Clang, but chose Doxygen.

1) Doxygen already used with Urho3D, so no new dependencies, sou user can just install doxygen and generate bindings on own PC
2) Prebuilded Clang for windows provide only limited and stable clang-c API
3) Building of Clang by himself is terrible - its required hunderts GB of HDD
4) Doxygen already provides convenient functionality that does not need to be implemented independently

weitjong:
I think you misunderstood me. I am not going to change what you have chosen to use in your proposed PR.

1vanK:
I think it makes no difference how the files were generated if they are no different

weitjong:
I have pulled your latest changes. The `BindGenerator` tool in this latest version and the original version has issue on Linux. It always generates "empty" `Generated.cpp` file as output, basically it is the template skeleton output file without any bindings in it. At first I thought may be it was related invocation order problem. But, after fixing the build script so that now the tool is invoked with deterministic order and invoke exactly only once when source tree is dirty, it still produces "empty" output file. Do you have any Linux VM to try the dev branch?

1vanK:
I added some debug output

https://travis-ci.org/github/urho3d/Urho3D/jobs/674114960

BindGen is called twice
```
BindGen: dir /home/travis/build/urho3d/Urho3D/build/ci/Source/Tools/BindingGenerator/generated/xml/ is opened
...
BindGen: dir /home/travis/build/urho3d/Urho3D/build/ci/Source/Tools/BindingGenerator/generated/xml/ is opened
```

GetXmlFiles() sucefully found xmls but stat() is failed
```
    while (dirent* de = readdir(dir))
    {
        string filePath = dirPath + de->d_name;

        cout << "BindGen: filePath = " << filePath << "\n";

        if (!stat(filePath.c_str(), &st))
            continue;

        cout << "BindGen: stat success\n";
```
`filePath =` is printed but `BindGen: stat success` not printed

weitjong:
I already call it a day. I will check in my WIP fix tomorrow. In my local branch the tool only invoked once and yet the output is still wrong. It should be something minor that causing it to fail since it works on Windows. I am fixing the build script using `vi’ editor, so I didn’t check deeper why it generated empty file yet.

1vanK:
Ok, I found my mistake. stat return 0 if success

weitjong:
The two problems in the build script you mentioned above should be fixed now. The source tree update is controlled by another existing build option called "URHO3D_UPDATE_SOURCE_TREE", which is set to disable right now in CI for dev branch to prevent any accident commit during development.

After merging with your commits in the dev branch, I can see a lot of debug output in my terminal while building locally. Unfortunately, the generated source file is still "empty". Tomorrow I will wear my C++ hat to help debug the BindingGenerator tool, if the issue is not fixed by then.

1vanK:
```
c:\Programs\Doxygen>doxygen.exe -v
1.8.17 (b5fa3cd1c6e6240e20d3b80a70e3f04040b32021*)`
```

1vanK:
I have bunch of errors with conflicted `#include` when try to build "generete-bindings" on my PC.
I see the same problem that forced me move Generated.cpp from BIN_DIR to SRC_DIR.

weitjong:
Just to be sure that you have tried the build with the latest commit that fix the exclusion pattern mistake that I made previously. CI from both AppVeyor and Travis seems to build fine, with the exception of Travis's CI run test still segfault on those that uses those new AngelScript binding.

Having the generated source files to be in the build tree instead of source tree is the approach I use to make the build order and dependency becomes deterministic. Otherwise, you have a runaway feedback loop and the binding tool will keep being invoked again and again on each build.

I also don't quite understand why you need those include in the first place. I thought the new host-tool only depends on pugixml?

1vanK:
> Just to be sure that you have tried the build with the latest commit that fix the exclusion pattern mistake 

yes

```
git log >../Log.txt

commit f781014ab642c1cfc00e8652f97be5954c984643
Author: Yao Wei Tjong 姚伟忠 <weitjong@gmail.com>
Date:   Wed Apr 15 01:05:53 2020 +0800

    Fix the exclusio pattern.
    Either use the generated file from the source tree (when binding is off) or from the build tree (when binding is on).
    [ci only: generate-bindings]
```

I download log: https://ci.appveyor.com/project/urho3d/urho3d/builds/32175551/job/9lyka5l61lhk5v3w

AngelScript-Generated.cpp not compiled at all, Generated.cpp is builded instead

On my local Visual Studio 2019 I have no Generated.cpp in build three, but have AngelScript-Generated.cpp and errors. I do not known why difference

weitjong:
Ok. Will enable the option for AppVeyor CI later to see whether it gets the same build error.

1vanK:
It seems different versions of doxygen generate xml files with different names

1vanK:
I reworked xml files reader, now it does not use file names at all, internal structure only. Travis works, appveyor can not find doxygen.

1vanK:
Any ideas why mingw do not found builded BindingGenerator? I have no problem with Mingw on my PC.

1vanK:
```
Running /home/travis/build/urho3d/Urho3D/build/ci/bin/tool/BindingGenerator.exe /home/travis/build/urho3d/Urho3D/build/ci/Source/Tools/BindingGenerator/generated/xml /home/travis/build/urho3d/Urho3D
```
ah yes, cross compiling

weitjong:
> ah yes, cross compiling

I have added the TODO line here.

https://github.com/urho3d/Urho3D/blob/generate-bindings/Source/Tools/CMakeLists.txt#L64-L66

> appveyor can not find doxygen.

Will try to install it first shortly.

weitjong:
I have also fixed the BindingGenerator binding for MinGW build in the dev branch. Since you already made further changes in your PR branch to fix the reader to cater for different version of Doxygen, I will let you perform the merge first.

1vanK:
I use this PR branch with old cmake config because I can not build generate-bindings branch on my PC with VS.

I see same conflicted includes https://ci.appveyor.com/project/urho3d/urho3d/builds/32226223/job/m0fs9b4j2ywrpypo
This PR compiled with AppVeyor successfully

Current version of generate-bindings branch should compile successfully with AppVeyor but generate empty file if doxygen have another version

weitjong:
Yes, now that I can see the error on AppVeyor, I can understand what you were talking about the other day. It should be fixable too I think, by adjusting the header search path accordingly. But like I said, I wait the changes to be merged in the dev branch before making further adjustment. I can do that too when I have some free time on the weekend later.

1vanK:
I just updated xml loader in generate-bindings to make travis happy

weitjong:
Sorry for the slow response. The last commit you made has fixed the generation on my local build and the generated file on the build tree is identical with the one on the source tree, which is good. However, on Linux/GCC I am not able to reproduce the error encountered on Windows/MSVC. The build error is not caused by an issue that I originally thought in my previous comment. I need a Windows build environment locally to figure out what causing this error. And, for that I am forced to salvage my old Win7 VM image from my old rig and finally I just did it. I will continue my investigation again tomorrow. Hopefully this does not stop you to keep pushing to your original PR branch.

1vanK:
I try to make the generated bindings match the existied 1:1, so Vector3::Length() binds as property, but IntVector3::Length() as function
```
    // float Length() const
    engine->RegisterObjectMethod("Vector3", "float get_length() const", asMETHODPR(Vector3, Length, () const, float), asCALL_THISCALL);
    ...
    // float Length() const
    engine->RegisterObjectMethod("IntVector3", "float Length() const", asMETHODPR(IntVector2, Length, () const, float), asCALL_THISCALL);
```
We can standardize this later

weitjong:
I believe originally Lasse would like to bind all the public getter and setter in scripting binding API, and additionally bind it as property too for convenience. I believe occasionally some bindings are missing due to mistake by different contributors or simply laziness. IMHO, in this particular case you don’t have to keep them verbatim in your new binding logic.

weitjong:
@1vanK The dev branch is yours now. The issue with VS should be fixed now.

1vanK:
Thanks! I have not tested it yet, but I assume that `#include "Core/Attribute.h"` is interpreted by compiler as `#include <Core/Attribute.h>` and base path is retrieved from compiler's variable INCLUDE_PATH. I can try to change generator to generate `#include <Core/Attribute.h>` instead `#include "../Core/Attribute.h>".

I also have a question. Where will be placed pregenerated `Generated.cpp`?

weitjong:
I have tested that `#include "Core/Attribute.h"` works in both source tree as `AngelScript/Generated.cpp` and build tree as `generated/Angelscript-Generated.cpp`. The current Lua script binding generator also output similar include directive without "../", most probably also due to the same reason to keep VS happy (I cannot really remember now). If you ask me, I would prefer to keep thing simple by not altering the include directive between the two locations. Thus, all we have to do is to just make a copy from the build tree to the source tree in one of the CI job. I have actually already scripted that.

https://github.com/urho3d/Urho3D/blob/26b8392d21ca06f9324bd7cb3d9fea9677bbb252/Source/Urho3D/CMakeLists.txt#L220-L222

https://github.com/urho3d/Urho3D/blob/26b8392d21ca06f9324bd7cb3d9fea9677bbb252/.travis.yml#L37

But as you can see in this dev branch I have intentionally disabled the `URHO3D_UPDATE_SOURCE_TREE` option for now. In the master branch later, the newly generated binding file will get pushed to GitHub repo automatically and a new fresh CI build jobs will be spawned.

1vanK:
If Generated.cpp with `#include "Core/Attribute.h"` will be committed to `SRC_DIR/Urho3D/AngelScript` it will be also interpreted as `#include <Core/Attribute.h>` because path `SRC_DIR/Urho3D/AngelScript/Core/Attribute.h` not exists

weitjong:
Ah OK. If you meant you will keep both places to use the angel bracket format then I am fine with it. Although, so far we usually use quotes format when referencing our own headers when building Urho3D library; and only use angle bracket when we are using the library. See below.

https://github.com/urho3d/Urho3D/blob/26b8392d21ca06f9324bd7cb3d9fea9677bbb252/Source/Urho3D/Container/Allocator.h#L25-L29

I agree with you that in this case the header in quotes could be found because the preprocessor has fallen back to use the header search path, i.e. interpreted as angle bracket directive. Therefore, I have no strong objection if we make an exception here.

1vanK:
Should we automatically delete pregenerated Urho3D/AngelScript/Generated.cpp if user build engine with `-DURHO3D_GENERATEBINDINGS=1` to force users manually copy BIN_DIR/.../AngelScript-Generated.cpp to SRC_DIR/Urho3D/AngelScript/Generated.cpp

This situation is currently possible:
1) User download repo
2) User modify engine and build engine with URHO3D_GENERATEBINDINGS=1
3) User disable URHO3D_GENERATEBINDINGS and forgot copy new file to SRC_DIR/Urho3D/AngelScript/Generated.cpp
4) User use new version of engine with old bindings

weitjong:
We could make the binding build option to auto-enable the source tree update build option later. So in your scenario, the copying happens automatically. However, I actually prefer user to turn on both options explicitly. I want to keep the source tree pristine. Only when user tells us to make the Source tree update then we dirty it. So, if we keep this thinking and going back to your scenario then user would have copy the generated files manually. There is no danger of losing work whatsoever. The file can always be regenerated too.

1vanK:
Autodetecting of Get/Set properties works perfect, but this is the internal behavior of the generator, which the programmer must always remember. Otherwise, he will not receive what he expects. So I just remove this feature: now programmer must manually mark function as property if required

1vanK:
Close this, because conflicts with commits in main repo

PR #2536, opened 2019-10-28
Add Open GL ES 3 support

orefkov:
Add support for GLES3 rendering API and it features:
- Instancing
- Uniform buffer object
- Multiple render targets
- Texture2DArray
- Texture3D
- Support for textures: ETC2, sRGB, floating point, depth, 1 & 2 channel (R & R/G).
Worked render pathes - Deferred, DeferredHWDepth, Prepass, PrepassHWDepth.

PR #2623, opened 2020-04-10
Add OpenGL ES 3 support (FIXED webgl build error)

kakashidinho:
This is the fix for #2536's webgl errors. Seems the original author doesn't want to continue that PR, so I have to create another PR. This PR is based on his PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -240,6 +240,8 @@ env:
   matrix:
     - ORG_GRADLE_PROJECT_URHO3D_LIB_TYPE=STATIC
     - ORG_GRADLE_PROJECT_URHO3D_LIB_TYPE=SHARED
+    - ORG_GRADLE_PROJECT_URHO3D_LIB_TYPE=STATIC ORG_GRADLE_PROJECT_URHO3D_GLES3=1
+    - ORG_GRADLE_PROJECT_URHO3D_LIB_TYPE=SHARED ORG_GRADLE_PROJECT_URHO3D_GLES3=1
 stage: universal build configuration
 before_script:
   - rake ci_timer
@@ -306,6 +308,9 @@ env:
     # ODROID-X2
     - URHO3D_64BIT=0 URHO3D_LIB_TYPE=STATIC ARM_ABI_FLAGS='-mcpu=cortex-a9 -mfpu=neon-fp16'
     - URHO3D_64BIT=0 URHO3D_LIB_TYPE=SHARED ARM_ABI_FLAGS='-mcpu=cortex-a9 -mfpu=neon-fp16'
+    # GLES3
+    - URHO3D_64BIT=0 URHO3D_LIB_TYPE=STATIC URHO3D_GLES3=1 ARM_ABI_FLAGS='-mcpu=cortex-a9 -mfpu=neon-fp16'
+    - URHO3D_64BIT=0 URHO3D_LIB_TYPE=SHARED URHO3D_GLES3=1 ARM_ABI_FLAGS='-mcpu=cortex-a9 -mfpu=neon-fp16'
 stage: release build configuration
 before_script:
   - rake ci_timer
--- a/CMake/Modules/UrhoCommon.cmake
+++ b/CMake/Modules/UrhoCommon.cmake
@@ -152,8 +152,15 @@ option (URHO3D_GLES2 "Enable GLES2" TRUE)
 option (URHO3D_GLES3 "Enable GLES3" FALSE)
 option (URHO3D_WEBP "Enable WebP support" TRUE)
 
-if ((ANDROID OR IOS) AND NOT URHO3D_GLES3)
-    set(URHO3D_GLES2 TRUE)
+# URHO3D_GLES3 and URHO3D_GLES2 are mutually exclusive
+if (ANDROID OR IOS)
+    if (URHO3D_GLES3)
+        set(URHO3D_GLES2 FALSE)
+    elseif (URHO3D_GLES2)
+        set(URHO3D_GLES3 FALSE)
+    else ()
+        set(URHO3D_GLES2 TRUE)
+    endif ()
 endif ()
 
 if (ARM AND NOT ANDROID AND NOT RPI AND NOT APPLE)
--- a/CMake/Modules/UrhoCommon.cmake
+++ b/CMake/Modules/UrhoCommon.cmake
@@ -910,7 +910,7 @@ macro (define_dependency_libs TARGET)
 
         # Graphics
         if (URHO3D_OPENGL)
-            if (NOT ANDROID AND NOT WEB)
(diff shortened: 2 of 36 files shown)


eugeneko:
So far I have only one issue with this PR.
We have _a lot_ of different obscure preprocess guards for graphics stuff.
Considering current messy state of Urho renderer, I'm afraid of it becoming even messier.

What exactly behavior is controlled by `DESKTOP_GRAPHICS`?
What behavior is controlled by `MOBILE_GRAPHICS`?
What behavior is controlled by `DESKTOP_GRAPHICS_OR_GLES3`?
What behavior is controlled by `URHO3D_GLES2` guard?
Do we have other guards and what is controlled by them?

Consider using a set of _specialized_ defines like `URHO3D_DIR_LIGHT_CASCADES` that is configured in specialized header basing on the platform defines.

Something like boost platform headers that define macros for available C++ features basing on compiler.

kakashidinho:
The the macro definitions are the serious issues of Urho3D's OpenGL renderer. I think the better way is omitting all the macros and just separate the code for Desktop GL and GLES like what irrlicht does (even if there would be duplicates, it would still be easier to maintain than macros).
However, the macros have been around before this PR.  #2536 introduced even more macros. IMO, there should be another PR that do the cleanup in the OGL renderer code. It is out of scope of this PR.

eugeneko:
It may be out of scope, but macro cleanup is needed _before_ we add even more macro mess and make such cleanup even more compilcated.

Anyway, I don't think it will go into current version, so we have time.

kakashidinho:
OK, I will try to cleanup the macros when I have time. In this PR, I'm just trying to help fix the issues introduced by the previous PR #2536 since looks like everyone is OK with that PR except the webgl issues. This PR actually doesn't add any new macros.
It's better that @orefkov continues this PR with @eugeneko's suggestion.

eugeneko:
I can revise graphics macros after release, if you don't do it already at this point.

weitjong:
Thanks for your PR. I suggest you to rebase it on top our dev branch (https://github.com/urho3d/Urho3D/commits/orefkov-GLES3_Support) instead of from the original orefkov's PR. 

Some of problem that you tried to fix are already fixed in the dev branch, including the float precision. But it is still not enough. There is still regression issue caused by original PR, the last time I checked. And, I believe the changes or the introduction of the new preprocessors are the caused of it. And, for the record, I also don't like those preprocessors changes.

kakashidinho:
I think I already fixed the float precision. Because I tested the web build, it can run fine. (I saw the precision errors before this fix). It was not caused by the new macros, rather it was caused by incomplete ifdef handling code (introduced by previous PR).

What are things that already fixed in the recent changes you mentioned? My changes only start from b3d5922. All of previous changes are from @orefkov‘s PR.

weitjong:
From the commit history in the PR I can clearly see what you have added and it is based on the original PR. But I am telling you that we already have a dev branch based on the original work and that your PR is made not based on this dev branch. Unless your intention to nullify my previous work, please rebase your PR against that dev branch and make further changes on top it. I hope I made myself clear this time.

kakashidinho:
OK, sorry I thought you were talking about the master branch. I tested the dev branch, it can run fine. My changes of moving of `URHO3D_GLES2` definition to cmake is not really important. Your dev branch's precision `ifdef` guard is enough for my test. It is just that it is missing from the `#else` block (but it is non-issue atm due to default condition). There are two blocks: 
```
#if !defined(GL3) || !defined(USE_CBUFFERS)
// ...
// your precision guard is here
#else
// OpenGL 3 uniforms (using constant buffers)
// Your precision guard is not here
#endif
```
But the thing is `USE_CBUFFERS` is disabled by default, @orefkov enabled it for GLES3 previously but he reverted it back.
Which examples that you still see the errors, so that I can double test again. If there is no issue, I don't think any changes in this PR is needed. This is the screenshot. Note this screenshot is from **WebGL1**, the **WebGL2** still having errors, but it is not related to precision. Unless you want this PR to become WebGL2 fix also, I don't think it is related to this PR.

Imgur

I also tested GLES3 on iOS and it run OK.

Regarding the macros, I don't really like them either. But they are not my cause, as an example some of the messy macros like `CBUFFER`, `GL3` above have been there for a long time all over the places. I just want to see GLES3 in Urho3D so that I can continue some of my other experiments. And it seems no body has any incentive to move the GLES3 PR forward. If you telling me that no one has incentive because it is messy, then it's fine to me. I will try to cleanup some of them in future if I have time.

P/S, IMO GLES2 should be enabled by default, since some android devices don't support GLES3. Also, iPhone 5 and below doesn't support GLES3 either (if the devs want to support them).

kakashidinho:
As a proof that I didn't run a wrong build, here is the screenshot of what happened if there is no precision's ifdef guard:
imgur

eugeneko:
> But the thing is USE_CBUFFERS is disabled by default

And they should stay disabled. Urho doesn't use cbuffers, it abuses them. So cbuffers end up slower than uniforms. Major rewrite of renderer is required before they can be enabled.

kakashidinho:
Forgive me if I am too frank. But honestly I think this topic still have some miscommunications. It is unclear what everyone wants to do with this PR. No proper finalization & discussion.
In previous PR, @weitjong said if anyone could help fixing the web errors with gles3 changes then he could move forward with the merge. Then I create this PR then @eugeneko is not very happy with the macros.
Then I suggested the macros to be cleaned in future PR since the code is not that pretty to begin with. Then @weitjong said he already made similar changes but there still errors.
Then I tested again with only changes in his dev branch but see no error. I asked him which samples he still sees error, no response. At least asking some 3rd persons to verify if the results two people seeing are inconsistent.

And for logical analysis, no I don’t believe the errors are caused by new macros. Because the explicit precision statement will be included in every fragment shader compiled by Urho. And it is not hidden by any new macros (not after the fix in this PR or @weitjong dev branch)

Finally can we have a conclusion of what should be done with this PR? Leave it to dust until the renderer code is refactored, cleaned up? Or until the confirmation that the webgl errors are indeed fixed? Do you expect people to auto understand the inplicit decision by no communication?

This honestly turned off most people’s enthusiasm for contributions.
I actually no longer care what will happen to this PR. I just hope this comment might be helpful for your future consideration when dealing with PRs and avoiding misunderstandings/miscommunications etc.

weitjong:
The problem is no one is taking charge of the overall thing after Lasse has left. I hesitate to merge the original PR even after my fix in the dev branch because I am still not happy with it personally. I could not comment on yours because I have not really reviewed it.

I have said earlier that I want someone else to continue to work on the existing dev branch. You created yet another PR based on the original and add on your own fixes. Therefore I asked you to rebase so that you work in on top the dev branch instead of original PR.

However, what I hear from you is that the dev branch is ok and merge-ready, which contradict with what I believe. If there is other people who will vouch for it, I/we may reconsider.

PR #2629, opened 2020-05-04
Initialize SSL when making a HTTP Request and URHO3D_SSL is defined

stievie:
Otherwise you'll get SSL not initialized error message when making a HTTPS request.

### DIFF
--- a/Source/Urho3D/Network/HttpRequest.cpp
+++ b/Source/Urho3D/Network/HttpRequest.cpp
@@ -35,6 +35,9 @@ namespace Urho3D
 
 static const unsigned ERROR_BUFFER_SIZE = 256;
 static const unsigned READ_BUFFER_SIZE = 65536; // Must be a power of two
+#ifdef URHO3D_SSL
+bool HttpRequest::sSSLInitialized = false;
+#endif;
 
 HttpRequest::HttpRequest(const String& url, const String& verb, const Vector<String>& headers, const String& postData) :
     url_(url.Trimmed()),
@@ -108,6 +111,14 @@ void HttpRequest::ThreadFunction()
         if (header.Length())
             headersStr += header + "\r\n";
     }
+
+#ifdef URHO3D_SSL
+    if (!sSSLInitialized)
+    {
+        mg_init_library(MG_FEATURES_TLS);
+        sSSLInitialized = true;
+    }
+#endif
 
     // Initiate the connection. This may block due to DNS query
     mg_connection* connection = nullptr;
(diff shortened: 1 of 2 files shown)


Source/Urho3D/Network/HttpRequest.h
```diff
@@ -45,6 +45,10 @@ enum HttpRequestState
 /// An HTTP connection with response data stream.
 class URHO3D_API HttpRequest : public RefCounted, public Deserializer, public Thread
 {
+private:
```
ArnisLielturks, line 48: There is already a `private` block at the end of the class definition and all variable names must start with lower case letter end with and underscore. So in this case it would be `sslInitialized_` + you could also initialize it in the header file with the brackets `sslInitialized_{false}`

Source/Urho3D/Network/HttpRequest.cpp
```diff
@@ -35,6 +35,9 @@ namespace Urho3D
 
 static const unsigned ERROR_BUFFER_SIZE = 256;
 static const unsigned READ_BUFFER_SIZE = 65536; // Must be a power of two
+#ifdef URHO3D_SSL
+bool HttpRequest::sSSLInitialized = false;
```
ArnisLielturks, line 39: Not needed if we initialize the variable in the header file

Source/Urho3D/Network/HttpRequest.cpp
```diff
@@ -109,6 +112,14 @@ void HttpRequest::ThreadFunction()
             headersStr += header + "\r\n";
     }
 
+#ifdef URHO3D_SSL
+    if (!sSSLInitialized)
+    {
+        mg_init_library(MG_FEATURES_TLS);
+        sSSLInitialized = true;
```
ArnisLielturks, line 119: variable name must be updated

Source/Urho3D/Network/HttpRequest.h
stievie: That's a static member variable and must be initialized in the source file. It's static because `mg_init_library` should only be called once, no matter how many requests you make.

I can change the name of the variable.

Source/Urho3D/Network/HttpRequest.cpp
stievie: But now that I think about it, it would be probably better to just put:
~~~cpp
#ifdef URHO3D_SSL
    static sslInitialized = false;
    if (!sslInitialized)
    {
        mg_init_library(MG_FEATURES_TLS);
        sslInitialized = true;
    }
#endif
~~~
in the Constructor.

ArnisLielturks: Not a bad idea.

PR #2632, opened 2020-05-10
Use MouseButtonFlags and QualifierFlags instead of int's

### LINKED ISSUE #2631: UI: Windows are drag moved with any mouse button
According to the following code, Windows should only be moved with the left mouse button (UI/Window.cpp Line 133-137):
~~~cpp
    if (buttons != MOUSEB_LEFT || !CheckAlignment())
    {
        dragMode_ = DRAG_NONE;
        return;
    }
~~~

# Reproduce

To reproduce, change the 37_UIDrag sample to make Windows instead of Buttons:
~~~patch
diff --git a/Source/Samples/37_UIDrag/UIDrag.cpp b/Source/Samples/37_UIDrag/UIDrag.cpp
index 79097cb34..e803613db 100644
--- a/Source/Samples/37_UIDrag/UIDrag.cpp
+++ b/Source/Samples/37_UIDrag/UIDrag.cpp
@@ -25,6 +25,7 @@
 #include <Urho3D/UI/Font.h>
 #include <Urho3D/UI/Text.h>
 #include <Urho3D/UI/UIEvents.h>
+#include <Urho3D/UI/Window.h>
 
 #include "UIDrag.h"
 
@@ -69,7 +70,9 @@ void UIDrag::CreateGUI()
 
     for (int i=0; i < 10; i++)
     {
-        auto* b = new Button(context_);
+        auto* b = new Window(context_);
+        b->SetMovable(true);
+        b->SetResizable(false);
         root->AddChild(b);
         // Reference a style from the style sheet loaded earlier:
         b->SetStyleAuto();
@@ -172,8 +175,9 @@ void UIDrag::HandleDragMove(StringHash eventType, VariantMap& eventData)
     auto* t = element->GetChildStaticCast<Text>("Event Touch", false);
     t->SetText("Drag Move Buttons: " + String(buttons));
 
-    if (buttons == BUTTONS)
-        element->SetPosition(IntVector2(X, Y));
+    // Windows are automatically positioned
+//    if (buttons == BUTTONS)
+//        element->SetPosition(IntVector2(X, Y));
 }
 
 void UIDrag::HandleDragCancel(StringHash eventType, VariantMap& eventData)
~~~

I think the problem is that UI.cpp `UI::ProcessMove()` the `buttons` argument is a `MouseButtonFlags` type, but `OnDragBegin` takes an `int`. I guess the compiler doesn't use the `operator Integer() const` operator of the `FlagSet` class to cast the `buttons` argument from `MouseButtonFlags` to `int`. Maybe because the underlying type of `MouseButtonFlags` is an `unsigned` not an `int`.

# Possible solutions

There are different possible solutions.

1. Do not let the compiler implicitly cast `MouseButtonFlags` to an `int`, but use `buttons.AsInteger()` (this would work, I tested it)
2. Or change the signature of `OnDragBegin()`, `OnDragMove()` and other functions to take a `MouseButtonFlags` type as `buttons` argument instead of an `int`. And then maybe also the `qualifiers` argument should be changed from `int` to `QualifierFlags`.
3. Maybe it also works when the buttons argument of `OnDragBegin()`, `OnDragMove()` etc. is changed from `int` to unsigned.

I can make a PR when you tell me what you prefer.

stievie:
Fixes #2631

### DIFF
--- a/Source/Urho3D/UI/UIElement.cpp
+++ b/Source/Urho3D/UI/UIElement.cpp
@@ -418,31 +418,31 @@ const IntVector2& UIElement::GetScreenPosition() const
     return screenPosition_;
 }
 
-void UIElement::OnHover(const IntVector2& position, const IntVector2& screenPosition, int buttons, int qualifiers, Cursor* cursor)
+void UIElement::OnHover(const IntVector2& position, const IntVector2& screenPosition, MouseButtonFlags buttons, QualifierFlags qualifiers, Cursor* cursor)
 {
     hovering_ = true;
 }
 
-void UIElement::OnDragBegin(const IntVector2& position, const IntVector2& screenPosition, int buttons, int qualifiers,
+void UIElement::OnDragBegin(const IntVector2& position, const IntVector2& screenPosition, MouseButtonFlags buttons, QualifierFlags qualifiers,
     Cursor* cursor)
 {
-    dragButtonCombo_ = buttons;
+    dragButtonCombo_ = buttons.AsInteger();
     dragButtonCount_ = CountSetBits((unsigned)dragButtonCombo_);
 }
 
-void UIElement::OnDragMove(const IntVector2& position, const IntVector2& screenPosition, const IntVector2& deltaPos, int buttons,
-    int qualifiers, Cursor* cursor)
+void UIElement::OnDragMove(const IntVector2& position, const IntVector2& screenPosition, const IntVector2& deltaPos, MouseButtonFlags buttons,
+    QualifierFlags qualifiers, Cursor* cursor)
 {
 }
 
-void UIElement::OnDragEnd(const IntVector2& position, const IntVector2& screenPosition, int /*dragButtons*/, int /*releaseButton*/,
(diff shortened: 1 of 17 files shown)


Source/Urho3D/UI/UIElement.cpp
```diff
@@ -418,31 +418,31 @@ const IntVector2& UIElement::GetScreenPosition() const
...
 {
     hovering_ = true;
 }
 
-void UIElement::OnDragBegin(const IntVector2& position, const IntVector2& screenPosition, int buttons, int qualifiers,
+void UIElement::OnDragBegin(const IntVector2& position, const IntVector2& screenPosition, MouseButtonFlags buttons, QualifierFlags qualifiers,
     Cursor* cursor)
 {
-    dragButtonCombo_ = buttons;
+    dragButtonCombo_ = buttons.AsInteger();
```
eugeneko, line 429: Can `dragButtonCombo_ ` be `MouseButtonFlags` as well?

stievie: Yes, it is possible. I changed it now.

eugeneko:
If no objections arise, I'll merge it in three days.

PR #2634, opened 2020-05-24
Emscripten shell mouse events

ArnisLielturks:
Mouse cursor visibility updates for the new html shell.

### DIFF
--- a/Source/Urho3D/Input/Input.cpp
+++ b/Source/Urho3D/Input/Input.cpp
@@ -153,8 +153,10 @@ EmscriptenInput::EmscriptenInput(Input* inputInst)
     inputInst_ = inputInst;
     auto* vInputInst = (void*)inputInst;
 
+#ifndef URHO3D_CUSTOM_SHELL
     // Handle pointer lock
     emscripten_set_pointerlockchange_callback(NULL, vInputInst, false, EmscriptenInput::HandlePointerLockChange);
+#endif
 
     // Handle mouse events to prevent mouse jumps
     emscripten_set_mousedown_callback(NULL, vInputInst, true, EmscriptenInput::HandleMouseJump);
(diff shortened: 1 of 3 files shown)


ArnisLielturks:
All samples available here: https://mouse-events.frameskippers.com/
Might be hard to notice any changes, but this basically fixes the rotation when mouse pointer is locked inside the canvas. Previously canvas captured the mouse pointer and sometimes you were not able to fully rotate the camera around the character. It allowed to rotate the camera until the cursor hit the canvas border.

eugeneko:
The issue with fullscreen-windowed transition became worse.

STR:

1) Open sample that always captures mouse (e.g. S13)
2) Capture mouse
3) Go to fullscreen (I use F11)

Now the mouse movement is completely disabled (although clicks work).
In current Urho samples it becomes alive when I click any mouse button.
In PR samples it doesn't and I have to Esc and re-capture mouse in order to make it work.

Win10, Chrome x64, 81.0.4044.138

Source/Urho3D/Input/Input.cpp
```diff
@@ -167,7 +174,13 @@ void EmscriptenInput::RequestPointerLock(MouseMode mode, bool suppressEvent)
 {
     requestedMouseMode_ = mode;
     suppressMouseModeEvent_ = suppressEvent;
+#ifdef URHO3D_CUSTOM_SHELL
```
eugeneko, line 177: When do we have "custom" shell and when is it "standard"?
Is "standard" shell even supported?

ArnisLielturks: I'll be honest. I'm trying to keep the support for the legacy emscripten shell as it was before until I resolve all the important issues with the new shell. Once that's done, I'm planning to remove it completely.

PR #2639, opened 2020-06-03
AS Autobindings

1vanK:


eugeneko:
What's the difference between MathAPI and ManualMath?
I also cannot find generated files...

1vanK:
> I also cannot find generated files...

All generated code in file `https://github.com/1vanK/Urho3D/blob/master/Source/Urho3D/AngelScript/Generated.cpp` This file can be deleted: this file is generated on the fly when `URHO3D_GENERATEBINDINGS=1`.

> What's the difference between MathAPI and ManualMath?

MathAPI - all bindings for Math

ManualMath - manual written part of Math bindings. I still working for it, so in future this will be minimized as possible. The main problem is that some functions consider structures as an array of numbers with a manually set size for optimization purposes. In other classes, there are no such hacks and the manually written part will not be required or will be minimal.

1vanK:
Travis used obsole version of Doxygen (1.8.13), that generated incorrect xml for XMLFile::ToString().
Is there a way to specify the version of tDoxygen that Travis  should use?

1vanK:
Help! Hown to install this? <https://ubuntu.pkgs.org/20.04/ubuntu-universe-amd64/doxygen_1.8.17-0ubuntu2_amd64.deb.html>

1vanK:
On Windows 10 official `doxygen-1.8.13.windows.x64.bin.zip` generate several xml files and crash... So I just set minimal doxygen version to 1.8.17

appveyor uses 1.8.17 but I still have no idea how to force travis use latest doxygen

1vanK:
Documentation also have error:
https://urho3d.github.io/documentation/HEAD/class_urho3_d_1_1_x_m_l_file.html

```
ToString (const String &indentation="\) const
```

weitjong:
Thanks for your hard work. I will check whether we could upgrade the deps version on our docker images later when I have time.

weitjong:
The upstream Ubuntu docker image has moved the "latest" tag to "Focal Fossa", which is the code name for "Ubuntu 20.04 LTS". The default Doxygen version for "Focal Fossa" is 1.8.17. So, the Doxygen dependency issue should be fixed by simply upgrading the base image. I will do that today and let you know again when it is done.

1vanK:
```
Fatal error: can't write 238 bytes to section .text of CMakeFiles/Urho3D.dir/AngelScript/Generated.cpp.obj: 'File too big'
```

May be exists some flags for compiler?

weitjong:
I recall we have to use “bigobj” flag for MSVC compiler for a few build targets. Similar flag should exist for MinGW too.

BTW, sorry that I got side track in upgrading the base OS for DBE. While fixing breaks, I take the opportunity to upgrade the compiler toolchains.

1vanK:
What is principal difference between successful and unsuccessful configurations?

weitjong:
The CI jobs that failed have setup to perform both the "build" and "test" steps. The "test" step currently run all the samples for about 5 seconds each. Based on the last travis log the failed sample test were Editor and NSW. So, it is easy to guess that at least one of the AS binding has caused the failure. Unfortunately though, the log could not show the exact root cause of the error. If only we could retrieve the corresponding Urho3D.log of the failed test then we should be able to get the clue from there. However since you know since when it start to fail, I hope you are able to quickly isolate the offending AS binding. Let me know if you need my help to run the failed test locally and reproduce the log.

1vanK:
I installed Linux Mint 20.04 on my old notebook, and successfully build and run Editor and other examples with this PR.

After this I downgraded Doxygen from 1.8.17 to 1.8.13.

I tried two prebuilded versions: `doxygen-1.8.13.linux.bin.tar.gz` from https://sourceforge.net/projects/doxygen/files/rel-1.8.13/ and `doxygen_1.8.13-10_amd64.deb` from https://ubuntu.pkgs.org/18.04/ubuntu-main-amd64/doxygen_1.8.13-10_amd64.deb.html and both crash when try generate xml files

```
[ 64%] Generating bindings to script languages
Generating XMLs with AST to /home/user/Urho3DFork/Build/Source/Tools/BindingGenerator/generated/xml
Segmentation fault (core dumped)
make[2]: *** [Source/Urho3D/CMakeFiles/Urho3D.dir/build.make:641: /home/user/Urho3DFork/Urho3D/Source/Urho3D/AngelScript/Generated.cpp] Error 139
make[1]: *** [CMakeFiles/Makefile2:2501: Source/Urho3D/CMakeFiles/Urho3D.dir/all] Error 2
```

weitjong:
I think those are different issue. One is the compilation error and the other is the runtime error. Anyway, if you really need the newer Doxygen now in your CI build then you can get it by adding one line in your `.travis.yml` like so:

https://github.com/urho3d/Urho3D/blob/ba629b81306f5afc6ae20bfb47708a99398c9efd/.travis.yml#L34

After the highlighted line, add another line below it.

```
  - DBE_TAG=latest
```
Note that though, the new DBE images are still in a flux especially those with cross-compiling toolchains. But I think the native DBE image should be stable enough for your usage.

1vanK:
Thanks, this fixed the problem with generating bindings. But caused a new problem inside SDL in 64-bit configs

weitjong:
> Thanks, this fixed the problem with generating bindings. But caused a new problem inside SDL in 64-bit configs

Those with build config with `URHO3D_64BIT=0` option are for 32-bit CI build actually. There was minor changes in how the 32-bit native DBE image is being prepared, however, in my own test branch the native DBE image seems to work correctly for both ABI. I have just rerun the 32-bit CI job on test branch and it was OK. https://travis-ci.org/github/urho3d/Urho3D/jobs/710794195. So, whatever it is, it only happened in your PR branch which is strange.

Let me manually clear the Travis build cache for your PR branch after all your CI jobs completed later on. The next time you commit then we check again if the issue reoccur or not.

weitjong:
Sorry that I didn't recall it earlier but your recent 32-bit build failure seems similar to what I have had before in my test branch, and I fixed that with this commit: https://github.com/urho3d/Urho3D/commit/c91c4ec0e4a6143c29f9d21beb60f46e8ba34cfd.

1vanK:
Thanks!

1vanK:
```
[ 76%] Building CXX object Source/Urho3D/CMakeFiles/Urho3D.dir/LibraryInfo.cpp.obj

..........................

cc1plus: out of memory allocating 65536 bytes after a total of 6360072192 bytes

make[2]: *** [Source/Urho3D/CMakeFiles/Urho3D.dir/build.make:1062: Source/Urho3D/CMakeFiles/Urho3D.dir/AngelScript/Generated.cpp.obj] Error 1

make[1]: *** [CMakeFiles/Makefile2:2842: Source/Urho3D/CMakeFiles/Urho3D.dir/all] Error 2
```

1vanK:
Code still ultra dirty and I still to work on it, but AutoBinder already cover 99% of the engine. The number of manual bindings is insignificant and will be reduced in the near future. So those interested can start testing this already.

weitjong:
Please kindly change your Travis CI to use back the "master" DBE tag again, i.e. remove the "DBE_TAG=latest" in the build config. Those DBE images that you are currently using have been re-tagged as "master" a few days ago.

I am planning to make a breaking change in the DBE that will only work with Github Actions. So, the new "latest" tag will break your CI in Travis if you continue to use it.

1vanK:
already deleted in e42b58c

1vanK:
There is still a lot of work to be done and I am continuing to work on the autobindings. But I propose to merge this as is because it fixes a huge amount mistakes in the current bindings, and also allows users to don't worry about AS bindings for new PRs.

P.S. URHO3D_UPDATE_SOURCE_TREE is not enabled and not tested

weitjong:
Thanks for the hard work. Could you merge the PR to the dev branch that I have created for you earlier first. Thanks.

1vanK:
That branch contains a lot of unnecessary commits, this PR has one commit at the top of the master. It will be difficult to resolve conflicts. Can I create a new branch?

weitjong:
Yes, you can. You can force push with this existing branch or just create a new one. It is all the same. Just make sure you don't lose any commits you want to keep.

1vanK:
https://github.com/urho3d/Urho3D/tree/generate-bindings-new

weitjong:
Let me know if you prefer I merge my dev branch into master first or yours first, if it is ready. Thanks.

1vanK:
I don't think it will be ready anytime soon, but it just works. I continue to make significant changes in the generator to make the code look less garbage, but it will generate the same bindings, so you can safely merge https://github.com/urho3d/Urho3D/tree/generate-bindings-new

weitjong:
Can confirm that you mean I should take the `generate-bindings-new` as it is now (last squashed on 19 Sept)?

1vanK:
yep

1vanK:
I have a lot of unfinished and experimental changes, but I will PR them later as one commit.

1vanK:
I only remind you that in generate-bindings-new URHO3D_UPDATE_SOURCE_TREE is not enabled and not tested

weitjong:
Thanks for the head up. I would not worry about "URHO3D_UPDATE_SOURCE_TREE" for now. The new CI/CD workflow after  migrating from Travis CI to Github Actions does not have this part scripted and tested also. Originally I plan to merge after I have a feature parity, but suddenly I am not in the mood to continue, at least for now.

weitjong:
I have merged it to the main branch with one more commit to disable the auto-generation until it is ready for prime time. Thanks for the hard work again. My eyes are too tired to go through all the changed files, but you have earned my trust. Since you have the full write access like the rest of the project members, you have my permission to merge your future changes from your dev branch to main branch.

weitjong:
There was issue with the Web build caused by the script API binding for the network subsystem. As you know currently network subsystem is disabled for Web build, so the script API binding failed. I see other network related binding are property guarded by the compiler preprocessor, except these 2. https://github.com/urho3d/Urho3D/runs/1320657829?check_suite_focus=true. Could you kindly help to sort it out. Thanks.

weitjong:
I think I am able to patch the generator to fix the problem I mentioned above. However, it appears that the exact output order of the binding generator is platform specific. On my Linux machine, it will regenerate the bindings with different order than the one you have provided. So, in the end I decided to wait for you to provide the fixed binding files.

Also I notice when the URHO3D_GENERATEBINDING is turned on, it actually generates the source files directly to the source tree[?]. The GCC does not like that. It detects a circular dependency. As the source tree is constantly dirtied, the successive 'make' command invocation detects there is work to do because the build tree timestamp is always older than the source tree, as soon as the build finished.

1vanK:
I will solve this problems soon

1vanK:
Currently I am experimenting with generating bindings as templates to significantly reduce the size of the source code. The problem with different ordering can be solved by sorting alphabetically

1vanK:
> It detects a circular dependency. 

I know this problem (VS show same warning), I believe this problem is in the generator's sources, but haven't looked into this question yet

1vanK:
Безымянный

weitjong:
I hope you won't be using Travis-CI and AppVeyor anymore. I am going to re-tag the current "latest" DBE image tag to become the new "master". As the change in the DBE is not backward compatible, this will break any PRs still using Travis-CI and Appveyor and pointing to the new "master" DBE tag.

1vanK:
No problems

1vanK:
Can we check doxygen version when projects generates and show warning when user uses untested version?
Currenly I test binding generator with doxygen 1.8.17 and 1.8.20 but there is no guarantee that in future versions Doxygen will be generate same xmls (as it was with version 1.8.14)

weitjong:
Yes, it is possible to request for a specific (minimum) version, if you want to do so. Modify the first highlighted line here

https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/CMakeLists.txt#L197-L201

Put the version string in between the package name and the REQUIRED keyword. See https://cmake.org/cmake/help/v3.18/command/find_package.html for detail.

1vanK:
As far as I understand, Cmake checks not minimum, but compatible version. Whether the version is compatible is determined by a separate script.

For boost it is minimum: https://github.com/Kitware/CMake/blob/master/Modules/FindBoost.cmake#L13

For doxygen I do not see any checks: https://github.com/Kitware/CMake/blob/master/Modules/FindDoxygen.cmake

weitjong:
The way I understood is, the corresponding module declares what was found and its version using another standard macro. https://github.com/Kitware/CMake/blob/master/Modules/FindDoxygen.cmake
The version string comparison is not the responsible of the module. The rest is the responsible of the “find_package” command. It should function as per the API doc, especially for the modules that they shipped as part of CMake.

1vanK:
Oh yeah you are right

1vanK:
Do you receive notifications about messages from draft PR https://github.com/urho3d/Urho3D/pull/2720 ?

PR #2643, opened 2020-06-06
Allow Camera::SetFlipVertical to be used externally for OpenGL

### LINKED ISSUE #2642: How to change from left to right handed coordinate system?
Is there a "proper" way to switch Urho3D to use a right handed coordinate system?
(As Urho defaults to a left handed system)

The reason I need this is for compatibility with a large system I am trying to adapt to use Urho3D for the 3D engine.

So far, I seem to have made good progress by misusing the `Camera::SetFlipVertical`  function in both DX9 and OGL modes. However, this is not ideal because that seems to really be intended for internal use, plus in OGL I have to keep re-enabling it because Urho resets it to false after each frame. I think there are probably other things that will bite me for this kludge too.

I did wonder if setting a custom projection via Camera::SetProjection, but that seems to disable the automatic aspect ratio adjustment. So I wondered if there is something I'm missing that would help me here?

ChrisDenham:
Allow `Camera::SetFlipVertical` to be used externally for OpenGL by toggling it either side of render to texture, i.e. instead of "set and clear" either side of render to texture.

The purpose of this pull request is to allow `Camera::SetFlipVertical` to be used for both DirectX and OpenGL to switch into a right handed coordinate system (see #2642). 

This pull request should not change the behaviour of any application that does not call `Camera::SetFlipVertical` directly. It also should correct the scenario where if the render to texture should not be flipped if the camera is already flipped (hence the toggling of state in the changes) 

Closes #2642

### DIFF
--- a/Source/Urho3D/Graphics/View.cpp
+++ b/Source/Urho3D/Graphics/View.cpp
@@ -601,6 +601,10 @@ void View::Render()
     {
         // On OpenGL, flip the projection if rendering to a texture so that the texture can be addressed in the same way
         // as a render texture produced on Direct3D9
+        // Note that the state of the FlipVertical mode is toggled here rather than enabled
+        // The reason for this is that we want the mode to be the opposite of what the user has currently set for the
+        // camera when rendering to texture for OpenGL
+        // This mode is returned to the original state by toggling it again below, after the render
         if (camera_)
             camera_->SetFlipVertical(!camera_->GetFlipVertical());
     }
--- a/Source/Urho3D/Graphics/View.cpp
+++ b/Source/Urho3D/Graphics/View.cpp
@@ -596,15 +596,15 @@ void View::Render()
     }
 #endif
 
+#ifdef URHO3D_OPENGL
     if (renderTarget_)
     {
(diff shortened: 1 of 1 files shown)


eugeneko:
If no objections arise, I'll merge it in three days

Lumak:
for every render target. flipping is toggled
```
camera_->SetFlipVertical(!camera_->GetFlipVertical());
```
so every other render target is flipped?

ChrisDenham:
No, I don't think so. My change toggles the flip camera state either side of the render for OGL. I believe this is equivalent to the original code which sets then clears the flipped state. The purpose of my PR is simply to retain the flipped state of the camera set by the user, which was previously being lost every frame for OGL.

1vanK:
Yes, the code does not look intuitive. May be add some more comment, that the value restored below.

Lumak:
Yes, the comment is very informative. I get it now.

ChrisDenham:
Can this PR be merged now? Or are there any outstanding concerns about it?

eugeneko:
Yeah, I tend to say "I'll merge it in 3 days" and forget about it for weeks.

PR #2645, opened 2020-06-12
Fix crash by fixing scene clean order

KatekovAnton:
I was creating temporary in-memory node, and then loading it with data. 
```
Urho3D::SharedPtr<Urho3D::Node> node;
node = new Urho3D::Node(m_DestinationNode->GetContext());
node->LoadXML(xmlFile.GetRoot());
m_DestinationNode->AddChild(node);
```

Project was crashing when I was closing it. Stack was showing path from `Scene::~Scene` to `Node::SetOwner` and node was looking very weird - like it was already deleted. Crash was here:
```
void Node::SetOwner(Connection* owner)
{
    impl_->owner_ = owner;
}

```
So I swapped scene destructor logic and put child data cleanup on first place and children destruction itself - onto the last place. 

Crash disappeared.

MacOS/Xcode 11.


eugeneko:
It's not obvious for me how this change fixes described bug.
It's unsafe to merge unless we understand the full scope of changes and consider it desirable.
I don't understand this change (yet), so I cannot verify its validity.

1vanK:
Can you provide small complete example to reproduce issue?

PR #2650, opened 2020-06-19
Default values for Variant Get methods

MystMagus:
Added default values to some other things earlier, so it makes sense to have it for Variant as well. Didn't include ones that I felt made no sense to have defaults for.


eugeneko:
I have only one concern about this PR.
Assuming it is merged, can you spot an error in the code below?
```
const Variant& var = eventData["Color"];
const Color& color = var.GetColor({ 1, 0.5, 1, 0.5 });
// use color
```

MystMagus:
Ah true, scope. I guess it'd be best to make separate functions for default values for classes, since they'll need to not be references and it'd be wasteful to copy stuff if not necessary.

That said, you really should just state the issue that you see right away. Easier that way for everyone in my opinion.

eugeneko:
This specific issue with returning const-ref parameter by const-ref is probably the most annoying part of C++ for me.
And I'm not sure if there's even good solution for it.

Probably adding separate function with different name that returns copy.
Otherwise it's too easy to make a mistake. Either by missing scoping and introducing crash, or by making copy by accident when you don't really need it.

MystMagus:
Indeed. I'll take care of it next week.

PR #2651, opened 2020-06-19
Ignore ScrollView ScrollBars on the first size calculation pass

MystMagus:
ScrollView could get stuck with a scrollbar on one side that didn't need it when resizing the scrollview, due to taking the scrollbar sizes into account too early.

### DIFF
--- a/Source/Urho3D/UI/ScrollView.cpp
+++ b/Source/Urho3D/UI/ScrollView.cpp
@@ -262,7 +262,7 @@ void ScrollView::OnKey(Key key, MouseButtonFlags buttons, QualifierFlags qualifi
 
 void ScrollView::OnResize(const IntVector2& newSize, const IntVector2& delta)
 {
-    UpdatePanelSize();
+    UpdatePanelSize(true);
     UpdateViewSize();
 
     // If scrollbar auto visibility is enabled, check whether scrollbars should be visible.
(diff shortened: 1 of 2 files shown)


MystMagus:
Am I doing something wrong when I attempt to contribute? If so then please tell me.

talarcon:
I think there is no one here to supervise this changes.

1vanK:
If you will provide a small buildable complete project to reproduce the problem so that I don't waste time to creating my own project, but can just check that this PR solves the problem, I can check and merge this commit.

MystMagus:
[ScrollerTest.zip](https://github.com/urho3d/Urho3D/files/4959911/ScrollerTest.zip)

A bit of a specific case, I know, but whatever the situation it does cause a state in which the scrollview has scrollbars that do nothing because there is no actual scroll. Resize the window down to the point that the scrollbars are supposed to appear and it will change to always having one scrollbar, even after increasing the size.

That aside, you didn't actually answer my question. I did get feedback on this pr, and I appreciate it, but I have several more that also time out without getting anything. Am I missing something in the submission process? It doesn't feel like there's a point to try to contribute if things just time out without anything said, even having them rejected is better.

1vanK:
Here everyone does something based on their desire/free time/competence. This is open source. Apparently your PR just didn't interest anyone enough.

1vanK:
With this change I can not scroll this panel

MystMagus:
Seems to bug out if they are manually set to visible. Will get back.

PR #2653, opened 2020-06-26
Added PODVector::Resize overload with default value

SirNate0:
Added PODVector::Resize overload that fills new elements with a default value.

I think it's useful as the user may prefer something like PODVector<Vector3> to be extended with a "nice" value like Vector3::ZERO in some circumstances while not caring in the general case.


1vanK:
But we already have this: https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/Container/Vector.h#L427

PR #2656, opened 2020-06-27
Image too dark for DX9 caused by mismatch between code PBRLitSolid.hlsl and PBRLitSolid.glsl  

ChrisDenham:
This change removes a line of HLSL code that does not occur in the GLSL version of the code.
In my experiments I noticed that the environment mapping was much darker in DX9 vs OGL versions.
I tracked to the difference to be purely a consequence of this line of code, which in effect squares the darkening effect of the specular color.  I suspect this difference may not have been noticed because it makes no difference if the material specular color is full white.

It may be that it is the GLSL version that needs correcting to make it consistent with the HLSL version, but I'm not really sure why the HLSL version of this code would need to effectively multiply a specular color by itself. But in anycase, I think the HLSL and GLSL code should at least be consistent in this respect.

Too dark in DX9:
PBRUrho3DDX9-envmap-too-dark

OK in OGL:
PBRUrho3DOpenGL

OK in DX9 with fix in this PR:
PBRUrho3DDX9

### DIFF
--- a/bin/CoreData/Shaders/HLSL/PBRLitSolid.hlsl
+++ b/bin/CoreData/Shaders/HLSL/PBRLitSolid.hlsl
@@ -222,7 +222,6 @@ void PS(
     metalness = clamp(metalness, METALNESS_FLOOR, 1.0);
 
     float3 specColor = lerp(0.08 * cMatSpecColor.rgb, diffColor.rgb, metalness);
-    specColor *= cMatSpecColor.rgb;
     diffColor.rgb = diffColor.rgb - diffColor.rgb * metalness; // Modulate down the diffuse
 
     // Get normal


gleblebedev:
I've also noticed it yesterday.

From my understanding there is no specular in pbr pipeline as Urho3D goes with metallic-roughness approach. I think your change is correct - there is no reason to multiply by specular color.

1vanK:
@dragonCASTjosh ?

ChrisDenham:
The things that I couldn't get my head around are that the effect seemed to only impact where it's multplied into envmap color in the ImageBasedLighting function called lower down, and also that specColor is a lerp between 8% of the specular color and 100% of the diffuse color, and then multiplying that result by 100% of specular color didn't seem to make mathematical sense to me.

1vanK:
As far as I remember, dragonCASTjosh first wrote hlsl version, and then ported to glsl. If we remove cMatSpecColor, this setting from material will be unused at all

ChrisDenham:
Are you suggesting removing use of cMatSpecColor entirely from the script? Or that it would be wrong to do so?

1vanK:
I'm not good at PBR shaders, but construction `float3 specColor = lerp(0.08 * cMatSpecColor.rgb, diffColor.rgb, metalness);` looks strange for me.

`0.08 * cMatSpecColor.rgb` looks like small value (~0) almost no affected to result

ChrisDenham:
Yes, I agree. And to then scale the result down again by multiplying by `cMatSpecColor` adds to rhe strangeness.

ChrisDenham:
Btw. regardless of whether the hlsl code is working as intended, the real issue here is that it's not consistent with the glsl version. Which suggests that either one or the other should be changed to match because they can't both be correct.

gleblebedev:
As I'm working on exporting materials from Unity to Urho I can confirm the glsl version looks close to Unity when hlsl doesn't.

Take the PR, it is good :)

1vanK:
I don't think we should delete this. You can just set MatSpecColor in material to white.
Without `specColor *= cMatSpecColor.rgb;` you lose additional control over the material.

dragonCASTjosh:
Sorry for being late to the party, I have not looked at this shader in a while but i can try take a look again.
From my memory is was not at all happy with the PBR shaders as i feel there is much to be desired along with not being able to do reflection probes in engine.

i do remember first developing the shaders on DX11 as part of my intention at the time was to bring Urhos rendering into the modern era so its likely that glsl has a mistake somewhere

ChrisDenham:
I don't really have a strong opinion about which has the most correct behaviour. The hlsl and glsl seemed very close in most respects apart from that one line missing from the glsl. So, if it's preferred to keep the line of code, then I can submit an alternative PR that adds it to the glsl? It doesn't make sense to me for them to behave differently.

dragonCASTjosh:
In this case looking at the image it seams like GL is showing the correct behavior. just need to find out why

dragonCASTjosh:
Actually you may be correct to remove the additional secular multiplication as the color is already controlled during the lerp from what i can tell

ChrisDenham:
Yes, that's what I thought was odd in the hlsl, i.e. that it did a lerp with it and then multiplied by it.

1vanK:
> In this case looking at the image it seams like GL is showing the correct behavior. just need to find out why

Probably because MatSpecColor in material is set to 0

> Actually you may be correct to remove the additional secular multiplication as the color is already controlled during the lerp from what i can tell

`0.08 * cMatSpecColor.rgb` very small value no effected on result (a tested this in editor)

What is this value 0.08?

gleblebedev:
I would guess it's a dielectric specular value. See some sample values from Unity:
https://docs.unity3d.com/Packages/com.unity.shadergraph@5.10/manual/Dielectric-Specular-Node.html

dragonCASTjosh:
If i remember correctly its an F0 term. quite a high one as normally dielectrics on 0.02-0.05 maybe something to look into again

PR #2658, opened 2020-07-06
Fix vegetation shader for vsm shadows

KatekovAnton:
Fix error when compiling vegetation shader and using vsm shadows


gleblebedev:
Probably you need to change HLSL file too.
    #ifdef VSM_SHADOW
        out float4 oTexCoord : TEXCOORD0,
    #else
        out float2 oTexCoord : TEXCOORD0,
    #endif


....

    #ifdef VSM_SHADOW
        oTexCoord = float4(GetTexCoord(iTexCoord), oPos.z, oPos.w);
    #else
        oTexCoord = GetTexCoord(iTexCoord);
    #endif

gleblebedev:
Extended the bugfix in https://github.com/urho3d/Urho3D/pull/2659

1vanK:
It works, thanks!

PR #2661, opened 2020-07-10
Check if scene exists

KatekovAnton:
Crash when creating node in memory (not in scene) and setting its tag. Or loading scene from file as prefab. Creating nodes in memory is correct operation afaik and is using to optimize loading process.

### DIFF
--- a/Source/Urho3D/Scene/Node.cpp
+++ b/Source/Urho3D/Scene/Node.cpp
@@ -360,17 +360,16 @@ void Node::AddTag(const String& tag)
 
     // Cache
     if (scene_) {
-       scene_->NodeTagAdded(this, tag);
+        scene_->NodeTagAdded(this, tag);
+    
+        // Send event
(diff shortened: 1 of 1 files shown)


Source/Urho3D/Scene/Node.cpp
```diff
@@ -360,17 +360,16 @@ void Node::AddTag(const String& tag)
 
     // Cache
     if (scene_) {
```
1vanK, line 362: use existed code style

KatekovAnton: what?

1vanK: `{` on a separate line

KatekovAnton: fixed!

KatekovAnton:
@1vanK github doesn't show new commits for some reason. closing this pr and creating new to include all changes.

PR #2663, opened 2020-07-15
Bugfix/fix crash creating inmemory nodes

KatekovAnton:
Crash when creating node in memory (not in scene) and setting its tag. Or loading scene from file as prefab. Creating nodes in memory is correct operation afaik and is using to optimize loading process.

previous pr:
https://github.com/urho3d/Urho3D/pull/2661


1vanK:
Thanks!

PR #2665, opened 2020-07-21
Create unit testing target

KatekovAnton:


eugeneko:
1) Consider postponing this PR until you have actual tests
2) Make tests auto-run in CI

KatekovAnton:
@eugeneko 
1 is ok - i will add unit tests for state machine
2 is separate feature. I will do later on another pr

ArnisLielturks:
Few commens:

ArnisLielturks:
Few comments from my side:

1. I'm all in for this
2. Add additional build flag `URHO3D_TESTS` or something similar to allow users to skip them. And it should be disabled by default I believe
3. Create actual tests for already existing systems. You don't have to do 100% coverage, but few tests would be ok from my side, we can always add new tests as we move further.
4. Regarding the CI I would suggest skipping it for now since we have a bit of uncertainty regarding current CI pipeline. There is already some discussions going on but nothing has been decided yet.

KatekovAnton:
Parallel discussion is happening here https://discourse.urho3d.io/t/ci-and-unittests/6280

KatekovAnton:
Unit tests are available in feature/state_machine branch
https://github.com/KatekovAnton/Urho3D/blob/feature/state_machine/Source/Tests/StateMachineTests.cpp

PR is here: https://github.com/urho3d/Urho3D/pull/2669

Lewis-ver1803 (approved):

PR #2668, opened 2020-07-25
Possibility to use BaseVertexIndex

tederis:
In some situations it would be handy to use BaseVertexIndex argument when rendering complicated models.


tederis:
What do you think? Is there something I should take into account?

PR #2669, opened 2020-07-25
Feature/state machine

KatekovAnton:
PR contains:
- Unit testing target (separate pr is opened here https://github.com/urho3d/Urho3D/pull/2665 )
- State machine classes
- Tests for state machine
- Sample scene project (55)
- Extra assets for good looking scene

The idea is - you modify parameters - state machine change its state


gleblebedev:
Preview

eugeneko:
Can we avoid copy-pasted shaders? We have more than enough shaders to maintain, we don't need more.
Also, what's the license of assets? If assets are yours, fine. If not, please link to asset source and license.

gleblebedev:
@eugeneko shaders and techniques are duplicates. @KatekovAnton you can remove shaders and techniques (except water) as they been merged into urho main branch already.

KatekovAnton:
@eugeneko done. it was not copy-pasted - it was created by exporter. @gleblebedev we can now only export BRDF and PBRWater since shaders are merged

KatekovAnton:
Hold on. I will add one more thing

1vanK:
What license is the content distributed under?

1vanK:
> bin/HouseScene/Content/Houses/CanyonHouse/testHouse0Стекла.005.mdl 

Cyrillic in the paths

KatekovAnton:
> Cyrillic in paths

Content creator did in this way. As far as I can see it is supported by the engine.

1vanK:
Yes, but agree, you would be surprised if you saw hieroglyphics or Arabic text.

Anyway, I think the external repository is the best place for your example.

KatekovAnton:
@1vanK I have removed Cyrillic from assets.
@eugeneko I have removed shaders and techniques.

Should be good now

PR #2672, opened 2020-07-27
Fix 3D texture mipmap generator.

ssinai1:
Missing assignment for alpha channel.


ssinai1:
Sorry, I should have written image instead of texture.

weitjong:
Thanks!

PR #2673, opened 2020-08-04
Fix missing argument in script binding

ssinai1:
Missing argument in angelscript binding of Image::SetSize().


1vanK:
At the moment I am creating automatic bindings and I notice a LOT of errors in the current bindings. So just wait when i finish work

PR #2675, opened 2020-08-07
An intersection test for IntRect

tederis:


eugeneko:
The check is incorrect: `IntRect{ 0, 0, 1, 1 }` and `IntRect{ 1, 1, 2, 2 }` don't intersect.

1vanK:
as far as I remember, in urho right and bottom borders do not belong to a rectangle, but left and upper belong to the rectangle

tederis:
I have made two separated commits contain different strateges: with and without implying right-bottom borders are owned by rect. I think the final variant is correct and best suitable with this particular Urho3D implementation.

eugeneko:
Looks good for me

PR #2678, opened 2020-09-09
Fix clicks on retina display

KatekovAnton:
Remove division by 2 for cursor position because it is already in correct metrics

Issue on my mac:
Image

### DIFF
--- a/Source/Urho3D/UI/UI.cpp
+++ b/Source/Urho3D/UI/UI.cpp
@@ -1321,9 +1321,6 @@ void UI::GetCursorPositionAndVisible(IntVector2& pos, bool& visible)
         if (!visible && cursor_)
             pos = cursor_->GetPosition();
     }
-
-    pos.x_ = (int)(pos.x_ / uiScale_);
-    pos.y_ = (int)(pos.y_ / uiScale_);
 }
 
 void UI::SetCursorShape(CursorShape shape)


weitjong:
That 2 lines are part of this commit https://github.com/urho3d/Urho3D/commit/4daf7fd3f0c384c95bfc10e6e883a9269fba7a7f. I don't have Mac with retina display to verify this PR. You will need at least one person to verify and vouch for this PR.

KatekovAnton:
@weitjong I am a mac user. I have attached a video that illustrates the issue on my mac. 

I also checked invisible cursor case 
```
в Input тоже все делится 

IntVector2 Input::GetMousePosition() const
{
    IntVector2 ret = IntVector2::ZERO;

    if (!initialized_)
        return ret;

    SDL_GetMouseState(&ret.x_, &ret.y_);
    ret.x_ = (int)(ret.x_ * inputScale_.x_);
    ret.y_ = (int)(ret.y_ * inputScale_.y_);

    return ret;
}
```

and `inputScale_` is (2;2) for my macbook retina display

KatekovAnton:
I created a discord server for Urho3D - https://discord.gg/g7qgG6 please join. lets discuss there

KatekovAnton:
up

1vanK:
As far as I understand, there is no difference a normal display is used or retina. It is just scaling of UI. So I test `ui.scale` on my Win 10 and can confirm the problem. Also I can confirm that your patch fixes the mouse coordinates. But I haven't done enough research to be sure it won't have any side effects.

1vanK:
Oh, it's harder than I thought, I don't use a Mac and I don't understand what's going on there. We have highDPI oprion. Is it retina related?

1vanK:
I just found it https://github.com/urho3d/Urho3D/issues/1252

KatekovAnton:
Hi @1vanK First yes - this was discovered on retina  display. related - I use 2 setups to test:
1. macbook (MacOS) with retina display
2. matebook x pro (Win10) with high quality display and scaling set to 200%

initially issue was discovered on 1st and I fixed it. Then I did some test on 2nd and confirmed that it also works fine with the fix. Last thing I did was test on mac on 2nd display that have different scaling factor and I can confirm it works correctly.

On windows this API called HightDPI (or поддержка высоких разрешений in Russian). I can share the code how I adapted my another project to this API on windows if you wish.

1vanK:
I believe that ui.scale should work correctly on any display and this patch fixes this issue. Therefore, I will accept this patch without proper verification. If anyone has a problem, they will let as know.

1vanK:
@KatekovAnton Please do not disappear,you are now in charge if someone finds a problem with retina display xD

KatekovAnton:
@1vanK you can find me in telegram chat and on Urho discord channel here https://discord.gg/3uW2aK8

1vanK:
I would prefer a bug tracker)

dprandle (commented):
Hi - this change breaks UI scaling for me. Reverted (added back dividing by scale) and works again. I use an IMac 2019 27" (retina display).

To reproduce, I add the following line to the drag UI example
Screen Shot 2021-02-11 at 10 22 50 PM

With the scaling lines removed (as it is currently), I get:
https://user-images.githubusercontent.com/8174404/107741072-ef546580-6cb8-11eb-97c8-ba55c80301e2.mov

Then I add the scaling lines back in, so the code now is restored to:
Screen Shot 2021-02-11 at 10 21 16 PM

And I have the following normal behavior:
https://user-images.githubusercontent.com/8174404/107741196-2a569900-6cb9-11eb-95b2-df6c0276b4b5.mov

I would do a pull request to revert but I currently have to many other minor changes in my fork.

Hope this helps

KatekovAnton:
@dprandle 
Hi. Do you have second monitor attached? The fact that you have to set hardcoded scale manually looks suspicious to me. 

We tested that on a few machines including windows with high dpi (with and without external monitor), with with different mac os versions and verified the fix.

please check if UI scale was determined correctly for you.

KatekovAnton:
@dprandle
Hi. Do you have second monitor attached? The fact that you have to set hardcoded scale manually looks suspicious to me.

We tested this fix on a few machines including windows with high dpi screen (with and without external monitor), with with different mac os versions and verified the fix.

Can you please check if UI scale was determined correctly for you?

Thank you

dprandle:
I did have a second monitor attached - so I just detached it, rebooted, went through the same process to get the same results.

I have to scale the UI manually because without scaling, the UI elements for the default UI style are too tiny on retina display... When making my own elements I could do without scaling, but then on other platforms the elements are too big.

The mouse position is fine in a global sense with or without the ui scaling - all of the object picking (which is using ray cast from mouse pos) is working fine, as well as dragging items following mouse.

With retina scaling (in high DPI mode), when I create a 1920 by 1080 window, for example, and then create a 1920 by 1080 button without UI scaling set to 2.0, the button is rendered as one screen quadrant instead of being the same size as the screen. So you could see how that might be a problem if the UI was meant for 1920 by 1080 - and with the mouse scaling removed as it was in the above merged change, the mouse seems to match the window coords and that is projected on to the UI coords.

For example, with the 1920 by 1080 button and window example above, clicking on the lower right corner of the window would generate a mouse click in the ui system on the lower right corner of the canvas.

Does that make sense?

KatekovAnton:
I see.

> 1920 by 1080 window and then create a 1920 by 1080 button without UI scaling set to 2.0, the button is rendered as one screen quadrant

Can you provide a code sample? I am guessing now that button takes its size from some image assigned to it. In this case it is a correct behavior - for 1920 by 1080 window you need 2 times larger retina-specific images. Here is the explanation from apple: https://developer.apple.com/design/human-interface-guidelines/ios/icons-and-images/image-size-and-resolution/

dprandle:
Sure - this is the code that makes the button
Screen Shot 2021-02-12 at 3 45 28 PM

And initializing the window size
Screen Shot 2021-02-12 at 3 46 38 PM

I don't have a problem if the button issue is simply that the coords specified for the window are not the window pixel size (and so the button being pixel sized is smaller than expected), but it is an issue that if I set the UI scale manually to 2, it messes up the mouse position and movement in the UI system. I would think scaling the UI manually would still result in correct mouse coord operation (which it does for me if you put back those two lines)

KatekovAnton:
Hi @dprandle 

I checked the current implementation and without modifications it works well to me.
9159fb33cdd7a85381860efaa9de6f32
I see your point. Setting the ui scale really breaks mouse pointer. In addition it breaks control drawing - fonts and images are not scaled properly and looks really bad. This test was only conducted on mac.

Considering this I think scaling elements manually is a right approach because it makes mouse pointer behavior and drawing resolution correct. This can be improved by respecting ui scale inside of the system and affecting control sizes... Considering existing project support, in order to avoid unexpected effects for urho users I imagine this new scale value should be a new property.

dprandle:
The main reason this whole thing popped up for me is the console and debug info HUDs are so tiny its hard for me to see them - so I tried to globally scale the UI

Not scaled:
Screen Shot 2021-02-12 at 10 49 45 PM

Scaled:
Screen Shot 2021-02-12 at 10 53 37 PM

The scaled ui version looks, size wise, what it looks like on my linux system.. but after the merged change I can't scale it like that anymore as the mouse doesn't work correctly with the console

Is there a way, currently, to scale the console or would I need to go in and create a new style/style image for it?

SuperWangKai:
I'm not very familiar with Urho3D's UI code, however, problem solved after a quick test (POC), I added these two lines
```c++
    positionCopy.x_ = int(positionCopy.x_ / uiScale_);
    positionCopy.y_ = int(positionCopy.y_ / uiScale_);
```

to `UIElement* UI::GetElementAt(UIElement* root, const IntVector2& position, bool enabledOnly)` before
https://github.com/urho3d/Urho3D/blob/0e488a902d89aa8e9534955313d2b0c103f28cdd/Source/Urho3D/UI/UI.cpp#L842

It seems `GetElementAt` does not take UI scale into caculation.

SuperWangKai:
https://github.com/urho3d/Urho3D/pull/2779

PR #2679, opened 2020-09-12
Update Bullet physics to 2.89.

rokups:
As title says.


1vanK:
Is this ready and why is it missing in rbfx?

1vanK:
I haven't fully explored this, but noticed that you forgot to restore some changes

```
SIMD_FORCE_INLINE   long    btVector3::maxDot( const btVector3 *array, long array_count, btScalar &dotOut ) const
{
#if (defined BT_USE_SSE && defined BT_USE_SIMD_VECTOR3 && defined BT_USE_SSE_IN_API) || defined (BT_USE_NEON)
    #ifdef BT_USE_SSE	// Urho3D - to be consistent with the function return below
```

is missing in your PR

1vanK:
I suggest updating libraries in a separate repository and move all our commits on top of upstream commits
```
git pull https://github.com/bulletphysics/bullet3 master --rebase
git push origin master --force
```
and the copy to Urho3D. This way we will always know which commits we made and why. At the moment it is rather difficult to understand how the code looked before our intervention.

1vanK:
Or just store old code in sources (comment but not delete)

1vanK:
By the way, there is version 3.06+ already

1vanK:
I am sorry that only now there was time to check the PR. Since there is already a newer version, I created another PR with a newer version.

rokups:
> Is this ready and why is it missing in rbfx?

I could add it to rbfx, or could add it to Urho3D and get it in rbfx too. Choice was to benefit two projects instead of one :)

> I suggest updating libraries in a separate repository and move all our commits on top of upstream commits

Anything would help. I was keeping plain patches in the repo along with source code myself.

> Or just store old code in sources (comment but not delete)

This is what i already suggested in the forum too.

1vanK:
> This is what i already suggested in the forum too.

Indeed, I found your message on the forum where you talked about it

PR #2680, opened 2020-09-24
Fix GetProperty method binding of some Urho2D classes

glhrmfrts:
The GetProperty AngelScript method of the PropertySet2D, Tile2D and TileMapLayer2D classes was wrongly bound to the HasProperty method. So I changed it to use the right one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,7 @@ Urho3D development, contributions and bugfixes by:
 - Katekov Anton
 - SirNate0
 - tederis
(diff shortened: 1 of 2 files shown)


1vanK:
already fixed in https://github.com/urho3d/Urho3D/blob/generate-bindings-new/Source/Urho3D/AngelScript/Generated_Members_P.cpp#L4310

glhrmfrts:
okay, I'll close this one then

PR #2685, opened 2020-10-19
Logging speedup

UkoeHB:
Speeds up logging by only constructing formatted messages when the log level justifies it. Previously all formatted log macros were calling `Urho3D::ToString()` regardless of log level, which is expensive. This patch completely solves performance issues that I had encountered in a test case where trace log messages were being created needlessly.

Also reduces duplication in the logging macros by adding two new macros that take log level as input instead of putting it in the macro name.
```
URHO3D_LOG(level, message)
URHO3D_LOGF(level, format, ...)
```

edit: I am not sure if this is thread safe... anyone have an opinion/idea?

### DIFF
--- a/Source/Urho3D/IO/Log.cpp
+++ b/Source/Urho3D/IO/Log.cpp
@@ -22,6 +22,7 @@
 
 #include "../Precompiled.h"
 
+#include "../Container/Str.h"
 #include "../Core/Context.h"
 #include "../Core/CoreEvents.h"
 #include "../Core/ProcessUtils.h"
@@ -133,6 +134,21 @@ void Log::SetTimeStamp(bool enable)
 void Log::SetQuiet(bool quiet)
 {
     quiet_ = quiet;
+}
+
+void Log::WriteFormat(int level, const char* format, ...)
+{
+    if (!logInstance || logInstance->level_ > level)
+        return;
+
+    // Forward to normal Write() after formatting the input
+    String message;
+    va_list args;
+    va_start(args, format);
+    message.AppendWithFormatArgs(format, args);
+    va_end(args);
+
+    Write(level, message);
 }
 
 void Log::Write(int level, const String& message)
--- a/Source/Urho3D/IO/Log.h
+++ b/Source/Urho3D/IO/Log.h
@@ -105,6 +105,8 @@ public:
 
     /// Write to the log. If logging level is higher than the level of the message, the message is ignored.
     static void Write(int level, const String& message);
+    /// Write formatted message to the log. If logging level is higher than the level of the message, the message is ignored.
+    static void WriteFormat(int level, const char* format, ...);
     /// Write raw output to the log.
     static void WriteRaw(const String& message, bool error = false);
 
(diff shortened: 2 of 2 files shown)


UkoeHB:
Moved the `GetLevelStatic()` declaration to follow [code conventions](https://urho3d.github.io/documentation/HEAD/_coding_conventions.html) around order of static members.

UkoeHB:
Um, stale huh. Any reason this can't be merged?

1vanK:
It looks like some kind of synthetic test. How the source code should look for this to have a performance impact? Have you tried the release build of the application. Because the compiler optimizer has to deal with such things. We cannot invent such clever ways for every function in the engine.

UkoeHB:
Um, whenever you have a formatted log message at any log level then `Urho3D::ToString()` is called every time the line is encountered (whether or not you want it to be called). This is very expensive for hot paths, and unnecessary.

1vanK:
> It looks like some kind of synthetic test. How the source code should look for this to have a performance impact?

How many times per frame do you call this function? You understand that every frame is called batch sorting, rendering, physics processing, game logic and other super heavy things. What percentage of overhead does it have?

> Don't pay for what you don't use...
https://www.oreilly.com/library/view/c-coding-standards/0321113586/ch09.html

In any case I think `static void WriteFormat(int level, const char* formatString, ...)` looks less hackly

UkoeHB:
Is it acceptable to add another ToString definition that accepts `va_list`? You can't [pass an ellipsis to an ellipsis](https://stackoverflow.com/questions/695982/passing-an-ellipsis-to-another-variadic-function).

1vanK:
Show me code and performance measurement

1vanK:
> Is it acceptable to add another ToString() definition that accepts va_list?

ToString() is just wrapper for `String::AppendWithFormatArgs(const char* formatString, va_list args)`

Copy content of ToString() to new function

1vanK:
Also it is strange that `va_list args` passed as value. Should it be passed as ref?

UkoeHB:
Re: performance. I am building a settings manager extension that is capable of handling key bindings (and other things, with a generic API). To do this requires a central `HandleSettingsEvent()` function that filters out arbitrary events to find which ones we want to call the appropriate handler for (according to the users' settings, etc.). It can potentially be a very hot path depending how the extension is used (which is impossible to estimate in advance), so I want to optimize it as much as possible. This function has a `LOGTRACEF()` in it, which I found made the function 2x (or higher) slower due to `ToString()` being called all the time. There is simply no reason logging should be a 'gotcha' for performance.

1vanK:
You don't need a wrapper around the wrapper.

UkoeHB:
Maybe not, but that's how I did it, and I'm not going to redo everything just because of logging.

1vanK:
I mean you do not need `String ToString(const char* formatString, va_list args)`

UkoeHB:
Ah my mistake, updated

Source/Urho3D/IO/Log.cpp
```diff
@@ -135,6 +136,21 @@ void Log::SetQuiet(bool quiet)
     quiet_ = quiet;
 }
 
+void Log::WriteFormat(int level, const char* format, ...)
+{
+    if (!logInstance || logInstance->level_ > level)
+        return;
+
+    // forward to normal write after formatting the input
```
1vanK, line 144: With a capital letter

UkoeHB: are you referring to the comment? i.e. Write vs write?

Source/Urho3D/IO/Log.h
```diff
@@ -131,31 +133,27 @@ class URHO3D_API Log : public Object
...
+#define URHO3D_LOG(level, message) Urho3D::Log::Write(level, message)
+#define URHO3D_LOGF(level, format, ...) Urho3D::Log::WriteFormat(level, format, ##__VA_ARGS__)
 #define URHO3D_LOGRAW(message) Urho3D::Log::WriteRaw(message)
-#define URHO3D_LOGTRACEF(format, ...) Urho3D::Log::Write(Urho3D::LOG_TRACE, Urho3D::ToString(format, ##__VA_ARGS__))
-#define URHO3D_LOGDEBUGF(format, ...) Urho3D::Log::Write(Urho3D::LOG_DEBUG, Urho3D::ToString(format, ##__VA_ARGS__))
-#define URHO3D_LOGINFOF(format, ...) Urho3D::Log::Write(Urho3D::LOG_INFO, Urho3D::ToString(format, ##__VA_ARGS__))
-#define URHO3D_LOGWARNINGF(format, ...) Urho3D::Log::Write(Urho3D::LOG_WARNING, Urho3D::ToString(format, ##__VA_ARGS__))
-#define URHO3D_LOGERRORF(format, ...) Urho3D::Log::Write(Urho3D::LOG_ERROR, Urho3D::ToString(format, ##__VA_ARGS__))
-#define URHO3D_LOGRAWF(format, ...) Urho3D::Log::WriteRaw(Urho3D::ToString(format, ##__VA_ARGS__))
+#define URHO3D_LOGRAWF(format, ...) URHO3D_LOGRAW(Urho3D::ToString(format, ##__VA_ARGS__))
```
1vanK, line 139: This have same issue?

Source/Urho3D/IO/Log.cpp
1vanK: forward -> Forward

1vanK:
Also simplify defines (return old variant). Defines that converted to defines illogical

Source/Urho3D/IO/Log.h
UkoeHB: Ah I didn't check the raw function. It doesn't need special treatment

UkoeHB:
This method seems more straightforward and involves much less duplication. Is it really a problem?

1vanK:
Man, we're in the middle of optimizing a function call. We have to go to the end. Let's also optimize the preprocessor.

UkoeHB:
Have it your way I guess

1vanK:
Have you tested this? Can I merge without testing.

UkoeHB:
Built and tested!

PR #2686, opened 2020-10-19
Silence compiler warnings

UkoeHB:
Silences various compiler warnings... missing enums in switch statement, esoteric associativity, intention safety. There are a few other warnings but these are the low hanging fruit.

### DIFF
--- a/Source/Urho3D/UI/ToolTip.cpp
+++ b/Source/Urho3D/UI/ToolTip.cpp
@@ -74,10 +74,15 @@ void ToolTip::Update(float timeStep)
             SharedPtr<UIElement> target = it->Lock();
             if (!target)
                 it = altTargets_.Erase(it);
-            else if (hovering = target->IsHovering() && target->IsVisibleEffective())
-                break;
             else
-                ++it;
+            {
+                hovering = target->IsHovering() && target->IsVisibleEffective();
+
+                if (hovering)
+                    break;
+                else
+                    ++it;
+            }
         }
     }
 
--- a/Source/Urho3D/Graphics/CustomGeometry.cpp
+++ b/Source/Urho3D/Graphics/CustomGeometry.cpp
@@ -321,7 +321,7 @@ void CustomGeometry::FillShape(const Vector<Vector3>& shapeList, bool connectTai
         auto nextIt = it;
         while (it != shapeList.End()) {
             nextIt = it + 1;
-            if (connectTail && nextIt == shapeList.End() || nextIt != shapeList.End())
+            if ((connectTail && nextIt == shapeList.End()) || nextIt != shapeList.End())
             {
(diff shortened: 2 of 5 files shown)


Source/Urho3D/UI/ToolTip.cpp
```diff
@@ -74,7 +74,7 @@ void ToolTip::Update(float timeStep)
             SharedPtr<UIElement> target = it->Lock();
             if (!target)
                 it = altTargets_.Erase(it);
-            else if (hovering = target->IsHovering() && target->IsVisibleEffective())
+            else if ((hovering = target->IsHovering() && target->IsVisibleEffective()))
```
1vanK, line 77: I don't understand what the extra parentheses around are changing

UkoeHB: See [this reference](https://stackoverflow.com/questions/4369446/using-the-result-of-an-assignment-as-a-condition-without-parentheses)

1vanK: Thanks

eugeneko: Both options are fine for me, but consider actually fixing `if` instead of shutting compiler warning with extra braces.
E.g. `else if (hovering = x) {` -> `else { hovering = x; if (hovering) {`
IMO it is bad practice to mix actions and conditions in single line for sake of shorter code.

UkoeHB: Good point @eugeneko, updated patch

Source/Urho3D/UI/ToolTip.cpp
```diff
@@ -74,10 +74,15 @@ void ToolTip::Update(float timeStep)
...
             if (!target)
                 it = altTargets_.Erase(it);
-            else if (hovering = target->IsHovering() && target->IsVisibleEffective())
-                break;
             else
-                ++it;
+            {
+                hovering = target->IsHovering()
+
+                if (hovering && target->IsVisibleEffective())
```
eugeneko, line 81: Bug injected here.

Before: `hovering = target->IsHovering() && target->IsVisibleEffective()`
After: `hovering = target->IsHovering()`

1vanK: it won't compile without `;`

UkoeHB: whoops

eugeneko: See, this is exactly the reason why I think old code is evil and has to be changed.

UkoeHB: Yeah I definitely got distracted by the syntax...

PR #2687, opened 2020-10-19
Fix build bugs for OSX and when dot dependency is missing

UkoeHB:
Fix bug in third party library that prevents the engine from building on El Capitan. [edit: [context](https://discourse.libsdl.org/t/sdl-2-0-10-old-mac-os-el-capitan-10-11-and-xcode7-3/26554/7) and [the patch](https://gist.githubusercontent.com/illume/ac2eb4daa1bd805255d5355806acf822/raw/7e3c83f760fe5bf61a9f6bcc196ed9be6f6e22da/sdl2-mac-10-11.patch)]

Also updates the gitignore for some temporary files that aren't consumed if the 'dot' dependency is missing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -116,3 +116,7 @@ configure*
 *.bak
 Thumbs.db
 .directory
+
+# Doxygen mess from absence of 'dot' dependency
+CMakeDoxyFile.in
+CMakeDoxygenDefaults.cmake
--- a/Source/ThirdParty/SDL/src/video/cocoa/SDL_cocoavideo.h
+++ b/Source/ThirdParty/SDL/src/video/cocoa/SDL_cocoavideo.h
@@ -18,6 +18,9 @@
      misrepresented as being the original software.
   3. This notice may not be removed or altered from any source distribution.
 */
+
+// Modified by UkoeHB for Urho3D
+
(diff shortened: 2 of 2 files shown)


UkoeHB:
Any reason this can't be merged?

1vanK:
If you make changes in third party libraries, you must write a comment at the beginning of file and beside changed line in order to don't lose those changes when update the library (search Urho3D in third party libraries for template)

UkoeHB:
I searched and found nothing. Would you mind giving some more hints?

1vanK:
`Urho3D\Source\ThirdParty\Bullet\src\BulletDynamics\ConstraintSolver\btSequentialImpulseConstraintSolver.cpp`

1vanK:
No ideas why new .gitignore breaks doc generation

UkoeHB:
@weitjong any ideas?

1vanK:
Perhaps the latest version of a environment is installed in a virtual machine and there have been some changes.

1vanK:
Ok I rerun previously successful old commit https://github.com/urho3d/Urho3D/runs/1455639197 and it is fail now. Some changes in evironment is source of the problem.

PR #2688, opened 2020-10-19
E2K: added use of x86 SIMD instruction set on MCST Elbrus 2000 CPU architecture

r-a-sattarov:
MCST e2k (Elbrus 2000) architecture has half native / half software support of most Intel/AMD SIMD
e.g. MMX/SSE/SSE2/SSE3/SSSE3/SSE4.1/SSE4.2/AES/AVX/AVX2 & 3DNow!/SSE4a/XOP/FMA4


weitjong:
Thanks!

PR #2689, opened 2020-10-20
Sprite: basic animation support

moneromooo:


Source/Urho3D/UI/Sprite.cpp
```diff
@@ -206,8 +225,16 @@ void Sprite::SetImageRect(const IntRect& rect)
...
-    if (texture_)
-        SetImageRect(IntRect(0, 0, texture_->GetWidth(), texture_->GetHeight()));
+    if (!textures_.Empty())
+    {
+        int width = 0, height = 0;
+        for (const auto &texture: textures_)
+        {
+            width = std::max(width, texture.first_->GetWidth());
+            height = std::max(height, texture.first_->GetHeight());
+            SetImageRect(IntRect(0, 0, width, height));
```
UkoeHB, line 235: Seems like `SetImageRect` should be outside the loop

Source/Urho3D/UI/Sprite.h
```diff
@@ -117,13 +125,16 @@ class URHO3D_API Sprite : public UIElement
     /// Rotation angle.
     float rotation_;
     /// Texture.
-    SharedPtr<Texture> texture_;
+    Vector<Pair<SharedPtr<Texture>, unsigned int>> textures_;
```
UkoeHB, line 128: Comment out of date

Source/Urho3D/UI/Sprite.h
```diff
@@ -72,6 +72,8 @@ class URHO3D_API Sprite : public UIElement
     void SetRotation(float angle);
     /// Set texture.
     void SetTexture(Texture* texture);
+    /// Set animation.
+    void SetAnimation(const Vector<Pair<SharedPtr<Texture>, unsigned int>> &textures);
```
UkoeHB, line 76: Maybe clarify the comment here about what the first parameter is. Seems like a set of frames, where the second part of the pair indicates approximately how long (in milliseconds?) the frame should be active before the next frame can appear.

Source/Urho3D/UI/Sprite.cpp
UkoeHB: Also maybe clarify intent
```
// set sprite image rectangle so it fits all frame textures
```

Source/Urho3D/UI/Sprite.cpp
```diff
@@ -289,7 +316,38 @@ void Sprite::SetTextureAttr(const ResourceRef& value)
...
+
+    unsigned ms = timer.GetMSec(false);
+    unsigned t = ms % total_time;
+    frame_ = 0;
+    for (const auto &texture: textures_)
+    {
+        if (t < texture.second_)
+            break;
+        t -= texture.second_;
+        ++frame_;
```
UkoeHB, line 349: Clarify intent, it's hard to understand this at first glance
```
// get the total time it takes to cycle through all frames
unsigned total_time = 0;
...
// if frames don't have timing information, then we just iterate through them with every update
if (total_time == 0)
...
// how far into this 'animation cycle' we are, based on time elapsed
unsigned ms = timer.GetMSec(false);
...
// figure out which animation frame this position in the cycle corresponds to
for (const auto &texture: textures_)
...
```

eugeneko:
I didn't check this code in-depth, but I see one critical issue so far:
This animation cannot be serialized, i.e. it cannot be configured from style file or from UI Editor.

Also, this functionality overlaps a bit with `ValueAnimation` that can animate everything, not just textures.
Can this code do anything that `ValueAnimation` cannot?

moneromooo:
No idea, I did not see ValueAnimation. There seems to be no point in this then.

PR #2691, opened 2020-10-20
DropDownList/ListView: allow selection by typed prefix

moneromooo:


eugeneko:
It's not responsibility of `ListView` to search its content, especially considering that this implementation is not very reliable and it cannot be.
I.e. if I store plain `Text` in `ListView` it works fine, but if I add tiny icons for elements (like Hierarchy view in Editor), it stops working.

User fills the content of `ListView`, user should implement search logic as well. `ListView` just doesn't have enough information to do it properly.

PR #2692, opened 2020-10-20
LineEdit: hide cursor when not editable

moneromooo:


eugeneko:
Explain this change. Why hide cursor for non-editable text element?
User still may want to copy contents.

moneromooo:
Because a cursor means input, no ? The patch changes the display only, it does not disable copying.

UkoeHB (approved):

UkoeHB:
The cursor still 'exists', it just never gets shown. If it didn't exist between blinks the code would be super broken.

eugeneko:
> Because a cursor means input, no ?

Nope. The cursor means current position for interaction, including selection and copying.
Editability is irrelevant.
image

moneromooo:
That's hardly good evidence, but hey, it's your software, so you decide.

eugeneko:
> That's hardly good evidence

I don't know if it's better to have cursor or not to have.
However, there should be a good reason (e.g. bug in code) to change _any_ kind of existing functionality.
And this specific case is not a bug.

moneromooo:
Fair. I guess it's subjective, and maybe down to whatever software one is used to,

PR #2693, opened 2020-10-20
core: add VariantMapVector to the variant types

moneromooo:


eugeneko:
Make sure that Variant of this new type is properly processed by generic facilities of the engine, e.g. Scene/XML/JSON serialization.

PS: Just curious, what kind of your code needs VariantMapVector?

moneromooo:
Internally passing complex structures. Looking at the code, it might take a bit to understand and test, I might just close this one once I've thought about it a bit.

moneromooo:
Not worth the extra work. I originally used it to pass complex data in events, but it was such a pain that I now pass pointers to the C++ data anyway, so it's unlikely it'll be useful to even a few people.

eugeneko:
As a side note, why use VariantMapVector instead of VariantVector of
VariantMap? You can stack VariantMaps and VariantVectors however you want.

вт, 20 окт. 2020 г. в 15:05, moneromooo <notifications@github.com>:

moneromooo:
Very good point! Thanks :)

PR #2694, opened 2020-10-20
Graphics: fix "infinite" loop when asking for an invalid monitor

moneromooo:


eugeneko:
Can you elaborate what exactly is the bug you fixed?
I cannot understand it from diff.

moneromooo:
An infinite loop. -1 (error from SDL_GetNumDisplayModes) cast to unsigned is 0xff...ff, so the loop would run 4 billion times before returning. Not infinite, but close.

rokups:
While change is good, it is a little bit lacking. Explicit error checking above loop would better communicate the fact that error may happen. Without that, if someone added more code and did not realize `SDL_GetNumDisplayModes()` could return an error - loop would be skipped, but following code would execute nevertheless and it may depend on `SDL_GetNumDisplayModes()` succeeding.

PR #2695, opened 2020-10-20
SoundSource: send E_SOUNDFINISHED to the source object too

moneromooo:
It might not have a node


UkoeHB:
If there is no node, wouldn't it crash trying to dereference the pointer?

eugeneko:
> wouldn't it crash trying to dereference the pointer?

It will. So this change doesn't actually change anything.

If you have node, you will get event both before and after.
If you don't have node, you will get a crash both before and after.

moneromooo:
True. I don't quite remember now, I'll look into old commits. It was background music, so it might be the node's parent (ie, not in the scene). I'll comment when I worked it out.

moneromooo:
This is not needed. I was registering to the sound source, but it works if I register to the node instead. It's confusing at start so that's why it did not work originally I think. I remember something about the sound source's node having no node itself, but since it works this patch is not needed.

PR #2696, opened 2020-10-20
PackageTool: fix empty string check in extension check

moneromooo:


1vanK:
Why you prevent checking of empty extension?

eugeneko:
@1vanK apparently the extension list is ""-terminated array.

1vanK:
In this case we can just remove "" from the list

1vanK:
Oh yeah thanks

PR #2697, opened 2020-10-20
PackageTool: add an exclude option

moneromooo:
One may set a number of glob patterns to exclude from the output

### DIFF
--- a/Source/Tools/PackageTool/PackageTool.cpp
+++ b/Source/Tools/PackageTool/PackageTool.cpp
@@ -20,6 +20,7 @@
 // THE SOFTWARE.
 //
 
+#include <regex>
 #include <Urho3D/Core/Context.h>
 #include <Urho3D/Container/ArrayPtr.h>
 #include <Urho3D/Core/ProcessUtils.h>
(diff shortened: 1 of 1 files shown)


Source/Tools/PackageTool/PackageTool.cpp
```diff
@@ -63,12 +64,33 @@ String ignoreExtensions_[] = {
...
 
+static std::string glob_to_regex(const std::string &val)
+{
+    std::string newval;
+
+    bool escape = false;
+    for (char c: val)
+    {
+      if (c == '*')
+        newval += escape ? "*" : ".*";
```
1vanK, line 83: and `escape` stay `true`?

1vanK:
Have you tested the code?

Source/Tools/PackageTool/PackageTool.cpp
```diff
@@ -63,12 +64,33 @@ String ignoreExtensions_[] = {
...
+
+    bool escape = false;
+    for (char c: val)
+    {
+      if (c == '*')
+        newval += escape ? "*" : ".*", escape = false;
+      else if (c == '?')
+        newval += escape ? "?" : ".", escape = false;
+      else if (c == '\\')
+        newval += '\\', escape = !escape;
```
1vanK, line 87: Will `bla\\bla` stay `bla\\bla` ? Or is it by design

moneromooo:
It escapes ? and *, which are the glob characters, and \ so it can be used. There's no canonical way to do that really.

1vanK:
After your change `\*` will be coverted to regex `.*`. Are you sure exactly what you have in mind?

1vanK:
`.` wiil be passed to regex as `.`

1vanK:
In fact, I see no point in checking code that didn't even run.

moneromooo:
Yes, that's a comon character. I don't really want to start escaping everything though. This commit can be dropped if not generic enough.

1vanK:
Common character `.` in regex not same `.` in path. In path dot is dot.

1vanK:
>This commit can be dropped if not generic enough.

The problem is that you're wasting people's time reviewing code that you haven't tested yourself.

P.s. code still incorrect. If a character is escaped, you don't really need to check what the character is, but just copy it.

moneromooo:
OK, fair enough.

PR #2698, opened 2020-10-20
Str: allow move assignment to self

moneromooo:
It's used for instance when shuffling a std::vector of String

### DIFF
--- a/Source/Urho3D/Container/HashBase.h
+++ b/Source/Urho3D/Container/HashBase.h
@@ -117,8 +117,6 @@ public:
     /// Swap with another hash set or map.
     void Swap(HashBase& rhs)
     {
-        if (&rhs == this)
-            return;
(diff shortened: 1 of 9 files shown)


eugeneko:
This is actually good point. Can you revise other move-assignments in Urho and fix them as well if they are broken? Shouldn't be too many.

eugeneko:
Thanks! I will merge it when build passes

1vanK:
Maybe just put the check to `Swap()`?

1vanK:
The question remains, which is more of an impact on performance: additional branching due by `if` or unnecessary copying of a variable.

eugeneko:
I don't believe that this branching have any significant contribuition in any case.

Having said that... probably branching has more impact than redundant swap just because statistically `this == &rhs` happens rarely. So maybe it would be slightly better to keep swaps w/o asserts.

PR #2699, opened 2020-10-20
StaticModel: optionally ignore E_RELOADFINISHED event

moneromooo:
It's much faster with lots of children on a node


eugeneko:
This change conflicts with virtual `SetModel`.
The worst part is that derived classes now have method `SetModel(model, flag)` that calls _base_ class instead of overload.
Is there any reason not to put this new flag is in its own separate setter?

moneromooo:
I don't think there is a reason, no. I can do that if you want the patch after the change.

PR #2700, opened 2020-10-20
Octree performance patches

moneromooo:


eugeneko:
I don't see any immediate issues, but this change feels risky.
Let's wait for others' feedback.

Source/Urho3D/Graphics/Octree.cpp
```diff
@@ -247,7 +247,7 @@ void Octant::GetDrawablesInternal(OctreeQuery& query, bool inside) const
         query.TestDrawables(start, end, inside);
     }
 
-    for (auto child : children_)
+    for (const auto &child : children_)
```
1vanK, line 250: Use [Coding conventions](https://urho3d.github.io/documentation/HEAD/_coding_conventions.html)
> Pointers and references append the * or & symbol to the type without a space in between. For example Drawable* drawable, Serializer& dest

eugeneko: > children_ is array of pointers. There is no optimization

SharedPtr copying is relatively expensive.
I would go either for `const auto&` or explicit `Octant*`

1vanK: Octant* children_[NUM_OCTANTS]{};

1vanK:
The commits are shipped in bulk at such a speed that I do not believe they are tested in any way)

moneromooo:
Weird. I must have confused it with another. I'll remove that patch then.

Source/Urho3D/Graphics/Octree.cpp
eugeneko: Oh, these are plaint pointers. Nasty.
Yep, in this case there's no optimization.

moneromooo:
Tested just for my own needs really. If there's any you don't like, it's not like you have to take them, they're here if you want them only.

PR #2701, opened 2020-10-20
Application: allow overriding the error dialog name

moneromooo:


1vanK:
This will cause confusion. When I see Application::GetTitle() I will expect this to be the title for the main window, but not for the dialog with an error message

PR #2702, opened 2020-10-20
Ray: fix wrong normal on HitDistance

moneromooo:
It could leak the normal from a hit behind the closest hit

### DIFF
--- a/Source/Urho3D/Math/Ray.cpp
+++ b/Source/Urho3D/Math/Ray.cpp
@@ -251,17 +251,20 @@ float Ray::HitDistance(const void* vertexData, unsigned vertexStride, unsigned v
     unsigned index = 0, nearestIdx = M_MAX_UNSIGNED;
     Vector3 barycentric;
     Vector3* outBary = outUV ? &barycentric : nullptr;
+    Vector3 tmpOutNormal, *tmpOutNormalPointer = outNormal ? &tmpOutNormal : nullptr;
 
     while (index + 2 < vertexCount)
     {
         const Vector3& v0 = *((const Vector3*)(&vertices[index * vertexStride]));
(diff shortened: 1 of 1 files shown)


Source/Urho3D/Math/Ray.cpp
```diff
@@ -251,17 +251,20 @@ float Ray::HitDistance(const void* vertexData, unsigned vertexStride, unsigned v
...
         const Vector3& v1 = *((const Vector3*)(&vertices[(index + 1) * vertexStride]));
         const Vector3& v2 = *((const Vector3*)(&vertices[(index + 2) * vertexStride]));
-        float distance = HitDistance(v0, v1, v2, outNormal, outBary);
+        float distance = HitDistance(v0, v1, v2, tmpOutNormalPointer, outBary);
         if (distance < nearest)
         {
             nearestIdx = index;
             nearest = distance;
+            if (outNormal)
+                *outNormal = tmpOutNormal;
```
UkoeHB, line 267: `nearest` is defined as `M_INFINITY` so won't `outNormal` always get defined if it exists (no matter what `distance` is for the first iteration, the conditional always evaluates `true`)? In other words, I can't figure out what behavior this changes.

moneromooo:
It (HitDistance) writes to *outNormal only if distance < nearest. It was writing every time.

moneromooo:
This fixes the case where there's more than one face intersecting the ray.

ChrisDenham:
The changes suggested here are almost the same as I had envisaged to solve this problem, however the changes may be complete. I say this because they do not appear to address the secondary issue, which is that if outUV is specified, it may still use the barycentric value computed for the wrong triangle when interpolating the UV coordinate result.

Just thought it worth mentioning here as I'm looking into the same issue in #2862

PR #2703, opened 2020-10-20
Misc terrain speedups

moneromooo:
This one has many patches, you may want to select the ones you want.


eugeneko:
That's... a lot of code.
I will try to look at it later.

Does it make rendering faster or just terrain creation?
Also, how much faster does it work?

moneromooo:
Terrain creation, I create chunks on the fly. I have no timings, but definitely noticeable when moving around and loading new chunks.

eugeneko:
It may be silly question, but I have to ask. Do you test release or debug builds?

moneromooo:
Release. I did an ASAN build a while ago, too, didn't find anything, though I'm not sure this was before or after the terrain patches.

PR #2704, opened 2020-10-20
Node: support derived components in GetComponent

moneromooo:


eugeneko:
This is breaking API change.
Why do you need it already having `GetDerivedComponent`?

moneromooo:
I don't recall why now. I have a feeling it was a call from urho3d itself, but since it changes the API as you point out, it should not be merged.

PR #2705, opened 2020-10-24
CMake/Modules: fixed SSE4 instruction set detection and use

r-a-sattarov:
Added SSSE3/SSE4.1/SSE4.2 instruction set check and usage.


weitjong:
The issue here is we could not set the target "SIMD level" based on the level we found on the build/host machine.

weitjong:
Perhaps we need another build option to specify the maximum level for the intended target? Or to enhance the detection mechanism to be based on the specified target machine (see URHO3D_DEPLOYMENT_TARGET build option) and not the "native" on the build/host machine, which the current `check_extension` macro does.

PR #2707, opened 2020-10-27
Fix Chinese input problems on Windows and Android platform

wu1274704958:
Fix Input Chinese on Windows will lose the candidate word selection box.
Fix Only input English on Android.
Ignore .vscode

### DIFF
--- a/Source/ThirdParty/SDL/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
+++ b/Source/ThirdParty/SDL/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
@@ -2098,7 +2098,7 @@ class DummyEdit extends View implements View.OnKeyListener {
     public InputConnection onCreateInputConnection(EditorInfo outAttrs) {
         ic = new SDLInputConnection(this, true);
 
-        outAttrs.inputType = InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD;
+        outAttrs.inputType = InputType.TYPE_CLASS_TEXT;//| InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD;
         outAttrs.imeOptions = EditorInfo.IME_FLAG_NO_EXTRACT_UI
                 | EditorInfo.IME_FLAG_NO_FULLSCREEN /* API 11 */;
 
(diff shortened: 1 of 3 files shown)


weitjong:
Thanks for your PR. We inherited those code from SDL library. It is strange SDL would have bug like this. On Linux the IME works fine for me.
IME on Linux

Could you show me the screenshots before and after the change. Thanks.

Source/ThirdParty/SDL/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
```diff
@@ -2098,7 +2098,7 @@ public boolean onKeyPreIme (int keyCode, KeyEvent event) {
     public InputConnection onCreateInputConnection(EditorInfo outAttrs) {
         ic = new SDLInputConnection(this, true);
 
-        outAttrs.inputType = InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD;
+        outAttrs.inputType = InputType.TYPE_CLASS_TEXT;//| InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD;
```
UkoeHB, line 2101: Seems a bit strange to remove anything with the word 'password' in it. What does this change do exactly?

wu1274704958: TYPE_TEXT_VARIATION_VISIBLE_PASSWORD make Android IME disable input wide byte characters.
image

weitjong: Now it makes sense. Could you kindly put the comment you have in the screenshot to the code itself. Prefix the comment with “Urho3D” as a marker. Usually, we also mark at the beginning of the source file to say “Modified for Urho3D by...”

weitjong:
Can you also do the same for the windows change? Screenshot as proof and proper comment. Thanks.

wu1274704958:
The change of Windows platform is not applicable to Windows 10.
It work fine on Windows 7.
The IME candidate list always invisible on Windows 10.

weitjong:
I only have Win7 VM. Vaguely I remember I have tested the IME on Linux and on Win7 too when I enabled the support after one of the SDL upgrade. Still, it is strange nobody has submitted a bug fix to the upstream SDL for this.

PR #2709, opened 2020-10-28
Mobile shaders fixes

gleblebedev:
@1vanK here are mobile fixes.

Also I've added my upgrade for PBR terrain shader. Now it supports up to 5 textures. Default behavior is the same as before. New behavior defined via TERRAINLAYERS* define where * is a number of textures.


gleblebedev:
Thanks! Sorry I didn't come back to this for a while.

PR #2711, opened 2020-10-29
AS Autobindings: Sorting classes by id to make order compiler indepen…

1vanK:
…dent


1vanK:
I tested this PR with VS and MimnGW64 on Windows.

@weitjong can you confirm that the problem is solved on Linux also?

weitjong:
It looks good to me. The newly generated binding files are identical this time. Well done.

Issue #2638, opened 2020-05-29
Some (TTF) fonts fail to display minus sign (-)


Lys0gen:
Specifically a regular minus "-" is not being displayed with Arial unless a huge font size is used (>=64), but the issue exists also many other fonts (and perhaps other characters?).

See https://discourse.urho3d.io/t/ui-not-displaying-minus-sign-with-most-fonts/6163

PR #2712, opened 2020-10-30
Quick fix for https://github.com/urho3d/Urho3D/issues/2638

1vanK:


1vanK:
This fixes the problem with spaces and minuses but kerning table still incorrect. I think incorrect kerning table is the reason for wide spaces of some fonts. Maybe someone will spend more time fixing this problem.

1vanK:
@iainmerrick

Source/Urho3D/UI/FontFaceFreeType.cpp
```diff
@@ -143,25 +141,9 @@ bool FontFaceFreeType::Load(const unsigned char* fontData, unsigned fontDataSize
 
     face_ = face;
 
-    auto numGlyphs = (unsigned)face->num_glyphs;
+    unsigned numGlyphs = (unsigned)face->num_glyphs;
```
eugeneko, line 144: I think the line "before" was automatic linter refactoring.
I wonder if we get red CI due to this change. Or we don't have linter in CI anymore?

1vanK: > we agreed to apply before

I do not remember voting or anything about this

eugeneko: > I do not remember voting or anything about this

It is irrelevant whether you voted or not. This line directly violates current coding standard and therefore should be reverted.

> Use `auto` for verbose, unknown or **redundant** types

@weitjong has explicitly applied this change (`auto` instead of redundant types) for whole codebase using clang-tidy several years ago. What's the purpose of manually reverting this change in single place?

1vanK: > It is irrelevant whether you voted or not. This line directly violates current coding standard and therefore should be reverted.

I also don't remember that there was a vote for the current coding standard. Even Cadaver preferred not to use auto: https://discourse.urho3d.io/t/moving-to-c-11/2788/50

eugeneko: > I also don't remember that there was a vote for the current coding standard

Nope, we didn't have vote, only discussion. I have written this new code standard (as conservative and non-modern as possible --  I don't really like `auto`, as you can see from that thread), there were no objections from other maintainers, and @weitjong enforced this new code standard using clang-tidy in following months.

I'm neither a fan of `auto` nor I hate it, but rules are rules. We have this code standard now, we had it for years, and it should be followed. If you don't like it -- you can start a topic on forums, propose your changes, discuss it, then change code standard, and only _then_ write code accoding to it.

eugeneko: It's not a question of someone's personal perferences in code, it's about following consistent rules regardless of whether you like them or not. For example, I absolutely hate that we don't use `virtual` for `override`-ed methods. Yet, I abide. And this is not even required in code standard!

1vanK: > I don't really like auto

then I have only one question, if everyone is against, why do we use it?

eugeneko: > then I have single question, if everyone is against, why do we use it?

Well, me & you & a couple of other ppl != everyone.
You know, I remember that some people proposed "almost always auto" paradigm.

I tried to make conservative rules that will encourage ppl to use `auto` in places where it's benefical for code.
I think it would be silly to switch to C++11 and keep 30-symbol-long iterator types when we have `auto` at hand.

Anyway, this is not a discussion for PR. If you want to change current code standard, make a proposal and let's discuss it on forums -- like I did years ago. Until it's changed, all PRs must follow it.

eugeneko: Also, it's not just about following rules.

This stuff with `auto` was automatic change.
I'm not 100% sure that this line would pass CI.
You may need to add `// NOLINT(modernize-use-auto)` after this line. Or maybe not, I don't know much about current CI.

1vanK: > make a proposal and let's discuss it on forums 

This has already been done, the problem is that we often target some developers who have left / are not interested.

https://discourse.urho3d.io/t/minimizing-usage-of-auto-in-sources/6449

weitjong: Here is the thing about the C++ standard. It seems it always promote heated debate among the developers. I don't want to get into one again here. We have spent too much time to talk about it already and nothing get done. At one point I decided to let the tool (clang-tidy) to take the first shot at modernizing our code base. It turned out not to be easy also. There are also dozens of decision to make on which checks to get enforced and which settings to flip. I was not sure I am the right person to do it but someone has to do it to end all the debates. I did it not because I want it. Each developer has tons of reasons that he/she is right, personal taste aside. But at the very least, I can argue that the default settings that I use came from the bright people from JetBrains who develop IntelliJ and CLion IDE. In respect to the "auto", the modernize-use-auto check is enabled. This replaces complex type in template and iterator use case with auto as we want it, unfortunately, it also replaces simple one too if it sees the surrounding context already make the type obvious to the reader (like in the case above). So, do you guys want to disable the check just because of this? Probably not. If you guys not happy about it, you can try to fine tune its setting if it is possible at all OR write your own custom check then substitute the check with yours. If you are able to do the latter (cause I cannot) then do something about it OR just shut up and move on.

The new CI/CD using GitHub Actions is actually not yet ready to be merged. It has not reached feature parity (although very close already) with the old scripts using Travis-CI. I am forced to merge it as is, after what have happened in the forum recently. But to answer your question, the clang-tidy job is still there (even better), but I have intentionally let the test always passed because after (re-)syncing the .clang-tidy configuration file with latest from CLion and I haven't applied the bulk update nor that I got the time to fine tune the config. My primary goal was to get the CI/CD jobs migrated first. But, like I said, I have no plan continue the work anymore for now.

eugeneko: @weitjong Thanks for clarification!
I think that current CI linter is as good as it can reasonably be, and there's no reason to fight against it.

1vanK: > If you guys not happy about it, you can try to fine tune its setting if it is possible at all OR write your own custom check then substitute the check with yours. If you are able to do the latter (cause I cannot) then do something about it OR just shut up and move on.

I will try to study this issue

1vanK: I still do not understand what the problem is. Automatic placing `auto` can be disbled by `-modernize-use-auto`. It is enough just to add `auto` manually only where it is really needed. There are few such places, because we do not use std. Even not all iterators need auto: https://github.com/urho3d/Urho3D/pull/2700

eugeneko: > Automatic placing auto can be disbled by -modernize-use-auto. It is enough just to add auto manually only where it is really needed

clang-tidy was _not_ designed by idiots and `modernize-use-auto` exists there for a reason.
In my opinion, it works reasonably good with current settings.
Why should we disable `-modernize-use-auto`? So far it did more good than bad.

1vanK: clang-tidy was designed for using with std

1vanK: when i see
```for (std::vector<std::map<int, std::string>>::const_iterator it = container.begin(); it != container.end(); ++it)```
i really want write `auto auto = auto { auto auto; kill c++ designers }` but this is a specific case

1vanK: > modernize-use-auto exists there for a reason

following your logic, the ability to disable this also exists for a reason

eugeneko: > clang-tidy was designed for using with std

What makes you think so? Most of lint rules have nothing to do with `std::*`. They worked just fine with Urho custom containers, probably due to some tinkering on WeiTjong side.

In this specific case, I like that linter is getting rid of copypaste in type name and type cast.
I obviously cannot make you like it: it's about personal preferences.
But "less copypaste makes code better" is just as valid opinion as yours.

I could have made some arguments about `unsigned x = (unsigned)y` being more error-prone than `auto x = (unsigned)y`... But I don't think it really matters. "code error-proneness" and "code readability" are way too vague criterias.

1vanK: Without auto to understand a type of a variable (except iterators) I don't need to read the whole line, and I can read a source code fluently

eugeneko: Without `auto` you have more chances to make a mistake if you decide to change variable type for whatever reason.
I don't know what's _better_, to be honest. I stick with `auto`-way because it's how Urho does things for years.

PS. I don't quite understand what you expect from this discussion.
There is codestyle (partially enforced with linter). Every PR must follow it. If code violates codestyle and/or linter, it must be fixed.

1vanK: > Without auto you have more chances to make a mistake if you decide to change variable type for whatever reason.

Reading code is an order of magnitude more frequent than writing. Changing the type of a variable is even rarer.

1vanK: In fact, I do not see people with whom I can argue about this. I appreciate the huge contribution of weitjong, but he is developing the build system and does not write C++ code. Eugene left to another project and continues to dictate terms in our project, which looks like sabotage. The activity of other developers is low. Who should I coordinate my actions with?

1vanK: Just look it 
https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/Audio/Audio.cpp#L209
https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/Audio/Audio.cpp#L166
https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/Audio/Audio.cpp#L249
there is places where is real need `auto`. This does not interest anyone. But Eugene blocks my PR.

Also Eugene knows about some bugs because he participates in their correction in rbfx but not but he doesn't care that the same bugs remain in Urho3D:
https://github.com/rokups/rbfx/commit/65a7f6ac3f079b706c4c3df417a5c1953d8bac9b
https://github.com/rokups/rbfx/commit/65a7f6ac3f079b706c4c3df417a5c1953d8bac9b

eugeneko: > Eugene left to another project and continues to dictate terms in our project, which looks like sabotage

I work with Urho3D codebase more than 99% of Urho users.
Are you saying that I should not care about code I work with on daily basis?

> But Eugene blocks my PR

I don't. You block your own PR by not following code style enforced by CI.
For example, I write the same kind of issues to @ArnisLielturks when he puts `{`-braces on wrong line.

Do you really blame _me_ for _your_ violation?

weitjong: Yes, the feeling is equal but extraordinary claim requires extraordinary proof. Some people just wish this project to go down.

The clang-tidy checks will not do EXACTLY what you want, unless you spent sufficient time to customize it. We didn't do that (yet). Just enable/disable checks will not achieve the desired result. But here is the thing you need to consider, without the tools (clang-tidy and clang-format) in the CI/CD, then you will be relying on another human person to catch the offending codes. Do you think you can dismiss the tools and find someone to perform that mundane task? And, potentially each time end up in another heated argument for something as simple as to-auto-or-not-to-auto.

As I said earlier, I am not the right person for this C++ standard discussion. Earlier when I said, "shut up", I was not really looking at you. But let fix thing one at a time. You can revisit this again when you have a better plan and/or better person to discuss it with.

eugeneko: I relocated this discussion on forum:
https://discourse.urho3d.io/t/minimizing-usage-of-auto-in-sources/6449/4

1vanK: > The clang-tidy checks will not do EXACTLY what you want, unless you spent sufficient time to customize it. 

We can just turn off check of `auto`, because it just doesn't work as seen in the examples.

> Do you think you can dismiss the tools and find someone to perform that mundane task? 

In any case, we check commits before merging. It is very easy to see unnecessary auto that do not significantly shorten the code.

weitjong: I would argue the result is acceptable. The modernize-use-auto only replaces the type to auto when the context is clear. Not based on how many letters it saved from typing. If it would replace in the case you sighted then the result would be “auto it = Obj.Find()”, which is bad use case for auto, at least to me.

1vanK: The type is clear from the context. Would you agree that discovering of a type from context requires extra effort instead of just reading the type? And this, on the contrary, slows down the study of the source code. The brain should be distracted by extraneous things, instead of looking for logical errors in the algorithm.

1vanK: For beginners who have just started learning the engine code, this is much more serious https://discourse.urho3d.io/t/minimizing-usage-of-auto-in-sources/6449/2

1vanK: Of course, you can infer the type from the context, but why do that?

weitjong: I think I have said enough about this manner. At the beginning I already said, somehow C++ Standard just promotes this kind of arguments. Now you have opened a can of worm.

1vanK: I don't argue that `auto` is very useful for iterators. I am not saying to get rid of `auto`.

1vanK: My last argument `auto* ui = GetSubsystem<UI>();`. If this is not an argument, then I cannot prove anything else.

weitjong: When the check sees the type is already clear in the context, it replaces with “auto” to avoid duplication. It does not matter if “auto” is actually longer than “UI” because the length is not its consideration.

1vanK: That's what I'm talking about

UkoeHB:
Was this really closed over such an inconsequential disagreement as coding standards?

1vanK:
Just wait a bug fix in the rbfx engine.

Source/Urho3D/UI/FontFaceFreeType.cpp
weitjong: If you refer back to one of my past comment. I actually said that is the "unfortunate" side of the modernize-use-auto check. Again, if you can find a way to customize it to eliminate this "drawback" then you have my blessing to go ahead. Disable the check outright has only one problem which I already pointed out. You will rely on another person to point out whether the code should have use "auto" when it is not; and vice versa. Believe me, the discussion in the new PR will soon get ugly. This PR is already a proof the case.

I am really sick by the "auto" discussions from the C++ community. You ask 10 persons, probably you would get 11 answers back.

1vanK: Disputes can arise on any ground, even over variable names.

PR #2714, opened 2020-11-01
Update Bullet to 3.06+

1vanK:


eugeneko:
As a side note, it would be wise to wait for green master before merging such big PRs.

weitjong:
OK. The automated CI/CD workflow also does not replace the normal regression test. However, Bullet upgrade usually does not cause unexpected issues like SDL upgrade. Personally the regression tests were what made me slow to push even when I have already the library upgraded successfully. I have the Bullet lib upgraded in my personal fork for months now but I lose interest half way through the testing for introducing/integrating new build option to enable OpenMP support.

There is easy and hard way to upgrade the underlying libraries. The hard way is to study what has changed and what's new, and trying to bring in those new bits to the Urho3D.

1vanK:
What's going on with github? It just shuffles the comments in a random order xD

weitjong:
@eugeneko I don't think you need to state the obvious. Also, I will give room to the active member, to make mistake and to learn.

1vanK:
I don't think it matters much, because in the changed places the code has no difference from the old version.

1vanK:
I restarted the tasks, but I get the feeling that the last commits from the main branch are not used.

weitjong:
The CI/CD workflow restart will just recover from technical errors. It will use the same commit SHA that triggered the initial workflow. You need to make another push if you want to change the result of the workflow.

Good work by the way in pushing the upgrade this fast.

1vanK:
Is there a way to restart a specific task?

weitjong:
No, it does not support of restarting specific "job", at least for now.

weitjong:
Just rerun the whole workflow. More than half the jobs are speed up by cache. Those will finish very fast.

weitjong:
Let me know if you need my help to troubleshoot. From the look of it, I suspect it was caused disk space issue. MinGW debug build and Android build takes so much disk space, they actually has used all the free disk space provided by the runner VM. I only managed to get it working again by "hacking" into the VM and removing one or two software components (like .net) that I don't need. It works without any issue since. But I recall the last time I have disk space issue I had the similar unexplained build error like you are facing now. I haven't got time to check out the new commit for Bullet upgrade yet, but I would double check if there is any mistake pulling more than what we need in from Bullet.

1vanK:
I have checked several times. I downloaded the old original version of Bullet library, compared it with the version from Urho and moved all changes to the new version. I also kept the original code as a commentary for easy future updates.

https://github.com/urho3d/Urho3D/blob/master/Source/ThirdParty/Bullet/src/LinearMath/btScalar.h#L77-L81

A also test VS and MinGW builds on my pc before PR

1vanK:
Can we use GLOB RECURSIVE

https://github.com/rokups/Urho3D-Upstream/blob/rk/bullet-update/Source/ThirdParty/Bullet/CMakeLists.txt#L32-L33

or does this impose version restrictions?

weitjong:
Yes, why not. Or just "cherry pick" the new CMakeLists.txt from my fork.

https://github.com/weitjong/Urho3D/blob/master/Source/ThirdParty/Bullet/CMakeLists.txt

1vanK:
Can incorrect building be related with cache?

weitjong:
It is possible, but unusually. You see, if the build failed then the build cache is not even saved. In other words, the saved build cache is coming from a previously successfully build or there is no cache if it has been invalidated. The faulty cache, if it unfortunately happened, would cause weird compilation error too. You can purposely invalidate all the build cache by adding a keyword "[cache clear]" somewhere in your commit message, if you really doubt the currently saved cache.

weitjong:
I got time to quickly glanced through the last commit in the main branch using a visual diff tool and I didn't see anything suspicious. For now may be we can rule out the problem is with the commit. It means the increase in the disk space utilization is unavoidable. And if so, I/we need to find a way to hack for more space.

weitjong:
Can you do another commit with "[cache clear]" to see if that is enough to make up the space. If not then after office I will try to remove some more unused software components from the runner VM to free up the space. I can't believe we have hit the envelope  in the GitHub Actions so fast :smiley:.

weitjong:
Hmm. I cannot even build cleanly locally for Android build and Web build. But let me fix the '[cache clear]' issue first.

weitjong:
It seems the magic keyword works for me. Some how GitHub did not see the keyword in the commit message. When the message has keyword then it works as as intended.
cache clear  works for me

So, now let me try to clean all cache and retry your commit in the other PR.

weitjong:
> Hmm. I cannot even build cleanly locally for Android build and Web build. But let me fix the '[cache clear]' issue first.

Build cleanly when using the modified CMakeLists.txt in the other PR. Tested the physic stress sample on Android emulator too. Web build also OK. I think can you merge that other PR regardless of what CI/CD said.

PR #2716, opened 2020-11-01
Simplify Bullet/CMakeLists.txt

1vanK:


1vanK:
Previous build never successful because linux-clang-format never successful. What version is the cache taken from?

1vanK:
Cache still used. What am I doing wrong?

1vanK:
```
Run actions/cache@v2
  with:
    path: build/cache/.ccache
  build/cache/.gradle/caches
  build/cache/.gradle/wrapper
  
    key: web-static-wasm-dbg-latest-6e1923611200b52c827c4c9f31c1eaaa85d59bc63636ba2db5142dec824fa0ca
  
    restore-keys: web-static-wasm-dbg-latest
  web-static-wasm-dbg
  
  env:
    HOST: linux
    PLATFORM: web
    BUILD_TYPE: dbg
    LIB_TYPE: static
    ARCH: wasm
    DBE_TAG: latest
Received 188743680 of 655720334 (28.8%), 179.8 MBs/sec
Received 436207616 of 655720334 (66.5%), 207.8 MBs/sec
Received 651526030 of 655720334 (99.4%), 207.0 MBs/sec
Received 655720334 of 655720334 (100.0%), 148.3 MBs/sec
Cache Size: ~625 MB (655720334 B)
/bin/tar --use-compress-program zstd -d -xf /home/runner/work/_temp/510572d5-5a8e-4b80-ad4c-1ffe116dc2c3/cache.tzst -P -C /home/runner/work/Urho3D/Urho3D
Cache restored from key: web-static-wasm-dbg-latest-d98110c0cc855a57c0a8ed11f86e40efb7be1938b77c5a97c1ef4ed8f4a42ca1
```

1vanK:
https://stackoverflow.com/questions/63521430/clear-cache-in-github-actions

```
Once you create a cache, you cannot change the contents of an existing 
cache but you can create a new cache with a new key.
```

weitjong:
I am still in "office", so I will be brief. The build cache has a unique scope for each job ID. It is not all or nothing per workflow. 

The magic keyword in the commit message did not do what it programmed to do yet for some reason. Strange that I have tested that before and it was ok. It is another teething problem. So please bare with it. Later I will see what's going for this separate issue too. 

You are right that GH Actions is immutable, but it is immutable per unique key. In the workflow I have not just put a unique key per job ID, but the complete key is also derived from the source code checksum. In short the key is auto renew and so does the cache is auto re-saved as necessary.

GH Actions does not provide a way to clear the cache. The magic keyword logic is our own implementation to workaround the limitation, which actually pre-exist in Travis CI. if it works as it is intended then it actually wipe the cache clean just before the build.

weitjong:
The clang-format error is expected. It is a new addition in the CI. Previously we don't have clang-format check before. But since it is new, I am not totally sure I got the config entirely correct yet. Probably need to fine tune further, but in this particular case I believe the result is desirable. Or may be not, depend on who you ask. Some like space after brace, some do not. If we cannot get into an agreement for this again, then just throw the new job away.

1vanK:
Space between brackets (if you are talking about `>>`) was necessary before C ++ 11, now it is not necessary (Autogenerator generate without space)

1vanK:
If you are about
```
if (...) {
```
vs
```
if (...)
{
```
then we had already established a second style

eugeneko:
I have just checked clang-format log and it looks way too agressive: it couldn't even print all the errors in single file, it just hits the limit of 100 errors.

On the one hand, I would really like to get uniform automatic formatting for whole codebase.
On the other, it would be huge pain in ass to work with such commit in history. Goodbye git blame.

weitjong:
No, I was referring to the the space in the brace initializer.

https://github.com/urho3d/Urho3D/runs/1342093953?check_suite_focus=true#step:9:447
https://github.com/urho3d/Urho3D/runs/1342093953?check_suite_focus=true#step:9:468

PR #2719, opened 2020-11-02
support loading shaders from memory

ArnisLielturks:
Sample code: https://github.com/ArnisLielturks/Urho3D-Tank-Web-IDE-Sample/blob/eeb2c8f932b421667813f6ceee4a192c6c8e86b2/Source/Samples/55_TankPrototype/WebResourceCache.cpp#L295-L307

Allows loading shaders from MemoryBuffer. Without the fix the `#includes` in the loaded shader are not loaded correctly since there is no way to set the name for the memory buffer.


1vanK:
I didn't spend too much time diving into this question, but why not add `fileName_` to `AbstractFile` class

1vanK:
Oh, the deserializer is used there

eugeneko:
Good point. If you add file name member to `MemoryBuffer`, it would be zero-cost unless used. Just a tiny bit of stack overhead.

1vanK:
Just move fileName_ from `File` to `AbstractFile`. This will also add name to `MemoryBuffer`.

ArnisLielturks:
Good idea. This would fix any resources that I haven't yet checked or used this way.

eugeneko:
@1vanK, @ArnisLielturks it sounds like really bad idea.
`AbstractFile` shouldn't have `SetName` because it cannot be implemented properly in all descedants (e.g. `NamedPipe`).
IF you add `AbstractFile::SetName`, what should `NamedPipe::SetName` do?

1vanK:
GetName() SetName() is virtual, just override in Pipe

1vanK:
we can even remove pipeName_ from pipe and use name_ instead fileName_ in `AbstractFile`

eugeneko:
@1vanK It solves nothing. What exactly should these classes do when you call virtual `SetName`?

1vanK:
name_ = name;

1vanK:
Or just override for Pipe with empty function

eugeneko:
@1vanK What's the _benefits_ of having `SetName` in `AbstractFile`, except saving 10 lines of code?
So far I see only disadvantage: user confusion.

If I call `file->SetName`, I expect to see this file with this name on HDD. Same with pipe.
If it doesn't, it's either a bug in code or bug in interface design.

If a class don't support reasonable `SetName`, then it shouldn't inherit interface with `SetName`. This is basic rule of interface design.

BTW, making `NamedAbstractFile` as descedant of `AbstractFile` and inherited by `MemoryFile` and some other "files" also will do.

1vanK:
Why can't an abstract file have a name?

eugeneko:
@1vanK It's a sign of error in interface design.
If derived class cannot implement `SetName`, then it shouldn't have `SetName` in its public API. If it has, it's design error.
It usually means that class hierarchy needs more interfaces.

> for pipe SetName() can be private

It will not make `SetName` actually public.
If you do `AbstractFile& file = GetSomething() /* returns pipe */`, `SetName` will become public without you ever noticing it.
What would happen if you call `file.SetName("foo"); assert(file.GetName() == "foo")` for this pipe?

1vanK:
> What would happen if you call 

nothing if it is an empty function

Source/Urho3D/IO/File.h
```diff
@@ -120,8 +115,6 @@ class URHO3D_API File : public Object, public AbstractFile
     /// Seek in file internally using either C standard IO functions or SDL RWops for Android asset files.
     void SeekInternal(unsigned newPosition);
 
-    /// File name.
-    String fileName_;
```
1vanK, line 124: I think need use name_ because NamedPipe use pipeName_

Source/Urho3D/IO/AbstractFile.cpp
```diff
@@ -0,0 +1,33 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+#include "AbstractFile.h"
+
+namespace Urho3D
+{
+
+    void AbstractFile::SetName(const String& name)
```
1vanK, line 28: just place it to header file `void SetName(const String& name) { name_ = name; }`

eugeneko:
> nothing if it is an empty function

But it would be assert, exactly because it's empty function.
I don't think it's good when you call `SetName` and then `GetName` right after returns something different.

What are your objections against adding one more interface into `AbstractFile.h`?

1vanK:
It seems to me that there is no need to overcomplicate the class hierarchy. Also, I don't understand the argument, what if I call a function on a class that I don't need. Then you don't need to call this function.

eugeneko:
Whoops, I forgot that `File::SetName` is needed for internal `ResourceCache` workings. I take that back.

@1vanK Okay, maybe it's not worth to add new interface just for `SetName`.
Sometimes it hard to ignore dayjob habits of coding for me, so I'm prone to occasinal overengineering.

@ArnisLielturks can you please revisit all the derived classes and add error reporting and/or assert for ones that cannot support `SetName`? NamedPipe at least, maybe someone else?

Source/Urho3D/IO/AbstractFile.cpp
1vanK: forgot about virtual keyword

1vanK: also need override GetName() override { return name_; }

ArnisLielturks:
Sorry, didn't saw your latest discussion regarding this. 

Okay, so there's 4 classes that are derived from AbstractFile:

```
File
MemoryBuffer
NamedPipe
VectorBuffer
```

@eugeneko your idea is to override the `SetName()` method in the ones that does not support it - `VectorBuffer` and `NamedPipe` and output an error if it's used?

1vanK:
There is no need to override functions for VectorBuffer, maybe someone will use the name for some purpose

eugeneko:
`VectorBuffer` should be fine -- it's not that different from `MemoryBuffer`. `SetName` is harmless there.
But `NamedPipe` cannot be renamed easily. If you call `SetName` for it, you would just get pipe with wrong name.

So, I suggest to add error reporting and assert to `NamedPipe`. Also, `SetName` obviously shouldn't change the name in this case.

Source/Urho3D/IO/NamedPipe.h
```diff
@@ -54,8 +54,8 @@ class URHO3D_API NamedPipe : public Object, public AbstractFile
     unsigned Write(const void* data, unsigned size) override;
     /// Return whether pipe has no data available.
     bool IsEof() const override;
-    /// Return the pipe name.
-    const String& GetName() const override { return pipeName_; }
+    /// Sets the name of the pipe
```
1vanK, line 57: dot in end

eugeneko, line 57: Let's explicitly say that it's not supported. This comment is quite confusing: as if we can actually set the name of the pipe.

Source/Urho3D/IO/NamedPipe.cpp
```diff
@@ -109,7 +109,7 @@ bool NamedPipe::Open(const String& pipeName, bool isServer)
         else
         {
             URHO3D_LOGDEBUG("Created named pipe " + pipeName);
-            pipeName_ = pipeName;
+            name_ = pipeName;
```
1vanK, line 112: may be rename also function parameter to equals field name

ArnisLielturks:
@eugeneko @1vanK Thanks for the help! 🚀

1vanK:
If you have doxygen installed you can build with `-DURHO3D_GENERATEBINDINGS=1` to update AS bindings. If not i will do it after merge

ArnisLielturks:
@1vanK I haven't, would be great if you could do it.

1vanK:
it seems you forgot replace all fileName_ to name_ in File.cpp

UkoeHB:
[This](https://stackoverflow.com/questions/24609872/delete-virtual-function-from-a-derived-class) has a compile-time solution classes that shouldn't use an inherited function.

ArnisLielturks:
I believe this is done! @1vanK @eugeneko any objections?

1vanK:
I just noticed that you have replaced `fileName_ = fileName;` to call virtual function `SetName(fileName);`. It seems to me that it is better use `name_ = fileName;`.

ArnisLielturks:
@1vanK ok, will update. Anything else?

1vanK:
I think that's all

PR #2720, opened 2020-11-03
Update AS Autobindings

1vanK:


1vanK:
Can we show free disk space after every build step to control disk space usage?

1vanK:
https://github.com/actions/virtual-environments/issues/709

> you can remove the part of pre-installed software in runtime. For example, this command will release 5+ of free space:
```
sudo rm -rf "/usr/local/share/boost"
sudo rm -rf "$AGENT_TOOLSDIRECTORY"
```

1vanK:
Why need remove `/usr/local/lib/android` when android is used? What's in this directory?

1vanK:
Removing packages takes a long time, but frees up 7 gigabytes

weitjong:
I have checked before the utilization's of those software components and I only keep the last two big fishes. I want to keep the list short because exactly it will unnecessary extra time to do the cleanup more than necessary. It does not make sense to take more than 5m to do the free while the actual build step only take less than a minute with the help from cache.

Android can be deleted because we are using DBE image. All prerequisite for building Urho3D are already in the docker image.

weitjong:
In fact those jobs that use DBE can actually run on a blank runner VM, provided the Docker Engine is installed. Alas, GH Actions does not have such runner option.

weitjong:
It was late last night and I didn't check whether we still have out of space issue or not before commenting. If more space is required then we have no choice to "hack" for more.

weitjong:
Based on my analysis on the log, it should be the out of memory issue instead of disk space for the remaining failed job. I have pushed a small change to increase the JVM size 4GB to see if it will solve the problem or not. It may be not easy to workaround this because the runner VM provided by GH Actions only has 7GB in total. Using self-hosted VM is out of the question for now.

1vanK:
Files `.appveyor.yml` and `.travis.yml` have been deleted. In which file do I need to enable `URHO3D_GENERATEBINDINGS` to test PR?

1vanK:
It seems circular dependency has been resolved in VS but not in MinGW

weitjong:
The CI/CD workflow is configured in the `.github/workflows/main.yml`. The `URHO3D_GENERATEBINDINGS` is already being recognized as one of the valid option in the new build environment, dockerized or not. So, what you need to do is to use the corresponding environment variable to set its value to 1 in the `env` key for each job id. It is as simple as that.

```
    env:
      URHO3D_GENERATEBINDINGS: 1
```

1vanK:
Thanks!

PR #2723, opened 2020-11-05
AS Autobindings: fix circular dependency

1vanK:


1vanK:
@weitjong can you confirm that this helps?

weitjong:
Well done. It works perfectly now.

PR #2726, opened 2020-11-16
Add support of 24-bit depth on mobiles.

eugeneko:
@SuperWangKai

### DIFF
--- a/Source/Urho3D/Graphics/OpenGL/OGLGraphics.cpp
+++ b/Source/Urho3D/Graphics/OpenGL/OGLGraphics.cpp
@@ -361,7 +361,6 @@ bool Graphics::SetScreenMode(int width, int height, const ScreenModeParams& para
         SDL_GL_SetAttribute(SDL_GL_DOUBLEBUFFER, 1);
 
 #ifndef GL_ES_VERSION_2_0
-        SDL_GL_SetAttribute(SDL_GL_DEPTH_SIZE, 24);
         SDL_GL_SetAttribute(SDL_GL_RED_SIZE, 8);
         SDL_GL_SetAttribute(SDL_GL_GREEN_SIZE, 8);
         SDL_GL_SetAttribute(SDL_GL_BLUE_SIZE, 8);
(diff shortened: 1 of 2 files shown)


Source/Urho3D/Graphics/OpenGL/OGLGraphics.cpp
```diff
@@ -423,36 +411,52 @@ bool Graphics::SetScreenMode(int width, int height, const ScreenModeParams& para
...
-                newParams.fullscreen_ = false;
-#endif
-            }
+            SDL_GL_SetAttribute(SDL_GL_DEPTH_SIZE, depthSize);
 
-            if (window_)
-                break;
-            else
+            // Try requested multisample level first, fallback to lower levels and no multisample
+            for (int multiSample = newParams.multiSample_; multiSample > 0; multiSample /= 2)
```
1vanK, line 420: is number of samples always power of two?

1vanK: Any reference in standart? I didn't find anything

eugeneko:
https://docs.microsoft.com/en-us/windows/win32/api/d3d11/ne-d3d11-d3d11_standard_multisample_quality_levels

This is the best doc I could find

Пн, 16 нояб. 2020 г. в 23:47, 1vanK <notifications@github.com>:

1vanK:
We can correct value before loop:

if (!IsPowerOfTwo(samples))
{
    samples = NextPowerOfTwo(samples);
    samples = samples >> 1u; // prev power of two
}

1vanK:
Or this way https://stackoverflow.com/questions/2679815/previous-power-of-2

SuperWangKai:
Pasted the original post link in Urho3D forum so people can get the more information of this PR -
https://discourse.urho3d.io/t/depth-stencil-buffer-issue-on-gles-2-0/6526/10

SuperWangKai:
@1vanK good point. Maybe we should be correct it or the developer should be responsible for the value since the original value is from Engine Parameters.

SuperWangKai:
This PR looks good to me and I have nothing to add.

1vanK:
```
params.multiSample_ = NextPowerOfTwo(Clamp(params.multiSample_, 1, 16));
```
Not sure this is correct. If multiSample_ is 2 it will be increased to 4

1vanK:
Or NextPowerOfTwo not increased value if already correct?

eugeneko:
It looks okay
```
0 -> 0
1 -> 1
2 -> 2
3 -> 4
4 -> 4
5 -> 8
6 -> 8
7 -> 8
8 -> 8
9 -> 16
10 -> 16
11 -> 16
12 -> 16
13 -> 16
14 -> 16
15 -> 16
16 -> 16
```

SuperWangKai:
I'm fine with current implementation. In our case it's even better if we can have kind of  `PreviousPowerOfTwo` and get this result:
~~0 -> 0~~ (will not happen in our case)
1 -> 1
2 -> 2
3 -> 2
4 -> 4
5 -> 4
6 -> 4
7 -> 4
8 -> 8
9 -> 8
10 -> 8
11 -> 8
12 -> 8
13 -> 8
14 -> 8
15 -> 8
16 -> 16

We already have `NextPowerOfTwo` and `ClosestPowerOfTwo`. https://github.com/urho3d/Urho3D/blob/024dc14aad0a545414cbe79e83a8cb5f2b7b7677/Source/Urho3D/Math/MathDefs.h#L280

SuperWangKai:
Something like:

```c++
/// Round down to the previous power of two.
inline unsigned PreviousPowerOfTwo(unsigned value)
{
    if (IsPowerOfTwo(value))
        return value;

    return NextPowerOfTwo(value) >> 1u;
}
```

eugeneko:
@SuperWangKai tbh I don't think `PreviousPowerOfTwo` will make any difference in this particular case.
Thanks for raising this issue and testing solution!

SuperWangKai:
@eugeneko Sure!
I've debugged this PR and found no issue till now.

Issue #2725, opened 2020-11-11
Pasting in LineEdit may exceed SetMaxLength() limit


tril0byte:
SetMaxLength() is a useful method which is correctly preventing typing text that would increase the widget's length over the limit specified. However, I am able to exceed the limit if textCopyable_ is true. 
Steps to reproduce:

1. Create a LineEdit widget
2. Set the Max Length with SetMaxLength()
3. Enter sufficient text into the widget so it's near the limit
4. Select a bunch of the text
5. Press Ctrl-C to copy text to the clipboard
6. Deselect the text (so that it doesn't just replace itself in the next step) by clicking to move the cursor anywhere inside the LineEdit
7. Press Ctrl-V. The length can now be longer then that specified by SetMaxLength().

I would guess the limit should be enforced in the KEY_V case in UI/LineEdit.cpp method LineEdit::OnKey(). https://github.com/urho3d/Urho3D/blob/f909775ca7d61e6291342b33921a22d837ac6b18/Source/Urho3D/UI/LineEdit.cpp#L237
Is there a reason not to implement this? Should anyone depend on the current behavior of being able to paste to increase text above the limit?

tril0byte:
I see a few ways to proceed if this is desired.
1. truncating the string and allowing the paste to possibly go partly through. should the characters be removed from the existing text starting from the end, or from the text being pasted? 
2. blocking the entire paste and triggering an exception or error to allow the caller to handle this

1vanK:
Yes, it looks like truncating the string at the end `case KEY_V:` the easiest way. Can you create PR?

PR #2727, opened 2020-11-17
UI/LineEdit: enforce length limit on paste

tril0byte:
created for https://github.com/urho3d/Urho3D/issues/2725

This is the obvious way and also very conservative. I don't truncate any characters, just do nothing (prevent the paste) if length would be exceeded.  Would be more complex to truncate text. In this patch I have kept things as simple as possible by only adding 1 line, and the behavior meets my use case.


1vanK:
This is incorrect behavior

tril0byte:
can you specify behavior you would like?   You asked for truncating the existing text- but I think it's incorrect to truncate the new text being pasted. I would think if the paste succeeds that the user expects the entire clipboard made it into the line.

1vanK:
I see it like end of  line should be cut

tril0byte:
Thank you for reviewing my request. I will look into it.

tril0byte:
will create new PR + branch for different approach

PR #2728, opened 2020-11-18
Revert PR #2476.

eugeneko:
I was reluctant to do it before, but we have Urho3D CE now. So this code would remain in some form, just not in main repo.

I'm removing a chunk of code that was merged by mistake and should have never been added in first place. Not into main Urho3D repo, at least. I think the reasons why I'm removing this code should be obvious for anyone. However, I will list them below for sake of history.

- `CustomGeometry` responsibility is to let user construct arbitrary geometry. `CustomGeometry` have never been responsible for helping in making geometry itself.

- These helpers don't need to be member functions at all -- they don't use any private members, only public interface. It means that they are basically _user code_ that was added inside `CustomGeometry` for whatever reason.

- Finally, these helpers have very generic names but they are not reusable enough. For example: there's function called `FillShape` which actually means `FillStarShapedPolygonWithCenterPointAtArithmeticAverageOfContourPoints`. Or `MakeSquare`, which in reality means `MakeAxisAlignedSquareInX0ZPlane`. All these helpers generate very specific shapes with almost non-existent customizability.


1vanK:
I agree

1vanK:
Is there a way to update a fork by excluding a specific commit?

PR #2729, opened 2020-11-18
LineEdit: enforce length limit on updating text

tril0byte:
Try #2 for fixing https://github.com/urho3d/Urho3D/issues/2725 in another way.

Here I enforce the length limit in method LineEdit::UpdateText(). This fixes pasting with Ctrl-V without modifying OnKey() at all, and also changes the behavior of SetText() to enforce the limit (previously SetText() would allow setting the line to contain more than the limit, until manually deleting text), as well as on any other functions current and future.

### DIFF
--- a/Source/Urho3D/UI/LineEdit.cpp
+++ b/Source/Urho3D/UI/LineEdit.cpp
@@ -566,8 +566,8 @@ void LineEdit::UpdateText()
 {
     unsigned utf8Length = line_.LengthUTF8();
 
-    // truncate line to enforce length limit
-    if (utf8Length > maxLength_)
+    // if maxLength_ nonzero, truncate line to enforce length
+    if (maxLength_ > 0 && utf8Length > maxLength_)
     {
       line_ = line_.SubstringUTF8(0, maxLength_);
       utf8Length = maxLength_;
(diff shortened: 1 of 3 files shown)


tril0byte:
I need to update this to skip the truncation if maxLength_ is 0

Source/Urho3D/UI/LineEdit.cpp
```diff
@@ -566,6 +566,13 @@ void LineEdit::UpdateText()
 {
     unsigned utf8Length = line_.LengthUTF8();
 
+    // if maxLength_ nonzero, truncate line to enforce length
```
1vanK, line 569: With a capital letter

Source/Urho3D/UI/LineEdit.cpp
```diff
@@ -566,6 +566,13 @@ void LineEdit::UpdateText()
 {
     unsigned utf8Length = line_.LengthUTF8();
 
+    // if maxLength_ nonzero, truncate line to enforce length
+    if (maxLength_ > 0 && utf8Length > maxLength_)
+    {
+      line_ = line_.SubstringUTF8(0, maxLength_);
+      utf8Length = maxLength_;
```
1vanK, line 573: Indent 4 spaces

PR #2736, opened 2020-12-06
Fix failing mingw32 build when there is space char in project path.

henu:
Running these commands

```
git clone https://github.com/urho3d/Urho3D.git "test space char"
cd test\ space\ char/
MINGW_PREFIX=/usr/bin/x86_64-w64-mingw32 script/cmake_mingw.sh .
make
```

gives me these errors:

```
x86_64-w64-mingw32-gcc: error: space: No such file or directory
x86_64-w64-mingw32-gcc: error: char/Source/ThirdParty/SDL/src/hidapi/hidapi: No such file or directory
make[2]: *** [Source/ThirdParty/SDL/CMakeFiles/SDL.dir/build.make:63: Source/ThirdParty/SDL/CMakeFiles/SDL.dir/src/SDL.c.obj] Error 1
make[1]: *** [CMakeFiles/Makefile2:2112: Source/ThirdParty/SDL/CMakeFiles/SDL.dir/all] Error 2
make: *** [Makefile:152: all] Error 2
```

This PR fixes that


1vanK:
Latest SDL also have this fix

https://github.com/spurious/SDL-mirror/blob/master/cmake/sdlchecks.cmake#L1094

https://github.com/spurious/SDL-mirror/commit/67fd29d55cad28b49acffe7a7a2ff81de35a44f3#diff-c81585f0a6d7862c453336c602f6828994fb61a2402f71c7f7e157a7a724d929

henu:
Ah, I didn't even though it would be fixed there. I actually didn't even realized this was third party code, although it says that in the path. :smile:

Should we still fix it in this project? I guess the earlier fix from SDL will eventually end up here too, so I assume there would be no harm fixing it already?

PR #2742, opened 2020-12-11
Fix crash when audio subsystem removed

ChrisDenham:
This PR simply adds a few extra conditions to avoid a crash in Urho3D when either not adding the Audio subsystem, or subsequent to removing it with `context->RemoveSubsystem<Urho3D::Audio>();`

The changes in the PR should have no effect unless `context->GetSubsystem<Audio>()` returns `nullptr`

You may well be wondering why I would want to remove/disable the Audio subsystem in Urho. I don't have very good reasons other than I am not using it, and that it seems to create a continuously running thread relating to audio devices (at least on Windows). So, it made sense to me to allow the disabling of the subsystem to avoid it having any potential impact.


1vanK:
Hmm, I think the ability to remove subsystems is a design flaw. Should we check for the existence of each subsystem on every use? There are hundreds of places. And what about scripts?

ChrisDenham:
Maybe, though it seemed to me that Audio should be an optional subsystem, like the other optional subsystems like profiling, logging,  network and database which are only added conditionally in Engine constructor. Therefore it seemed to me that removal of an optional subsystem should be equivalent to not adding it in the first place?

1vanK:
> optional subsystems like profiling, logging, network and database

this subsystems is designed to be disabled on build step

ChrisDenham:
Yes, I know. I was thinking Audio should be like that, but I thought that would be a larger change and even less likely to be accepted. It looked as though there were actually very few calls to GetSubsystem<Audio> that might need to be covered, so seemed like my PR was an easy win to achieve the objective of optional audio, which I don't need in my application. I only wanted to remove it because I noticed it was creating an unnecessary process thread.

1vanK:
Does SDL create this thread?

1vanK:
> very few calls to GetSubsystem that might need to be covered

you forgot about scripts

ChrisDenham:
> Does SDL create this thread?

Yes, I think so, but didn't look too deeply into it other than noticing it continuously hitting breakpoints I put in https://github.com/urho3d/Urho3D/blob/master/Source/ThirdParty/SDL/src/audio/SDL_audio.c  and that stops happening when I remove the Audio Subsystem from Urho.

ChrisDenham:
> > very few calls to GetSubsystem that might need to be covered
> 
> you forgot about scripts

Ah, ok. I guess that's true. Is the point then that there is no point in inserting these few nullptr checks unless we also cover all cases? I mean my almost trivial PR only fixes a crash I noticed when not adding audio subsystem. The changes do not encourage anyone to remove the audio subsystem in any other application.

ChrisDenham:
> > Does SDL create this thread?
> 
> Yes, I think so, but didn't look too deeply into it other than noticing it continuously hitting breakpoints I put in https://github.com/urho3d/Urho3D/blob/master/Source/ThirdParty/SDL/src/audio/SDL_audio.c and that stops happening when I remove the Audio Subsystem from Urho.

I mean in: https://github.com/urho3d/Urho3D/blob/master/Source/ThirdParty/SDL/src/audio/wasapi/SDL_wasapi.c

1vanK:
> The changes do not encourage anyone to remove the audio subsystem in any other application.

The engine either expects the subsystem to always exist, or allows its deletion. But for special cases, you can use your modified version of the engine.

1vanK:
When a person comes and says - "Hello, I disabled audio, but I have a problem with scripts", will you fix this bug, or will another person have to do it?

ChrisDenham:
> When a person comes and says - "Hello, I disabled audio, but I have a problem with scripts", will you fix this bug, or will another person have to do it?

Wait, hold on there. You are implying my PR introduces a problem, but it is only an incomplete fix for a problem that was already present.  But, never mind. I'll close the PR as your replies suggest no one but  me might find it helpful.

1vanK:
In fact, I don't like the idea of subsystems at all. It's just a singleton. On paper, you have a nice "x-> GetSubsystem<ClassName>()" mechanism, but in fact you still need to include the header file with the subsystem class. Why not just let you access the class directly.

ChrisDenham:
Interesting, I had assumed subsystems were attached to the Context object , so they wouldn't be singletons unless Context object is a singleton?

I do wish no one had invented the term singleton as it always seems to legitimise what is essentially a global variable, and ignores the huge code changes required in the cases where we get the decision about global state wrong. But this is not a place for my singleton war stories. ha ha.

1vanK:
I suspect that nothing good will come of it if you try to initialize the audio system or graphics twice in two different objects.

PR #2753, opened 2021-01-04
Fixed texture loading failure caused by Heightmap typo

SuperWangKai:
There is a case issue with the heightmap texture name in `PBRExample.xml` scene file. Texture loading fails due to this issue.

https://github.com/urho3d/Urho3D/blob/master/bin/Data/Textures/HeightMap.png


SuperWangKai:
It seems there is something wrong with the CI/CD.

PR #2763, opened 2021-01-20
Update boost::preprocessor to 1.75.0

1vanK:


eugeneko:
image
Ouch. That's a lot of code. Can you remind me how is it used?

1vanK:
Used only for Angel Script in Web build. I updated this because I plan to add boost::optional which will be used in autobinding and the header files were from the same version of the library.

1vanK:
I changed my mind. It turned out that boost::optional depends on many other libraries. Perhaps sometime in the future we will add Boost library to the engine instead of EASTL, but not now.

iSLC:
Just throwing it out there. Because it will be said eventually. But usually boost in a game engine doesn't go well.

1vanK:
Perhaps, but in any case, the boost is better than eastl.

eugeneko:
> I changed my mind. It turned out that boost::optional depends on many other libraries. Perhaps sometime in the future we will add Boost library to the engine instead of EASTL, but not now.

_Whole_ EASTL is only 80k lines of code, which is already smaller than just boost::preprocessor.
If you need just `optional`, why use Boost? There are a lot of much better options, including but not limited to EASTL.

> I changed my mind

Are you going to revert this PR?
Unless you plan to actually _use_ these 200+ files and 100k+ lines of code, of course.

iSLC:
If only `std::optional` is needed then there are a couple standalone implementations of that. Probably in less than 1k/loc. And since it would mostly be used for AngelScript, it wouldn't weigh so heavy on dependencies.

Anyway, I'm probably ranting more than necessary. I guess what I'm trying to say is that we shouldn't bring an entire STL implementation if we're not going to use it exclusively. Even EASTL is debatable if you don't make use of it fully.

1vanK:
> Are you going to revert this PR?

I have not touched this part of the bindings yet, perhaps there is a way to drop out boost::preprocessor altogether. For now, let it stay.

> Whole EASTL is only 80k lines of code, which is already smaller than just boost::preprocessor.

In my opinion, the size of the library is not the most important thing. Much more important is performance, support for all possible compilers and other

eugeneko:
> Even EASTL is debatable if you don't make use of it fully.

@iSLC Sure. If you need only `std::optional`, you can just take standalone implementation and then abandon it when you migrate to newer C++ standard.

> In my opinion, the size of the library is not the most important thing

@1vanK It's not the most important, but it's still important, because it affects build time, checkout time, file system and IDE performance. It's quite easy to turn library into bloatware by adding boost.

I use boost in my both projects at job and I like it, but we have build times measured in hours.
Every time I need to check out project into new folder -- triple that.

> Much more important is performance, support for all possible compilers

Given that EASTL was specifically created and optimized for using in high-performance game engines on PC and consoles, I am more confident in EASTL performance than in Boost. Having said that, EASTL has much less users than Boost and therefore tends to have more bugs.

1vanK:
> I use boost in my both projects at my dayjob and I like it, but we have build times measured in hours.

It seems that your project is just huge. Are you sure that compilation would be faster with a different library implementation?

Issue #2718, opened 2020-11-02
Multichannel audio issue


gleblebedev:
When Urho3D requests an audio device from SDL it doesn't restrict number of channels. I have headphones with surround sound that emulates multichannel setup. When I use them with Urho3D the request returns a device with 8 channels but Urho3D treats it as a mono setup (8!=2) which breaks all the logic. I've mitigated the issue by restricting number of channels:
https://github.com/gleblebedev/rbfx/commit/17aba42cdcb39c02beeed1cb2516a443d7e91973

But this code won't fall back in a mono playback if stereo is not available. I didn't find a way to ask SDL about stereo compatibility so I think the correct fix would be to try to create device with stereo sound and if it fails retry with mono.

gleblebedev:
Let's fix it. How can I reopen the ticket?

PR #2769, opened 2021-01-30
Limit number of channels to avoid issues on multichannel audio device

gleblebedev:
Limit number of channels to avoid issues on multichannel audio device (5.1, 7.1, etc)

The issue is present on, for example, Logitech G933 headphones with surround sound enabled. 

Bug: https://github.com/urho3d/Urho3D/issues/2718

### DIFF
--- a/Source/Urho3D/Audio/Audio.cpp
+++ b/Source/Urho3D/Audio/Audio.cpp
@@ -97,11 +97,26 @@ bool Audio::SetMode(int bufferLengthMSec, int mixRate, bool stereo, bool interpo
 
     // Intentionally disallow format change so that the obtained format will always be the desired format, even though that format
     // is not matching the device format, however in doing it will enable the SDL's internal audio stream with audio conversion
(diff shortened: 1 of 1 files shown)


1vanK:
I see problems here and discuss them in private messages with Gleb

eugeneko:
Code looks okay tho. What's wrong with it?

1vanK:
1. The source of the bug is unknown
2. Even without delving into the search for a bug, the initialization should occur to other way
2.1 If stereo required request exactly 2 channels
2.2 If wailed or mono required request exactly 1 channel

But even this way is not very good, because it would be to find the source of the bug and save access to 8 channels for future use in SoundSource3D.

Currently correct bugfix should be using 2 channels only if more than 2 are available

gleblebedev:
`Currently correct bugfix should be using 2 channels only if more than 2 are available`

I don't agree with that. 

The Logitech g933 headphones report to system as 7.1 setup but actually then mix 8 channels into 2 applying some variation of HRTF to provide surround sound effect. Using only 2 channels may apply unnecessary distortion to the sound.

When played via proper 7.1 setup and using only 2 channels would result in playing only on left and right speakers. Initializing stereo device on 7.1 setup would enable single processing where lower sounds would go to subwoofer and the rest would be distributed between all speakers.

Let hardware do it's thing.

1vanK:
Ok, then just fix initialization

PR #2779, opened 2021-02-18
Fix UI scale issues.

SuperWangKai:
Should work now, did not tested on Retina though. TODO: `SetFixedXXX` of class `UI` shoube be re-worked.

### DIFF
--- a/Source/Urho3D/UI/UI.cpp
+++ b/Source/Urho3D/UI/UI.cpp
@@ -383,8 +383,7 @@ void UI::Update(float timeStep)
     {
         TouchState* touch = input->GetTouch(i);
         IntVector2 touchPos = touch->position_;
-        touchPos.x_ = (int)(touchPos.x_ / uiScale_);
-        touchPos.y_ = (int)(touchPos.y_ / uiScale_);
+        touchPos = ConvertSystemToUI(touchPos);
         ProcessHover(touchPos, MakeTouchIDMask(touch->touchID_), QUAL_NONE, nullptr);
     }
 
@@ -1321,7 +1320,10 @@ void UI::GetCursorPositionAndVisible(IntVector2& pos, bool& visible)
         visible = input->IsMouseVisible();
 
         if (!visible && cursor_)
+        {
             pos = cursor_->GetPosition();
+            pos = ConvertUIToSystem(pos);
+        }
     }
 }
 
@@ -1892,8 +1894,7 @@ void UI::HandleTouchBegin(StringHash eventType, VariantMap& eventData)
     using namespace TouchBegin;
 
     IntVector2 pos(eventData[P_X].GetInt(), eventData[P_Y].GetInt());
-    pos.x_ = int(pos.x_ / uiScale_);
-    pos.y_ = int(pos.y_ / uiScale_);
+    pos = ConvertSystemToUI(pos);
     usingTouchInput_ = true;
 
     const MouseButton touchId = MakeTouchIDMask(eventData[P_TOUCHID].GetInt());
@@ -1913,8 +1914,7 @@ void UI::HandleTouchEnd(StringHash eventType, VariantMap& eventData)
     using namespace TouchEnd;
 
     IntVector2 pos(eventData[P_X].GetInt(), eventData[P_Y].GetInt());
-    pos.x_ = int(pos.x_ / uiScale_);
-    pos.y_ = int(pos.y_ / uiScale_);
+    pos = ConvertSystemToUI(pos);
 
     // Get the touch index
     const MouseButton touchId = MakeTouchIDMask(eventData[P_TOUCHID].GetInt());
@@ -1944,10 +1944,8 @@ void UI::HandleTouchMove(StringHash eventType, VariantMap& eventData)
 
     IntVector2 pos(eventData[P_X].GetInt(), eventData[P_Y].GetInt());
     IntVector2 deltaPos(eventData[P_DX].GetInt(), eventData[P_DY].GetInt());
-    pos.x_ = int(pos.x_ / uiScale_);
-    pos.y_ = int(pos.y_ / uiScale_);
-    deltaPos.x_ = int(deltaPos.x_ / uiScale_);
-    deltaPos.y_ = int(deltaPos.y_ / uiScale_);
+    pos = ConvertSystemToUI(pos);
+    deltaPos = ConvertSystemToUI(deltaPos);
     usingTouchInput_ = true;
 
     const MouseButton touchId = MakeTouchIDMask(eventData[P_TOUCHID].GetInt());
@@ -2067,8 +2065,7 @@ void UI::HandleDropFile(StringHash eventType, VariantMap& eventData)
     if (input->IsMouseVisible())
     {
         IntVector2 screenPos = input->GetMousePosition();
-        screenPos.x_ = int(screenPos.x_ / uiScale_);
-        screenPos.y_ = int(screenPos.y_ / uiScale_);
+        screenPos = ConvertSystemToUI(screenPos);
 
         UIElement* element = GetElementAt(screenPos);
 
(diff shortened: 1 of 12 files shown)


dprandle:
I can verify - this fix works for me. Urho3DPlayer has UI scale set to 2.0 and mouse behaves correctly, and UIDrag example mouse behaves correctly for both UI scale of 1.0 and 2.0.

Source/Urho3D/UI/UI.cpp
```diff
@@ -1308,6 +1309,7 @@ void UI::GetCursorPositionAndVisible(IntVector2& pos, bool& visible)
     if (cursor_ && cursor_->IsVisible())
     {
         pos = cursor_->GetPosition();
+        pos = ConvertUIToSystem(pos);
```
1vanK, line 1312: Cursor is UI element and should works in the same coordinates as other UI elements. After your modification, there will be confusion, since the user will never be sure in which coordinates he is working

1vanK:
Actually we have logical error when used OS-specific cursor:
```
void UI::GetCursorPositionAndVisible(IntVector2& pos, bool& visible)
{
    // Prefer software cursor then OS-specific cursor
    if (cursor_ && cursor_->IsVisible())
    {
        pos = cursor_->GetPosition();
        visible = true;
    }
    else if (GetSubsystem<Input>()->GetMouseMode() == MM_RELATIVE)
        visible = true;
    else
    {
        auto* input = GetSubsystem<Input>();
        pos = input->GetMousePosition(); // -------------------------- Screens coord used here 
        visible = input->IsMouseVisible();

        if (!visible && cursor_)
            pos = cursor_->GetPosition();
    }
}
```

Source/Urho3D/UI/UI.cpp
SuperWangKai: The problem here is: If there is no cursor, i.e. system mouse cursor is used, the coordindate is "system", rather than "scaled UI position". In this case, there will inconsistent definition of the coordinate.

https://github.com/urho3d/Urho3D/blob/84913ce65b3aa0b8704a2e155d15e196dc572bdb/Source/Urho3D/UI/UI.cpp#L774
https://github.com/urho3d/Urho3D/blob/84913ce65b3aa0b8704a2e155d15e196dc572bdb/Source/Urho3D/UI/UI.cpp#L1318

Let's keep the discussion open.

1vanK:
In addition, we have inputScale_, which modifies "system cursor" position. It's hard to say anything without tests on a real retina display. @KatekovAnton what do you think?

bin/Data/Scripts/Editor/EditorGizmo.as
```diff
@@ -215,7 +215,7 @@ void UseGizmo()
         return;
     }
 
-    IntVector2 pos = ui.cursorPosition;
+    IntVector2 pos = ui.ConvertSystemToUI(ui.cursorPosition);
```
1vanK, line 218: cursorPosition already should be in ui coords. If u need real mouse pos, then we have input->GetMousePosition

SuperWangKai:
Let's keep this PR open, at least we have something to refer and work on.

dprandle:
I have retina display - if you need specific checks I can check as well.. I mentioned above with SuperWangKai's first commit, the editor (which has inputScale_ as [2,2] and uiScale_ as 2) mouse worked correctly, and if ui scale is changed for UIDrag example to uiScale_ = 2 with inputScale_ as [2,2] (due to HighDPI), it also worked. Before the commit, the editor mouse position worked correctly, but setting uiScale_ = 2 in the UIDrag example did not.

SuperWangKai:
Just unified cursor position to "Scaled UI position". Also works like the first version. Modifications seem simpler.
Also fixed mouse click issues in `15_Natigation` and `39_CrowdNavigation` samples if we disable cursor.

SuperWangKai:
So here are the total changes till now -
https://github.com/SuperWangKai/Urho3D/compare/84913ce65b3aa0b8704a2e155d15e196dc572bdb...0467ba09a8d986ce2d06af2b9323d1e187dd4d5d

bin/Data/Scripts/15_Navigation.as
```diff
@@ -314,6 +314,8 @@ bool Raycast(float maxDistance, Vector3& hitPos, Drawable@& hitDrawable)
     if (!ui.cursor.visible || ui.GetElementAt(pos, true) !is null)
         return false;
 
+    pos = ui.ConvertUIToSystem(ui.cursorPosition);
```
1vanK, line 317: This is scene related stuff, just use input->mousePos

SuperWangKai: Nope. According to my tests, `Input::GetMousePosition()` won't show the correct position if `MM_FREE ` is NOT used.

You could check `15_Natigation` and `39_CrowdNavigation`, the system mouse position remains in the center of the window while UI cursor is moving around, which leads to functional failure in our cases.

It seems to me that `UI::GetCursorPosition()` is origianlly designed to be the interface of get cursor position, just as the name shows, and scaled UI introduces some vagueness.

SuperWangKai:
Due to lack of web version setup, https://github.com/urho3d/Urho3D/pull/2779/commits/bc2801f87d10786764a65aa96b06e044698e2732 is not tested.

bin/Data/Scripts/15_Navigation.as
1vanK: Yes, it looks like you are right

bin/Data/Scripts/Editor/EditorUI.as
```diff
@@ -131,7 +131,7 @@ void ResizeUI()
     }
 
     // Relayout root UI element
-    editorUIElement.SetSize(graphics.width, graphics.height);
+    editorUIElement.SetSize(graphics.width, graphics.height);
```
1vanK, line 134: Space at the end

1vanK:
Since ui.scale  is used to increase a picture on retina displays, should we also scale the cursor sprite like other elements?

SuperWangKai:
UI is scaled in the rendering, Cursor, as a UIElement, should already be scaled - 
https://github.com/urho3d/Urho3D/blob/84913ce65b3aa0b8704a2e155d15e196dc572bdb/Source/Urho3D/UI/UI.cpp#L1019-L1026

1vanK:
I just run editor, opened console (F1) and write `ui.scale = 0.5`



Does the editor use a hardware cursor that will be automatically scaled or retina displays?

1vanK:
Yes, it looks like the editor is using a hardware cursor because it is not in the screenshot  xD

1vanK:
I confirm, the cursor is normally scaled in example 15

SuperWangKai:
> I just run editor, opened console (F1) and write `ui.scale = 0.5`

I'm afraid you cannot just set `ui.scale` in the console, since there are many hard-coded `SetFixedWidth` in the Editor's initialization as well as ModeChanged event handlers.

---------
Now, the Editor should work relatively normal with an initial `ui.scale` value (You can add `ui.scale = my_value` in the Editor's `void Start()` function)

It seems the Editor never worked normal with UI scale set. I hope people can use it after this PR.

1vanK:
I have no more notes about this PR. Just wait tests from retina display users.

SuperWangKai:
Do you have a chance to test it on the web version? Thanks!

1vanK:
I tested masrter version and couldn't reproduce the problem on my computer so don't think my tests can be helpful

SuperWangKai:
> 
> 
> I tested master version and couldn't reproduce the problem on my computer so don't think my tests can be helpful

Thanks!

dprandle:
Can't vouch for web versions - but desktop for Mac OS big sur - issues I was having are resolved.. see my above comments for what I mean specifically.

In game i'm working, I use UI element as selection rectangle (like desktop selection rectangle) and it is working as expected with this PR.

As retina display urho user, let me know if there are any other things that need testing to merge... Not sure what else needs to be looked at

KatekovAnton:
Hi, I tested with following setup:
- Macbook/macos catalina - retina display
- Macbook/macos catalina - retina display + external monitor (tested on both monitors)
- Macbook/macos catalina - only external monitor
- laptop with highdpi display/windows 10 - only highdi display
- laptop with highdpi display/windows 10 - highdi display + external monitor (tested on both monitors)
- laptop with highdpi display/windows 10 - only external monitor

works correctly in all combinations!

PR #2782, opened 2021-02-21
AS Bindings: Allows using generic calling convention on any platform

1vanK:


1vanK:
CI/CD: errors with C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\um
my pc: no problems with c:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0

PR #2785, opened 2021-02-26
Add VIDEO_WAYLAND build option to getting started doc

dmlo:


1vanK:
This is a build option for SDL. We don't have build options for SDL in the documentation yet. There are a lot of them. Should we add just one? Moreover, the description should not contain things like "needed to avoid some kind of bug".

dmlo:
You are right of course. It is not a great PR.

The problem I was trying to solve is that when someone comes to build the
repo for the first time this SDL option causes a cryptic 300-line linker
failure message.

If we shouldn't put such build hints in a document called 'Building Urho3D
library', how would you like to handle this? The Wayland situation is a
mess everywhere but it is going to take years for the Linux community to
resolve. In the meantime, can we offer new Urho3D users some guidance?

I considered submitting a PR to disable Wayland by default but that
solution is part of what is causing the Wayland problem - by not nudging
people to know about and integrate (or at least deliberately ignore)
Wayland the pain of transition is being extended endlessly.

I'll work in whatever direction you think best. I'm new here.

1vanK:
>If we shouldn't put such build hints in a document called 'Building Urho3D library', how would you like to handle this?

I expect linux usersto fix the build bug wit Wayland. If it suddenly turns out that this is not possible for some reason, this option should be disabled by default.

dmlo:
Understood.

PR #2793, opened 2021-04-01
Upgrade assimp to 5.0.1

PencilAmazing:
Why not? It has some things I wanted that the older version didn't have. Main edits were only in the two CMakeLists.txt files and deleted unnecessary folders (test, coverage, android).


PencilAmazing:
wikipedia.org/wiki/ISO_10303-21
STEP files are apparently an ISO specified CAD format? I'm not sure why it's failing for MinGW on Linux but not for MinGW Windows but I am setting up a VM to experiment
MacOS is most probably me deleting something wrong.

PencilAmazing:
I consider this branch ready for master. Unless there are any objections that is. I didn't change much but I'm open for any discussion.

1vanK:
I need some time to finish my current task, after that I will handle your PR

1vanK:
You need to keep the original code. This will facilitate future updates. Example: https://github.com/urho3d/Urho3D/blob/master/Source/ThirdParty/Bullet/src/BulletDynamics/ConstraintSolver/btSequentialImpulseConstraintSolver.cpp#L675-L681

Source/ThirdParty/Assimp/CMakeLists.txt
```diff
@@ -38,7 +38,9 @@ SET(CMAKE_POLICY_DEFAULT_CMP0074 NEW)
 
 CMAKE_MINIMUM_REQUIRED( VERSION 3.0 )
 
-# Modified for Urho3D
+# Modified for Urho3D: Removed Hunter, Android, tests, forced in-house builds,
```
1vanK, line 41: Just write "Modified for Urho3D". All details should be near lines with changes

Source/ThirdParty/Assimp/code/CMakeLists.txt
```diff
@@ -854,12 +858,12 @@ ADD_ASSIMP_EXPORTER( STEP
 )
 
 if ((NOT ASSIMP_NO_EXPORT) OR (NOT ASSIMP_EXPORTERS_ENABLED STREQUAL ""))
-	SET( Exporter_SRCS
-	  Common/Exporter.cpp
-	  CApi/AssimpCExport.cpp
-	  ${HEADER_PATH}/BlobIOSystem.h
-	)
-	SOURCE_GROUP( Exporter FILES ${Exporter_SRCS})
+    SET( Exporter_SRCS
```
1vanK, line 862: Try to minimize changes of the original files. Otherwise, you will have to repeat these changes with every upgrade. Style changes are not necessary.

1vanK:
Try to minimize changes of the original files. Otherwise, you will have to repeat these changes with every upgrade. Style changes (spaces for example) are not necessary.

Source/ThirdParty/Assimp/CMakeLists.txt
```diff
@@ -33,14 +34,18 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #----------------------------------------------------------------------
+SET(CMAKE_POLICY_DEFAULT_CMP0074 NEW)
 
-# Modified by Yao Wei Tjong for Urho3D
+CMAKE_MINIMUM_REQUIRED( VERSION 3.0 )
 
-PROJECT( Assimp )
+# Modified for Urho3D
```
1vanK: "Modified for Urho3D" must be found after the original license agreement

Source/ThirdParty/Assimp/CMakeLists.txt
1vanK: "Modified for Urho3D" must be found after the original license agreement

1vanK:
Are you using the most recent version of the library? Add file like https://github.com/urho3d/Urho3D/blob/master/Source/ThirdParty/Bullet/Urho3DNotes.txt to know when and from where the library was downloaded

PencilAmazing:
I will make a second pass on the files once I have time. Is it fine if I put a comment where I deleted things? There's a lot of basically duplicated code because of hunter integration and I think adding a single comment at the top would be more appropriate.

PencilAmazing:
There's a lot of lines deleted that come after each other. Is it fine if I leave a comment saying what was removed?

1vanK:
Why did you use such an old version?

1vanK:
You deleted files in subdirectories, but you kept the folder structure. This will complicate the next library upgrades. All changes to the third party library must be documented. You can delete some folders in the root directory, but you don't need to delete individual files in subdirectories.

1vanK:
You also haven't updated the files in the root directory (CHANGES for example)

PencilAmazing:
Old version? I downloaded the assimp 5.0.1 release from the github which as far as I know is what's recommended. As for the documentation I'll put them in the Urho3DNotes.txt

1vanK:
assimp 5.0.1 was released 12 Jan 2020

PencilAmazing:
The CHANGES file included with the zip and tar (and even the master!) file both had it up to 4.1.0. I'm 100% sure I got the 5.0.1 release. You're right I forgot some root files too.
I could redo everything if you're not content with my PR.

PR #2794, opened 2021-04-06
fix shader parameter cache bug

aimoonchen:


eugeneko:
So, what's the bug?

iSLC:
I don't think he means bug. More like an optimisation.

eugeneko:
@iSLC More like pessimization, because now it does more work.

1vanK:
> I don't think he means bug. More like an optimisation.

```
if (globalFrameNumber != frameNumber_ || graphics_->GetLastShaderProgram() != this)
```

No, it looks like trying to fix some kind of bug, because added new activation conditions

aimoonchen:
> @iSLC More like pessimization, because now it does more work.

@iSLC @eugeneko It doesn't work when i change the graphic backend to bgfx. I guess the unifom cache policy doesn't work well when shader changes.

eugeneko:
> It doesn't work when i change the graphic backend to bgfx

I don't get it... Urho doesn't have bgfx backend.

eugeneko:
Oh, I see.
What purpose this code serves in Urho master branch? Except possible performance degradation.
I mean, PRs are supposed to make things better, not worse. It says "fix bug", but there's no bug to be fixed.

aimoonchen:
Sorry, this is not a bug in urho master branch, the OpenGL specification says : "... Uniforms are program object-specific state. They retain their values once loaded, and their values are restored whenever a program object is used, as long as the program object has not been re-linked ...".
But in bgfx, the uniform is not program object-specific state, it is global state, so it does't work in my bgfx branch.

PR #2816, opened 2021-08-01
Static check for PODVector

1vanK:
@eugeneko ?

### DIFF
--- a/Source/Urho3D/AngelScript/APITemplates.h
+++ b/Source/Urho3D/AngelScript/APITemplates.h
@@ -95,6 +95,24 @@ template <class T> CScriptArray* VectorToArray(const PODVector<T>& vector, const
         return nullptr;
 }
 
+/// I don't know why in some cases compilers confuses Vector and PODVector and use incorrect overloading
+template <class T> CScriptArray* NotPODVectorToArray(const Vector<T>& vector, const char* arrayName)
+{
+    Context* context = GetScriptContext();
+    if (context)
+    {
+        asITypeInfo* type = context->GetSubsystem<Script>()->GetObjectType(arrayName);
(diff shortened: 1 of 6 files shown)


eugeneko:
Do you have clear steps how it can be reproduced?
Compiler cannot just confuse two different types. If it "does", you probably need to investigate why and find the root cause.

So far I don't see what the bug is and how this PR fixes it.

PS. No need to have 3 exact copies of the code next to each other, 1 would be enough.

1vanK:
> Do you have clear steps how it can be reproduced?

remove NotPODVectorToArray and rename calls NotPODVectorToArray to VectorToArray

eugeneko:
@1vanK I believe you have hit this "feature" of C++:
https://stackoverflow.com/questions/36298211/static-assert-triggering-during-function-name-resolution

Still, solution is really ugly. Just make first parameter of `VectorToArray` template and remove second (and third) copy of this function.

1vanK:
https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/AngelScript/APITemplates.h#L72

eugeneko:
If you wanted to make a point by the latest comment, sorry, I didn't get it.

My point is that both versions of `VectorToArray` are identical and therefore you don't need two of them, only one.

The "propper" fix for this compiler "bug" would be to use `enable_if` and template specializations instead of `static_assert`, but it's ugly as hell and is overkill in this specific case.

1vanK:
> Just make first parameter of VectorToArray template

T is used and I can not just replace it to something else

eugeneko:
> T is used and I can not just replace it to something else

Functions may have more than one template parameter

PR #2825, opened 2021-09-05
Fix Connection support JSON scene format on a client side

YuriAzathoth:
As we know, Urho3D supports the scenes in the JSON format (even the standard editor can export the scenes in JSON), but after connecting to the server when the scene is loading on the client side, the engine gives an error like "ERROR: Scene.json is not a valid scene file". This fix solves this problem.

### DIFF
--- a/Source/Urho3D/Network/Connection.cpp
+++ b/Source/Urho3D/Network/Connection.cpp
@@ -1619,6 +1619,8 @@ void Connection::OnPackagesReady()
 
         if (extension == ".xml")
             success = scene_->LoadAsyncXML(file);
+        else if (extension == ".json")
+            success = scene_->LoadAsyncJSON(file);
         else
             success = scene_->LoadAsync(file);
 


eugeneko:
Just for a record, I would advice against using JSON scenes and nodes.
Due to [bug in serialization](https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/Scene/Serializable.cpp#L446) _there's no gurarntee that components will be successfully loaded from JSON_ becuase attributes are loaded in random order and strict order is required for some components.

weitjong:
Component that requires the attributes to be fully loaded before applying all or part of the attributes should override the `ApplyAttribute()` method. I will merge this PR shortly.

weitjong:
Thanks for submitting the fix!

eugeneko:
> Component that requires the attributes to be fully loaded before applying all or part of the attributes should override the ApplyAttribute() method.

In perfect world? I agree. In reality...

Current `StaticModel` and its descendants rely on strict attribute order (`Model` goes before `Materials`). _If_ `StaticModel` serialization to JSON works now (I didn't test it), it's solely due to lucky hash function and lucky number of buckets. And I don't see an easy way to fix it w/o allocating extra memory in already fat `StaticModel`.

There may or may not be other components which rely on strict attribute order, and they may or may not work with JSON now (I have no clue).

Sorry for offtopic, I just wanted to ensure that author knows about possible issues.

YuriAzathoth:
I think that this fix is much better than incomprehensible mistakes and many hours attempts to understand what's wrong with the scene in JSON format and why it is loaded without any problems in singleplayer and gives strange mistakes in the multiplayer.

PR #2828, opened 2021-09-28
Remove FakeAddRef and FakeReleaseRef

orefkov:
Registration in AngelScript of classes marked as @fakeref with the flag "asOBJ_NOCOUNT", without registration empty AddRef and Release. This optimizes the AngelScript code, as it disables the call to reference counting for such objects. In general, I think that for such objects, the registration of the factory should also be removed, since the objects created in the script through the factory will never be deleted.

### DIFF
--- a/Source/Urho3D/AngelScript/APITemplates.h
+++ b/Source/Urho3D/AngelScript/APITemplates.h
@@ -232,19 +232,22 @@ template <class BaseType, class DerivedType> void RegisterSubclass(asIScriptEngi
 {
     if (!strcmp(baseClassName, derivedClassName))
         return;
-
-    String declReturnBase(String(baseClassName) + "@+ opImplCast()");
+    String declReturnBase = String(baseClassName) +
+                            ((engine->GetTypeInfoByName(baseClassName)->GetFlags() & asOBJ_NOCOUNT) ? "@" : "@+") +
+                            " opImplCast()";
     engine->RegisterObjectMethod(derivedClassName, declReturnBase.CString(), AS_FUNCTION_OBJLAST((RefCast<DerivedType, BaseType>)), AS_CALL_CDECL_OBJLAST);
 
-    String declReturnBaseConst("const " + String(baseClassName) + "@+ opImplCast() const");
+    String declReturnBaseConst("const " + declReturnBase + " const");
     engine->RegisterObjectMethod(derivedClassName, declReturnBaseConst.CString(), AS_FUNCTION_OBJLAST((RefCast<DerivedType, BaseType>)), AS_CALL_CDECL_OBJLAST);
 
     //String declReturnDerived(String(derivedClassName) + "@+ opCast()");
-    String declReturnDerived(String(derivedClassName) + "@+ opImplCast()");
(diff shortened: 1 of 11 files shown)


1vanK:
When I checked your request, I regenerate the bindings and got a lot of hanges in files Generated_Classes.h and Generated_Members.h. Perhaps when generating the bindings, something is platform dependent. What compiler / OS are you using?

1vanK:
Probably also affected by Doxygen version

orefkov:
> Probably also affected by Doxygen version

I build on Win10, VS2019, Doxygen 1.9.2 (caa4e3de211fbbef2c3adf58a6bd4c86d0eb7cb8)

1vanK:
Ok, something broke in Doxygen 1.9+

1vanK:
https://github.com/doxygen/doxygen/issues/8805

1vanK:
Can you update your PR with latest master?

Source/Urho3D/AngelScript/APITemplates.h
```diff
@@ -232,19 +232,20 @@ template <class BaseType, class DerivedType> void RegisterSubclass(asIScriptEngi
...
 
-    //String declReturnDerived(String(derivedClassName) + "@+ opCast()");
-    String declReturnDerived(String(derivedClassName) + "@+ opImplCast()");
-    engine->RegisterObjectMethod(baseClassName, declReturnDerived.CString(), AS_FUNCTION_OBJLAST((RefCast<BaseType, DerivedType>)), AS_CALL_CDECL_OBJLAST);
+    strCast = String(derivedClassName) +
+              ((engine->GetTypeInfoByName(derivedClassName)->GetFlags() & asOBJ_NOCOUNT) ? "@" : "@+") +
+              " opImplCast()";
+    engine->RegisterObjectMethod(baseClassName, strCast.CString(), AS_FUNCTION_OBJLAST((RefCast<BaseType, DerivedType>)), AS_CALL_CDECL_OBJLAST);
 
-    //String declReturnDerivedConst("const " + String(derivedClassName) + "@+ opCast() const");
```
1vanK, line 246: I was planning to do an explicit conversion from base to derived as recommended here https://www.angelcode.com/angelscript/sdk/docs/manual/doc_adv_class_hierarchy.html (but this require fix many scripts)

1vanK: May be some build option to backward compatibility

1vanK:
Not tested yet, but is this possible to convert between objects "asOBJ_REF | asOBJ_NOCOUNT"
 and "asOBJ_REF"? Not sure if this is used somewhere, but Serializer and Deserializer is base class for some others

1vanK:
Functions `void FakeAddRef(void* ptr)` and `void FakeReleaseRef(void* ptr)` can be removed

>  I think that for such objects, the registration of the factory should also be removed, since the objects created in the script through the factory will never be deleted.

sounds reasonable

P.S. Feel free to make any changes if they are tested enough, as I do not use AS in my practice and will hardly notice regressions.

PR #2835, opened 2021-10-07
Bug fixes and improves in AngelScript using

orefkov:
Fixed a bug when creating script objects - when the factory was executed, the execution nesting level did not increase, and if the constructor of the script object called the execution of the script function, it did not receive a free context for execution.
When compiling a script into bytecode, the initialization of the global variables of the script is no longer called.
Bug fixed - "scriptFile" was not available in the constructors of global script objects.
Improved script api dump generation - added output to the header file "operator XXX", casting to other types, added output of method's arguments name.


1vanK:
Not tested yet. If this commit change `AngelScriptAPI.h` place new version to Docs

PR #2841, opened 2021-11-05
Fixed GLTF import

niansa:
I noticed glTF2Importer::CanRead not returning the right value on success :wink:

### DIFF
--- a/Source/ThirdParty/Assimp/code/glTF2Importer.cpp
+++ b/Source/ThirdParty/Assimp/code/glTF2Importer.cpp
@@ -117,7 +117,7 @@ bool glTF2Importer::CanRead(const std::string& pFile, IOSystem* pIOHandler, bool
         }
     }
 
-    return false;
+    return true;
 }
 
 


weitjong:
Thanks!

1vanK:
need comment for changes in 3rd-libs

weitjong:
Ideally, yes. However, this bug fix is so glaring that I would not be surprised if it has been fixed in upstream long time ago.

1vanK:
https://github.com/assimp/assimp/blob/master/code/AssetLib/glTF2/glTF2Importer.cpp#L114-L126

1vanK:
Actually I don't understand why the function returns true if no check was successful

weitjong:
The upstream is preparing for the Release 5.1.0. If no one has done so, I/we should submit the PR for this bug fix to upstream.

weitjong:
I think we just have to keep this fix locally as the version of code has changed in the upstream's master branch. See the thread in the upstream for the funny story. Morale of the story: don't just trust the tool with CLI, eye ball the code manually :smiley:. For some reason my eye also just locked in to the last return statement when 1vank showed the code 2 days ago here. 

p.s. It seems the 5.1.0 is in fact already out. Anyone, who is interested in upgrading the Assimp, can just safely override this local (unmarked) patch of ours with the newer release.

Issue #2826, opened 2021-09-09
Material Shader Parameter Animation Not Reset In ResetToDefaults()


SirNate0:
As best I can tell, `Material::ResetToDefaults()` clears and resets the `shaderParameters_` but it does not clear the `shaderParameterAnimationInfos_`.

weitjong:
Most probably it is  a bug since when the shader parameter animation was introduced by Aster 8 years ago. From the commit history you can see that Aster's commit did not touch the `Material::ResetToDefaults()`. The latter was last changed by Lasse about 9 to 10 years ago. My point is, the `shaderParameterAnimationInfos_` was not added by the original author of the engine. I agree with you that it should be cleared like the rest. If you submit as PR, I will help to merge your fix.

PencilAmazing:
I could give it a try.

SirNate0:
Sorry, it got pushed to the back burner. I think Material should also update (clear) its event subscriptions when reset, otherwise it would be listening to events it no longer acts on. Added a fix for that with #2843.

PR #2843, opened 2021-11-07
Material unsubscribe from animation events on ResetToDefault

SirNate0:
I believe Material should also unsubscribe from the update events when we clear the shader animations in addition to clearing them. An additional fix for #2826


weitjong:
I agree. Thanks!

Issue #2859, opened 2022-01-03
UI - Default TAB order should be left-to-right


Lys0gen:
For some reason the default tab order of UI elements is right-to-left. Flipping the ternary evaluation result in this line should fix it.
(I don't know if the current behaviour is correct for RTL language environments but if it is: maybe make the direction a setting somewhere in the engine?)

https://github.com/urho3d/Urho3D/blob/c65eaab8a31d2f09c1dead987e1d2a9072922157/Source/Urho3D/UI/UI.cpp#L2010

1vanK:
How do I reproduce the problem?

Lys0gen:
Have several focusable UI Elements (e.g. LineEdit). Select one, press TAB on your keyboard. It will go to the previous element instead of the next.

1vanK:
Tested in Editor: View > Editor preferences. Everything works fine.

Lys0gen:
Indeed you are right, that wasn't the actual cause.

The problem comes from UIElement::SortChildren (which is enabled by default on all elements).
It keeps messing up the order of the child elements when they have the same priority. Very annoying.

I don't know if the default being true is necessary for anything or if that may be changed to false.

For now my personal solution will have to be to do SetSortChildren(false) on all my elements.
Regardless, thanks for your response and sorry for potentially wasting your time.

PR #2861, opened 2022-01-09
StableSort for UIElement::SortChildren

Lys0gen:
UIElements of equal priority will keep their existing order. Primarily avoids messing up the TAB selection order of elements.

Credit goes to @ab4daa who implemented this a while back
https://gist.github.com/ab4daa/5932716ef369f110ef31b95ca94d6564

### DIFF
--- a/Source/Urho3D/Container/Sort.h
+++ b/Source/Urho3D/Container/Sort.h
@@ -141,4 +141,234 @@ template <class T, class U> void Sort(RandomAccessIterator<T> begin, RandomAcces
     InsertionSort(begin, end, compare);
 }
 
+
+/// Stable sort is based on http://thomas.baudel.name/Visualisation/VisuTri/inplacestablesort.html
+/// Benchmark https://www.codeproject.com/Articles/26048/Fastest-In-Place-Stable-Sort
+template <class T> void StableSortRotate(RandomAccessIterator<T> begin, RandomAccessIterator<T> mid, 
+    RandomAccessIterator<T> end)
+{
+/*  a less sophisticated but costlier version:
+  reverse(from, mid-1);
+  reverse(mid, to-1);
+  reverse(from, to-1);
+*/
+    if(begin == mid || mid == end) return;
+
+    //int n = gcd(end - begin, mid - begin);
+    int n1 = end - begin;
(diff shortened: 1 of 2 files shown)


eugeneko:
Any reason _not_ to use `std::stable_sort`?

Lys0gen:
Well if that were up to me then I don't care. In my personal code I use std-stuff all the time.

But
1. I was under the impression that Urho doesn't use std containers/algorithms anywhere
2. Maybe I am doing something wrong, but I can not use std::stable_sort on Urho3D containers without compiler errors (which in this case is necessary to sort UIElement children...)
3. It doesn't fit your own code conventions??

eugeneko:
1. Urho doesn't use std containers in _public_ API because there are no bindings for it, and this is pretty much the only practical reason. Your code doesn't care about bindings, so it doesn't matter here.
2. Urho containers have non-standard iterators, which sucks. You can fix it by making Vector iterators standard-compliant, if you wish (it's very easy). Or you can just use pointers instead of iterators (`iter.ptr_`), or just sort `vector.Data(), vector.Data() + vector.Size()` instead of Begin/End.
3. tbh I don't find anything in coding conventions that forbids using std library.

Lys0gen:
Okay. You're a maintainer here so it's up to you whether you want this integrated or not.

But just in case you didn't notice this is a fix for todo comment that has been in UIElement::SortChildren for literally years. 

[> /// \todo Order is not stable when children have same priorities](https://github.com/urho3d/Urho3D/blob/5aa443ae89c0f7324d5d47060c713ed42f6e8b44/Source/Urho3D/UI/UIElement.cpp#L1795)

If you deem it better done with std then please fix it that way. I only noticed this recently because the current sorting messed up my element order: #2859 
And with this change everything works perfectly.

eugeneko:
1. `ptr_` [is totally public](https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/Base/Iter.h#L123)
2. Okay, it's called `Buffer`, not `Data`. Didn't use Urho3D::Vector for years.

> Okay. You're a maintainer here so it's up to you whether you want this integrated or not.

Fair enough.

PR #2863, opened 2022-01-12
Fix bug in Ray::HitDistance functions for geometry.

### LINKED ISSUE #2862: Potential bug in Ray::HitDistance - unreliable results in `outNormal` and `outUV`
I think I may have found a bug in at least two of the `Ray::HitDistance` functions in Urho3D.

The bug is that they appear to return the normal of the last intersected triangle rather the nearest one.
In addition, what also looks wrong to me is that the barycentre used to interpolate the returned UV coordinate may not necessarily be calculated for the same triangle for which the UV coordinate is interpolated.
 
Of course, if there is only one ray intersection with the mesh, the result is reliable, but generally speaking the function appears to be assuming that the last intersection found is also the nearest, and thus the resultant normal and UV coord are potentially dependent on the ordering of the triangles in the mesh. Please correct me if I'm wrong here.

I think I can see how to fix it simply by only assigning to `outNormal`, `outUV` & `outBary` at the point of assigning `nearestIndices = indices`. However I'm a bit hesitant to submit that change as a PR in case I've misunderstood something here and wanted to check with the masters of Urho before doing that. :-)

https://github.com/urho3d/Urho3D/blob/928b5b1b068c4e525d4d8274033d3633eabacf7f/Source/Urho3D/Math/Ray.cpp#L287-L369

ChrisDenham:
Fixes bug in Ray::HitDistance functions which may return incorrect normal and UV when encountering multiple geometry triangle hits.

Essentially the fix is simply ensuring that the normal and UV returned by RayHitDistance belongs to the nearest triangle rather than for the last intersection encountered.

Fixes #2862 Closes #2862
Fixes #2702 Closes #2702

### DIFF
--- a/Source/Urho3D/Math/Ray.cpp
+++ b/Source/Urho3D/Math/Ray.cpp
@@ -250,16 +250,27 @@ float Ray::HitDistance(const void* vertexData, unsigned vertexStride, unsigned v
     const unsigned char* vertices = ((const unsigned char*)vertexData) + vertexStart * vertexStride;
     unsigned index = 0, nearestIdx = M_MAX_UNSIGNED;
     Vector3 barycentric;
-    Vector3* outBary = outUV ? &barycentric : nullptr;
+
+    Vector3 tempNormal_, tempBary_; 
+    Vector3* tempNormal = outNormal ? &tempNormal_ : nullptr;
+    Vector3* tempBary = outUV ? &tempBary_ : nullptr;
 
     while (index + 2 < vertexCount)
     {
         const Vector3& v0 = *((const Vector3*)(&vertices[index * vertexStride]));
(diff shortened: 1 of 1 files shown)


1vanK:
The underscore at the end of an identifier is used for class members, do not use it for local variables

ChrisDenham:
> The underscore at the end of an identifier is used for class members, do not use it for local variables

Good point. Fixed now.

1vanK:
Can you test this option and I'll merge it.

ChrisDenham:
> Can you test this variant and I'll merge it.

Looks good to me.
I tested again with your changes in VS2019/WIN32/DX9&OGL
Thanks very much for the speedy replies and edits.

PR #2864, opened 2022-01-13
Split Urho2D to Urho2D and Physics2D

1vanK:


eugeneko:
You know, when I first saw this PR, I had this gut reaction: "why break compilation of user code without any real gain?"

But I actually think this is great idea to have these two subsystems split, as long as they are actually separate, without dangling bits inbetween (like Urho2D including Physics2D and vice versa).

1vanK:
Physics2D - just wrap around box2d (like Physics - wrap around Bullet)

PR #2867, opened 2022-01-15
Add template friend class declaration for UniquePtr

luchu1993:


eugeneko:
What's the purpose of this change?

luchu1993:
> What's the purpose of this change?
Different type of UniquePtr template member function `operator ==` will compile failed. 
UniquePtr<T> can not access private member `ptrs_` in UniquePtr<U>.

SharedPtr has declared the template friend class, but UniquePtr not.

1vanK:
Several Unique_Ptr cannot contains same raw pointer. Because they are unique.

eugeneko:
@1vanK however, _same_ unique pointer may contain the same raw pointer.
And you may want to know if two UniquePtr& are the same pointer.

@luchu1993 good point, thanks.

1vanK:
If two UniquePtr are the same pointer the error has already occurred. Because destruction of one of them corrupt all the others.

1vanK:
same unique pointer can be compared without friend

eugeneko:
> same unique pointer can be compared without friend

Oh. I didn't think of that. Heh. Perhaps template in comparison should be removed instead -- because different types of UniquePtr can never be the same, and therefore heterogeneous comparison is pointless.

luchu1993:
> > same unique pointer can be compared without friend
> 
> Oh. I didn't think of that. Heh. Perhaps template in comparison should be removed instead -- because different types of UniquePtr can never be the same, and therefore heterogeneous comparison is pointless.

I agree with you. The comparison template method should be removed.

PR #2868, opened 2022-01-24
Replace Vector and PODVector by std::vector

1vanK:


1vanK:
At the end of the work, a performance comparison will be made and only then will a decision be made to merge.

iSLC:
Instead of going file by file and replacing `Vector` with `std::vector` can I suggest that you just write a simple wrapper around a `std::vector`?

```cpp
template < class T > struct Vector {
	using Storage = std::vector< T >;
	//...
	Storage::reference Front() { return mStorage.front(); }
	Storage::const_reference Front() const { return mStorage.front(); }
	//...
	Storage::size_type Size() { return mStorage.size(); }
	//...
	void Clear() { mStorage.clear(); }
	//...
	Storage mStorage;
};
```

The compiler is most likely to inline the function calls of the wrapper and it'll be as if the wrapper didn't even exist in the end.

Should save you from wasting time on this (_probably_) unnecessary conversion.

If you just want to see if it makes any difference (_i'm guessing_).

1vanK:
How does this help get rid of the two kinds of vectors in the engine?

PR #2873, opened 2022-02-03
Benchmark

1vanK:


eugeneko:
FYI: JSON scenes are unstable in general because Urho loads component attributes in _random_ order from JSON.
If you get unlucky, your component will fail to load and you will be unable to do anything about it.

1vanK:
Thanks for the information. The scenes are just stubs now, so it won't be a problem to replace them with xml

1vanK:
@eugeneko I haven't done any tests, but it seemed to me that xml scenes load much faster than json. I assumed it would be the other way around since xml is a much more complex and verbose format.

1vanK:
https://www.youtube.com/watch?v=bw3lsSqlxgM

1vanK:
The idea taken from here https://www.youtube.com/watch?v=71rhWjM6aMo

PR #2876, opened 2022-02-14
Calc StringHash compile time if possible

1vanK:


1vanK:
Expressions like `"abcd"_hash` return compiled time StringHash

PR #2888, opened 2022-04-10
Update PackageTool.cpp

lokimx88:
packagetool can decompress  .pak file


1vanK:
Thanks, I'll check it out when I'm done with my PR.

1vanK:
Some notes:

1) Use existed code style
2) Use informative names for commits
3) Separate your comments from the license agreement
4) Write doc to https://github.com/urho3d/Urho3D/blob/master/Docs/Reference.dox

Issue #2865, opened 2022-01-13
49_Urho2DIsometricDemo.lua not completed


1vanK:
collisions are not handled

1vanK:
Fixed in https://github.com/urho3d/Urho3D/pull/2896

PR #2896, opened 2022-04-15
Complete 49_Urho2DIsometricDemo.lua (#2865)

lucasvinbr:
- added collision handling
- added debug drawing of tileMap stuff, as in the AS version
- also edited SpawnEffect in Sample2D.lua to match the AS version
(without the edit, the effect would spawn behind the scenario in sample 49)


1vanK:
Thanks!

PR #2897, opened 2022-04-17
Short String Optimization

1vanK:


1vanK:
If a string length <= 10 (11 with end zero) on x32 or <= 14 (15 with endzero) on x64 then the string stores in stack

1vanK:
It's hard to measure the usefulness of this. But it didn't get worse XD

iSLC:
It shouldn't get worse. Maybe a few extra instructions here and there but you save a bunch when you look up short names (like components for example). Or convert integers etc.

SSO string implementations are complemented by string_view and span types. That's when they shine most and the cost of constantly having to figure out which type of buffer it uses gets removed almost completely.

I think I have an implementation of those from past experiments with urho, now that c++17 is on the table.

PR #2901, opened 2022-04-19
Replace UniquePtr by std::unique_ptr [cache clear]

1vanK:
Related:
* https://github.com/urho3d/Urho3D/pull/2867
* https://github.com/urho3d/Urho3D/pull/2451

std::unique_ptr supports arrays, custom deleters. Our own implementation will not have any advantages over the standard one.


gleblebedev:
Maybe eastl unique ptr?

1vanK:
No, I use the std class in this case only because our own implementation will completely repeat what is already there. Using eastl will not solve all problems (for example I plan to add ToLower() & ToUpper() for UTF-8 text). Therefore, there is no reason to drag a library into the project, which also needs to be modified.

PR #2902, opened 2022-04-20
Move reference counter into object [cache clear]

1vanK:
Only weak reference counter located outside objects and allocates when Weak_Ptr used first time.


eugeneko:
I have concerns about thread-safety of this code.

1vanK:
Could you explain what has changed in terms of thread safety?

eugeneko:
Current Urho implementation is not, strictly speaking, thread safe (unlike eastl/std::shared_ptr/weak_ptr which are thread safe).
Which sucks, but 1) it was simple enough to be _practicaly_ thread-safe in many scenarios, and 2) it was easily patched to become thread safe.

This PR looks deliberately thread-unsafe without an easy way to make it thread-safe.

Let's ignore the scenario when object is destoryed in one thread and locked by WeakPtr from another.

Can we freely create and destroy SharedPtrs and WeakPtrs from arbitrary threads?
The answer _was_ "yes" if we assumed that refcouters are atomic.
Now the answer is clear "no".

1vanK:
I mean nothing has changed from the previous version. The old version was not thread safe and this version is not thread safe. It just became less memory allocations.

1vanK:
> The answer _was_ "maaaybe yes" if we assumed that refcouters are atomic.

If we talk about code modification, then thia version can be returned to the previous one by allocating memory for the weak reference counter in the constructor

1vanK:
https://www.angelcode.com/angelscript/sdk/docs/manual/doc_adv_weakref.html
https://www.angelcode.com/angelscript/sdk/docs/manual/doc_reg_basicref.html

eugeneko:
Before, I could safely create and destroy weak pointers in thread A while I destroy object in thread B. After, I cannot do that anymore due to this sign flipping thingy.

I know this sounds like a very specific example, but it’s just an illustration of generic issue. This change *may* start silently crashing user code.

Having said that, I am not going to use this code anyway, so my opinion here is not very important.

1vanK:
> Having said that, I am not going to use this code anyway, so my opinion here is not very important.

Will rbfx no longer take code from the main repository?

1vanK:
Suspended for additional research. It might be worth making the implementation AS-compatible.

eugeneko:
Before, I could safely create and destroy weak pointers in thread A while I
destroy object in thread B. After, I cannot do that anymore.

I know this sounds like a very specific example, but it’s just an
illustration. This change *may* start silently crashing user code.

PR #2920, opened 2022-04-30
Move VertexBuffer to GraphicsAPI folder [cache clear]

1vanK:
Currently we have all graphics stuff in one place. It would be nice to separate Graphics API wrapper from everything else. This allows disable existed scene graph and add alternative scene graphs like Quake or [Tesseract](http://tesseract.gg/).


1vanK:
This commit contains changes that can be obtained by simply moving files. Further changes require splitting some classes into several classes.

PR #2926, opened 2022-05-07
BigInt (WIP)

1vanK:
Clickers operate on extremely large numbers, so I think that such a class should be exists in the engine (along with an example and bindings). Those who are interested can follow the progress and suggest changes.


1vanK:
The division operation will be added later.
Also no AS bindings yet: https://www.gamedev.net/forums/topic/712254-vs-as-type-detection/

PR #2936, opened 2022-05-16
Check MinGW compatibility - fix https://github.com/urho3d/Urho3D/issu…

1vanK:
…es/2887


1vanK:
Current DBE contains obsolete MinGW

PR #2938, opened 2022-05-18
Fix C++ version detection

1vanK:


1vanK:
https://github.com/urho3d/Urho3D/pull/2487

Current minimal version is c++17, so `URHO_CPP14_STANDARD` and `URHO_CPP11_STANDARD` have been removed

PR #2939, opened 2022-05-18
Make RandomAccessIterator std compatible | Replace LowerBound & Upper…

1vanK:
…Bound by std functions [cache clear]


Source/Urho3D/Base/Iter.h
```diff
@@ -4,13 +4,22 @@
...
+#include <iterator>
 
 namespace Urho3D
 {
 
 /// Random access iterator.
 template <class T> struct RandomAccessIterator
 {
+    using iterator_category = std::random_access_iterator_tag;
+    using difference_type = i64;
```
eugeneko, line 18: I think this should be ptrdiff_t, as well as `constexpr int operator -` below

1vanK: We use unsigned for containers indices. i64 can contains -i32.. +i32

eugeneko: About `operator -`: `Vector` of 1'000'000'000 of i64 will have `end-begin=8'000'000'000` which doesn't fit in `int`, so `end-begin` will just return invalid result in this case.

About `difference_type`: I think having it as i64 may generate warnings on 32 bit systems due to narrowing conversions and stuff.

1vanK: (actually ptrdiff_t is i64 on VS x64)

eugeneko: I thought about it again, and ideally, ptrdiff_t should be the same type as operators + and - (int in this case)

eugeneko: Because the std algos work like this:

```
difference_type d = iter1 - iter2;
iter1 += d;
```

So if you have different types you may get annoying warinigs in `std::`

1vanK: actually Scott Meyers recommeds use signed types in interfaces https://www.aristeia.com/Papers/C++ReportColumns/sep95.pdf so for size of containers we also can use int instead unsigned

iSLC: > (actually ptrdiff_t is i64 on VS x64)

`ptrdiff_t` should be signed regardless of the compiler https://en.cppreference.com/w/cpp/types/ptrdiff_t

"*std::ptrdiff_t is the signed integer type of the result of subtracting two pointers.*"

Issue #2940, opened 2022-05-19
Use int for containers


1vanK:
We use `unsigned` to store sizes.
If user write `size1 - size2`, then overflow occurs if `size1 < size2`.
Those, sizes still need to be converted to `int` before being used in calculations, and user can easily forget about it

Cycle `for (unsigned i = 9; i >= 0; i--)` will never stop, because `i >= 0` is always true.
User can workaround this in many ways, but why? Again, this is an additional source of error.

We [disable warning](https://github.com/urho3d/Urho3D/commit/3dcab4c039374515ac3d5e66621c7e17aff25dde) to allow using cycles like
```
void Audio::UpdateInternal(float timeStep)
{
    URHO3D_PROFILE(UpdateAudio);

    // Update in reverse order, because sound sources might remove themselves
    for (u32 i = soundSources_.Size() - 1; i < soundSources_.Size(); --i)
    {
```

Scott Meyers recommeds use signed types in interfaces <https://www.aristeia.com/Papers/C++ReportColumns/sep95.pdf>

Bullet uses int for capacity & size & indices <https://github.com/urho3d/Urho3D/blob/master/Source/ThirdParty/Bullet/src/LinearMath/btAlignedObjectArray.h>

Related: https://github.com/urho3d/Urho3D/pull/2939

1vanK:
C# uses int for sizes https://docs.microsoft.com/en-us/dotnet/api/system.array.length?view=net-6.0

1vanK:
Bjarne Stroustrup recommeds use signed types for sizes: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1428r0.pdf

PR #2941, opened 2022-05-23
Str: replace size and index types from unsigned to signed [cache clear]

1vanK:


1vanK:
Related https://github.com/urho3d/Urho3D/issues/2940

jayrulez:
What's the motivation for this change?

PR #2953, opened 2022-05-27
Make CountSetBits() signed

1vanK:


1vanK:

P.S. Similar function from Bullet also returns int `int countSetBits(ULONG64 bits)`

PR #3025, opened 2022-06-30
Log::WriteFormat fails for LOG_RAW

SirNate0:
Adds special case in Log::WriteFormat for LOG_RAW to match the behavior of Log::Write.

The introduction of WriteFormat in #2685 left LOG_RAW messages unprinted (messages like those produced by LuaScript::Print). Another alternate fix, though it would leave WriteFormat and Write with inconsistent behaviors, would be to return the URHO3D_LOGRAWF and URHO3D_LOGRAW macros to their old definitions from before #2685.

### DIFF
--- a/Source/Urho3D/IO/Log.cpp
+++ b/Source/Urho3D/IO/Log.cpp
@@ -119,9 +119,15 @@ void Log::SetQuiet(bool quiet)
 
 void Log::WriteFormat(int level, const char* format, ...)
 {
-    // No-op if illegal level
-    if (!logInstance || (logInstance->level_ > level && level != LOG_RAW) || level >= LOG_NONE)
+    if (!logInstance)
         return;
+
+    if (level != LOG_RAW)
+    {
+        // No-op if illegal level
+        if (level < LOG_TRACE || level >= LOG_NONE || logInstance->level_ > level)
+            return;
+    }
 
     // Forward to normal Write() after formatting the input
     String message;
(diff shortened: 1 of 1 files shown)


SirNate0:
Sorry, thought I mixed up the comparison for LOG_NONE, and since I had to fix it anyways I was going to add "Fix" to the title (still not used to creating commits directly from github). But I actually had it right, so I'll just leave this as is. If you can, feel free to adjust the title to improve the message in the history.

1vanK:
> Adds special case in Log::WriteFormat for LOG_RAW to match the behavior of Log::Write.

`Log::Write()` prints a message even with `level >= LOG_NONE`

1vanK:
```
void Log::WriteFormat(int level, const char* format, ...)
{
    if (!logInstance)
        return;

    // No-op if illegal level
    if ((level < LOG_TRACE || level >= LOG_NONE) && level != LOG_RAW)
        return;
```

?

SirNate0:
```
void Log::WriteFormat(int level, const char* format, ...)
{
    if (!logInstance)
        return;

    // No-op if illegal level
    if (level != LOG_RAW)
        if (level < LOG_TRACE || level >= LOG_NONE
            || logInstance->level_ > level)
            return;

    // Forward to normal Write() after formatting the input
    String message;
    va_list args;
    va_start(args, format);
    message.AppendWithFormatArgs(format, args);
    va_end(args);

    Write(level, message);
}
```
I generally prefer avoiding goto and I think this is as clear. Thoughts? I'm writing from my phone, so perhaps it would make sense to have the inner if all on one line rather than two.

PR #3051, opened 2022-08-14
Fix: samples not builded when compiled through Android Studio

1vanK:
Perhaps there is a better way to set the environment variable in Android Studio, but I do not know it.


1vanK:
Actually this can be in `android\launcher-app\build.gradle.kts`

PR #3069, opened 2022-08-28
IO: Make stream size and pos i64

1vanK:


1vanK:

1vanK:
i64 is used for a stream pos/size and i32 for read/write data blocks from the steam. This allows to read large files in chunks.  Same way is used in C#: <https://docs.microsoft.com/en-us/dotnet/api/system.io.stream?view=net-6.0> Actually we can combine Serializer + Deserializer + AbstractFile to single Stream class

1vanK:
Not really tested

Issue #3040, opened 2022-07-20
Segfault on SetModel in headless mode


ProtonPunk:
Trying to load and set Model or GetResource<Model> results in SegFault when there is no window open. 

I'm seeing this migrating from 1.7 to latest w/ my headless server app. This works on 1.8-tag but not on master (df3d93ce8). I'm willing to find the commit which introduced it if needed.

How to reproduce: `rake new [crash,tmpdir]`.
Create Setup func in .h and in .cpp use `void crash::Setup() {  engineParameters_ ["Headless"] = true; }`.

```
// This line will segfault when there is no window
object->SetModel(cache->GetResource<Model>("Models/Mushroom.mdl"));
```

```
[Tue Jul 19 21:17:06 2022] INFO: Initialized engine
[Tue Jul 19 21:17:06 2022] DEBUG: Loading resource Models/Mushroom.mdl

Thread 1 "rep-server" received signal SIGSEGV, Segmentation fault.
__memmove_avx_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:416
416     ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S: No such file or directory.
(gdb) bt
#0  __memmove_avx_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:416
#1  0x00007ffff7a3d35e in __GI__IO_file_xsgetn (fp=0xda8870, data=<optimized out>, n=50640) at fileops.c:1304
#2  0x00007ffff7a30ee3 in __GI__IO_fread (buf=0x0, size=50640, count=1, fp=0xda8870) at iofread.c:38
```

1vanK:
> I'm willing to find the commit which introduced it if needed.

It would help a lot

eugeneko:
@ProtonPunk Try version before this commit:
https://github.com/urho3d/Urho3D/commit/653d12f952ee83704f7aa4f34013370e7ebb2a8f
It looks somewhat relevant.

1vanK:
I think this caused by https://github.com/urho3d/Urho3D/pull/2885

Error in

```
bool Model::BeginLoad(Deserializer& source)
{
   ...
            void* dest = buffer->Lock(0, desc.vertexCount_);
```

This return null because gapi == GAPI_NONE in headless mode

1vanK:
Reopen this issue if the problem remains

PR #3070, opened 2022-08-28
Set GAPI even in headless mode

1vanK:


1vanK:
https://github.com/urho3d/Urho3D/issues/3040

PR #3096, opened 2022-09-21
Variant: add c32 (char32_t) type support | URHO3D_ATTRIBUTE: auto detection type of variable

1vanK:


1vanK:
@eugeneko ?

eugeneko:
Initially I wanted to complain that now users cannot use custom types convertible to variant types (e.g. serialize u8 as int), but I think it's not that important, and it's great to have less to write.

1vanK:
I also thought about it, so I decided to consult. Anyway, if such a need arises, it will be possible to return the original define with a different name.

PR #3118, opened 2022-10-18
Fix setWindowTitle in emscripten SDL

dprandle:
In upstream SDL newer versions this is fixed - so updating SDL would also fix this issue.

The docs say the SDK is tested against emscripten 2.0.8, version 2.0.8 (all the way back to around 1.39 I believe?) require the setWindowTitle EM_ASM in SDL to be updated as is here.

Without this, building urho with emscripten 2.0.8 works, but all of the examples crash in chrome and firefox.

Also added a git ignore entry for vscode project files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,6 +60,9 @@ build/
 # Qt Creator project settings
 /CMakeLists.txt.user
 
+# VSCode project settings
+/.vscode/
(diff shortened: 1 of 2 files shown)


1vanK:
Thanks!

PR #3165, opened 2022-12-24
Fix Mac Build

SirNate0:
Changes to get the Mac build working again. Mainly removing a lot of the tricks to speed up and/or workaround bugs with older versions of xcode that break the build the build on newer versions.


SirNate0:
Hopefully I got everything important, as the U3D fork has altered the workflow setup compared to Urho3D.

1vanK:
```
/Applications/Xcode_14.0.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.3.sdk/usr/include/stdio.h:320:6: note: 'fseeko' declared here
int      fseeko(FILE * __stream, off_t __offset, int __whence);
         ^
/Users/runner/work/urho3d/urho3d/Source/Urho3D/IO/File.cpp:63:12: error: use of undeclared identifier 'ftello64'; did you mean 'ftello'?
    return ftello64(stream);
           ^~~~~~~~
           ftello
```

Urho3D uses 64-bit file functions to allow works with files > 4GB (pak for example)

https://github.com/urho3d/urho3d/blob/master/Source/Urho3D/IO/File.cpp#L49-L66

SirNate0:
Thanks, will update. Looks like Mac uses 64 bit file operations by default without having to use different functions:

https://stackoverflow.com/questions/4003479/how-to-enable-large-file-support-under-darwin

1vanK:
Can I merge this?

SirNate0:
Should be good to go unless you want me to squash the last few commits getting the large file sorry working into one (I made a couple of mistakes the first times, hence it's multiple commits now). It at least builds now, which is a big improvement. Further changes could be made (the deployment target is set too high, leading to a number of warnings, but they seem not too be errors).

Regarding testing the Mac stuff, sure, I'd be happy to help out. Just understand that:
1. I don't actually use the Mac often, it just sits in a corner for me to occasionally remote into it to run a build or test. So my tests may be a bit infrequent, probably no more than once a month.
2. I don't have an iOS device to test that side of things with. Hopefully I'll be able to set up the simulator in xcode, but I haven't tried it yet.

SirNate0:
If you have any idea where the deployment target is set, let me know. I wasn't able to find where its set to 12.6 by searching through the Urho directory.

```
/Users/runner/work/urho3d/urho3d/build/ci/Urho3D.xcodeproj: warning: The macOS deployment target 
'MACOSX_DEPLOYMENT_TARGET' is set to 12.6, but the range of supported deployment target versions 
is 10.9 to 12.3.99. (in target 'SpritePacker' from project 'Urho3D')
```

1vanK:
> Should be good to go unless you want me to squash the last few commits getting the large file sorry working into one

I can squash all commits into one when merge PR

> MACOSX_DEPLOYMENT_TARGET

CMAKE_OSX_DEPLOYMENT_TARGET

1vanK:
> So my tests may be a bit infrequent, probably no more than once a month.

That's quite enough. GitHub workflow runs examples on GitHub servers. It's just that sometimes there are errors, for example, that the application starts successfully, but does not show anything except a black screen.

